Issue #1519, opened 2022-04-07
server.config: secerts and keys can not be loaded from config files


dnephin:
## Describe the bug

Secrets and keys can be added to the server config file, but loading those values does not work. It results in an error about an unexpected type.  Config loading for those types depended on a custom `UnmarshalYAML` method. When we switched from YAML decoding to viper (which uses mapstructure) those methods are no longer called because config is first loaded from yaml into a `map[string]interface{}`. The yaml decoder has no knowledge of these custom `UnmarshalYAML`.  Then mapstructure is used to convert from `map[string]interface{}` to the appropriate type, but it doesn't know how to use the `UnmarshalYAML` method.

## Expected behavior

Adding secrets and keys to a config file works.

## Additional context

We can fix this by adding a decode hook and changing the `UnmarshalYAML` methods to implement something the decode hook can use.

jmorganca:
Resolved in #1600 and #1520 (thanks @dnephin !)

PR #1520, opened 2022-04-07
Fix loading of secret providers from config file

dnephin:
## Summary

Initially I thought the problem was simply the structure (a missing `config:` field), so the first few commits fix that problem and the docs. Then I realized it was the `UnmarshalYAML` methods, so I started converting those.

This PR changes the UnmarshalYAML hook to one that mapstructure understands.

## Related Issues

Related to #1519 (but keys need the same fix, which I will do in a follow up PR)

### DIFF
--- a/docs/install/configure/secrets.md
+++ b/docs/install/configure/secrets.md
@@ -30,7 +30,8 @@ Kubernetes takes configuration, like so:
 secrets:
   - name: kubernetes # can optionally provide a custom name
     kind: kubernetes
-    namespace: mynamespace
+    config:
+      namespace: mynamespace
 ```
 
 namespace defaults to whatever is defined in `/var/run/secrets/kubernetes.io/serviceaccount/namespace`, or the `default` namespace.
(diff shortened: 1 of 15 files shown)


dnephin:
Ok, this should be ready to review now.  

I found an easier way to decode polymorphic config without needing to pass the decode function. Mapstructure will take care of decoding from a map, all we need to do is set up some concrete types for the `Config interface{}` field first.

I will follow up with a PR for loading of secret keys (to resolve the TODOs here).

mxyng (commented):
This would be a breaking change since it changes config format correct?

dnephin:
> This would be a breaking change since it changes config format correct?

Maybe! Today there is no way to configure secret providers or key providers. The documented config results in an error.

If this was working today then I'd say yes it's definitely a breaking change. Since nothing works today I think it's safe to say that no one is using this and so no one will be broken by use changing the format.

I'm happy to include "BREAKING CHANGE" in a commit message to call it out, if we want to.

mxyng:
Probably OK to omit the `!` for now

PR #1565, opened 2022-04-11
feat: grants for inactive identities (#1536) (#1564)

### LINKED ISSUE #1536: Users should be able to get grants before they exist in Infra
## Describe the bug
When creating a grant it should be possible to specify a user, machine, or group that do not yet exist in Infra. This will be a common case because users/groups come from identity providers "just in time". So it will be common to create a grant for a user/group before they login to infra for the first time.

## Expected behavior
I'm able to create a grant for a user/group based on their name and have that resolve later when the user logs in for the first time.

## Additional context

This should work, hello@example.com may exist in Okta, but has not signed in yet.
```
infra grants add hello@example.com --role view infra                 
ERROR  No identity of name hello@example.com exists
```

### LINKED ISSUE #1564: `infra list` output is confusing before destination is added
## Describe the bug
If a user is granted access to a destination before and it is not yet connected they get a confusing output from `infra list` that tells them they have no grants.

```
$ infra grants add hi@example.com kubernetes.docker-desktop --role cluster-admin
Access granted!

$ infra login localhost
? Select a login method: Login as a local user
? Email: hi@example.com
? Password: **********

$ infra list           
No grants found for user "hi@example.com"
```

## Expected behavior
The output from `infra list` tells the user that they do not have access to any destinations.

BruceMacD:
- allow adding grants for identities before they are known to infra
- hide identities that are not connected to an identity provider
- allow creating an identity linked to an identity provider
- output when a new identity is granted access

## Summary
When creating a grant it should be possible to specify a user, machine, or group that do not yet exist in Infra. This will be a common case because users/groups come from identity providers "just in time". So it will be common to create a grant for a user/group before they login to infra for the first time.

This change allows grants to be created before a user is associated with an identity provider. Before an user is associated with an identity provider they will be considered "inactive".

CLI flow:
```
$ infra grants add test@example.com kubernetes.docker-desktop --role cluster-admin
New unlinked identity "test@example.com" added to Infra
Access granted!

$ infra id list                                                                   
  NAME (4)                     TYPE     
  admin                           machine  
  connector                    machine  
  test@example.com     user     

$ infra id add test@example.com  
Created user identity.
Email: test@example.com
Password: something

$ infra id list                                                                   
  NAME (4)                     TYPE     
  admin                           machine  
  connector                    machine  
  test@example.com     user     
```

## Checklist

## Related Issues

Resolves #1536 
Resolves #1564

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -199,7 +199,7 @@ func delete(client Client, path string) error {
 }
 
 func (c Client) ListIdentities(req ListIdentitiesRequest) ([]Identity, error) {
-	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "showInactive": strconv.FormatBool(req.ShowInactive)})
+	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "includeUnlinked": strconv.FormatBool(req.IncludeUnlinked)})
 }
 
 func (c Client) GetIdentity(id uid.ID) (*Identity, error) {
--- a/api/client.go
+++ b/api/client.go
@@ -199,7 +199,7 @@ func delete(client Client, path string) error {
 }
 
 func (c Client) ListIdentities(req ListIdentitiesRequest) ([]Identity, error) {
-	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "includeUnlinked": strconv.FormatBool(req.IncludeUnlinked)})
+	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "all": strconv.FormatBool(req.All)})
 }
 
 func (c Client) GetIdentity(id uid.ID) (*Identity, error) {
--- a/api/client.go
+++ b/api/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 
 	"github.com/infrahq/infra/uid"
 )
@@ -198,7 +199,7 @@ func delete(client Client, path string) error {
 }
 
 func (c Client) ListIdentities(req ListIdentitiesRequest) ([]Identity, error) {
-	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name})
+	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "showInactive": strconv.FormatBool(req.ShowInactive)})
 }
 
 func (c Client) GetIdentity(id uid.ID) (*Identity, error) {
--- a/api/client.go
+++ b/api/client.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strconv"
 
 	"github.com/infrahq/infra/uid"
 )
@@ -199,7 +198,7 @@ func delete(client Client, path string) error {
 }
 
 func (c Client) ListIdentities(req ListIdentitiesRequest) ([]Identity, error) {
-	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "all": strconv.FormatBool(req.All)})
+	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name})
 }
 
 func (c Client) GetIdentity(id uid.ID) (*Identity, error) {
--- a/api/identity.go
+++ b/api/identity.go
@@ -14,12 +14,14 @@ type Identity struct {
 }
 
 type ListIdentitiesRequest struct {
-	Name string `form:"name"`
+	Name         string `form:"name"`
+	ShowInactive bool   `form:"showInactive"`
 }
 
 type CreateIdentityRequest struct {
-	Name string `json:"name" validate:"required"`
-	Kind string `json:"kind" validate:"required,oneof=user machine"`
+	Name       string  `json:"name" validate:"required"`
+	Kind       string  `json:"kind" validate:"required,oneof=user machine"`
+	ProviderID *uid.ID `json:"providerID"`
 }
 
 type UpdateIdentityRequest struct {
@@ -30,6 +32,6 @@ type UpdateIdentityRequest struct {
 type CreateIdentityResponse struct {
 	ID              uid.ID `json:"id"`
 	Name            string `json:"name" validate:"required"`
-	ProviderID      uid.ID `json:"providerID" validate:"required"`
+	ProviderID      uid.ID `json:"providerID,omitempty" `
 	OneTimePassword string `json:"oneTimePassword,omitempty"`
 }
--- a/api/identity.go
+++ b/api/identity.go
@@ -19,9 +19,9 @@ type ListIdentitiesRequest struct {
 }
 
 type CreateIdentityRequest struct {
-	Name         string `json:"name" validate:"required"`
-	Kind         string `json:"kind" validate:"required,oneof=user machine"`
-	ProviderName string `json:"providerName"`
+	Name       string  `json:"name" validate:"required"`
+	Kind       string  `json:"kind" validate:"required,oneof=user machine"`
+	ProviderID *uid.ID `json:"providerID"`
 }
 
 type UpdateIdentityRequest struct {
@@ -32,6 +32,6 @@ type UpdateIdentityRequest struct {
 type CreateIdentityResponse struct {
 	ID              uid.ID `json:"id"`
 	Name            string `json:"name" validate:"required"`
-	ProviderName    string `json:"providerName,omitempty" `
+	ProviderID      uid.ID `json:"providerID,omitempty" `
 	OneTimePassword string `json:"oneTimePassword,omitempty"`
 }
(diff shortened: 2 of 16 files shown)


BruceMacD:
@dnephin I've added a handler test for identity creation, this is good for final reviews now 👍

internal/cmd/identities.go
```diff
@@ -57,7 +58,13 @@ EMAIL must contain a valid email address in the form of "local@domain".
...
 			}
 
-			createResp, err := CreateLocalIdentity(name)
+			// an identity created with this command always exists in the local infra provider
+			infraProvider, err := GetInfraProvider()
+			if err != nil {
+				return err
+			}
+
+			createResp, err := CreateProviderIdentity(name, infraProvider)
```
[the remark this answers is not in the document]
BruceMacD: I made this change too

internal/access/identity.go
```diff
@@ -86,13 +86,47 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
...
+	if includeUnlinked {
+		return identities, nil
+	}
+
+	// filter out identities that do not have a linked provider
+
+	providerIdentities, err := data.ListProviderUsers(db)
+	if err != nil {
+		return nil, err
+	}
```
ssoroka, line 109: this is potentially too expensive. you should instead query for ProviderUsers that are in the list of returned users and return only those.

internal/access/identity.go
```diff
@@ -86,13 +86,47 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
...
+
+	for _, active := range providerIdentities {
+		activeIdentity := idToIdentity[active.IdentityID]
+		if activeIdentity != nil {
+			activeIdentities = append(activeIdentities, *activeIdentity)
+
+			// remove from the map so we don't add identities multiple times
+			idToIdentity[active.IdentityID] = nil
+		}
+	}
```
ssoroka, line 127: might be simpler with a map of providerUsers and iterate identities instead

internal/cmd/grants.go
```diff
@@ -282,7 +282,13 @@ func getIDByName(client *api.Client, name string, identityType identityType) (ui
...
 		switch len(groups) {
 		case 0:
-			return "", fmt.Errorf("No group of name %s exists", name)
+			created, err := client.CreateGroup(&api.CreateGroupRequest{Name: name})
+			if err != nil {
+				return "", err
+			}
+			fmt.Printf("New group %q added to Infra\n", name)
+
+			id = uid.NewGroupPolymorphicID(created.ID)
```
ssoroka, line 291: fun side effect. mistyping a grant remove can add that group

internal/cmd/grants.go
```diff
@@ -296,7 +302,13 @@ func getIDByName(client *api.Client, name string, identityType identityType) (ui
...
 		switch len(identities) {
 		case 0:
-			return "", fmt.Errorf("No identity of name %s exists", name)
+			created, err := CreateIdentity(name)
+			if err != nil {
+				return "", err
+			}
+			fmt.Printf("New unlinked identity %q added to Infra\n", name)
+
+			id = uid.NewIdentityPolymorphicID(created.ID)
```
ssoroka, line 311: removing grants for identities can inadvertently add that identity. 
also cli should probably confirm before adding this?

internal/cmd/identities_test.go
```diff
@@ -117,9 +117,9 @@ func TestIdentities(t *testing.T) {
 					assert.NilError(t, err)
 
 					respBody := api.CreateIdentityResponse{
-						ID:         uid.New(),
-						Name:       createIdentityReq.Name,
-						ProviderID: providerID,
+						ID:           uid.New(),
+						Name:         createIdentityReq.Name,
+						ProviderName: "infra",
```
ssoroka, line 122: not sure I understand this. Seems like the only valid value is "infra", and the identity isn't tied to a provider anyway

internal/cmd/list.go
```diff
@@ -114,7 +114,7 @@ func list() error {
 	if len(rows) > 0 {
 		printTable(rows)
 	} else {
-		fmt.Printf("No grants found for user %q\n", config.Name)
+		fmt.Println("You have not been granted access to any active destinations")
```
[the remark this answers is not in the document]
ssoroka: listing grants for other people will still say "You..."

internal/cmd/providers.go
```diff
@@ -162,6 +163,15 @@ func newProvidersRemoveCmd() *cobra.Command {
 	}
 }
 
+func GetInfraProvider() (*api.Provider, error) {
+	client, err := defaultAPIClient()
+	if err != nil {
+		return nil, err
+	}
+
+	return GetProviderByName(client, models.InternalInfraProviderName)
```
ssoroka, line 172: not great that it references the models package. probably should use the string "infra"

internal/server/handlers.go
```diff
@@ -58,33 +58,67 @@ func (a *API) CreateIdentity(c *gin.Context, r *api.CreateIdentityRequest) (*api
 		Kind: kind,
 	}
 
-	if err := access.CreateIdentity(c, identity); err != nil {
-		return nil, err
-	}
+	// identity creation linked to a provider should be attempted even if an identity is already known
+	if r.ProviderName != "" {
```
ssoroka, line 62: I don't think CreateIdentity should be concerned with providers. Identity is a stand-alone infra-owned thing now

internal/server/handlers.go
```diff
@@ -58,33 +58,67 @@ func (a *API) CreateIdentity(c *gin.Context, r *api.CreateIdentityRequest) (*api
...
 		}
 
-		resp.OneTimePassword = oneTimePassword
+		if len(providers) != 1 {
+			// should not happen
+			return nil, fmt.Errorf("multiple providers match the specified name")
+		}
+
+		// this is a placeholder for when the user logs in using this provider
+		_, err = access.CreateProviderUser(c, &providers[0], identity)
```
ssoroka, line 102: could just always create the infra provider here? Not sure we need any

internal/server/handlers.go
```diff
@@ -58,33 +58,67 @@ func (a *API) CreateIdentity(c *gin.Context, r *api.CreateIdentityRequest) (*api
...
+
+		resp.ProviderName = providers[0].Name
+
+		if (identity.Kind == models.UserKind) && (models.InternalInfraProviderName == providers[0].Name) {
+			oneTimePassword, err := access.CreateCredential(c, *identity)
+			if err != nil {
+				return nil, fmt.Errorf("create credential: %w", err)
+			}
+
+			resp.OneTimePassword = oneTimePassword
```
ssoroka, line 115: seems very weird that we might be creating login info for a temp user

internal/cmd/identities.go
```diff
@@ -163,6 +169,9 @@ func newIdentitiesListCmd() *cobra.Command {
 			return nil
 		},
 	}
+
+	cmd.Flags().Bool("unlinked", false, "Show identities that exist in grants but are not linked to an identity provider")
```
jmorganca, line 173: `--all` show all users

api/client.go
```diff
@@ -198,7 +199,7 @@ func delete(client Client, path string) error {
 }
 
 func (c Client) ListIdentities(req ListIdentitiesRequest) ([]Identity, error) {
-	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name})
+	return list[Identity](c, "/v1/identities", map[string]string{"name": req.Name, "includeUnlinked": strconv.FormatBool(req.IncludeUnlinked)})
```
jmorganca, line 202: "all"

api/identity.go
```diff
@@ -14,12 +14,14 @@ type Identity struct {
 }
 
 type ListIdentitiesRequest struct {
-	Name string `form:"name"`
+	Name            string `form:"name"`
+	IncludeUnlinked bool   `form:"includeUnlinked" note:"Show identities that exist in grants but are not linked to an identity provider"`
```
jmorganca, line 18: `All` bool "show all users, even those who haven't logged in yet"

internal/cmd/identities.go
```diff
@@ -34,7 +34,8 @@ func newIdentitiesCmd() *cobra.Command {
 }
 
 type identityCmdOptions struct {
```
jmorganca, line 36: Should use structs for separate commands

```
type identityAddCmdOptions struct {
  all bool `mapstructure:"all"`
}
```

internal/cmd/identities_test.go
BruceMacD: Changed to a "SetOneTimePassword" flag in the request that sets credentials in the Infra provider. Dropped the ProviderName

internal/cmd/list.go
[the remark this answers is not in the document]
BruceMacD: This is only used in the `infra list` command

jmorganca (approved):

PR #1569, opened 2022-04-11
fix: migrate identity provider_id change

### LINKED ISSUE #1570: Infra users that exist across multiple providers need migration from 0.9.0 to current
## Describe the bug
If a user was created associated with 2 different identity providers duplicate records are created in the identity table.

## Expected behavior
Only one user is created for both records and migrated accordingly.

## Steps to reproduce
In v0.9.0:
```
$ infra grants add --user hi@example.com kubernetes.docker-desktop --role cluster-admin --provider okta
Access granted!
$ infra grants add --user hi@example.com kubernetes.docker-desktop --role cluster-admin --provider infra
Access granted!
```

Then upgrade Infra and there will be two users with the same name in the `infra id list` output.

ssoroka:
## Checklist

## Related Issues

Resolves #1570


internal/server/data/migrations_test.go
```diff
@@ -0,0 +1,65 @@
...
+	// fp := secrets.NewFileSecretProviderFromConfig(secrets.FileConfig{})
+
+	// kp := secrets.NewNativeSecretProvider(fp)
+	// key, err := kp.DecryptDataKey("migrationdata/202204111503.key", keyRec.Encrypted)
+	// assert.NilError(t, err)
+
+	// models.SymmetricKey = key
+
+	logging.L = zaptest.NewLogger(t)
+	logging.S = logging.L.Sugar()
```
dnephin, line 62: I think this needs to be reset with a `t.Cleanup`, otherwise subsequent tests could panic trying to write to the wrong `testing.T`.

There is an example of that here: https://github.com/infrahq/infra/blob/6c04f37699628bed5bbe515131e8b9728c10b5fd/internal/server/server_test.go#L52-L59

internal/server/data/migrations_test.go
```diff
@@ -0,0 +1,65 @@
...
+	// keyRec, err := GetEncryptionKey(db, ByName("dbkey"))
+	// assert.NilError(t, err)
+
+	// fp := secrets.NewFileSecretProviderFromConfig(secrets.FileConfig{})
+
+	// kp := secrets.NewNativeSecretProvider(fp)
+	// key, err := kp.DecryptDataKey("migrationdata/202204111503.key", keyRec.Encrypted)
+	// assert.NilError(t, err)
+
+	// models.SymmetricKey = key
```
dnephin, line 59: debugging to be removed?

internal/cmd/cmd.go
```diff
@@ -306,9 +306,11 @@ func NewRootCmd() *cobra.Command {
 			}
 			if rootOptions.Info {
 				if err := mustBeLoggedIn(); err != nil {
-					return err
+					return fmt.Errorf("login check: %w", err)
+				}
+				if err := info(); err != nil {
+					return fmt.Errorf("info: %w", err)
 				}
-				return info()
```
dnephin, line 311: Missing `return nil` here? I don't think we want to print help when `--info` is used, do we?

internal/server/data/data.go
```diff
@@ -21,6 +21,19 @@ import (
 )
 
 func NewDB(connection gorm.Dialector) (*gorm.DB, error) {
+	db, err := newRawDB(connection)
```
dnephin, line 24: We could make the name imply that it runs migration (`InitializeDB` ?), but maybe a godoc comment would be enough?

```suggestion
// NewDB creates a new database connection and runs any required database migrations
// before returning the connection.
func NewDB(connection gorm.Dialector) (*gorm.DB, error) {
	db, err := newRawDB(connection)
```

internal/server/data/migrations_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/internal/server/models"
+	"go.uber.org/zap/zaptest"
+	"gorm.io/gorm"
+	"gotest.tools/v3/assert"
+)
+
+func Test202204111503(t *testing.T) {
+	db := setupWithNoMigrations(t, func(db *gorm.DB) {
+		loadSQL(t, db, "202204111503")
```
dnephin, line 17: Any thoughts on where we can document the steps to create this file? Maybe as a comment in this file somewhere?

dnephin (commented):
Nice! Being able to test migrations is awesome!

internal/server/data/data.go
ssoroka: yeah, since that includes automigrate and the memory db is usually empty, it's pretty useless to return a non-migrated db.. except in the case where you're trying to test the migrations! :D

dnephin (approved):
Nice! Tests for migrations are great! I didn't read over the migration logic itself, but the testing and other code changes LGTM

PR #1571, opened 2022-04-11
fix: allow use by name alone

### LINKED ISSUE #1552: Infra use command not checking for type
```
mchiang@Michaels-MacBook-Pro infra % ./infra use ubernetes.development-72f9584e 
Switched to context "infra:development-72f9584e".
```
```
Switched to context "infra:development-72f9584e".
mchiang@Michaels-MacBook-Pro infra % ./infra use development-72f9584e  
ERROR  invalid argument
```

We should just allow the name of the cluster, and only require the type if there are duplicate names from multiple infrastructures (i.e. kubernetes cluster named dev  and a database named dev )

mxyng:
## Summary

`infra use` should allow referencing by name, i.e. without the `kubernetes.` prefix

## Checklist

- [ ] Commit message conforms to [Conventional Commit][1]

## Related Issues

Resolves #1552

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -171,19 +171,18 @@ func newUseCmd() *cobra.Command {
 		Use:   "use DESTINATION",
 		Short: "Access a destination",
 		Example: `
-# Connect to a Kubernetes cluster
-$ infra use kubernetes.development
+# Use a Kubernetes context
+$ infra use development
 
-# Connect to a Kubernetes namespace
-$ infra use kubernetes.development.kube-system
-		`,
+# Use a Kubernetes namespace context
+$ infra use development.kube-system`,
 		Args:  cobra.ExactArgs(1),
 		Group: "Core commands:",
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			return mustBeLoggedIn()
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			name := args[0]
+			destination := args[0]
 
 			client, err := defaultAPIClient()
 			if err != nil {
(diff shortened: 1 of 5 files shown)


internal/cmd/kubernetes.go
```diff
@@ -23,25 +23,35 @@ func clientConfig() clientcmd.ClientConfig {
 	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
 }
 
-func kubernetesSetContext(name string) error {
+func kubernetesSetContext(cluster, namespace string) error {
```
dnephin, line 26: Nice! I better understand how this works now!

A couple tests for this function to show it builds the expected config might be nice, but not blocking.

internal/cmd/cmd.go
```diff
@@ -200,17 +199,22 @@ $ infra use kubernetes.development.kube-system
 				return err
 			}
 
-			parts := strings.Split(name, ".")
+			parts := strings.Split(destination, ".")
```
dnephin, line 202: Seems we rely on there being at most 2 parts, so maybe this?

```suggestion
			parts := strings.SplitN(destination, ".", 2)
```

dnephin (approved):
Good fix! Some test coverage for `kubernetesSetContext` and/or the `infra use`  command would be great, but could be a follow up.

internal/cmd/cmd.go
mxyng: Ah it's possible the user supplies more than 2 parts, e.g. `kubernetes.cluster.namespace`, in which case we want to discard the first part and pass the second and third part to `SetContext`

PR #1586, opened 2022-04-12
fix: docs build

jmorganca:
Fixes the doc build

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,7 +87,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: readmeio/rdme@7.0.3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -99,7 +99,7 @@ jobs:
           repository: infrahq/web
           token: ${{ secrets.DOCS_TOKEN }}
       - run: |
-          echo $GITHUB_REF  > ./docs-ref
+          echo $GITHUB_SHA  > ./docs-ref
```
mxyng, line 102: maybe `tee` here so it also prints out the input?

```suggestion
          echo $GITHUB_SHA  | tee ./docs-ref
```

mxyng (approved):

PR #1593, opened 2022-04-12
maintain: remove relogin

mxyng:
## Summary

Remove `relogin()` since even tokens already exits with an error if the session is no longer valid.

## Checklist

- [ ] Commit message conforms to [Conventional Commit][1]

## Related Issues

Resolves #


BruceMacD (approved):
Tested and validated that a helpful error message is still output after expiry:
```
$ kubectl get pods -n infrahq
ERROR  Session expired. Run 'infra login' to start a new session.
```

PR #1594, opened 2022-04-12
feat: add request timeouts

ssoroka:
## Summary

timeout both requests and db transactions.

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -32,7 +32,6 @@ func TimeoutMiddleware(timeout time.Duration) gin.HandlerFunc {
 
 		start := time.Now()
 
-		c.Set("ctx", ctx)
 		c.Next()
 
 		if elapsed := time.Since(start); elapsed > timeout {
@@ -44,15 +43,7 @@ func TimeoutMiddleware(timeout time.Duration) gin.HandlerFunc {
 // DatabaseMiddleware injects a `db` object into the Gin context.
 func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		ctx := context.Background()
-
-		if ctxIntf, ok := c.Get("ctx"); ok {
-			if existingCtx, ok := ctxIntf.(context.Context); ok {
-				ctx = existingCtx
-			}
-		}
-
-		err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		err := db.WithContext(c.Request.Context()).Transaction(func(tx *gorm.DB) error {
 			c.Set("db", tx)
 			c.Next()
 			return nil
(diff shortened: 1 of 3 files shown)


internal/server/middleware.go
```diff
@@ -15,37 +15,37 @@ import (
...
 func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		err := db.Transaction(func(tx *gorm.DB) error {
+		ctx := context.Background()
+
+		if ctxIntf, ok := c.Get("ctx"); ok {
+			if existingCtx, ok := ctxIntf.(context.Context); ok {
+				ctx = existingCtx
+			}
+		}
```
dnephin, line 46: Shouldn't this use `c.Request.Context()` as it did before? Having two contexts in the `gin.Context` seems like it will be prone to error, and unexpected.

dnephin (commented):
Thanks for fixing this! Looks like we accidentally removed this middleware while fixing another problem. That might have been my mistake, sorry about that.

Maybe we should set this timeout using `net/http.Server.ConnContext` (https://pkg.go.dev/net/http#Server), instead of a middleware? 

I think that would remove the need for the middleware, and would also allow us to set `net.Conn.SetWriteDeadline` (https://pkg.go.dev/net#Conn) to prevent malicious clients from keeping a connection open. The context deadline won't prevent that I don't think.

internal/server/middleware.go
mxyng: +1 on this. Having a single context will make it much easier to reason about the code

ssoroka: the internal gin.Context actually defers to c.Request.Context but does so pretty elegantly with checks and reasonable defaults if it's not available, so the two are nearly equivalent.

mxyng: Maybe I'm missing something but can't you call `c.Request.Context()` to get the `ctx`?

ssoroka:
> remove the need for the middleware
I suspect we'll still need the middleware to cancel the db context. I wasn't convinced gin was relaying that properly.

> SetDeadline is probably the right one to timeout both reads and writes
I think we can add this as a later change? seems like connection timeouts are their own beast

dnephin:
Keeping `SetDeadline` for a future change sounds good. I think it would be good to at least use `request.Context()`. We can look at switching from middleware to `net/http.Server.ConnContext` when we look at using `SetDeadline`.

dnephin:
> You still can't move away from the middleware because closing the connection will not close on-going db queries.

Hmm, I thought that closing the connection will cancel the context. We can still set a deadline on the context as well. Doesn't that stop the DB query ?

dnephin (approved):

dnephin:
https://github.com/golang/go/commit/faf882d1d427e8c8a9a1be00d8ddcab81d1e848e suggests this works the way I would expect. If the server fails to write, I think it cancels the context passed to the server handler.

Issue #1159, opened 2022-03-09
SQL statements printed on error


jmorganca:
If unique constraint fails, the entire SQL INSERT statement will be printed in the logs. See #1160

PR #1595, opened 2022-04-12
improve: put sql logs in debug level logs

ssoroka:
## Summary

it can be really handy to see the SQL queries happening to debug an issue. This PR turns them on when log level is set to debug. Most sensitive fields are encrypted anyway, so there's not much to leak to the logs. I don't believe this raises security concerns.

Example output looks like this:
```
2022-04-12T15:36:49.435Z	DEBUG	logging/gorm.go:51	SQL (1 rows, 108.625µs): SELECT * FROM `identities` WHERE id = 36836655819464704 AND `identities`.`deleted_at` IS NULL ORDER BY `identities`.`id` LIMIT 1
2022-04-12T15:36:49.436Z	DEBUG	logging/gorm.go:51	SQL (1 rows, 281.25µs): UPDATE `identities` SET `created_at`="2022-04-12 15:35:43.139",`updated_at`="2022-04-12 15:36:49.435",`deleted_at`=NULL,`kind`="machine",`name`="admin",`last_seen_at`="2022-04-12 15:36:49.435" WHERE `identities`.`deleted_at` IS NULL AND `id` = 36836655819464704
2022-04-12T15:36:49.437Z	DEBUG	logging/gorm.go:51	SQL (0 rows, 224µs): SELECT `groups`.`id`,`groups`.`created_at`,`groups`.`updated_at`,`groups`.`deleted_at`,`groups`.`name` FROM `groups` JOIN `identities_groups` ON `identities_groups`.`group_id` = `groups`.`id` AND `identities_groups`.`identity_id` = 36836655819464704 WHERE `groups`.`deleted_at` IS NULL
2022-04-12T16:35:43.203Z	DEBUG	logging/gorm.go:51	SQL (1 rows, 40.354791ms): SELECT count(*) FROM `identities` WHERE `identities`.`deleted_at` IS NULL
2022-04-12T16:35:43.207Z	DEBUG	logging/gorm.go:51	SQL (1 rows, 182.958µs): SELECT count(*) FROM `groups` WHERE `groups`.`deleted_at` IS NULL
2022-04-12T16:35:43.208Z	DEBUG	logging/gorm.go:51	SQL (1 rows, 522.209µs): SELECT count(*) FROM `grants` WHERE `grants`.`deleted_at` IS NULL
2022-04-12T16:35:43.208Z	DEBUG	logging/gorm.go:51	SQL (1 rows, 82.125µs): SELECT count(*) FROM `providers` WHERE `providers`.`deleted_at` IS NULL
```

## Checklist


internal/logging/gorm.go
```diff
@@ -47,6 +47,6 @@ func (l *gormLogger) Error(ctx context.Context, s string, args ...interface{}) {
 }
 
 func (l *gormLogger) Trace(ctx context.Context, begin time.Time, fc func() (sql string, rowsAffected int64), err error) {
-	// don't log traces for now
-	_, _ = fc()
+	sql, rows := fc()
+	l.l.Debugf("SQL (%d rows, %s): %s", rows, time.Since(begin).String(), sql)
```
dnephin, line 51: If https://github.com/uber-go/zap/issues/1076 happens maybe we can use the structured fields for this. Until then a single string is probably better.

dnephin, line 51: minor optimization: we can avoid doing some work when debug logging is off by removing the `String()`. The logger will call that for us when debug logs are enabled, but the work is skipped when debug logs are off.

```suggestion
	l.l.Debugf("SQL (%d rows, %s): %s", rows, time.Since(begin), sql)
```

We could go even further and optimize away the `time.Since()` call in that case with our own type, but doesn't seem worth it right now.

dnephin (approved):
This seems safe to me. Debug logs should be off in production generally as well.

Minor suggestion for a minor optimization, but likely isn't important right now.

mxyng:
Need to keep https://github.com/infrahq/infra/issues/1159 in mind. There are definitely cases were the SQL contains sensitive information.

ssoroka:
So do we want to merge or no? Currently any sensitive fields are EncryptedAtRest fields, which means the logged query will not have readable sensitive fields. This just leaves fields which we haven't marked as encrypted. Are any of concern?

ssoroka:
I've reviewed the models and there's no others that contain sensitive information. All sensitive fields are now encrypted at rest, and thus unreadable to the sql logger.

internal/server/data/data.go
```diff
@@ -35,7 +35,16 @@ func NewDB(connection gorm.Dialector) (*gorm.DB, error) {
...
 }
 
-func newRawDB(connection gorm.Dialector) (*gorm.DB, error) {
+func PreMigrate(db *gorm.DB) error {
+	return premigrate(db)
+}
+
+func Migrate(db *gorm.DB) error {
+	return migrate(db)
+}
```
dnephin, line 44: We can remove these and export `migrate` and `premigrate` since they are the same function

```suggestion
```

internal/server/data/migrations.go
```diff
@@ -458,6 +479,15 @@ func migrate(db *gorm.DB) error {
 	return automigrate(db)
 }
 
+func premigrate(db *gorm.DB) error {
```
dnephin, line 482:
```suggestion
func PreMigrate(db *gorm.DB) error {
```

dnephin, line 482:
```suggestion
func PreMigrate(db *gorm.DB) error {
```

internal/server/models/encryption.go
```diff
@@ -58,3 +58,49 @@ func (s *EncryptedAtRest) Scan(v interface{}) error {
 
 	return nil
 }
+
+// EncryptedAtRestBytes defines a field that knows how to encrypt and decrypt itself with Gorm
+// it depends on the SymmetricKey being set for this package.
+type EncryptedAtRestBytes []byte
```
dnephin, line 64: Since `[]byte` and `string` are effectively the same , I wonder if we could have only a single one of these (`EncryptedAtRestBytes` and `EncryptedAtRest`), and add a `String()` or `Bytes()` method to convert them?

Looks like the string version is already doing `[]byte(v)`, so the convert is already happening in one case.

dnephin, line 64: Since `[]byte` and `string` are effectively the same , I wonder if we could have only a single one of these (`EncryptedAtRestBytes` and `EncryptedAtRest`), and add a `String()` or `Bytes()` method to convert them?

dnephin (commented):
I guess I was late to review, but I'll submit my feedback anyway.

Some godoc comments for `PreMigrate`, `Migrate`, and `automigrate` would be great. I don't think I follow exactly what each of those does.

PR #1597, opened 2022-04-12
Remove the infra user role

### LINKED ISSUE #1585: User should only have one `infra` grant
```
jmorgan@mbp infra % infra grants list
  IDENTITY          ACCESS  DESTINATION  
  jeff@infrahq.com  user    infra        
  jeff@infrahq.com  admin   infra 
```

In this case – is Jeff a `user` or an `admin`?

dnephin:
## Summary

Instead of a user role that we grant to all identities, we can allow authenticated requests to list and get destinations.

This is effectively the same behaviour, without showing the extra grant in the list of grants.

## Related Issues

Resolves #1585

## Open questions

* Do we need a migration for this to remove any existing grants for "infra user role" ?

### DIFF
--- a/internal/access/destination.go
+++ b/internal/access/destination.go
@@ -27,20 +27,12 @@ func SaveDestination(c *gin.Context, destination *models.Destination) error {
 }
 
 func GetDestination(c *gin.Context, id uid.ID) (*models.Destination, error) {
-	db, err := RequireInfraRole(c, models.InfraAdminRole, models.InfraViewRole, models.InfraConnectorRole, models.InfraUserRole)
-	if err != nil {
-		return nil, err
-	}
-
+	db := getDB(c)
 	return data.GetDestination(db, data.ByID(id))
 }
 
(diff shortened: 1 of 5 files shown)


internal/access/destination.go
```diff
@@ -27,20 +27,12 @@ func SaveDestination(c *gin.Context, destination *models.Destination) error {
 }
 
 func GetDestination(c *gin.Context, id uid.ID) (*models.Destination, error) {
-	db, err := RequireInfraRole(c, models.InfraAdminRole, models.InfraViewRole, models.InfraConnectorRole, models.InfraUserRole)
```
dnephin, line 30: Previously every identity was granted `InfraUserRole`, so this statement would be a no-op unless someone had deleted the grant for a particular user (which I assume would have caused other problems).

Now we simply allow any authenticated user to perform these two operations, which should be the same as before.

BruceMacD (approved):
Good refactor here!

Migration is a nice to have, but the user grants that exist in the database won't cause any harm.

PR #1599, opened 2022-04-12
docs: OpenAPI enum for oneof (#1496)

### LINKED ISSUE #1496: docs: OpenAPI spec should be enum for `oneof` fields
## Describe the bug
If `oneof` is specified on a field it should be an enum in the OpenAPI docs.

BruceMacD:
- generate an enum for one-of tags in open api spec
- refactor some duplicated tag reading code
- add "one of" to identity api struct
- fix /v1/identites in the right category

## Summary
API structs that are validated to be "one of" set values should be enums in the OpenAPI spec.

## Checklist

## Related Issues

Resolves #1496

### DIFF
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -24,6 +24,7 @@ var (
 	pathIDReplacer            = regexp.MustCompile(`:\w+`)
 	funcPartialNameToTagNames = map[string]string{
 		"Grant":       "Grants",
+		"Identities":  "Identities",
 		"Identity":    "Identities",
 		"Group":       "Groups",
 		"AccessKey":   "Authentication",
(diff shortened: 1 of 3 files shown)


internal/server/openapi.go
```diff
@@ -24,6 +24,7 @@ var (
 	pathIDReplacer            = regexp.MustCompile(`:\w+`)
 	funcPartialNameToTagNames = map[string]string{
 		"Grant":       "Grants",
+		"Identities":  "Identities",
```
mxyng, line 27: Is this necessarily? From the context, it seems to be setting up aliases so `Identities->Identities` is redundant

mxyng (approved):

internal/server/openapi.go
BruceMacD: This one is for categorizing names, I missed it the first time. Without it this happens:
image

PR #1603, opened 2022-04-12
fix: not logging api calls

mxyng:
## Summary

Server is not currently logging incoming requests. This was due to the logging middleware being (accidentally?) removed. 

Replacing the `IdentityAwareLoggingMiddleware` which doesn't work in the current state. It requires `identity` to exist in the context but `identity` is not set until much later in the `AuthenticationMiddleware`

## Checklist

## Related Issues

Resolves #


dnephin (approved):

mxyng:
Good call. I'm also adding `RequestTimeoutMiddleware()` back in. This will change in #1594 but worth having until then

mxyng:
I'm specifically not applying the logging middleware to the UI group as I feel it doesn't add much value. Can attach it earlier in the future if we need it though

PR #1604, opened 2022-04-12
improve(api): move /debug/pprof to /v1/debug/pprof

dnephin:
## Summary

I found this `http.StripPrefix` which makes it easy to move the pprof handlers to `/v1`. This is probably good because any paths that start with a different prefix are assumed to be paths for serving the UI.

Would be a breaking change if anyone was using these, but I don't think anyone has a use for them yet.

### DIFF
--- a/internal/server/debug.go
+++ b/internal/server/debug.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"net/http"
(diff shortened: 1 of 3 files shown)


jmorganca (approved):
Awesome!

PR #1605, opened 2022-04-12
improve: log identity name for logged in API request

dnephin:
## Summary

Removes the IdentityAwareMiddleware as it did not work, and because we
can do this with a single middleware. It did not work for a few reasons.
It attempted to read an identity from the context before it existed, and
because the identity in the context is a model.Identity, not just an ID.

This new approach works because we read the identity out after the
request handler runs.

Tested manually and saw authenticated requests with an identity field,
and unauthenticated requests without.

### DIFF
--- a/internal/logging/logger.go
+++ b/internal/logging/logger.go
@@ -14,7 +14,7 @@ import (
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/term"
 
-	"github.com/infrahq/infra/uid"
+	"github.com/infrahq/infra/internal/server/models"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


internal/logging/logger.go
```diff
@@ -148,25 +148,6 @@ func (w *filteredWriterSyncer) Sync() error {
...
-			}
-		}
-
-		c.Next()
-	}
-}
-
 // Logger gets the request-specific logger from the context
 // If a request-specific logger cannot be found, use the default logger
 func Logger(c *gin.Context) *zap.Logger {
```
mxyng, line 153: This can likely be removed if `logger` isn't being set. The only place it should be called in `errors.go:sendAPIError()`

internal/logging/logger.go
```diff
@@ -212,7 +193,14 @@ func Middleware() gin.HandlerFunc {
 			c.Request.ContentLength,
 		)
 
-		SugarLogger(c).Infow(
+		logger := SugarLogger(c)
```
mxyng, line 196: This is not very useful since `identity` is not being set so it reverts to using `logging.L`. Might as well use `logging.L` directly

dnephin: Good call, seems like there is more that can be cleaned up here. I'll do that now. 

I think longer term receiving a logger from either `gin.Context` or `context.Context` may be desirable though. It makes it much easier to test that logging is being applied correctly. We've had a few regression with these middleware, and being able to test some of them (ex: logging) are wired up correctly would be nice.

mxyng (approved):

Issue #1609, opened 2022-04-13
proposal: unexport global logger and use a logger from context


dnephin:
Currently most log messages are emitted using `logging.L`, or `logging.S`.

This issue proposes that we move to passing a logger as part of `context.Context`. Any code that needs to log would receive a logger by calling `logging.FromContext(ctx)`. `gin.Context` includes `http.Request.Context()`, so anything that has access to the gin context should also have access to `context.Context`.

By passing a logger through context we can do a few things:
1. we can augment the context with details that will get logged as part of any subsequent log lines (ex: the ID of the athenticated user making the request)
2. it makes it much easier to test that important logs message are emitted correctly (error logs or access logs)
3. this pattern is exactly what we would need for distributed application tracing. In the future it will be much easier to add tracing if we establish this pattern now.

ssoroka:
Notes of things to add to the context-specific logs: orgID, userID, requestID

ssoroka:
related to https://github.com/infrahq/infra/issues/3031

PR #1611, opened 2022-04-13
api: more robust structure for building routes

dnephin:
(Best viewed by individual commit because it required moving a function to a different file).

## Summary

We've had a couple regressions with the API middleware either applying to the wrong routes (ui routes), or not being applied to routes (missing logging on API routes). Even when we notice the problem it has been a challenge to fix it properly (requiring lots of back-and-forth or pairing).

I think there are a few constraints that make this challenging:
1. Our UI routes are applied as catch-all middleware with no path prefix. This makes it really easily to accidentally add a middleware to those routes without noticing.
2. The interface gin providers for constructing these routes is overly flexible and not well documented (I had to read the source to understand how/when middleware would be applied to a route. See the godoc I added for details).
3. It's very difficult to test things are constructed properly in any meaningful way. We could potentially write a test that checks the number or list of middleware applied (although I'm not sure if that is possible with Gin), however that is just mirroring the implementation. Testing that a crash middleware or timeout middleware is applied to the routes we want is challenging at best.

The approach I've taken here is to:
1. consolidate - move all of the HTTP handler building to a single function. This makes it easier to reason about because we can see it all together (instead of it being split between 2+ functions across 2 files.
2. document - both the godoc on the function and inline comments should help clarify why things are this way, and what to look out for when it changes.

There is already some pretty reasonable test coverage of the "not found" route,  the UI routes, and the healthz endpoint. We should be adding tests for all the other endpoints which will further improve our confidence that at least all the routes exist (but we still don't have a good way of testing the middleware is applied correctly). Changes like #1609 will make it easier to test the logging middleware at least.

### DIFF
--- a/internal/server/routes.go
+++ b/internal/server/routes.go
@@ -16,114 +16,111 @@ import (
 	"github.com/infrahq/infra/metrics"
 )
 
+// GenerateRoutes constructs a http.Handler for the primary http and https servers.
+// The handler includes gin middleware, API routes, UI routes, and others.
+//
+// As a side effect of building the handler all API endpoints are registered in
+// the global openAPISchema.
+//
+// The order of routes in this function is important! Gin saves a route along
+// with all the middleware that will apply to the route when the
+// Router.{GET,POST,etc} method is called.
 func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) (*gin.Engine, error) {
+	a := &API{t: s.tel, server: s}
 	router := gin.New()
+	router.NoRoute(a.notFoundHandler)
 
 	router.Use(gin.Recovery())
-	a := &API{
-		t:      s.tel,
-		server: s,
-	}
+	router.GET("/healthz", healthHandler)
 
-	a.registerRoutes(router.Group("/"), promRegistry)
-
-	if err := s.registerUIRoutes(router); err != nil {
-		return nil, err
-	}
-
-	return router, nil
-}
-
-func (a *API) registerRoutes(router *gin.RouterGroup, promRegistry prometheus.Registerer) {
-	router.GET("/healthz", a.healthHandler)
-
+	// This group of middleware will apply to everything, including the UI
 	router.Use(
 		logging.Middleware(),
 		RequestTimeoutMiddleware(),
 	)
 
-	router.GET("/.well-known/jwks.json", DatabaseMiddleware(a.server.db), a.wellKnownJWKsHandler)
-
-	router.Use(
+	// This group of middleware only applies to non-ui routes
+	api := router.Group("/",
 		sentrygin.New(sentrygin.Options{}),
 		metrics.Middleware(promRegistry),
 		DatabaseMiddleware(a.server.db),
 	)
+	api.GET("/.well-known/jwks.json", a.wellKnownJWKsHandler)
 
-	v1 := router.Group("/v1")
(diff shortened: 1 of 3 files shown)


internal/server/routes.go
```diff
@@ -112,6 +126,10 @@ func (a *API) registerRoutes(router *gin.RouterGroup, promRegistry prometheus.Re
 	v1.GET("/machines/:id/grants", removed("v0.9.0"))
 }
 
+type ReqHandlerFunc[Req any] func(c *gin.Context, req *Req) error
+type ResHandlerFunc[Res any] func(c *gin.Context) (Res, error)
+type ReqResHandlerFunc[Req, Res any] func(c *gin.Context, req *Req) (Res, error)
```
dnephin, line 128: I moved these down to just above the first place they are referenced.

There are different approaches to ordering things in source files. I know some languages prefer to put all the types or constants at the top of a file, but personally I think that's a mistake.

Source code is for humans, and the less humans have to context switch, the better off we are.  

I think files should be ordered from the main entrypoint (in this case `GenerateRoutes`) followed by all the functions and types it uses. Sometimes if the main entrypoint accepts a large `Options` struct, it might make sense to start with that, but in this case these types are implementation details of building routes, so I moved them down.

This approach allows files to be read easily from top-to-bottom, and reduces the amount of jumping around (context switching) that is required to read the source.

internal/server/routes.go
```diff
@@ -16,114 +16,111 @@ import (
...
-	router.GET("/.well-known/jwks.json", DatabaseMiddleware(a.server.db), a.wellKnownJWKsHandler)
-
-	router.Use(
+	// This group of middleware only applies to non-ui routes
+	api := router.Group("/",
 		sentrygin.New(sentrygin.Options{}),
 		metrics.Middleware(promRegistry),
 		DatabaseMiddleware(a.server.db),
 	)
+	api.GET("/.well-known/jwks.json", a.wellKnownJWKsHandler)
```
dnephin, line 48: Small changes in functionality here. Previously we applied the database middleware but not metrics or sentry. With this change we apply those extra two middleware.

I think anything that uses the database should be part of our metrics. The sentry middleware is probably safe to apply to anything so I included it to reduce the number of middleware groups we have.

internal/server/routes.go
```diff
@@ -16,114 +16,111 @@ import (
...
+	get(a, authn, "/v1/destinations/:id", a.GetDestination)
+	post(a, authn, "/v1/destinations", a.CreateDestination)
+	put(a, authn, "/v1/destinations/:id", a.UpdateDestination)
+	delete(a, authn, "/v1/destinations/:id", a.DeleteDestination)
 
-	// these endpoints are left unauthenticated
-	unauthorized := v1.Group("/")
+	get(a, authn, "/v1/introspect", a.Introspect)
+	post(a, authn, "/v1/tokens", a.CreateToken)
+	post(a, authn, "/v1/logout", a.Logout)
```
dnephin, line 85: You may notice these 3 moved lines. I moved some of these miscellaneous handlers into a section together, instead of having 3 ungrouped.

mxyng (approved):

mxyng (commented):
I like this organization of router groups. Giving each group a name makes it easier to understand

internal/server/routes.go
jmorganca: Good catch!

jmorganca (approved):

PR #1613, opened 2022-04-13
fix: re-index identities after de-duplication

### LINKED ISSUE #1570: Infra users that exist across multiple providers need migration from 0.9.0 to current

BruceMacD:
## Summary
Move re-indexing identity names to after they are de-duplicated in migration.

## Checklist

## Related Issues

Resolves #1570

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -347,13 +347,6 @@ func migrate(db *gorm.DB) error {
 				}
 				identityTable := &Identity{}
 				logging.S.Debug("starting migration")
-				if tx.Migrator().HasIndex(identityTable, "idx_identities_name_provider_id") {
-					logging.S.Debug("has idx_identities_name_provider_id index")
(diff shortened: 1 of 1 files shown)


mxyng (approved):

dnephin:
Are there any changes we can make to the test for this migration that would have caught this problem? Can we add extra data or check the number of results?

PR #1615, opened 2022-04-13
improve: remove error return from registerUIRoutes

dnephin:
Branched from #1611, so that will need to merge first

This PR does a bit of cleanup from previous changes. I had to introduce an error return value to a few methods that shouldn't need error returns.

To remove the error return I introduced a type for `url.Url` that can be decoded from command line flag or config file. This allows us to handle the parse error earlier. Handling it earlier should also be better for users, as it keeps our errors consistent, and would allow us to implement a validate config command later.

### DIFF
--- a/internal/cmd/about.go
+++ b/internal/cmd/about.go
@@ -14,7 +14,6 @@ type triangle struct {
 	B     *point3D
 	C     *point3D
 	Color rune
-	surf  *sprite.Surface
 }
 
 func newAboutCmd() *cobra.Command {
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -45,7 +45,7 @@ func newServerCmd() *cobra.Command {
 
 			options.DBEncryptionKey = dbEncryptionKey
 
-			srv, err := server.New(options)
+			srv, err := newServer(options)
 			if err != nil {
 				return fmt.Errorf("creating server: %w", err)
 			}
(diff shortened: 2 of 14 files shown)


internal/cmd/about.go
```diff
@@ -14,7 +14,6 @@ type triangle struct {
 	B     *point3D
 	C     *point3D
 	Color rune
-	surf  *sprite.Surface
```
dnephin, line 17: This was failing linting for me locally. I'm not sure why it doesn't fail the linter in CI, but it appears to be unused.

pdevine: Weird. It didn't pick up w/ lint for me either. It's safe to remove.

internal/cmd/server.go
```diff
@@ -95,3 +90,6 @@ func defaultServerOptions() server.Options {
 var runServer = func(ctx context.Context, srv *server.Server) error {
 	return srv.Run(ctx)
 }
+
+// newServer is a shim for testing.
+var newServer = server.New
```
BruceMacD, line 95: What's the overhead of this? Does it mean instantiating 2 servers every time we want to run one?

internal/server/server.go
```diff
@@ -81,6 +80,11 @@ type ListenerOptions struct {
 	Metrics string
 }
 
+type UIOptions struct {
+	Enabled  bool
```
BruceMacD, line 84: These should be exposed in the mapstructure for the config file. 

I think changing the structure here also makes this a breaking change.

internal/cmd/server.go
dnephin: There shouldn't be any extra overhead (definitely not creating two servers).

This line is capturing a reference to the `server.New` function in the variable `newServer`. Functionally it would be similar to this:

```go
func newServer(opts server.Options) (*Server, error) {
    return server.New(opts)
}
```

The only difference is with `var newServer = server.New` we have one less function call. Calling `newServer(...)` is exactly the same as calling `server.New(...)`, it's the same function, just captured to a different variable.

Does that make sense? I might have misunderstood your question

internal/server/server.go
dnephin: Good call out! It is a breaking change, I will call it out in the commit message!

I have a test showing this works from the config file in this PR, the test case name is `parse ui-proxy-url from config file`. Although it looks like I'm only testing the proxy url, not enabled as well. I'll update the test to show that both are working.

We shouldn't need the `mapstructure` struct field tags on every config field. Mapstructure actually compares fields names case-insensitive by default, so most of those tags are not doing anything. I think maybe the reason we needed some of them is because of `viper`? I'm planning on removing viper so that we don't have to add all these struct tags and things will work as expected. We can always add `mapstructure` struct tags if we want to customize something.

dnephin:
Ok, I improved the test case to include the `enabled` field as well, and I added `BREAKING CHANGE` to the commit message since the config fields have changed.

internal/cmd/server.go
BruceMacD: That makes sense, thanks

BruceMacD (approved):

PR #1616, opened 2022-04-13
fix: infra providers add set client ID/secret

### LINKED ISSUE #1612: Cannot set providers using Infra CLI
## Describe the bug

`infra providers add` creates a provider resource with incomplete information making the provider unusable. Specifically, the `clientID` and `clientSecret` options are not being set.

## Expected behavior

Providers added through `infra provider add` can be used to login to Infra.

## Environment

```
$ infra version
Client: 0.10.2
```

mxyng:
## Summary

Field tags for providers options was incorrect so the values wasn't being set correctly. Add a test to check for this

## Checklist

## Related Issues

Resolves #1612

### DIFF
--- a/internal/cmd/providers_test.go
+++ b/internal/cmd/providers_test.go
@@ -1,0 +1,80 @@
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/uid"
+)
+
+func TestProviders(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	t.Setenv("USERPROFILE", home)
+
+	id := uid.New()
+
+	setup := func(t *testing.T, handler func(http.ResponseWriter, *http.Request)) {
+		svc := httptest.NewTLSServer(http.HandlerFunc(handler))
+		t.Cleanup(svc.Close)
+
+		cfg := ClientConfig{
+			Version: "0.3",
+			Hosts: []ClientHostConfig{
+				{
+					PolymorphicID: uid.NewIdentityPolymorphicID(id),
+					Name:          "test",
+					Host:          svc.Listener.Addr().String(),
+					SkipTLSVerify: true,
+					AccessKey:     "access-key",
+					Expires:       api.Time(time.Now().Add(time.Hour)),
+					Current:       true,
+				},
+			},
+		}
+
+		err := writeConfig(&cfg)
+		assert.NilError(t, err)
+	}
+
+	t.Run("AddOktaProvider", func(t *testing.T) {
+		setup(t, func(resp http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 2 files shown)


BruceMacD (approved):

internal/cmd/providers_test.go
```diff
@@ -0,0 +1,80 @@
...
+					Current:       true,
+				},
+			},
+		}
+
+		err := writeConfig(&cfg)
+		assert.NilError(t, err)
+	}
+
+	t.Run("ProvidersAdd", func(t *testing.T) {
```
dnephin, line 47: Maybe this case could be called "add okta provider" ?

internal/cmd/providers_test.go
```diff
@@ -0,0 +1,80 @@
...
+			if req.Method == http.MethodPost && req.URL.Path == "/v1/providers" {
+				var createProviderRequest api.CreateProviderRequest
+
+				err := json.NewDecoder(req.Body).Decode(&createProviderRequest)
+				assert.NilError(t, err)
+
+				assert.Equal(t, "okta", createProviderRequest.Name)
+				assert.Equal(t, "okta-url", createProviderRequest.URL)
+				assert.Equal(t, "okta-client-id", createProviderRequest.ClientID)
+				assert.Equal(t, "okta-client-secret", createProviderRequest.ClientSecret)
```
dnephin, line 58: This handler will run in a goroutine, and any calls to `t.Fatalf` (or `t.FailNow`) will cause a full panic, instead of a helpful error message. A panic in tests is not as bad as in production code, but it does stop the rest of the tests from running, and can make it difficult to see the failure (because the stack trace ends up being most of the output).

Almost all of the `assert` methods call `t.Fatalf`, the only expcetion is `assert.Check`, which calls `t.Errorf` (`t.Fail`). That will still fail the test, but won't stop the goroutine from running.

I think generally we probably want assertions to be in the main part of the test, and this handler (a test fixture) could simply return a 500 error code any time there is a problem. I guess in this case that's not easy to do because we want to test the server receives the values as expected.

I think this could be fixed either by changing the assertions to use `assert.Check`, or by returning an error from the handler.

dnephin (commented):
Thanks for fixing this! If I understand the problem, it has to do with the CLI not accepting command line flag arguments, and this test should show that it's working correctly. Just some small suggestions for the test.

internal/cmd/providers_test.go
mxyng: Noted. Ideally it checks the response but unfortunately the command discards the response from the server

mxyng:
> If I understand the problem, it has to do with the CLI not accepting command line flag arguments, and this test should show that it's working correctly. Just some small suggestions for the test.

That's partially the problem. The CLI was accepting the flags but it wasn't mapping the command line flags, e.g. `--client-id`, to the options field mapped by `mapstructure`. `parseOptions` converts the kebab case of command line parameters to the camel case of options. In this case options sets kebab case which means after conversion, the values are not mapped to the right field.

internal/cmd/providers_test.go
dnephin: I'll update the godoc for the assert functions as well to call out this restriction. I think it's not really clear right now from the godoc, and it should be.

PR #1624, opened 2022-04-14
fix: do not set a otp on machine identities

### LINKED ISSUE #1622: Do not create OTP for machine identity
## Describe the bug
When a machine identity is created in the CLI it should not get a one time password.

## Expected behavior
When a machine identity is created no authentication mechanism is set by default

## Environment

```
$ infra version
```

```
$ kubectl version
```

BruceMacD:
## Summary
There was a bug in identity creation through the CLI that would set a one time password on machine identities by default. Fixing this behavior to not set a one time password on machine identity creation.

## Checklist

## Related Issues

Resolves #1622

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -58,8 +58,10 @@ func (a *API) CreateIdentity(c *gin.Context, r *api.CreateIdentityRequest) (*api
 		Kind: kind,
 	}
 
+	setOTP := r.SetOneTimePassword && (identity.Kind == models.UserKind)
+
 	// infra identity creation should be attempted even if an identity is already known
-	if r.SetOneTimePassword {
+	if setOTP {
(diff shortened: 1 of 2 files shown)


internal/cmd/identities.go
```diff
@@ -247,6 +247,10 @@ func CreateIdentity(req *api.CreateIdentityRequest) (*api.CreateIdentityResponse
 
 	req.Kind = kind.String()
 
+	if kind == "machine" {
+		req.SetOneTimePassword = false
+	}
```
mxyng, line 252: This does not prevent the user from send a request for a machine user with `SetOneTimePassword=true`. Would it be better to have this check/mutation server-side?

BruceMacD: Made this change

dnephin (approved):
LGTM!

jmorganca (approved):

PR #1628, opened 2022-04-14
fix(cli): fix the extension-deadline flag

dnephin:
## Summary

I started by changing the fields from string to `time.Duration` to avoid the need to `ParseDuration`. Then I wrote a test for the command. The test was failing because `--extension-deadline` was not being passed to the API. I'm not sure why, the struct field tags look right. I don't think it was the change to the field types, because the `ttl` field was working and it had the same type.

I changed the command to use `DurationVar`, which got the test to pass.

This change removes support for setting these fields from environment variables, but that should be an easy thing to restore. I am working on a proposal for how we can do that with a single function, without the need for viper.

Since this is a CLI command that we expect humans to run, I guess no one is trying to use it with environment variables right now. The command line flags will be a lot more convenient for humans in most cases.

### DIFF
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -32,11 +32,13 @@ func newKeysCmd() *cobra.Command {
 }
 
 type keyCreateOptions struct {
-	TTL               string `mapstructure:"ttl"`
-	ExtensionDeadline string `mapstructure:"extension-deadline"`
+	TTL               time.Duration
+	ExtensionDeadline time.Duration
 }
 
 func newKeysAddCmd() *cobra.Command {
+	var options keyCreateOptions
+
 	cmd := &cobra.Command{
 		Use:   "add ACCESS_KEY_NAME MACHINE_NAME",
 		Short: "Create an access key for authentication",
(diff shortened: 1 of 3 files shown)


internal/cmd/keys.go
```diff
@@ -68,35 +65,23 @@ infra keys add first-key bot --ttl=12h --extension-deadline=1h
...
 			fmt.Printf("key: %s \n", resp.AccessKey)
-
 			return nil
 		},
 	}
 
-	cmd.Flags().String("ttl", "", "The total time that an access key will be valid for, defaults to 30 days")
-	cmd.Flags().String("extension-deadline", "", "A specified deadline that an access key must be used within to remain valid, defaults to 30 days")
+	cmd.Flags().DurationVar(&options.TTL, "ttl", ThirtyDays, "The total time that an access key will be valid for, defaults to 30 days")
+	cmd.Flags().DurationVar(&options.ExtensionDeadline, "extension-deadline", ThirtyDays, "A specified deadline that an access key must be used within to remain valid, defaults to 30 days")
```
dnephin, line 84: This change also improves the help text of the command, because now the default value (30 days) is visible from the help  text.

BruceMacD (approved):
Much better, thanks.

PR #1630, opened 2022-04-14
fix: help prevent bad provider config

### LINKED ISSUE #1623: Make a misconfigured identity provider more obvious
If an identity provider is mis-configured (for example, the groups included in the user info response are not correct) it is not obvious. There is a log in the server but it isn't bubbled up anywhere.

On provider import or creation it would be nice if we could test the info available to make sure that we have all the information we need.

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

BruceMacD:
- require client ID in API
- require client secret in API
- log empty groups response

## Summary
It should not be possible to create an IDP without a client ID and client secret. Validate these in the API. 

Also add a bit more logging to help with the case where groups are not set correctly. 

## Checklist

## Related Issues

Resolves #1623


internal/server/authn/oidc.go
```diff
@@ -192,7 +192,7 @@ func (o *oidcImplementation) GetUserInfo(providerUser *models.ProviderUser) (*Us
 		return nil, fmt.Errorf("decode user info response: %w", err)
 	}
 
-	if info.Groups == nil {
+	if info.Groups == nil || len(*info.Groups) == 0 {
```
dnephin, line 195: `*[]Groups` is an interesting type for this field. Usually pointer to slice isn't necessary because a slice itself can be nil. Do you know why we use this type of field?

If we used `[]Groups`, I think we'd only need to check `len(info.Groups) == 0`, since a nil slice also has zero length.

api/provider.go
```diff
@@ -15,17 +15,17 @@ type Provider struct {
 
 type CreateProviderRequest struct {
 	Name         string `json:"name" validate:"required" example:"okta"`
-	URL          string `json:"url" validate:"required" example:"infrahq.okta.com"`
-	ClientID     string `json:"clientID" example:"0oapn0qwiQPiMIyR35d6"`
-	ClientSecret string `json:"clientSecret" example:"jmda5eG93ax3jMDxTGrbHd_TBGT6kgNZtrCugLbU"`
+	URL          string `json:"url" validate:"fqdn,required" example:"infrahq.okta.com"`
```
dnephin, line 18: Can this field accept a port and/or path?

api/provider.go
```diff
@@ -15,17 +15,17 @@ type Provider struct {
...
+	ClientID     string `json:"clientID" validate:"required" example:"0oapn0qwiQPiMIyR35d6"`
+	ClientSecret string `json:"clientSecret" validate:"required" example:"jmda5eG93ax3jMDxTGrbHd_TBGT6kgNZtrCugLbU"`
 }
 
 type UpdateProviderRequest struct {
 	ID           uid.ID `uri:"id" json:"-" validate:"required"`
-	Name         string `json:"name" example:"okta"`
+	Name         string `json:"name" validate:"fqdn,required" example:"okta"`
```
dnephin, line 25: The example here doesn't appear to match the validation format (`fqdn`). Should the example be updated?

dnephin (commented):
Nice! Mostly just questions from me, to make sure I understand the change.

We could probably test this with an API test for the `CreateProvider` endpoint, to show that a missing ClientID or ClientSecret results in a 400 status code. Not the most interesting test, but it might help prevent an accidental regression.

internal/server/authn/oidc.go
BruceMacD: Good call out, originally I was using a pointer because I only wanted to log a warning if there was no groups field on a response at all (rather than an empty groups field, which means the application was configured correctly, but the user wasn't in any groups). This update means we will log both of those cases so I changed it to a slice rather than a pointer to a slice.

api/provider.go
BruceMacD: I tested it and it's a bit wonky, removed `fqdn`

api/provider.go
BruceMacD: fixed

dnephin (approved):
LGTM!

PR #1633, opened 2022-04-14
feat: login and logout of current or all servers, also affects clear

### LINKED ISSUE #1395: Review `help` (descriptions and examples) for all commands.
To do list: 

1. `infra list` help description is not very useful. This needs to be more descriptive since there are similar commands, e.g. `infra destinations list`, `infra grants list`, which may confuse the user.
  `infra list` should have a long description where it describes that it outputs the destinations/resources/roles/permissions the logged in user has access to.
2. Go through all descriptions
3. Go through all examples 
4. Go through helpful messages after running each command. EG after `dest add`, output suggestions (eg. `grants add`)
5. Headers should always be printed even with no results

### LINKED ISSUE #1437: No easy way to delete an old infra server from infra login
## Describe the bug

An infrastructure user installed another instance of Infra, and there was no easy way to delete my old infra instance from infra login 

## Expected behavior

There is an easy way to delete an infra instance that's no longer there from `infra login`

kimskimchi:
## Summary

1. Logout can be specific to current server, a specific server with `[URL]`, or all servers with `--all`. 

2. Clear is the exact same - just append `--clear` and will delete whatever server(s) was provided. 

3. Behaviors:
   - `infra logout` -> logout of current server
   - `infra logout URL` -> logout of server at URL
   - `infra logout --all` -> logout of all servers
   - `infra logout --clear` -> logout and clear current server
   - `infra logout URL --clear` -> logout and clear server at URL
   - `infra logout --all --clear` -> logout and clear all servers 
   - `infra logout URL --all` -> **error** out, cannot specify both
   - `infra logout` not logged into any server -> output success msg 
   - `infra logout URL` & not logged in -> no output

4. Changes/fixes/updates:5. 
   - docs
   - clearing session on client side now additionally removes user info (name & pid)
   - isLoggedIn() now additionally checks user info (name&pid)
   - logout errors out IF api call fails
   - logout errors out if both [SERVER] and [--all] are specified
   - adds a bunch of debug logs in logout

4. Doc update:5. 
   - update help in all places for consistency (eg. same example names, urls, punctuation, etc)

## Checklist

   - Logs server url

## Related Issues

Resolves #1437
Resolves #1395


internal/cmd/logout.go
```diff
@@ -2,31 +2,76 @@ package cmd
...
+func logoutOfServer(hostConfig *ClientHostConfig) {
+	client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
+	if err != nil {
+		logging.S.Warn(err.Error())
+	}
+
+	hostConfig.AccessKey = ""
+	hostConfig.PolymorphicID = ""
+	hostConfig.Name = ""
+
```
kimskimchi, line 69: logout now clears user information as well.

internal/cmd/logout.go
```diff
@@ -36,20 +81,91 @@ func logout(purge bool) error {
...
 	}
 
-	for i, hostConfig := range config.Hosts {
-		config.Hosts[i].AccessKey = ""
-
-		client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
-		if err != nil {
-			logging.S.Warn(err.Error())
-			continue
+	switch {
```
kimskimchi, line 84: switch: all, current, or specific server

Prints relevant output messages so user can observe any state changes.

docs/reference/cli-reference.md
```diff
@@ -29,37 +29,33 @@
...
 ```
 
 ### Options
 
 ```
       --key string        Login with an access key
       --provider string   Login with an identity provider
-      --server string     Infra server to login to
       --skip-tls-verify   Skip verifying server TLS certificates
+      --url string        Infra server URL
```
jmorganca, line 58: No need for both a flag and an argument for this – would remove the flag

docs/reference/cli-reference.md
```diff
@@ -29,37 +29,33 @@
...
 
-# Login to a specified server
-$ infra login SERVER
-$ infra login --server SERVER
-
-# Login with an access key
-$ infra login --key KEY
+# Login to a specific server
+$ infra login infraexampleserver.com
+$ infra login --url infraexampleserver.com
```
jmorganca, line 43: See below – would remove the `--url` flag

docs/reference/cli-reference.md
```diff
@@ -75,19 +71,36 @@ $ infra login --provider NAME
...
 ```
 
 ### Examples
 
 ```
+# Log out of current server
 $ infra logout
+		
+# Log out of a specific server
+$ infra logout INFRA_URL
```
jmorganca, line 84: would put a real URL here instead of `INFRA_URL`

internal/cmd/login.go
```diff
@@ -45,24 +45,20 @@ const cliLoginRedirectURL = "http://localhost:8301"
...
-# By default, login will prompt for all required information.
+		Example: `# By default, login will prompt for all required information.
 $ infra login
 
-# Login to a specified server
-$ infra login SERVER
-$ infra login --server SERVER
+# Login to a specific server
+$ infra login infraexampleserver.com
+$ infra login --url infraexampleserver.com
```
jmorganca, line 55: Remove me RE comment above

internal/cmd/login.go
```diff
@@ -74,18 +70,18 @@ $ infra login --provider NAME
 			}
 
 			if len(args) == 1 {
-				if options.Server != "" {
-					fmt.Fprintf(os.Stderr, "Server is specified twice. Ignoring flag [--server] and proceeding with %s", options.Server)
+				if options.URL != "" {
+					fmt.Fprintf(os.Stderr, "Server is specified twice. Ignoring flag [--url] and logging in to server %s", options.URL)
```
jmorganca, line 74: No need for this is we remove `--url` as per above :-)

internal/cmd/logout.go
```diff
@@ -2,31 +2,76 @@ package cmd
...
 }
 
-func logout(purge bool) error {
+func logoutOfServer(hostConfig *ClientHostConfig) {
+	client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
+	if err != nil {
+		logging.S.Warn(err.Error())
+	}
+
+	hostConfig.AccessKey = ""
```
jmorganca, line 66: This should delete the entry entirely from the hostconfig vs clearing out the fields

jmorganca: Oops, I misread this!

internal/cmd/logout.go
```diff
@@ -36,20 +81,94 @@ func logout(purge bool) error {
...
+	case all:
+		stateChanged := false
+		for i := range config.Hosts {
+			if config.Hosts[i].isLoggedIn() {
+				logoutOfServer(&config.Hosts[i])
+				stateChanged = true
+			}
+		}
+		if stateChanged {
+			fmt.Fprintf(os.Stderr, "  Logged out of all servers.\n")
```
jmorganca, line 94: Let's just always print this (vs the other message which implies an error)

internal/cmd/logout.go
```diff
@@ -36,20 +81,94 @@ func logout(purge bool) error {
...
+		} else {
+			fmt.Fprintf(os.Stderr, "  Not logged in to any servers.\n")
 		}
 
-		_ = client.Logout()
-	}
+		if clear {
+			if config.Hosts != nil {
+				config.Hosts = nil
+				fmt.Fprintf(os.Stderr, "  Cleared list of all servers\n")
```
jmorganca, line 102: More of a suggestion, but we don't need to let the user know this as they specified it with `--clear` and it may be too much information

internal/cmd/logout.go
```diff
@@ -36,20 +81,94 @@ func logout(purge bool) error {
...
+			} else {
+				fmt.Fprintf(os.Stderr, "  No servers to clear.\n")
+			}
+		}
+	case url == "":
```
jmorganca, line 107: This case and the next one should be merged into the same code path – it's fundamentally doing the same thing.

internal/cmd/logout.go
```diff
@@ -36,20 +81,94 @@ func logout(purge bool) error {
...
+		serverFound := false
+		for i := range config.Hosts {
+			if config.Hosts[i].Current {
+				serverFound = true
+				if config.Hosts[i].isLoggedIn() {
+					logoutOfServer(&config.Hosts[i])
+					fmt.Fprintf(os.Stderr, "  Logged out of server %s.\n", config.Hosts[i].Host)
+				} else {
+					fmt.Fprintf(os.Stderr, "  Not logged in to server %s.\n", config.Hosts[i].Host)
```
jmorganca, line 116: Similar to above - no need to print this

internal/cmd/logout.go
```diff
@@ -36,20 +81,94 @@ func logout(purge bool) error {
...
 	}
 
-	for i, hostConfig := range config.Hosts {
-		config.Hosts[i].AccessKey = ""
-
-		client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
-		if err != nil {
-			logging.S.Warn(err.Error())
-			continue
+	switch {
```
mxyng, line 84: This is hard to understand. It would be easier to read and understand if some of the shared functionality is split out:

```go
func logout(server, all, clear) error {
  config, err := readConfig()
  if err != nil {
    return err
  }

  for i := range config.Servers {
    if all || config.Servers[i].Server == server || config.Servers.[i].Current {
      if err := logoutServer(&config.Servers[i], clear); err != nil {
        return err
      }
    }

    if config.Servers[i] == nil {
      config.Servers = append(config.Servers[:i], config.Servers[i+1:]...)
    }
  }

  if err := writeConfig(config); err != nil {
    return err
  }

  return nil
}

func logoutServer(config *ClientHostConfig, clear bool) error {
  client, err := apiClient(config.Server, config.AccessKey, config.SkipTLSVerify)
  if err != nil {
    return err
  }

  _ = client.Logout()

  if clear {
    client = nil
    return nil
  }

  client.AccessKey = ""
  return nil
}
```

internal/cmd/logout_test.go
```diff
@@ -90,9 +92,9 @@ func TestLogout(t *testing.T) {
 		},
 	}
 
-	t.Run("default", func(t *testing.T) {
+	t.Run("with all", func(t *testing.T) {
```
mxyng, line 95: should add test cases for current, server, and clear variants

internal/cmd/login.go
```diff
@@ -507,32 +498,37 @@ func promptServer() (string, error) {
...
 	}
 
-	return host, nil
+	return server, nil
 }
 
-func promptExistingHosts(hosts []string) (string, error) {
-	const defaultOption string = "Connect to a different host"
-	hosts = append(hosts, defaultOption)
+func promptExistingServers(servers []ClientHostConfig) (string, error) {
```
mxyng, line 520: function name isn't reflective of the options since adding a new server is an option so "Existing" is misleading

internal/cmd/logout.go
```diff
@@ -2,31 +2,76 @@ package cmd
...
+	"os"
 
 	"github.com/spf13/cobra"
 
 	"github.com/infrahq/infra/internal/logging"
 )
 
 func newLogoutCmd() *cobra.Command {
-	var purge bool
+	var (
```
mxyng, line 14: Might want to consider using a struct to hold the options since you're passing them all to `logout`

internal/cmd/logout.go
```diff
@@ -2,31 +2,76 @@ package cmd
...
+	cmd.Flags().BoolVar(&all, "all", false, "logout of all servers")
 
 	return cmd
 }
 
-func logout(purge bool) error {
+func logoutOfServer(hostConfig *ClientHostConfig) {
+	client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
+	if err != nil {
+		logging.S.Warn(err.Error())
```
mxyng, line 63: This error should probably be debug.

```suggestion
		logging.S.Debugf("%s", err)
```

internal/cmd/logout.go
```diff
@@ -2,31 +2,76 @@ package cmd
...
+		
+# Logout and clear list of all servers 
+$ infra logout --all --clear`,
+		Args:  cobra.MaximumNArgs(1),
+		Group: "Core commands:",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return logout(purge)
+			if len(args) == 1 {
+				if all {
+					fmt.Fprintf(os.Stderr, "  Server is already specified. Ignoring flag [--all] and logging out of server %s.\n", args[0])
```
mxyng, line 45: ignoring user input is usually not a good thing. it's possible they're adding `--all` to `logout SERVER` expecting it to logout of all servers

internal/cmd/logout_test.go
kimskimchi: done

internal/cmd/logout.go
kimskimchi: done

internal/cmd/login.go
```diff
@@ -507,32 +498,37 @@ func promptServer() (string, error) {
...
+	var promptOptions []string
+	for _, server := range servers {
+		promptOptions = append(promptOptions, server.Host)
+	}
+
+	const defaultOption string = "Connect to a new server"
+	promptOptions = append(promptOptions, "Connect to a new server")
```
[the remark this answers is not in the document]
kimskimchi: done

internal/cmd/logout.go
kimskimchi: done

internal/cmd/logout.go
kimskimchi: done

internal/cmd/logout.go
kimskimchi: done

internal/cmd/logout.go
```diff
@@ -83,92 +98,43 @@ func logout(clear bool, server string, all bool) error {
...
-				var newHostConfigs []ClientHostConfig
 
-				// New slice of hosts except the one to clear
-				for i := range config.Hosts {
-					if config.Hosts[i].Current {
-						fmt.Fprintf(os.Stderr, "  Cleared [%s] from list of servers\n", config.Hosts[i].Host)
-						continue
+				if clear {
+					serverUrl := config.Hosts[i].Host
+					if len(config.Hosts) < 2 {
```
jmorganca, line 142: Is there a way to just make it an empty slice and avoid this `if` branch?

internal/cmd/logout.go
```diff
@@ -42,11 +44,17 @@ $ infra logout --all --clear`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 1 {
 				if all {
-					fmt.Fprintf(os.Stderr, "  Server is already specified. Ignoring flag [--all] and logging out of server %s.\n", args[0])
-					all = false
+					return fmt.Errorf("Argument [SERVER] and flag [--all] cannot be both specified.")
 				}
 				server = args[0]
 			}
+			logging.S.Debugf("flags set:")
```
jmorganca, line 52: We can remove these debug statements, since the flags being set is part of the input and printing this may not help the user much

internal/cmd/logout.go
```diff
@@ -42,11 +44,17 @@ $ infra logout --all --clear`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 1 {
 				if all {
-					fmt.Fprintf(os.Stderr, "  Server is already specified. Ignoring flag [--all] and logging out of server %s.\n", args[0])
-					all = false
+					return fmt.Errorf("Argument [SERVER] and flag [--all] cannot be both specified.")
```
jmorganca, line 48: Since `logout` is generally a safe thing to do – we can just default to logging out of all, even if the user specified a `SERVER` argument

internal/cmd/logout.go
kimskimchi: we could, but that requires more refactoring because it triggers a different error flow in a bunch of places (ie. we just check if its nil rather than len=0 and error out if so)

internal/cmd/logout.go
kimskimchi: done

internal/cmd/grants.go
```diff
@@ -169,21 +171,19 @@ func removeGrant(cmdOptions grantsCmdOptions) error {
...
-		Long: `Grant one or more identities access to a destination. 
+		Short: "Grant an identity access to a destination",
+		Example: `# Grant an identity access to a destination
+$ infra grants add johndoe@example.com kubernetes.docker-desktop 
+$ infra grants add machineA kubernetes.docker-desktop
 
-IDENTITY is the subject that is being given access.
-DESTINATION is what the identity will gain access to. 
+# Grant a group access to a destination 
+$ infra grants add groupA kubernetes.staging --group
```
ssoroka, line 180: I really dislike the --group boolean flag.

internal/cmd/logout.go
ssoroka: seems like that would complicate things again.

internal/cmd/logout_test.go
```diff
@@ -43,16 +49,70 @@ func TestLogout(t *testing.T) {
...
+				resp.WriteHeader(http.StatusBadRequest)
+				return
+			}
+			atomic.AddInt32(&count, 1)
+			resp.WriteHeader(http.StatusUnauthorized)
+			_, _ = resp.Write([]byte(`{}`)) // API client requires a JSON response
+		}
+
+		srv := httptest.NewTLSServer(http.HandlerFunc(handler))
+		unauthorizedSrvURL = srv.Listener.Addr().String()
```
ssoroka, line 103: return this from the function instead of using shared variables

internal/cmd/logout.go
kimskimchi: discussed/resolved, keeping as is

internal/cmd/logout_test.go
kimskimchi: done

internal/cmd/logout.go
kimskimchi: agreed, although this should be standardized for all cmds in a single pr (some uses parseOptions, some don't, some use vars). I believe Daniel was discussing this somewhere too

internal/cmd/login.go
ssoroka: perhaps something like `promptForServerFromList`?

internal/cmd/login.go
```diff
@@ -537,7 +537,7 @@ func promptExistingServers(servers []ClientHostConfig) (string, error) {
 	}
 
 	const defaultOption string = "Connect to a new server"
```
ssoroka, line 539: don't use const strings, that's not very go-like.
```suggestion
	defaultOption := "Connect to a new server"
```

ssoroka (approved):
I don't really like the pattern of returning errors as user-facing text. This makes tests brittle as you can't change the text without breaking a test. Instead we should be defining package-level error variables, and checking for specific variables to display specific text to the user. Functions need to decide if they're returning errors or printing text, and try not to do both.

internal/cmd/logout.go
dnephin: Ya, I think these can be an options struct, `logoutOptions`

internal/cmd/logout_test.go
```diff
@@ -127,4 +239,30 @@ func TestLogout(t *testing.T) {
...
+
+		updatedCfg, err := readConfig()
+		assert.NilError(t, err)
+		assert.DeepEqual(t, &testFields.config, updatedCfg)
+	})
+
+	t.Run("error", func(t *testing.T) {
+		testFields := setupError(t)
+		err := Run(context.Background(), "logout", testFields.serverURLs[0])
+		assert.Error(t, err, fmt.Sprintf(`Failed to logout of server %s due to an internal error: POST "/v1/logout" responded 500: Internal Server Error.`, testFields.serverURLs[0]))
```
dnephin, line 258: It would probably be ok to be less strict about the error here, and use something like this:

```suggestion
		assert.ErrorContains(t, err, "Failed to logout of server "+testFields.serverURLs[0])
```

internal/cmd/logout_test.go
```diff
@@ -91,31 +159,75 @@ func TestLogout(t *testing.T) {
...
+		assert.NilError(t, err)
+
+		assert.Equal(t, int32(1), atomic.LoadInt32(testFields.count), "calls to API")
+
+		updatedCfg, err := readConfig()
+		assert.NilError(t, err)
+
+		assert.Equal(t, int32(1), int32(len(updatedCfg.Hosts)))
+		assert.DeepEqual(t, testFields.config.Hosts[1], updatedCfg.Hosts[0])
+		// assert.DeepEqual(t, &expected, updatedCfg)
```
dnephin, line 194: Should this commented out line be deleted?

dnephin (approved):
LGTM! Some small suggestions, but could easily be a follow up PR

PR #1634, opened 2022-04-14
fix(ci): actions not running correctly

mxyng:
## Summary

`on.pull_request.branches` refers to the base ref of the pull request, e.g. `main`, so the filter cannot be used to filter jobs based on the head ref, e.g. `release-please--branches--main`.

### DIFF
--- a/.github/workflows/human_pull_request.yml
+++ b/.github/workflows/human_pull_request.yml
@@ -2,11 +2,10 @@ name: pull request
 
 on:
   pull_request:
-    branches-ignore:
-      - 'dependabot/*'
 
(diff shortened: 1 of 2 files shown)


dnephin (approved):
Looks right to me!

Reading the [docs here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-based-on-the-head-or-base-branch-of-a-pull-request) it seems like we might be able to remove the `== true`, but not necessary

mxyng:
iirc the inverse doesn't work so I wanted to be explicit

PR #1644, opened 2022-04-18
improve: updates help for all commands

kimskimchi:
## Summary

Updates help messages 

## Checklist

None, only cmd help messages were modified

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -97,17 +97,19 @@ func newGrantRemoveCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "remove IDENTITY DESTINATION",
 		Aliases: []string{"rm"},
-		Short:   "Revoke access to a destination",
-		Long: `Revokes access that user has to the destination.
+		Short:   "Revoke access from a destination",
+		Example: `# Remove all grants of an identity in a destination
+$ infra grants remove janedoe@example.com kubernetes.docker-desktop 
+$ infra grants remove machineA kubernetes.docker-desktop
 
-IDENTITY is one that was being given access.
-DESTINATION is what the identity will lose access to. 
+# Remove all grants of a group in a destination
+$ infra grants remove groupA kubernetes.staging --group
 
-Use [--role] to specify the exact grant being deleted. 
-If not specified, it will revoke all roles for that user within the destination. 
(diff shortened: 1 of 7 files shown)


internal/cmd/grants.go
```diff
@@ -97,17 +97,19 @@ func newGrantRemoveCmd() *cobra.Command {
...
 		Aliases: []string{"rm"},
-		Short:   "Revoke access to a destination",
-		Long: `Revokes access that user has to the destination.
+		Short:   "Revoke access from a destination",
+		Example: `# Remove all grants of an identity in a destination
+$ infra grants remove janedoe@example.com kubernetes.docker-desktop 
+$ infra grants remove machineA kubernetes.docker-desktop
 
-IDENTITY is one that was being given access.
-DESTINATION is what the identity will lose access to.
```
mxyng, line 104: This is still important information the user should have. It describes what the command does rather than how to use it. There's no reason not that have both a long description and examples

internal/cmd/identities.go
```diff
@@ -45,14 +45,16 @@ func newIdentitiesAddCmd() *cobra.Command {
...
-		Long: `Create a machine identity with NAME or a user identity with EMAIL.
+		Long: `Create an identity.
 
-NAME must only contain alphanumeric characters ('a-z', 'A-Z', '0-9') or the
-special characters '-', '_', or '/' and has a maximum length of 256 characters.
-
-EMAIL must contain a valid email address in the form of "local@domain".
-		`,
+If an email is provided, a user is created; otherwise, a machine is created.
+A new user must change their one time password before further usage.`,
```
mxyng, line 51: The new description isn't entirely accurate since there are non-email that cannot be used as names, e.g. `a@`. I would suggest keeping the previous description

internal/cmd/grants.go
jmorganca: Would add a mention of identity to the description vs outlining the arguments (examples do this well)

i.e. `Revoke an identity's access to a destination`

internal/cmd/identities.go
jmorganca: For now we can error if there's a strange case (e.g. trying to name a user `a@`)

kimskimchi: reworded it as 'detects a valid email / username'

Issue #1632, opened 2022-04-14
Sporadic test failure loading database key


BruceMacD:
## Describe the bug
```
=== Failed
=== FAIL: internal/cmd TestServerCmd_WithSecretsConfig (0.25s)
    server_test.go:191: assertion failed: error is not nil: creating server: loading database key: saving root key: setenv: setenv: invalid argument
```

## Expected behavior
Test does not passes consistently.

## Environment

```
$ infra version
```

```
$ kubectl version
```

dnephin:
Thanks for creating this issue! I've seen this occasionally as well. I'll try to reproduce to find the cause.

dnephin:
I think the error is coming from here: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/syscall/env_unix.go;l=109-110

Apparently a 0 byte (`\0`) is not allowed in the value for an environment variable, and when we generate a random key that includes the 0 byte it will fail.

dnephin:
We could force the test to use a different secret provider with `base64: true`, but I think there is a real underlying bug here.

I guess the underlying bug is that the default `env` secret provider (not base64-encoded) is subtly incompatible with the `native` key provider. 

I'm not entirely sure of the right way to fix this. Maybe the `NativeKeyProvider` could always base64 encode the root key. We would need to document that the root key must always be base64 encoded.

PR #1650, opened 2022-04-18
fix: flakey test fix

ssoroka:
## Summary

fixes test that fails periodically. Test was misconfigured.

## Checklist

## Related Issues

Resolves https://github.com/infrahq/infra/issues/1632


dnephin (approved):
Thanks for fixing the flake!  I created #1651 to track the underlying issue, since that may impact users as well.

PR #1652, opened 2022-04-18
Add a test for CreateGrant API endpoint

dnephin:
## Summary

This PR adds a test for `CreateGrant` (`POST /v1/grants`). I believe this is the first API test that is strict about the API response body, so it includes some new helper functions. We don't necessarily need to write every API test in this way, but I think we should have at least one test for each API endpoint that makes assertions about the body of the response. 

There are other ways of testing APIs. We could decode the response into the `api` type. The problem with that approach is that we will mask regressions in the API. If we change the `api` type, the test doesn't fail! It is easy to unintentionally change API responses, and tests like this help prevent API regressions.  It does mean that when we add fields we will have to update these tests. That's probably a good reason to have a small number of tests that are strict about the API response, so we only have to update one or two tests when we add fields to the response.

As we write more of these we may want to extract them into an `apitest` package.

There are a few fields here that are not predictable (uid and times), which makes writing a strict test more difficult. The approach taken here is to do an approximate comparison. An alternative would be to patch the uid generator and the database `time.Now` to set specific values. I think the approximate comparison might be easier to maintain than a lot of patching.

### DIFF
--- a/internal/server/handlers_test.go
+++ b/internal/server/handlers_test.go
@@ -6,17 +6,19 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/gin-gonic/gin"
-	"github.com/google/go-cmp/cmp"
+	gocmp "github.com/google/go-cmp/cmp"
 	"github.com/prometheus/client_golang/prometheus"
 	"gotest.tools/v3/assert"
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
 )
 
 func TestAPI_ListIdentities(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/server/handlers_test.go
```diff
@@ -249,3 +252,113 @@ func TestDeleteIdentity_NoDeleteSelf(t *testing.T) {
...
+
+	expected := jsonUnmarshal(t, fmt.Sprintf(`
+		{
+		  "id": "<any-valid-uid>",
+		  "created_by": "%[1]v",
+		  "privilege": "admin-role",
+		  "resource": "kubernetes.some-cluster",
+		  "subject": "i:12345",
+		  "created": "%[2]v",
+		  "updated": "%[2]v"
```
dnephin, line 286: We could take a different approach here, something more like `"<any-valid-uid>"`, and instead make this

```suggestion
		  "created": "<time-now>",
		  "updated": "<time-now>"
```

Which would remove the need to pass `time.Now().UTC().Format(time.RFC3339)` to `fmt.Sprintf`. 

Thoughts?

dnephin: Also, we could use a different placeholder syntax, these angle brackets are just the first thing that came to mind. We could also do:
* `"[any-valid-uid]"`,  `"[time-now]"`
* `"uid.ID(any)"`, `"time.Now()"`

These are exact string matches, so it can be anything really.

dnephin:
Ya, I don't think we should have more than 1 of these kinds of tests for each API endpoint. Maybe a couple if there are different possible responses, but likely only one per endpoint.

PR #1653, opened 2022-04-19
maintain: move secrets out to its own repo

ssoroka:
## Summary

Moves secrets out to its own repo. This sets the groundwork for focusing Infra's use of secrets if we want.

## Checklist


pki/certificates_test.go
```diff
@@ -41,8 +39,6 @@ func TestMain(m *testing.M) {
 	os.Exit(result)
 }
 
-var containerIDs []string
```
dnephin, line 44: Do we still need the `TestMain` above this, and the `setup` and `teardown` below this? Seems like they don't do anything now.

internal/server/config.go
```diff
@@ -46,10 +46,58 @@ type KeyProvider struct {
...
+	AWSConfig `mapstructure:",squash"`
+
+	EncryptionAlgorithm string `mapstructure:"encryptionAlgorithm"`
+	// aws tags?
+}
+
+type AWSSecretsManagerConfig struct {
+	AWSConfig `mapstructure:",squash"`
+
+	UseSecretMaps bool `mapstructure:"useSecretMaps"`
```
dnephin, line 70: I believe this `UseSecretMaps` field is never used. We now pass it to the secrets package, but I don't think the secrets package uses it either.  Might be a good opportunity to remove it until we need it.

internal/server/config.go
```diff
@@ -46,10 +46,58 @@ type KeyProvider struct {
...
+	GenericConfig `mapstructure:",squash"`
+	Path          string `mapstructure:"path" validate:"required"`
+}
+
+type KubernetesConfig struct {
+	Namespace string `mapstructure:"namespace"`
+}
+
+type VaultConfig struct {
+	TransitMount string `mapstructure:"transitMount"`              // mounting point. defaults to /transit
```
dnephin, line 94: Does it still default to `/transit` ? The expected value in the tests changed to empty string.

dnephin (approved):
LGTM! A few questions, mostly about things we might be able to remove at the same time.

internal/server/config.go
ssoroka: yes but the default comes from the secrets package, not this struct. So it just needs to set the transit only when it's not "" to use the default /transit value. No functionality should have changed.

dnephin (approved):

Issue #1606, opened 2022-04-13
CLI list ordering 


mchiang0610:
Current: 
<img width="454" alt="image" src="https://user-images.githubusercontent.com/3325447/163076675-126dec32-bb1c-4897-a1f7-4588b89b4aef.png">

Expected: 
- Alphabetical ordering

technovangelist:
To be clear, the current ordering is consistent but not based on order of creation, order of last login, or any other order I can think of.

jmorganca:
closing for #1341

PR #1654, opened 2022-04-19
api: sort list identities by name

dnephin:
## Summary

While working on something else I thought I had found a bug in list identities. I wrote a test for the API endpoint, and then realized the problem was I had typed the identity name wrong.

This commit:
* orders identities by name ascending
* adds a test for `ListIdentities` API endpoint
* improves the tests for `data.ListIdentities` by asserting an order

## Checklist

## Related Issues

Related to #1606, but only for list identities.


ssoroka (approved):
seems fine.

PR #1655, opened 2022-04-19
cli: use DefaultsFromEnv - part 1

dnephin:
(best viewed by individual commit, more details in the commit messages)

## Summary

This PR adds a new `cliopts.DefaultsFromEnv` for loading default values for flags from environment variables. The new function is used in two commands for now: `infra login`, and `infra identities edit`. This should be enough to demonstrate the pattern. I'll add it to other places once everyone is happy with this approach.

Also fixes a bug related to `cobra.Command.PersistentPreRunE` in a parent command clobbering the one from the root command. Fixing that bug was necessary to write tests for this, because `DefaultsFromEnv` is called from `PersistentPreRunE`.

Adds another test case for `infra id edit`, but nothing for `infra login` yet.

### DIFF
--- a/internal/cmd/cliopts/flags.go
+++ b/internal/cmd/cliopts/flags.go
@@ -1,0 +1,68 @@
+package cliopts
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+type FlagSet interface {
+	VisitAll(fn func(*pflag.Flag))
+}
+
+// DefaultsFromEnv looks for an environment variable for any unset flags. When
+// an environment variable is found, the flag value is set from the environment
+// variable.
+//
+// The environment variable for a flag has the prefix prepended, dashes replaced
+// with underscores, and lowercase converted to uppercase (ex:
+// --my-flag would be set from PREFIX_MY_FLAG).
+//
+// DefaultsFromEnv should be called after FlagSet.Parse, but before any flags
+// are used.
+func DefaultsFromEnv(prefix string, flags FlagSet) error {
+	replacer := strings.NewReplacer("-", "_")
+	prefix = prefix + "_"
+
+	var errs []error
+	flags.VisitAll(func(flag *pflag.Flag) {
+		if flag.Changed {
+			return
+		}
+
+		key := strings.ToUpper(prefix + replacer.Replace(flag.Name))
+		v, exists := os.LookupEnv(key)
+		if !exists {
+			return
+		}
+		if err := flag.Value.Set(v); err != nil {
+			err = fmt.Errorf("failed to set %v from environment variable: %w", flag.Name, err)
+			errs = append(errs, err)
+		}
+	})
+
+	if len(errs) > 0 {
+		return MultiError(errs)
+	}
+	return nil
+}
+
+type MultiError []error
+
+func (e MultiError) Error() string {
+	errs := ([]error)(e)
+	switch len(errs) {
+	case 1:
+		return errs[0].Error()
+	default:
+		var sb strings.Builder
+		sb.WriteString("multiple errors:")
+		for _, err := range errs {
+			sb.WriteString("\n    " + err.Error())
+		}
+		sb.WriteString("\n")
+		return sb.String()
+	}
+}
--- a/internal/cmd/cliopts/flags.go
+++ b/internal/cmd/cliopts/flags.go
@@ -24,7 +24,7 @@ type FlagSet interface {
 // are used.
 func DefaultsFromEnv(prefix string, flags FlagSet) error {
 	replacer := strings.NewReplacer("-", "_")
-	prefix = prefix + "_"
+	prefix += "_"
 
 	var errs []error
 	flags.VisitAll(func(flag *pflag.Flag) {
--- a/internal/cmd/cliopts/flags_test.go
+++ b/internal/cmd/cliopts/flags_test.go
@@ -1,0 +1,208 @@
+package cliopts
+
+import (
+	"fmt"
+	"testing"
+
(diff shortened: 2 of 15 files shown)


internal/cmd/cmd.go
```diff
@@ -370,6 +373,10 @@ func NewRootCmd() *cobra.Command {
 	return rootCmd
 }
 
+func addNonInteractiveFlag(flags *pflag.FlagSet, bind *bool) {
+	flags.BoolVar(bind, "non-interactive", false, "Disable all prompts for input")
+}
```
dnephin, line 387: I moved this out of `rootOptions` to be an option that we can add to any command that has interactive prompts. That way we don't have to pass around options from the root command.

internal/cmd/login.go
```diff
@@ -437,10 +441,6 @@ func listProviders(client *api.Client) ([]api.Provider, error) {
 }
 
 func promptLoginOptions(client *api.Client) (loginMethod loginMethod, provider *api.Provider, err error) {
-	if isNonInteractiveMode() {
-		return 0, nil, fmt.Errorf("Non-interactive login requires key, instead run: 'infra login SERVER --non-interactive --key KEY")
-	}
```
dnephin, line 442: I moved a couple of these to the caller so that we don't have to pass in `options`. 

This was part of making the options local references instead of a static global.

internal/cmd/login.go
```diff
@@ -231,8 +230,8 @@ func updateInfraConfig(client *api.Client, loginReq *api.LoginRequest, loginRes
 	return nil
 }
 
-func isNonInteractiveMode() bool {
-	return rootOptions.NonInteractive || os.Stdin == nil || !term.IsTerminal(int(os.Stdin.Fd()))
+func (o loginCmdOptions) isNonInteractiveMode() bool {
+	return o.NonInteractive || os.Stdin == nil || !term.IsTerminal(int(os.Stdin.Fd()))
```
ssoroka, line 234: unfortunately this will end up getting replicated for each options type

internal/cmd/cliopts/flags.go
```diff
@@ -0,0 +1,68 @@
...
+// DefaultsFromEnv looks for an environment variable for any unset flags. When
+// an environment variable is found, the flag value is set from the environment
+// variable.
+//
+// The environment variable for a flag has the prefix prepended, dashes replaced
+// with underscores, and lowercase converted to uppercase (ex:
+// --my-flag would be set from PREFIX_MY_FLAG).
+//
+// DefaultsFromEnv should be called after FlagSet.Parse, but before any flags
+// are used.
```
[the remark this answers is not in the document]
dnephin: Check out the tests! The second case shows what happens with default values.  The order of precedence is:

```
1. value specified by user from command line flag (highest)
2. defaults set in command line flag definitions
3. values from env variables (lowest)
```

internal/cmd/login.go
dnephin: True. Right now there are only two commands that use this, and the second was not using this logic.

I think we will want to share this differently.

In the future I expect us to have `commonOptions` (or something like that) with flags for overriding what is in the client config. The interactive flag might be good on `commonOptions`.  I didn't want to make that change now so that the PR was more focused on just the changes to loading values from flags.

I think for now we could accept the boolean flag value to `isNonInteractiveMode`, so that this can be called from other places.

internal/cmd/cliopts/flags.go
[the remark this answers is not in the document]
ssoroka: shouldn't 2 and 3 be switched? I would think a command line default is the lowest.

internal/cmd/login.go
ssoroka: yeah, or move o.NonInteractive out of this check, and require the caller to check it themselves

internal/cmd/cliopts/flags_test.go
```diff
@@ -0,0 +1,208 @@
...
+		Many    []string
+	}
+
+	setup := func(tg *target) *pflag.FlagSet {
+		flags := pflag.NewFlagSet("testing", pflag.ContinueOnError)
+		flags.String("word", "default-value", "")
+		flags.Int("count", 12, "")
+		flags.String("two-parts", "", "")
+		flags.StringVar(&tg.OneMore, "one-more-extra-long", "", "")
+		flags.Int32Var(&tg.Next, "next", 0, "")
```
dnephin, line 24: I'll update this to use a non-zero default, to cover both the "bind to variable", and "lookup flag value" cases with default values.

internal/cmd/cliopts/flags.go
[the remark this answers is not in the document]
dnephin: Ya! Sorry I described it wrong the first time. Fixed it in an edit

internal/cmd/login.go
dnephin: Thinking about this more, the `os.Stdin == nil || !term.IsTerminal(int(os.Stdin.Fd())` should probably be the default set on the flag.

That way `--non-inactive=false` will work to override the stdin check. Otherwise there's no way to force it to be interactive when running without a terminal (which we might need for testing).

That has another plus, that now simply using the field from options is correct, and the contributor doesn't need to remember to call this function.

dnephin: fixed in the latest commit

internal/cmd/cmd.go
dnephin: In a subsequent commit I also set a default here based on "is stdin a terminal", which removes the need to call other functions later to determine the interactivity mode, and allows us to override the value when necessary.

ssoroka (approved):

PR #1660, opened 2022-04-20
feat: load identities from config

### LINKED ISSUE #1275: Ability to declare users in server config
Users should be declarable in the Infra configuration rather than manually creating them through the CLI.

Declaring a user should accept an email and a one time password specified in a secret provider or the environment.
```
users:
  - email: admin@example.com
     oneTimePassword: vault:infra-admin-otp
```

mxyng:
## Summary

* remove option to load admin access key from cli
* load identities and credentials/access keys from config
* upgrade users which set server.config.adminAccessKey to the new identities config
  * add a warning in NOTES.txt informing users to update their values
* bundle internal provider and identity loading into `loadProviders` and `loadIdentities`
* move identity creation in `loadGrants` to using `loadIdentities` instead since it shares functionality. it also allows the code to correct track what is created by config and what is not
* refactor the code preventing users from updating/removing internal resources to using `data.InfraProvider()` and `data.InfraConnectorIdentity()`

## Checklist

## Related Issues

Resolves #1275


internal/server/models/grant.go
```diff
@@ -11,11 +11,6 @@ const (
 	InfraConnectorRole = "connector"
 )
 
-const (
-	CreatedBySystem = 0
-	CreatedByConfig = 1
```
mxyng, line 16: consolidating this under `CreatedBySystem` since the server now loads the internal providers and identities using the same mechanism as config providers/identities

internal/server/models/grant.go
```diff
@@ -42,8 +37,6 @@ type Grant struct {
 	Subject   uid.PolymorphicID `validate:"required"` // usually an identity, but could be a role definition
 	Privilege string            `validate:"required"` // role or permission
 	Resource  string            `validate:"required"` // Universal Resource Notation
-
-	CreatedBy uid.ID
```
mxyng, line 46: making this a global field so it can be used by all models

internal/server/models/model.go
```diff
@@ -13,15 +13,20 @@ type Modelable interface {
...
+	ID        uid.ID
+	DeletedAt gorm.DeletedAt
+
+	// Set by GORM to time.Now() when a record is created.
 	// See https://gorm.io/docs/conventions.html#Timestamp-Tracking
 	CreatedAt time.Time
-	// CreatedAt is set by GORM to time.Now when a record is updated.
+	// Set by GORM to time.Now() when a record is updated.
 	// See https://gorm.io/docs/conventions.html#Timestamp-Tracking
 	UpdatedAt time.Time
```
dnephin, line 27: What's the reason for this godoc change? I don't think this matches the recommended style.

Godoc comments are supposed to start with the field name (like they were before).

internal/server/config.go
```diff
@@ -725,3 +593,146 @@ func loadGrant(db *gorm.DB, input Grant) (*models.Grant, error) {
...
+	identity, err := loadIdentity(db, Identity{Name: models.InternalInfraConnectorIdentityName})
+	if err != nil {
+		return err
+	}
+
+	keep = append(keep, identity.ID)
+
+	if err := data.DeleteIdentities(db, data.NotIDs(keep), data.CreatedBy(models.CreatedBySystem)); err != nil {
+		return err
+	}
```
dnephin, line 622: I'm still weary of deleting things just because they aren't in the config anymore. I know that's expected for some tools, but I think it's a bit strange for API-drive systems.

The config is a way of bootstrapping entities, it's not the primary mechanism for managing entities.

dnephin (commented):
Looks cool! Took a quick look since it's still in draft, to get an idea of how it all works.

internal/server/models/model.go
mxyng: Wasn't aware of the recommended style which seems repetitive. Will revert

internal/server/config.go
mxyng: Not sure I agree with this. I would expect any resource created by configs to be removed if they're left out of a config update. I can see where you're coming from but I don't know if keeping the resource is expected

ssoroka: as long as `data.CreatedBy(models.CreatedBySystem)` always points to identities created by the config and not to ones created by the API, we should be fine here. 

I sort of side with daniel on this though: If you're doing IAC it's going to be weird to have a UI/API where you can't change anything (because it's always overwritten by config). I'd really like these sources to play nice together.

internal/server/models/model.go
dnephin: Looking for some docs or examples to make sure I'm not mistaken. Some examples here: https://pkg.go.dev/net/http#Client, https://pkg.go.dev/net/http#PushOptions

I couldn't find it explicitly documented in effective Go for struct fields.

internal/server/config.go
dnephin: I think there are a few difference between IAC tools like `helm` and `terraform` and what we have here.

With `helm` and `terraform`:
1. the config is exclusively the API operations you want to perform
2. the config is read and applied explicitly for that reason
3. the config is read by a client and sent to an API server as API requests

With our server config:
1. it's a mix of operational settings and API operations
2. the server could be restarted for many other reasons, not necessarily because someone wants to make changes to the entities
3. the config is read by the server when a process starts or restarts

I think it would be a lot safer to print a warning in the logs, saying "entity X was removed from the config, if you no longer want it then delete it using the CLI or API", rather than deleting things.

I think it's too easy to accidentally start the process with the wrong config, and cause unintended problems. That's going to be even more of a risk when someone runs an external postgresql DB and multiple replicas of the server.

In the future if we have an `infra apply` command that reads a config to apply it as API calls, that would be a safer way to manage these entities as code.

dnephin: It's also quite possible I'm overestimating the risk, and that as long as we document it properly people will be careful.

internal/access/identity.go
```diff
@@ -61,6 +78,15 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return fmt.Errorf("cannot delete self: %w", internal.ErrForbidden)
 	}
 
+	internalIdentity, err := isInternalIdentity(c, id)
```
BruceMacD, line 81: What does this guard against? Is it just for preventing internal identities from making external requests for their own data?

internal/access/provider.go
```diff
@@ -15,6 +15,17 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
+func isInternalProvider(c *gin.Context, requestedID uid.ID) (bool, error) {
+	db := getDB(c)
+
+	internalProvider, err := data.GetProvider(db, data.ByName(models.InternalInfraProviderName))
```
BruceMacD, line 21: You can use the `InfraProvider()` function here to get the internal provider simply

internal/server/config.go
```diff
@@ -678,23 +548,34 @@ func loadConfig(db *gorm.DB, config Config) error {
 }
 
 func loadProviders(db *gorm.DB, providers []Provider) error {
-	toKeep := make([]uid.ID, 0)
+	keep := make([]uid.ID, 0, len(providers)+1)
```
BruceMacD, line 551: We should make this `+1` a named constant rather than a magic number

mxyng: Wasn't a fan of this either. Instead, internal configurations will be injected into the config at an earlier stage so this won't be necessary anymore

internal/access/identity.go
mxyng: This is to prevent the user from removing internal identities as that will have unexpected consequences

internal/server/models/grant.go
mxyng: Reverted this since there were some unexpected side effects such as adding `CreatedBy` to `Settings`

BruceMacD (commented):
Looking good, just a couple more questions

helm/charts/infra/templates/connector/_helpers.tpl
```diff
@@ -158,6 +157,20 @@ existing secret and use its password. If the secret does not exist, randomly gen
 {{- end }}
 {{- end }}
 
+{{/*
+Infra connector 'env' values. Merges global and local values.
```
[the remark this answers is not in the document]
mxyng: These are two different ways of setting environment variables. One takes a key value from configmap or secret while the other takes the entire object. There's a reasonable use case for both

BruceMacD (approved):

dnephin (commented):
Oops, sorry forgot to hit submit on this comment

internal/server/models/model.go
```diff
@@ -13,15 +13,20 @@ type Modelable interface {
 	IsAModel() // there's nothing specific about this function except that all Model structs will have it.
 }
 
+const CreatedBySystem = 1
```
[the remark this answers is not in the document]
dnephin: Or do we need a migration for this? From another discussion it sounded like making this `1` would probably be good, so that `0` is an invalid ID.

PR #1661, opened 2022-04-20
fix: fix 404 and proxy ui routes

jmorganca:
Fixes:
* ui routes falling through to the NotFound handler in development
* Embedded FS not detecting `.html` files exist – (e.g. when the user loads `/a/path` it should check if `/a/path.html` exists as well as `/a/path`

### DIFF
--- a/internal/server/static_fs.go
+++ b/internal/server/static_fs.go
@@ -35,8 +35,13 @@ const uiFilePathPrefix = "ui"
 
 func (sfs StaticFileSystem) Exists(prefix string, filepath string) bool {
 	if p := strings.TrimPrefix(filepath, prefix); len(p) < len(filepath) {
-		_, err := sfs.base.Open(path.Join(uiFilePathPrefix, p))
-		return err == nil
+		if _, err := sfs.base.Open(path.Join(uiFilePathPrefix, p)); err == nil {
+			return true
+		}
+
+		if _, err := sfs.base.Open(path.Join(uiFilePathPrefix+".html", p)); err == nil {
+			return true
+		}
 	}
 
 	return false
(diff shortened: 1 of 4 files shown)


internal/server/static_fs.go
```diff
@@ -35,8 +35,13 @@ const uiFilePathPrefix = "ui"
...
 	if p := strings.TrimPrefix(filepath, prefix); len(p) < len(filepath) {
-		_, err := sfs.base.Open(path.Join(uiFilePathPrefix, p))
-		return err == nil
+		if _, err := sfs.base.Open(path.Join(uiFilePathPrefix, p)); err == nil {
+			return true
+		}
+
+		if _, err := sfs.base.Open(path.Join(uiFilePathPrefix+".html", p)); err == nil {
+			return true
+		}
```
dnephin, line 44: Hmm, I thought `TestServer_GenerateRoutes_UI/page without .html suffix` was already testing this case. 

I guess that test is wrong, because `/providers` also exists as a directory. Oops!

Changing the path in that test case to `/providers/add/admins` shows the problem i think. That change should be enough to verify this fix.

dnephin (commented):
Thanks for fixing these!

Change LGTM, and I left a suggestion for how to get test coverage of the static fs fix.

Maybe we could have `Exists` use `sfs.Open` (instead of `base.Open`), but it didn't work for me, so maybe not.

dnephin (approved):
LGTM!

`TestServer_GenerateRoutes_UI/page without .html suffix` could still be fixed to cover this case, but not necessary to do that now.

dnephin (approved):

PR #1665, opened 2022-04-21
draft: include identity names in API response

dnephin:
Related to #1548

The CLI should not have to perform more queries to provide a reasonable response. If our CLI needs this, there's a good chance that other API consumers will need it as well.

This PR augments the list keys API response to include the identity name. That way the CLI can use it without have to query for the identity.

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -401,11 +401,18 @@ func (a *API) ListAccessKeys(c *gin.Context, r *api.ListAccessKeysRequest) ([]ap
 	results := make([]api.AccessKey, len(accessKeys))
 
 	for i, a := range accessKeys {
+		// TODO: this would be better as a JOIN query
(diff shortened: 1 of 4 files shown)


internal/server/handlers.go
```diff
@@ -401,11 +401,18 @@ func (a *API) ListAccessKeys(c *gin.Context, r *api.ListAccessKeysRequest) ([]ap
 	results := make([]api.AccessKey, len(accessKeys))
 
 	for i, a := range accessKeys {
+		// TODO: this would be better as a JOIN query
+		identity, err := access.GetIdentity(c, a.IssuedFor)
```
dnephin, line 405: This is still iterating and querying once for ever key, which is not ideal.

The way I am used to doing this would be to instead have a join query between access keys and identities, so that we only need one query. I'm not sure how to do that with GORM.

dnephin:
This was done in #1666

PR #1666, opened 2022-04-21
feat: format infra keys list with user name

### LINKED ISSUE #1255: `infra keys list` output needs formatting
1. Remove`ID` header
2. Issued for should be the name as shown `infra id list`
3. Date/times should be cleaned up to a cleaner format

```
  ID          NAME                                                    ISSUED FOR    CREATED                                  EXPIRES                                  EXTENSION DEADLINE             
  4Lu3ui8Y5h  4Lu3ui8Y5h                                              m:4Lu3ueNuG1  2022-03-19 05:01:13.184144253 +0000 UTC  2022-03-19 17:01:13.183881837 +0000 UTC  0001-01-01 00:00:00 +0000 UTC  
  4Lu4peujGq  access key presented by desktop kubernetes destination  m:4Lu4p9ES2N  2022-03-19 05:01:21.468754424 +0000 UTC  2023-03-19 05:01:21.468612132 +0000 UTC  0001-01-01 00:00:00 +0000 UTC  
  4Lu3ufaZvA  admin access key                                        m:4Lu3ueNuG1  2022-03-19 05:01:13.176149253 +0000 UTC  2314-06-29 04:48:30.030859227 +0000 UTC  0001-01-01 00:00:00 +0000 UTC  
```

ssoroka:
## Summary

Format keys list output
```
$ ./infra keys list     
  NAME        ISSUED FOR                 CREATED        EXPIRES            EXTENSION DEADLINE  
  6pmZ7QVjoU  steven.soroka@infrahq.com  6 minutes ago  12 hours from now  never               
```

## Checklist

## Related Issues

Resolves #1255

### DIFF
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -58,7 +58,7 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
 		resp.Code = http.StatusBadRequest
 		resp.Message = err.Error()
 	case errors.Is(err, context.DeadlineExceeded):
-		resp.Code = http.StatusBadGateway // not ideal, but StatusRequestTimeout isn't intended for this.
+		resp.Code = http.StatusGatewayTimeout // not ideal, but StatusRequestTimeout isn't intended for this.
 		resp.Message = "request timed out"
 	default:
 		log = logging.L.WithOptions(zap.AddCallerSkip(1)).Error
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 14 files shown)


kimskimchi (approved):

internal/server/errors.go
```diff
@@ -56,7 +57,9 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
 	case errors.Is(err, (*validator.InvalidValidationError)(nil)):
 		resp.Code = http.StatusBadRequest
 		resp.Message = err.Error()
-
+	case errors.Is(err, context.DeadlineExceeded):
+		resp.Code = http.StatusBadGateway // not ideal, but StatusRequestTimeout isn't intended for this.
```
dnephin, line 61: Maybe [504 Gateway Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) ?

dnephin (approved):
LGTM!

One suggestion for the timeout status code change.

I think it might still be worth exploring an approach like #1665 in the future, to remove the need for additional HTTP request and DB queries, but I think this is a safe approach too. Supporting filtering by IDs in list identities is a nice feature either way.

internal/server/errors.go
ssoroka: yeah. hard to know what's good here. Gateway timeout seemed a lot like we were pointing the finger at another service, but it's probably a timeout between us and the DB

dnephin: Ya, none of them seem like a perfect fit. "Gateway Timeout" is nice in that at least it mentions timeout.

"Bad Gateway" is a risk because that's what we special case that for "there was a problem talking to the identity provider". So we probably don't want to use it again for this.

ssoroka: I actually thought I selected GatewayTimeout 🤣

PR #1667, opened 2022-04-21
improve: add a test helper for creating a ClientConfig

dnephin:
## Summary

Most CLI commands require a login first. In tests we simulate a login by writing out a `ClientConfig`. This PR extracts a test helper for building that `ClientConfig`, so that if we change the format of the config it'll be less work to migrate the tests.

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -103,24 +103,10 @@ XlW7KilKI5YkcszGoPB4RePiHsH+7trf7l8IQq5r5kRq7SKsZ41BI6s1E1PQVW93
 			}
 		}
 
-		svc := httptest.NewTLSServer(http.HandlerFunc(handler))
-		t.Cleanup(svc.Close)
+		srv := httptest.NewTLSServer(http.HandlerFunc(handler))
+		t.Cleanup(srv.Close)
 
-		cfg := ClientConfig{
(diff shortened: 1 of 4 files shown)


internal/cmd/cmd_test.go
```diff
@@ -197,3 +183,33 @@ XlW7KilKI5YkcszGoPB4RePiHsH+7trf7l8IQq5r5kRq7SKsZ41BI6s1E1PQVW93
...
 }
+
+// newTestClientConfig returns a ClientConfig that can be used to test CLI
+// commands. Most CLI commands require a login first, which saves a ClientConfig
+// to a file.
+// newTestClientConfig provides a reasonable default for most cases, removing
+// the need to perform a full login. The returned value may be modified, and then
+// should be saved to a file with writeConfig.
+// If any fields in identity are not set, they will be set to default values.
+func newTestClientConfig(srv *httptest.Server, identity api.Identity) ClientConfig {
```
dnephin, line 194: I wasn't sure about the `*httptest.Server` arg. We could pass in just a string instead, but in the future if we want to use the TLS settings from the server we might need more of it.

dnephin (commented):
I left the logout tests for now because it's more of a special case, and there is another open PR that changes logout, so I don't want to conflict.

mxyng (approved):

PR #1668, opened 2022-04-21
cli: use DefaultsFromEnv - part 2

dnephin:
Branched from #1655, and includes a commit with #1667, so those two will probably need to merge first

## Summary

Converts the remaining "interactive" (short lived) CLI commands to remove the calls to `parseOptions`. The only two calls that are left are for the "long running" commands (`server` and `connector`) which will need a different approach.

Also adds some tests for the commands that changes, to show the options are passed to the API correctly.

### DIFF
--- a/internal/cmd/identities.go
+++ b/internal/cmd/identities.go
@@ -36,10 +36,6 @@ func newIdentitiesCmd() *cobra.Command {
 	return cmd
 }
 
-type addIdentityCmdOptions struct {
-	Password bool `mapstructure:"password"`
-}
-
 func newIdentitiesAddCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "add IDENTITY",
@@ -54,11 +50,6 @@ EMAIL must contain a valid email address in the form of "local@domain".
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			name := args[0]
-
-			var options addIdentityCmdOptions
-			if err := parseOptions(cmd, &options, ""); err != nil {
-				return err
-			}
 
 			createResp, err := CreateIdentity(&api.CreateIdentityRequest{Name: name, SetOneTimePassword: true})
 			if err != nil {
--- a/internal/cmd/providers.go
+++ b/internal/cmd/providers.go
@@ -3,10 +3,12 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/infrahq/infra/api"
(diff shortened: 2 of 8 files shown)


internal/cmd/identities.go
```diff
@@ -36,10 +36,6 @@ func newIdentitiesCmd() *cobra.Command {
 	return cmd
 }
 
-type addIdentityCmdOptions struct {
-	Password bool `mapstructure:"password"`
-}
```
dnephin, line 41: This wasn't being used, so I removed it. Should it be used? I don't see any flags on `identity add`, so I'm guessing they were removed.

internal/cmd/providers.go
```diff
@@ -97,34 +120,22 @@ PROVIDER is a short unique name of the identity provider bieng added (eg. okta)
...
-
-	if err := cmd.MarkFlagRequired("url"); err != nil {
-		panic("cannot set flag [--url] as required")
-	}
-	if err := cmd.MarkFlagRequired("client-id"); err != nil {
-		panic("cannot set flag [--client-id] as required")
-	}
-	if err := cmd.MarkFlagRequired("client-secret"); err != nil {
-		panic("cannot set flag [--client-secret] as required")
-	}
```
dnephin, line 126: Requiring flags this way does not allow us to set values from env vars. We could choose to not accept values from env vars. This was a problem when we used `parseOptions` as well.

The code to replace this (the `Validate` method above) is only 2 extra lines of code, and it allows us to customize the message. I think even if we don't use env vars we can ignore this feature of `cobra` and do it ourselves.

uid/snowflake.go
```diff
@@ -68,6 +68,7 @@ func (u *ID) UnmarshalText(b []byte) error {
 	return nil
 }
 
-func (u *ID) MarshalText() ([]byte, error) {
+// TODO: test cases
+func (u ID) MarshalText() ([]byte, error) {
```
dnephin, line 72: I noticed this while testing. It's the same problem as #1625.

From [effective Go](https://go.dev/doc/effective_go#pointers_vs_values):
> The rule about pointers vs. values for receivers is that value methods can be invoked on pointers and values, but pointer methods can only be invoked on pointers.

By using a value receiver, we can marshal api types directly without them needing to be pointers. We don't use pointers to `uid.ID`, but I think what ends up happening is that `json.Marshal` calls `reflect.Value.CanAddr`. When that returns false it is not able to use the method with a pointer receiver.

I can write a similar set of tests for this in another PR.

kimskimchi (approved):
looks good, thank you!

ssoroka (approved):

PR #1674, opened 2022-04-22
improve(cli): include usage in error message

### LINKED ISSUE #1551: CLI errors should print usage
```
mchiang@Michaels-MacBook-Pro infra % infra grant add                                  
ERROR  accepts 2 arg(s), received 0
``` 

Example of a good experience: 

```
mchiang@Michaels-MacBook-Pro infra % docker run
"docker run" requires at least 1 argument.
See 'docker run --help'.

Usage:  docker run [OPTIONS] IMAGE [COMMAND] [ARG...]

Run a command in a new container
```

dnephin:
## Summary

Include the usage line when a command fails because the wrong number of arguments were provided.

## Checklist

## Related Issues

Resolves #1551


docs/reference/cli-reference.md
```diff
@@ -348,7 +348,7 @@ infra identities list [flags]
 Delete an identity
 
 ```
-infra identities remove NAME [flags]
+infra identities remove IDENTITY [flags]
```
dnephin, line 351: I changed this because as I was writing tests for it I noticed it was inconsistent with the other `infra identities` commands. The usage string for both `add` and `edit` use `IDENTITY` as the name of the positional argument.

internal/cmd/args.go
```diff
@@ -0,0 +1,49 @@
...
+		if len(args) == number {
+			return nil
+		}
+		return fmt.Errorf(
+			"%q requires exactly %d %s.\nSee '%s --help'.\n\nUsage:  %s\n",
+			cmd.CommandPath(),
+			number,
+			pluralize("argument", number),
+			cmd.CommandPath(),
+			cmd.UseLine())
```
dnephin, line 23: Example:
```
ERROR  "infra grants add" requires exactly 2 arguments.
See 'infra grants add --help'.

Usage:  infra grants add IDENTITY DESTINATION [flags]

```

internal/cmd/args.go
```diff
@@ -0,0 +1,49 @@
...
+		if len(args) <= max {
+			return nil
+		}
+		return fmt.Errorf(
+			"%q accepts at most %d %s.\nSee '%s --help'.\n\nUsage:  %s\n",
+			cmd.CommandPath(),
+			max,
+			pluralize("argument", max),
+			cmd.CommandPath(),
+			cmd.UseLine())
```
dnephin, line 47: Example:
```
ERROR  "infra login" accepts at most 1 argument.
See 'infra login --help'.

Usage:  infra login [SERVER] [flags]

```

internal/cmd/args.go
dnephin: Hmm, maybe this should use only single quotes or only double quotes, instead of both

jmorganca (approved):

PR #1675, opened 2022-04-22
improve: remove NotCreatedBy system filter for grants list

### LINKED ISSUE #1669: Admin cannot introspect their `infra` role
Admin user cannot see their role since it's created by the "system"

dnephin:
## Summary

Show all grants when listing grants. This was fixed on the CLI side in #1533.

I kept the filter for `DeleteGrants` for now. I wanted to handle it explicitly with a more specific error, but that would break `infra grant delete` because it iterates over all grants. I think we should revisit that separately, it works well for now.

I tested it manually and saw this:

```
$ infra grant list
  IDENTITY   ACCESS     DESTINATION  
  admin      admin      infra        
  connector  connector  infra        
```

To automate a test for this we need more of an integration test (with both CLI and a real API) and some way to capture the stdout of the CLI. We don't have a pattern for that yet, but it's something I'm going to look at adding soon.

## Checklist

## Related Issues

Resolves #1669

### DIFF
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -64,6 +64,7 @@ func Signup(c *gin.Context, name, password string) (*models.Identity, error) {
 		Subject:   uid.NewIdentityPolymorphicID(identity.ID),
 		Privilege: models.InfraAdminRole,
 		Resource:  "infra",
+		CreatedBy: models.CreatedBySystem,
 	}
 
 	if err := data.CreateGrant(db, grant); err != nil {
--- a/internal/access/grant.go
+++ b/internal/access/grant.go
@@ -14,7 +14,7 @@ func GetGrant(c *gin.Context, id uid.ID) (*models.Grant, error) {
 		return nil, err
 	}
 
-	return data.GetGrant(db, data.ByID(id), data.NotCreatedBy(models.CreatedBySystem))
+	return data.GetGrant(db, data.ByID(id))
 }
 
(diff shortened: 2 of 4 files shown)


internal/access/signup.go
```diff
@@ -64,6 +64,7 @@ func Signup(c *gin.Context, name, password string) (*models.Identity, error) {
 		Subject:   uid.NewIdentityPolymorphicID(identity.ID),
 		Privilege: models.InfraAdminRole,
 		Resource:  "infra",
+		CreatedBy: models.CreatedBySystem,
```
dnephin, line 67: This is a no-op change, because `models.CreatedBySystem` has a value of `0`, but I added this here and in another place to be more explicit about the value. And so that if we change the value in the future (we might in #1660), then it'll be correct.

I wonder if we actually want to use a non-zero value for this, so that we van error if `CreatedBy` is set to 0, to catch bugs in the future. Although I'm not sure how we ensure that `uid.New` doesn't generate the same ID.

ssoroka: There was some weirdness around the zero value of uid.ID (unintentionally from my point of view) defaulting to 1 instead of 0. Not sure what effect that would have here. I'd probably like CreatedBySystem to have a value > 1 and a validation rule that doesn't accept <= 1

ssoroka (approved):

jmorganca (approved):

PR #1687, opened 2022-04-25
fix: cluster-specific ca for generating connector certificates

### LINKED ISSUE #1637: Avoid use of SNI
Many Kubernetes clients don't have great SNI support (e.g. javascript-based ones)

We should generate a CA with the connectors and generate self-signed certs from that CA with the proper SAN fields instead of relying on SNI since many clients don't have this.

Potentially helpful: https://helm.sh/docs/chart_template_guide/function_list/#genca

jmorganca:
Fixes #1637

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"os"
 	"strings"
 	"sync"
 	"time"
@@ -20,7 +21,6 @@ import (
 	"github.com/goware/urlx"
 	"github.com/infrahq/secrets"
 	"github.com/prometheus/client_golang/prometheus"
-	"golang.org/x/crypto/acme/autocert"
 	"gopkg.in/square/go-jose.v2"
 	"gopkg.in/square/go-jose.v2/jwt"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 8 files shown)


internal/connector/connector.go
```diff
@@ -320,37 +318,47 @@ func Run(ctx context.Context, options Options) error {
...
+	validServernames.Store("127.0.0.1", true)
+	validServernames.Store("infra-connector", true)
 
-		if err := manager.Cache.Put(context.TODO(), serverName, certBytes); err != nil {
-			return err
+	// Generate TLS certificates on the fly for clients
+	// GenerateCertificate caches certificates
+	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
+	tlsConfig.GetCertificate = func(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+		servername := hello.ServerName
```
ssoroka, line 355: servername might be missing if client doesn't support it

ssoroka (approved):

internal/connector/connector.go
jmorganca: Great catch! Updated since the prior approach wouldn't work for IP addresses

mxyng (approved):

jmorganca:
@mxyng @ssoroka  thanks for the review!

PR #1702, opened 2022-04-25
fix: do not look up k8s name when provided

### LINKED ISSUE #1308: connector: don't lookup kubernetes cluster name when one is provided
In `internal/kubernetes.Kubernetes.Name()` we return both the name and the checksum.

The checksum is always used by the caller, but the name is thrown away if options contained a name.  The collector is making a few unnecessary API calls to the cloud provider or kubernetes in this case.

We could fix this by splitting up `Name()`, and having it always return the checksum, but only doing the name lookup when there is none provided.

BruceMacD:
- separate checksum lookup and autoname functions

## Summary
If the name is provided for the kubernetes connector we don't need to do the name look-up. Separate the function so that we can skip this and still get the checksum.

## Checklist

## Related Issues

Resolves #1308

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -306,13 +306,18 @@ func Run(ctx context.Context, options Options) error {
 		return err
 	}
 
-	autoname, chksm, err := k8s.Name()
+	chksm, err := k8s.Checksum()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


dnephin (approved):
LGTM! Thanks for making this change

PR #1704, opened 2022-04-25
UI milestone 1 

### LINKED ISSUE #1617: ui: setting up admin users should use `/v1/identities`
## Describe the bug

The UI uses `/v1/users` when setting up an admin user. This API is deprecated and has been replaced with `/v1/identities`

hoyyeva:
## Summary
This PR contains all the features that are included in the milestone 1 of UI which are:

- dashboard
1. a vertical navigation that has two options (infrastructures & identity providers)
2. settings on the bottom of the navigation close to the log out button
3. both infrastructures and identity providers have empty table page and the layout of the two pages are similar

- infrastructures
1. able to add multiple clusters
2. able to remove the cluster from the list ( but this doesn't mean the cluster is uninstalled, the uninstall should still be done through CLI)
3. able to grant access and update the role

- identity providers
1. able to add okta providers
2. able to delete okta providers 

- setting page
1. add an admin section
2. the admin section is able to add and revoke infra admin

## Related Issues
Resolves #1617


ui/lib/email.js
```diff
@@ -0,0 +1,7 @@
+export const validateEmail = (email) => {
```
jmorganca, line 1: Is there a way we can do this via `<input type='email' required />`?

ui/pages/destinations/add.js
```diff
@@ -0,0 +1,181 @@
...
+    <Fullscreen closeHref='/destinations' verticalCenteredContent={false}>
+      <Head>
+        <title>Infra - Destinations</title>
+      </Head>
+      <div className='flex flex-col mb-10 w-full max-w-md'>
+        <HeaderIcon iconPath='/destinations-color.svg' position='center' />
+        <h1 className='text-xl font-bold tracking-tight text-center'>Connect a Cluster</h1>
+        <div className='flex gap-1 mt-8 mb-5'>
+          <div className='flex-1 w-full'>
+            <InputDropdown
```
jmorganca, line 129: What's the advantage of using an `InputDropdown` here vs simply an `<input>`?

ui/pages/destinations/grantAccessContent.js
```diff
@@ -0,0 +1,151 @@
+import useSWR, { useSWRConfig } from 'swr'
```
jmorganca, line 1: This file should be in `components` and not under `pages`. Even better would be to put it in the `destinations/index.js` page since it isn't that much code and this is the only place where it is used.

ui/pages/destinations/grantAccessContent.js
```diff
@@ -0,0 +1,151 @@
...
+    fetch('/v1/grants', {
+      method: 'POST',
+      body: JSON.stringify({ subject: id, resource: destination.name, privilege })
+    })
+      .then((response) => response.json())
+      .then(() => mutate(`/v1/grants?resource=${destination.name}`))
+      .finally(() => setEmail(''))
+  }
+
+  const handleInputChang = (value) => {
```
jmorganca, line 53: Typo - also no need for the `()` paranthesis: `value => {`

ui/pages/destinations/grantAccessContent.js
```diff
@@ -0,0 +1,151 @@
...
+  const handleKeyDownEvent = (key) => {
+    if (key === 'Enter' && email.length > 0) {
+      handleShareGrant()
+    }
+  }
+
+  const handleShareGrant = () => {
+    if (validateEmail(email)) {
+      setError('')
+      fetch(`/v1/identities?name=${email}`)
```
jmorganca, line 67: Let's use `async` and `await` here

ui/lib/email.js
hoyyeva: I tried to do that first, but for some reason it did not verify the input value. So that is why I do this way. I was planning to make an investigate ticket to see why it did not work after we release the ui

ui/pages/destinations/add.js
hoyyeva: The input dropdown comes with the dropdown inside the input (the one in the share dialog on destination page). The reason why I turned it into a component is because I think this design pattern will be used in other area of the ui. However, we don't have that right now in the ui just yet. So it doesn't really need to be a component

ui/pages/destinations/grantAccessContent.js
```diff
@@ -0,0 +1,151 @@
...
+    } else {
+      setError('Invalid email')
+    }
+  }
+
+  const handleUpdateGrant = (privilege, grantId, userId) => {
+    fetch(`/v1/grants/${grantId}`, { method: 'DELETE' })
+      .then(() => {
+        if (privilege === 'remove') {
+          mutate(`/v1/grants?resource=${destination.name}`)
```
jmorganca, line 92: Better to put the fetch in `mutate` so we can have an [optimistic update](https://swr.vercel.app/docs/mutation#optimistic-updates) (instant to the user):

ui/pages/destinations/grantAccessContent.js
hoyyeva: I don't know if it is good idea to put it in the `destinations/index.js` page because I think it would make the `destinations/index,js` too long to maintain. But I think it is good idea to move it to `component`

ui/pages/destinations/add.js
```diff
@@ -0,0 +1,181 @@
+import Head from 'next/head'
+import Router from 'next/router'
+import { useState, useEffect } from 'react'
+import useSWR from 'swr'
+
+import Fullscreen from '../../components/modals/fullscreen'
+import HeaderIcon from '../../components/header-icon'
+import InputDropdown from '../../components/input'
+
+const CommandInput = ({ enabledCommandInput, accessKey, currentDestinationName }) => {
```
jmorganca, line 10: nit: `function CommandInput(` for consistency

ui/pages/destinations/add.js
```diff
@@ -0,0 +1,181 @@
...
+  const server = window.location.host
+  const isHttps = window.location.origin.includes('https')
```
jmorganca, line 12: Does this work while using self-signed certificates? We may have to use a [trick](https://stackoverflow.com/questions/2618944/check-in-javascript-if-an-ssl-certificate-is-valid) to detect if the certificates are valid (i.e. the user has set up TLS certificates) or not

ui/lib/providers.js
```diff
@@ -0,0 +1,37 @@
+export const providers = [{
+  name: 'Okta',
+  kind: 'okta',
+  available: true,
+  validate: {
+    url: url => {
```
hoyyeva, line 6: is this enough to check if the url is valid? I think it is possible to customize the okta URL domain (https://developer.okta.com/docs/guides/custom-url-domain/main/). wondering In that case how are we going to handle that?

ui/lib/providers.js
```diff
@@ -0,0 +1,37 @@
...
+      if (!url.endsWith('okta.com')) {
```
hoyyeva, line 7: are we checking if the url is ended with `okta.com` or `.okta.com`?

ui/pages/destinations/index.js
```diff
@@ -0,0 +1,201 @@
...
+                <button onClick={() => setOpen(true)} className='py-1 px-2 -mr-2 cursor-pointer'>
+                  <XIcon className='w-5 h-5 text-gray-500 hover:text-white' />
+                </button>
+
+                {/* delete modal */}
+                <DeleteModal
+                  open={open}
+                  setOpen={setOpen}
+                  onSubmit={async () => {
+                    await fetch(`/v1/destinations/${id}`, { method: 'DELETE' })
```
hoyyeva, line 119: ☝🏻 is this here by accident? why are we calling delete endpoint twice?

jmorganca: Great catch. Fixed.

jmorganca:
Note to reviewers: while this appears to be a large change, it's mostly generated files since the UI is statically generated.

docs/guides/identity-providers/okta.md
```diff
@@ -15,27 +15,38 @@ infra providers add okta \
...
+   c. Click **Next**.
 
 ![App Type](../../images/connect-users-okta-okta2.png)
 
-4. On the **General** tab, **note** the **Client ID**, **Client Secret**, and **Okta domain** for adding your Okta information to Infra later.
+4. Configure your new Okta App:
+  - For **App integration name** write **Infra**.
+  - Under **General Settings** > **Grant type** select **Authorization Code** and **Refresh Token**.
+  - For **Sign-in redirect URIs** write:
+    - `http://localhost:8301` (for CLI login)
```
jmorganca, line 33: This is re-worked to add UI support.

Note: we should consider moving all OIDC flows to the UI once `--enable-ui` is `true` by default.

For two reasons:
1.  Users only have to add a single redirect URL
2. Users will be able to log in environments without a browser (e.g. ssh, a docker container)

jmorganca (approved):

PR #1708, opened 2022-04-25
cli: support testing the stdout produced by CLI commands

dnephin:
## Summary

One of the primary responsibilities of the CLI is to output things in a format that is easy for humans to consume. We should have some test coverage of that output, to prevent regressions. When we intentionally make change to the output it also gives us an easy way to preview the changes by looking at the tests.

This PR implements an approach that I've used previously to test CLI output.
1. the first commit adds a `CLI` type for passing around input and output streams. I think it's likely we'll add more fields to this type in the future. The `api.Client` could be one of the fields, common CLI flags that may be used in multiple commands is another one.
2. the second commit adds a test for `infra keys list` that uses the support added in 1

This PR uses [golden files](https://softwareengineering.stackexchange.com/a/358792) (called [Output comparison testing in wikipedia](https://en.wikipedia.org/wiki/Software_testing#Output_comparison_testing)) to store the expected values. Golden files are nice because:
* they are strict about the expected output
* they allow both the contributor and reviewers to easily see what the output will look like
* they are easy to update when they change

To update a golden file, run `go test ./internal/cmd -test.update-golden`, then look at the `git diff` to make sure the change is what you expect.

### DIFF
--- a/internal/cmd/cli.go
+++ b/internal/cmd/cli.go
@@ -1,0 +1,51 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+)
+
+// CLI exposes common dependencies to commands.
+type CLI struct {
+	Stdin  io.Reader
+	Stdout io.Writer
+	Stderr io.Writer
+}
+
+// Output a string to CLI.Stdout. Output is like fmt.Printf except that it always
+// adds a trailing newline.
+// To write output without a trailing newline use CLI.Stdout directly.
+func (c *CLI) Output(format string, args ...interface{}) {
+	fmt.Fprintf(c.Stdout, format+"\n", args...)
+}
+
+// TODO indicates a place where os.Stdout is being used instead of CLI.Stdout.
+// It is temporary, and can be removed once everything is ported to use CLI
+// for output.
+var TODO = os.Stdout
+
+// key is a type to ensure no other package can access the CLI value in context.
+type key struct{}
+
+// ctxKey used to store CLI in the context.
+var ctxKey = key{}
+
+// newCli looks for a CLI stores in context. If one exists, the CLI from
+// context is returned, otherwise a new CLI is created with streams set to the
+// standard input and output streams.
+//
+// newCLI is a shim for testing, allowing tests to use a buffer instead of the
+// standard streams.
+func newCLI(ctx context.Context) *CLI {
+	cli, ok := ctx.Value(ctxKey).(*CLI)
+	if ok {
+		return cli
+	}
+	return &CLI{
+		Stdin:  os.Stdin,
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+	}
+}
--- a/internal/cmd/keys_test.go
+++ b/internal/cmd/keys_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"gotest.tools/v3/assert"
+	"gotest.tools/v3/golden"
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/uid"
@@ -25,6 +26,10 @@ func TestKeysAddCmd(t *testing.T) {
 		handler := func(resp http.ResponseWriter, req *http.Request) {
 			// the command does a lookup for machine ID
 			if requestMatches(req, http.MethodGet, "/v1/identities") {
+				if req.URL.Query().Get("name") != "my-machine" {
+					resp.WriteHeader(http.StatusBadRequest)
+					return
+				}
 				resp.WriteHeader(http.StatusOK)
 				err := json.NewEncoder(resp).Encode([]*api.Identity{
 					{ID: uid.ID(12345678)},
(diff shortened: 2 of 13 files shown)


internal/cmd/cli.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+)
+
+// CLI exposes common dependencies to commands.
+type CLI struct {
```
dnephin, line 12: I expect we will very likely add to this type. Probably common CLI options  and maybe an `api.Client`.

We should also do something with logging. Right now patching `stderr` doesn't do much because the logger would have already grabbed a reference to `os.Stderr` before this is called.

internal/cmd/testdata/TestKeysListCmd/list_all
```diff
@@ -0,0 +1,4 @@
+  NAME        ISSUED FOR  CREATED       EXPIRES       EXTENSION DEADLINE  
+  front-door  admin       4 months ago  never         never               
+  side-door   admin       4 months ago  4 months ago  4 months ago        
+  storage     clerk       4 months ago  4 months ago  never
```
dnephin, line 4: For now these values are stored in separate files.  I think it would be possible to instead store them in package-level variables in the test file, next to the test. That might work better for our use case.

I think that should be possible, I'll see about adding support for that to `gotest.tools/v3/golden`.

I'd also like to change the update flag to `-update` instead of `-test.update-golden`, to make it easier to remember.

internal/cmd/keys_test.go
```diff
@@ -88,3 +93,102 @@ func TestKeysAddCmd(t *testing.T) {
 func requestMatches(req *http.Request, method string, path string) bool {
 	return req.Method == method && req.URL.Path == path
 }
+
+func TestKeysListCmd(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	t.Setenv("USERPROFILE", home) // for windows
+
+	base := time.Date(2022, 1, 2, 3, 4, 5, 6, time.UTC)
```
dnephin, line 102: This isn't going to work with our relative date output. I'll change this to not be a fixed date.

internal/cmd/cli.go
```diff
@@ -0,0 +1,73 @@
...
+// adds a trailing newline.
+// To write output without a trailing newline use CLI.Stdout directly.
+func (c *CLI) Output(format string, args ...interface{}) {
+	fmt.Fprintf(c.Stdout, format+"\n", args...)
+}
+
+// TODO indicates a place where os.Stdout is being used instead of CLI.Stdout.
+// It is temporary, and can be removed once everything is ported to use CLI
+// for output.
+var TODO = os.Stdout
```
BruceMacD, line 28: Were there some constraints you met that required adding this TODO or just trying to keep the size of the PR down?

internal/cmd/cli.go
```diff
@@ -0,0 +1,73 @@
...
+		Stdin:  os.Stdin,
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+	}
+}
+
+// PatchCLI returns a context which contains a CLI value with the output streams
+// set to buffers. PatchCLI is used by tests to record the output produced by
+// CLI commands.
+func PatchCLI(ctx context.Context) (context.Context, BufferedStreams) {
```
BruceMacD, line 57: Should this live in a test file where it is only ever used by tests?

internal/cmd/keys_test.go
```diff
@@ -25,6 +26,10 @@ func TestKeysAddCmd(t *testing.T) {
 		handler := func(resp http.ResponseWriter, req *http.Request) {
 			// the command does a lookup for machine ID
 			if requestMatches(req, http.MethodGet, "/v1/identities") {
+				if req.URL.Query().Get("name") != "my-machine" {
```
BruceMacD, line 29: Any ideas for how we can simplify these CLI test handlers? I could see a lot of code building up here. Not needed for this PR though.

internal/cmd/cli.go
dnephin: That would work for now. I think it's likely that we'll end up with more than one package for the CLI commands, which will require this to be in the live code, but we could still find a way to put it into a test-only package.

I'll move it into a test file for now, since we don't have that constraint. We can figure out where to move it when we end up with more packages for CLI commands.

internal/cmd/keys_test.go
dnephin: Ya, great question! I'm honestly not sure what to do with these yet. 

I'm guessing you are referring to the stub implementations of the API handlers? I wrote a few last week that seemed like they would be a challenge to maintain, so I have also been thinking we should look for other options there.

One option is to run a full `server.Server`, which removes the need to return canned data, but introduces the need to create the entities in the DB. I'm not sure how much extra time that will add to tests, but it's an option we can try out. That approach would also makes it much harder to write these tests, because we can't predict the IDs or times in the API responses.

If our API handlers accepted an interface for the "backend" (what is currently `access` and `data`) we could run a real API router, with a fake backend to inject the data. That might be less work than what we have here. That would be a much larger change, and probably not something we can easily do just for this purpose.

Another option would be to extract some of this, and make a more complete "fake" API. We could use an actual router that simulates our real API and put that in a more central place. That might allow us to remove some of the boilerplate path matching and JSON encoding, and allow these setup methods to only specify canned responses for specific inputs.

That last option seems the most viable at this point. Thoughts?

BruceMacD (approved):

internal/cmd/keys_test.go
BruceMacD: (moved discussion to our engineering channel)

mxyng (approved):

PR #1712, opened 2022-04-26
fix: helm `values.yaml` documentation for ui

jmorganca:


pdevine (approved):

internal/cmd/server.go
```diff
@@ -46,6 +47,9 @@ func newServerCmd() *cobra.Command {
 
 			options.DBEncryptionKey = dbEncryptionKey
 
+			// override default strcase.ToLowerCamel behaviour
+			strcase.ConfigureAcronym("enable-ui", "enableUI")
```
dnephin, line 51: Doesn't this need to be called before `parseOptions`? I thought that's where `strcase` was used.

dnephin (commented):
I believe it already works to enable the UI from config using:

```yaml
ui:
  enabled: true
```

https://github.com/infrahq/infra/blob/main/internal/cmd/server_test.go#L123-L125

dnephin (approved):
LGTM, thanks!

Do we expect people to use the UI proxy ? We might consider removing it from the example config  if it's mostly for dev. But also seems fine to leave it in.

PR #1721, opened 2022-04-26
fix: user list is sorted on create and update

### LINKED ISSUE #1711: ui: changing cluster role causes row ordering to change
Screen Shot 2022-04-25 at 9 40 51 PM

hoyyeva:
## Summary
- the cluster share user list is sorted by user id, so the list will not reorder when update the role
- change text grant to share

## Related Issues
Resolves #1711

### DIFF
--- a/ui/components/grant.js
+++ b/ui/components/grant.js
@@ -107,9 +107,10 @@ export default function ({ id, modalOpen, handleCloseModal }) {
     }, { optimisticData: list.filter(item => item?.id !== grantId) })
   }
 
+
   return (
     <InfoModal
-      header='Grant'
+      header='Share'
       handleCloseModal={handleCloseModal}
       modalOpen={modalOpen}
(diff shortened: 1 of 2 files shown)


ui/components/grant.js
```diff
@@ -8,11 +8,14 @@ import ErrorMessage from '../components/error-message'
 import InfoModal from './modals/info'
 
 function Grant ({ id }) {
-  const { data: user } = useSWR(`/v1/identities/${id.replace('i:', '')}`, { fallbackData: { name: '' } })
-
-  return (
-    <p>{user.name}</p>
-  )
+  if (id) {
```
jmorganca, line 11: Would fast-return here:

```
if (!id) {
  return null
}
```

jmorganca (approved):

jmorganca (approved):

PR #1725, opened 2022-04-26
fix: only use color logging with a terminal

### LINKED ISSUE #1719: cli: do not print colors if in non-interactive
In non-interactive environments infra shouldn't print color codes.

Example:

```
Command failed: /opt/homebrew/bin/infra tokens add [31mERROR[0m Session expired. Run 'infra login' to start a new session.
```

dnephin:
## Summary

Only use color when stdin is a terminal. Tested with

```
./infra id list < /dev/null
```

## Related Issues

Resolves #1719


BruceMacD (approved):
👍

PR #1727, opened 2022-04-26
improve: generalize oidc client config

### LINKED ISSUE #1703: Generalize OIDC Client Configuration
OIDC Provider clients make some assumptions about the endpoints they should be connecting based on the Okta endpoints. 

We can get this information from the OIDC .well-known endpoint instead. 

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

BruceMacD:
- set auth URL from provider well known endpoint
- make user info endpoint dynamic

## Summary
OIDC Provider clients make some assumptions about the endpoints they should be connecting based on the Okta endpoints. We can prepare for general OIDC by automatically finding these endpoints at the `.well-known/openid-configuration` endpoint that OIDC providers publicly expose.

This change does not have any impact on the current user experience or backwards compatibility.

## Checklist

## Related Issues

Resolves #1703

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -14,6 +14,7 @@ import (
 	survey "github.com/AlecAivazis/survey/v2"
 	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/cli/browser"
+	"github.com/coreos/go-oidc/v3/oidc"
 	"github.com/goware/urlx"
 	"github.com/muesli/termenv"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


internal/cmd/login.go
```diff
@@ -225,13 +226,44 @@ func updateInfraConfig(client *api.Client, loginReq *api.LoginRequest, loginRes
 }
 
 func oidcflow(host string, clientId string) (string, error) {
+	// find out what the authorization endpoint is
```
BruceMacD, line 229: We can use this library to automatically find the authorization endpoint that the CLI should open in the browser for the provider. It uses the `.well-known/oidc` endpoint to do this.

internal/cmd/login.go
```diff
@@ -225,13 +226,44 @@ func updateInfraConfig(client *api.Client, loginReq *api.LoginRequest, loginRes
...
+	}
+
+	if err := provider.Claims(&claims); err != nil {
+		return "", fmt.Errorf("parsing claims: %w", err)
+	}
+
+	scopes := []string{"openid", "email"} // openid and email are required scopes for login to work
+
+	// we want to be able to use these scopes to access groups, but they are not needed
+	wantScope := map[string]bool{
```
BruceMacD, line 247: making these scopes optional will allow us to login using IDPs that do not support the groups claim

internal/server/authn/oidc.go
```diff
@@ -132,10 +122,10 @@ func (o *oidcImplementation) ExchangeAuthCodeForProviderTokens(code string) (raw
 	}
 
 	if err := idToken.Claims(&claims); err != nil {
-		return "", "", time.Time{}, "", fmt.Errorf("id cliams: %w", err)
+		return "", "", time.Time{}, "", fmt.Errorf("id claims: %w", err)
 	}
 
-	return rawAccessToken, rawRefreshToken, exp, claims.Email, nil
+	return rawAccessToken, rawRefreshToken, exchanged.Expiry, claims.Email, nil
```
BruceMacD, line 128: Some IDPs do not return the access token as a JWT, so we can rely on this response expiry instead of checking the actual JWT instead

internal/server/handlers.go
```diff
@@ -623,20 +623,17 @@ func (a *API) UpdateIdentityInfoFromProvider(c *gin.Context) error {
...
-
-			deleteAuthCookie(c)
-		}
-
 		if errors.Is(err, context.DeadlineExceeded) {
 			return fmt.Errorf("%w: %s", internal.ErrBadGateway, err.Error())
 		}
 
-		return fmt.Errorf("update user info: %w", err)
+		if nestedErr := access.DeleteAllIdentityAccessKeys(c); nestedErr != nil {
```
BruceMacD, line 630: Err on the side of caution on update errors. Revoke sessions when something goes wrong.

mxyng (approved):

PR #1728, opened 2022-04-26
Use CLI.Stdout more consistently

dnephin:
## Summary

Follow up to #1708

This PR removes the `TODO` type, and uses the new `CLI.Stdout` and `CLI.Stderr` in most places. There are a few more calls to `fmt.Printf` which will require a bit more work to change. We can probably address those when we write tests that require it. I'm hoping this passes the `CLI` struct to enough places to make it easy to use for new commands or new output.


mxyng (approved):

mxyng:
It'd be nice to use `CLI` as a central point for formatting outputs, both user outputs and informational outputs currently outputing on stderr

PR #1729, opened 2022-04-26
feat(cmd): optional access key name

### LINKED ISSUE #1479: `key` name should be optional

Giving names to keys is a nice to have and provides a reader friendly way of identifying different keys but it should not be a requirement. Users should always be able to reference keys, and other resources, by ID.

Instead of `infra keys add KEY_NAME IDENTITY_NAME` it should be `infra keys add IDENTITY_NAME [KEY_NAME]`

mxyng:
## Summary

Move name positional parameter to a keyword parameter for infra keys add. If name is not specified, generate one based on the identity issuing the key and the current timestamp

Move `infraProviderCache` and `infraConnectorIdentityCache` to `data/data.go` and a way to reset the cache value. This is mainly used for tests as the database gets recreated frequently which causes the cache to point to a nonexistent object

## Checklist

## Related Issues

Resolves #1479

### DIFF
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -41,6 +41,10 @@ func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey, identityID uid
 		return "", fmt.Errorf("get access key identity: %w", err)
 	}
 
+	if accessKey.Name == "" {
+		accessKey.Name = fmt.Sprintf("%s-%s", identity.Name, time.Now().UTC().Format("2006-01-02T15:04:05.000000"))
+	}
+
 	if identity.Kind != models.MachineKind {
 		// direct access key creation isn't supported for user identities yet
 		return "", internal.ErrNotImplemented
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -34,6 +35,7 @@ func newKeysCmd(cli *CLI) *cobra.Command {
 }
 
 type keyCreateOptions struct {
+	Name              string
 	TTL               time.Duration
 	ExtensionDeadline time.Duration
 }
@@ -42,20 +44,21 @@ func newKeysAddCmd(cli *CLI) *cobra.Command {
 	var options keyCreateOptions
 
 	cmd := &cobra.Command{
-		Use:   "add KEY IDENTITY",
+		Use:   "add IDENTITY",
 		Short: "Create an access key",
 		Long:  `Create an access key. Only machine identities are supported at this time.`,
 		Example: `
 # Create an access key named 'example-key' that expires in 12 hours
 $ infra keys add example-key machine-a --ttl=12h
 `,
-		Args: ExactArgs(2),
+		Args: ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			keyName := args[0]
-			machineName := args[1]
+			machineName := args[0]
 
-			if strings.Contains(keyName, " ") {
-				return fmt.Errorf("key name cannot contain spaces")
+			if options.Name != "" {
+				if strings.Contains(options.Name, " ") {
+					return fmt.Errorf("key name cannot contain spaces")
+				}
 			}
 
 			client, err := defaultAPIClient()
@@ -70,7 +73,7 @@ $ infra keys add example-key machine-a --ttl=12h
 
 			resp, err := client.CreateAccessKey(&api.CreateAccessKeyRequest{
 				IdentityID:        machine.ID,
-				Name:              keyName,
+				Name:              options.Name,
 				TTL:               api.Duration(options.TTL),
 				ExtensionDeadline: api.Duration(options.ExtensionDeadline),
 			})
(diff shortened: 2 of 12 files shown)


internal/cmd/keys.go
```diff
@@ -114,11 +122,20 @@ func newKeysRemoveCmd() *cobra.Command {
...
 			if err != nil {
 				return err
 			}
 
+			issuedFor := key.IssuedForName
+			if issuedFor == "" {
+				issuedFor = key.IssuedFor.String()
+			}
+
+			fmt.Fprintf(os.Stderr, "Deleted access key %q issued for %q\n", key.Name, issuedFor)
```
mxyng, line 137: This isn't related to optional access key name but remove doesn't currently have any outputs

internal/cmd/keys.go
```diff
@@ -42,19 +44,22 @@ func newKeysAddCmd() *cobra.Command {
...
-			keyName := args[0]
-			machineName := args[1]
+			machineName := args[0]
 
-			if strings.Contains(keyName, " ") {
+			if options.Name == "" {
+				options.Name = fmt.Sprintf("%s-%s", machineName, time.Now().Format("20060102150405.000000"))
+			}
+
+			if strings.Contains(options.Name, " ") {
```
dnephin, line 62: Should the defaulting of the name be part of the API? Would other API consumers want this same behaviour?

I suspect the validation of the name should be part of the API.

dnephin (commented):
Nice! Just one question about CLI vs API for this behaviour

internal/cmd/keys.go
mxyng: That's a good question. I was on the fence but I think there's a good case for moving name generation to the API

internal/cmd/keys.go
```diff
@@ -42,19 +44,22 @@ func newKeysAddCmd() *cobra.Command {
...
-		Args: ExactArgs(2),
+		Args: ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			keyName := args[0]
-			machineName := args[1]
+			machineName := args[0]
 
-			if strings.Contains(keyName, " ") {
+			if options.Name == "" {
+				options.Name = fmt.Sprintf("%s-%s", machineName, time.Now().Format("20060102150405.000000"))
```
ssoroka, line 59: I was hoping for names in the format of <adjective>-<noun> :D

ssoroka (commented):
I'm worried this change encourages users not to put a descriptive name for the keys, then they will have no idea what these keys are for.

mxyng:
> I'm worried this change encourages users not to put a descriptive name for the keys, then they will have no idea what these keys are for.

I think that's an issue regardless if we generate a name or not. It's still possible to create a key with a descriptive name if user wants through `--name` but forcing users to name the access key will result in many nondescriptive names

internal/cmd/keys.go
mxyng: I'm not a fan of pet names. It might be more reader friendly but it's not very descriptive. This format was picked because it conveys some information to the user if they elect to not give the key a more descriptive name

internal/access/access_key.go
```diff
@@ -41,6 +41,10 @@ func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey, identityID uid
 		return "", fmt.Errorf("get access key identity: %w", err)
 	}
 
+	if accessKey.Name == "" {
+		accessKey.Name = fmt.Sprintf("%s-%s", identity.Name, time.Now().Format("2006-01-02-15:04:05.000000"))
```
dnephin, line 45: This format looks like it is almost `time.RFC3339` except that it uses `-` instead of `T` to separate the date and time, and doesn't have a time zone.

Excluding the timezone seems like a nice way of keeping it more readable, but maybe we should call `.UTC()` on the time first, to make sure the timestamp is in UTC?

I'm also wondering if we should use the `T` to separate the two parts instead of the `-`. 

(Not blocking)

internal/cmd/keys.go
```diff
@@ -70,21 +73,24 @@ $ infra keys add example-key machine-a --ttl=12h
...
 				ExtensionDeadline: api.Duration(options.ExtensionDeadline),
 			})
 			if err != nil {
 				return err
 			}
 
-			cli.Output("key: %s", resp.AccessKey)
+			fmt.Fprintf(os.Stderr, "Created access key for %q\n", machineName)
+			fmt.Printf("Name: %s\n", resp.Name)
+			fmt.Printf("Key: %s\n", resp.AccessKey)
```
dnephin, line 86: So that we can capture these in tests:

```suggestion
			fmt.Fprintf(cli.Stderr, "Created access key for %q\n", machineName)
			cli.Output("Name: %s", resp.Name)
			cli.Output("Key: %s", resp.AccessKey)
```

internal/server/server_test.go
```diff
@@ -40,6 +41,8 @@ func setupServer(t *testing.T, ops ...func(*testing.T, *Options)) *Server {
 	err = s.loadConfig(s.options.Config)
 	assert.NilError(t, err)
 
+	data.InvalidateCache()
```
dnephin, line 44: This should be part of cleanup? So that we can use the cache while the test runs, but we clean it up when the test ends?

```suggestion
	t.Cleanup(data.InvalidateCache)
```

Or maybe we want it to do both (clear in initially in case some other test did not, and also using `t.Cleanup` to clean up properly after itself) ?

dnephin (approved):
LGTM! Nice API tests!

Some minor thoughts, but nothing blocking a merge.

internal/server/server_test.go
mxyng: Calling this in the setup is great to ensure the state is as expected before the test runs. Calling is cleanup is fine if all other tests are well behaving but leaves more room for errors from side effects of running other tests. In this case, I'll add `InvalidateCache` to both setup and cleanup

dnephin (approved):
:ship:

internal/server/config.go
```diff
@@ -844,7 +844,7 @@ func (s Server) loadAccessKey(db *gorm.DB, identity *models.Identity, provider *
 		}
 
 		accessKey := &models.AccessKey{
-			Name:       fmt.Sprintf("%s-%s", identity.Name, time.Now().Format("20060102150405")),
+			Name:       fmt.Sprintf("%s-%s", identity.Name, time.Now().UTC().Format("2006-01-02T15:04:05.000000")),
```
mxyng, line 847: Making this change so the name generated by API and config are consistent. This also fixes a race where the same name can be generated if an operator decides to import two access keys for the same user

PR #1738, opened 2022-04-26
Update snowflake dependency and add more tests

dnephin:
## Summary

While working on a change to our API, I ran into some unexpected behaviour. I traced it to [this bug fix](https://github.com/bwmarrin/snowflake/commit/7511dd2b95883fdcd3fd35494451124fdc9b06cb), which we don't have yet because there has not been a release of this project for 3 years.

This PR adds some more tests, and updates the dependency.

### DIFF
--- a/uid/snowflake_test.go
+++ b/uid/snowflake_test.go
@@ -2,7 +2,6 @@ package uid_test
 
 import (
 	"encoding/json"
-	"fmt"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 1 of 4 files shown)


uid/snowflake_test.go
```diff
@@ -25,26 +24,51 @@ func TestJSONCanUnmarshal(t *testing.T) {
...
+		},
+		{
+			id:          "npL6MjP8Qfd", // 0x7fffffffffffffff + 1
+			expectedErr: `invalid base58 id "npL6MjP8Qfd"`,
+		},
+		{
+			id:          "JPwcyDCgEuqJPwcyDCgEuq",
+			expectedErr: `invalid base58 id "JPwcyDCgEuqJPwcyDCgEuq"`,
+		},
+		// does not result in an error, but probably should
```
jmorganca, line 58: Should we keep this commented out?

jmorganca (approved):

ssoroka (approved):

PR #1739, opened 2022-04-26
api: remove /v1/introspect for now

### LINKED ISSUE #1425: api: replace `GET /introspect` with `GET /identities/self`
The `/introspect` API endpoint is quite similar to `GET /identities/:id`. I think we could remove it by accepting a `self` keyword in place of the ID. That would allow a user to inspect their identity.

We may need to prevent anyone from creating an `Identity`  with an `ID` of `self` (if we don't already).

dnephin:
Branched from #1738, that will need to merge first
 
## Summary

This PR writes a bunch of tests for `GetIdentity` and adds support for `/v1/identities/self` to query for details about the current identity. By supporting `self` we remove the need for `/v1/introspect` for now, because all it returns is less information about the identity.

It is quite possible that in the future we will re-introduce an endpoint similar to `/v1/introspect`. By removing it now we won't have to worry about changing the structure of the response. We'll have more information about the use case it is intended to solve, which lets us design the endpoint from scratch.

## Checklist

## Related Issues

Resolves #1425


internal/server/server_test.go
```diff
@@ -197,6 +197,9 @@ func TestSignupEnabled(t *testing.T) {
 }
 
 func TestServer_Run(t *testing.T) {
+	if testing.Short() {
```
jmorganca, line 200: How "slow" is this?

jmorganca, line 200: How "slow" is this?

jmorganca (approved):

docs/api/openapi3.json
```diff
@@ -2203,14 +2187,14 @@
...
         "parameters": [
           {
-            "example": "4yJ3n3D8E2",
             "in": "path",
             "name": "id",
             "required": true,
             "schema": {
+              "description": "a uid or the literal self",
               "example": "4yJ3n3D8E2",
               "format": "uid",
```
BruceMacD, line 2196: I think this format specifier should be `"uid|self" too

BruceMacD (approved):

docs/api/openapi3.json
dnephin: I'm not exactly sure how to represent a description of a format vs a literal, I'll use `uid|self` for now and we can figure it out later.

jmorganca (approved):

PR #1740, opened 2022-04-26
api: replace c.Set(path) with FullPath

dnephin:
## Summary

This PR replaces `c.Set("path", ...)` with `c.FullPath()`, which should be the same string.

Also removes some path joining logic, and replaces it with `path.Join`.  `gin`'s `FullPath` is constructed with [joinPaths](https://github.com/gin-gonic/gin/blob/master/utils.go#L128-L138), which is mostly `path.Join` with some special cases for adding a trailing `/`, and handling empty relative paths. We don't have either of those problems here because we don't use trailing slash, and the router always starts with a base path of `/`.

The intermediate commit has a test that shows that `path.Join` works the way we want.


internal/server/routes.go
```diff
@@ -155,7 +155,6 @@ func post[Req, Res any](a *API, r *gin.RouterGroup, path string, handler ReqResH
 	fullPathStr := fullPath(r, path)
 
 	r.POST(path, func(c *gin.Context) {
-		c.Set("path", fullPathStr)
```
dnephin, line 158: These are the calls I was removing

internal/server/server_test.go
```diff
@@ -197,6 +197,9 @@ func TestSignupEnabled(t *testing.T) {
 }
 
 func TestServer_Run(t *testing.T) {
+	if testing.Short() {
```
jmorganca, line 200: Out of curiosity, how slow are these?

jmorganca (approved):

internal/server/server_test.go
dnephin: Seems to be anywhere from 2s to 3.5s. All the time is spent in the HTTPS sub test, and this got a lot slower recently. I suspect the time is generating private keys for the CA.  We may be able to speed up this test by generating the private key and certs ahead of time,  including them in git under `./testdata/`, and having the server use those.

internal/server/routes.go
ssoroka: used in

		a.t.Event(c, "error", Properties{
			"code": resp.Code,
			"path": c.GetString("path"),
		})

ssoroka (changes requested):

internal/server/routes.go
dnephin: Ahhh, I did not realize there were multiple ways of getting values out of context!  I was searching for `c.Get("path")`.

That's good to know.

dnephin:
I'll move the tests out of this PR, and change this to explore a different approach

dnephin:
Ok, I've updated this PR to use `c.FullPath` from `sendAPIError`, so that we don't need the context variable. Also removed some other path joining so that we can do it in fewer places.

internal/server/routes.go
jmorganca: Good catch @ssoroka !

mxyng (approved):

ssoroka (approved):

PR #1745, opened 2022-04-27
improve: validate provider fields

### LINKED ISSUE #1715: api: validate okta information on submit
Currently we don't validate the Okta `url`, `clientID`, `clientSecret` and other information when users add Okta.

### LINKED ISSUE #1736: api: validate `POST /v1/providers` `url` field

BruceMacD:
- check provider url and client are valid
- return better error message for a secret that does not exist

## Summary
When an identity provider is added to Infra through the API or CLI check that the URL, client ID and client secret are valid. This relies on some information being present in the response from the identity provider. We will probably iterate on this as more identity providers are supported.

Example outputs:
```
$ infra providers add okta --url invalid-example.okta.com --client-id ${CLIENT_ID} --client-secret ${CLIENT_SECRET}
ERROR  POST "/v1/providers" responded 400: bad request: invalid provider url

$ infra providers add okta --url example.okta.com --client-id ${INVALID_CLIENT_ID} --client-secret ${CLIENT_SECRET}
ERROR  POST "/v1/providers" responded 400: bad request: invalid client id

$ infra providers add okta --url example.okta.com --client-id ${CLIENT_ID} --client-secret ${INVALID_CLIENT_SECRET}
ERROR  POST "/v1/providers" responded 400: bad request: invalid client secret

$ infra providers add okta --url example.okta.com --client-id ${CLIENT_ID} --client-secret kubernetes:${CLIENT_SECRET}
ERROR  POST "/v1/providers" responded 400: bad request: client secret not found
```

## Checklist

## Related Issues

Resolves #1715 
Resolves #1736

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -41,9 +41,9 @@ func checkError(status int, body []byte) error {
 	case http.StatusNotFound:
 		return ErrNotFound
 	case http.StatusBadRequest:
-		return fmt.Errorf("%w: %s", ErrBadRequest, apiError.Message)
+		return fmt.Errorf(apiError.Message)
 	case http.StatusBadGateway:
-		// this errror should be displayed to the user so they can see its an external problem
+		// this error should be displayed to the user so they can see its an external problem
 		return fmt.Errorf(apiError.Message)
 	case http.StatusInternalServerError:
 		return ErrInternal
--- a/internal/server/authn/errors.go
+++ b/internal/server/authn/errors.go
@@ -1,0 +1,9 @@
+package authn
+
+import "fmt"
+
+var (
+	ErrInvalidProviderURL          = fmt.Errorf("invalid provider url")
+	ErrInvalidProviderClientID     = fmt.Errorf("invalid client id")
+	ErrInvalidProviderClientSecret = fmt.Errorf("invalid client secret")
+)
--- a/internal/server/authn/errors.go
+++ b/internal/server/authn/errors.go
@@ -2,12 +2,11 @@ package authn
 
 import (
 	"fmt"
-
-	"github.com/infrahq/infra/internal"
 )
 
 var (
-	ErrInvalidProviderURL          = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider url")
-	ErrInvalidProviderClientID     = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider client id")
-	ErrInvalidProviderClientSecret = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider client secret")
+	ErrValidation                  = fmt.Errorf("validation failed")
+	ErrInvalidProviderURL          = fmt.Errorf("%w: invalid provider url", ErrValidation)
+	ErrInvalidProviderClientID     = fmt.Errorf("%w: invalid provider client id", ErrValidation)
+	ErrInvalidProviderClientSecret = fmt.Errorf("%w: invalid provider client secret", ErrValidation)
 )
--- a/internal/server/authn/errors.go
+++ b/internal/server/authn/errors.go
@@ -1,9 +1,13 @@
 package authn
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/infrahq/infra/internal"
+)
 
 var (
-	ErrInvalidProviderURL          = fmt.Errorf("invalid provider url")
-	ErrInvalidProviderClientID     = fmt.Errorf("invalid client id")
-	ErrInvalidProviderClientSecret = fmt.Errorf("invalid client secret")
+	ErrInvalidProviderURL          = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider url")
+	ErrInvalidProviderClientID     = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider client id")
+	ErrInvalidProviderClientSecret = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider client secret")
 )
--- a/internal/server/authn/oidc.go
+++ b/internal/server/authn/oidc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/coreos/go-oidc/v3/oidc"
@@ -22,6 +23,7 @@ type InfoClaims struct {
 }
 
 type OIDC interface {
+	Validate() error
 	ExchangeAuthCodeForProviderTokens(code string) (accessToken, refreshToken string, accessTokenExpiry time.Time, email string, err error)
 	RefreshAccessToken(providerUser *models.ProviderUser) (accessToken string, expiry *time.Time, err error)
 	GetUserInfo(providerUser *models.ProviderUser) (*InfoClaims, error)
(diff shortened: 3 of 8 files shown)


internal/server/authn/oidc_test.go
```diff
@@ -0,0 +1,14 @@
+package authn
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestValidateInvalidURL(t *testing.T) {
```
BruceMacD, line 9: Can't test an invalid client ID and client secret without relying on external URLs that could change, so skipped testing those for now.

internal/server/authn/errors.go
```diff
@@ -0,0 +1,9 @@
+package authn
+
+import "fmt"
+
+var (
+	ErrInvalidProviderURL          = fmt.Errorf("invalid provider url")
+	ErrInvalidProviderClientID     = fmt.Errorf("invalid client id")
+	ErrInvalidProviderClientSecret = fmt.Errorf("invalid client secret")
```
ssoroka, line 8: do we want to inherit from errors that we know how to render to the user?

BruceMacD: Good call-out, made this change:
https://github.com/infrahq/infra/pull/1745/commits/6c9c255725cabde8cc6436b4e0678bd26d0bea91

api/client.go
```diff
@@ -41,9 +41,9 @@ func checkError(status int, body []byte) error {
 	case http.StatusNotFound:
 		return ErrNotFound
 	case http.StatusBadRequest:
-		return fmt.Errorf("%w: %s", ErrBadRequest, apiError.Message)
+		return fmt.Errorf(apiError.Message)
```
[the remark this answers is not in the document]
dnephin: I think we'll eventually want to change how `checkError` works, to return an error that allows the caller to retrieve the underlying exit status in all cases. 

Right now we can in some cases, but not in others. And I guess this change makes it so that the caller can no longer determine if it was a 400 status or some other status.

internal/server/authn/errors.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"fmt"
+
+	"github.com/infrahq/infra/internal"
+)
+
+var (
+	ErrInvalidProviderURL          = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider url")
+	ErrInvalidProviderClientID     = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider client id")
+	ErrInvalidProviderClientSecret = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider client secret")
```
dnephin, line 12: I'm not sure how well this pattern will scale. Usually I would want packages like `authn` to return their own errors, and then the api layer would translate those errors into the appropriate status code. That way all of our domain logic packages don't need to import the `internal` package.

That might look something like:

```go
var ErrValidation = fmt.Errorf("validation failed")

var ErrInvalidProviderURL = fmt.Errorf("%w: invalid provider url", ErrValidation)

```

And then `sendAPIError` would translate `authn.ErrValidation` to a 400 status code.

But I guess that's a big change from our current pattern, so not something we need to change right now. I think this approach works well (except for the very minor problem you encountered with repeated messages in the error string, which is solved well enough in this PR).

internal/server/handlers.go
```diff
@@ -619,6 +626,16 @@ func (a *API) UpdateIdentityInfoFromProvider(c *gin.Context) error {
 	return access.UpdateUserInfoFromProvider(c, info, user, provider)
 }
 
+// validateProvider checks that a provider being modified is valid
+func (a *API) validateProvider(c *gin.Context, provider *models.Provider) error {
+	oidc, err := a.providerClient(c, provider, "http://localhost:8301")
```
dnephin, line 631: This url is ignored because validate doesn't look at it, is that right? Or is this url being used?

I was a bit surprised by it, so maybe a comment about it being ignored?

It'd be nice if we could avoid needing to use it. I'm not sure if that's possible. Maybe not worth the refactor right now.

internal/server/errors.go
```diff
@@ -41,10 +41,9 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
 		resp.Message = err.Error()
 	case errors.As(err, validationErrors):
 		resp.Code = http.StatusBadRequest
-		resp.Message = err.Error()
+		resp.Message = fmt.Sprintf("%s: %s", internal.ErrBadRequest, err)
```
dnephin, line 44: What's the reason for adding "bad request" to this message? Won't we end up with duplicate messages on the client side, because of the 400 status code?

internal/server/authn/oidc.go
```diff
@@ -43,6 +45,31 @@ func NewOIDC(domain, clientID, clientSecret, redirectURL string) OIDC {
...
+	ctx := context.Background()
+	conf, _, err := o.clientConfig(ctx)
+	if err != nil {
+		logging.S.Debugf("error validating oidc provider: %s", err)
+		return ErrInvalidProviderURL
+	}
+
+	_, err = conf.Exchange(ctx, "test-code") // 'test-code' is a placeholder for a valid authorization code, it will always fail
+	if err != nil {
+		if strings.Contains(err.Error(), "client_id") || strings.Contains(err.Error(), "client id") {
```
dnephin, line 59: I'm wondering if there is any way we could be a little more strict about this. I guess we'll eventually need to test with a bunch of oidc providers to make sure they return the response we expect.

If I'm reading the code correctly, I think the errors we care about from `conf.Exchange` will be https://pkg.go.dev/golang.org/x/oauth2#RetrieveError.

To be a little more strict, we could do something like this:

```go
var errRetrieve *oauth2.RetrieveError
if errors.As(err, &errRetrieve) {
    ...
}
```

That would give us access to both the `http.Request` (to check if the error code is 400), and to the body of the response. Any chance the body of this response is JSON that we could parse? Even if not, checking `strings.Contains(string(errRetrieve.Body)), "client_id")` is a bit more strict.

dnephin (commented):
Overall this approach makes sense to me. This is my first time reading over the oidc code, so mostly questions from me, to make sure I understand

internal/server/authn/oidc.go
```diff
@@ -43,6 +45,31 @@ func NewOIDC(domain, clientID, clientSecret, redirectURL string) OIDC {
...
+func (o *oidcImplementation) Validate() error {
+	ctx := context.Background()
+	conf, _, err := o.clientConfig(ctx)
+	if err != nil {
+		logging.S.Debugf("error validating oidc provider: %s", err)
+		return ErrInvalidProviderURL
+	}
+
+	_, err = conf.Exchange(ctx, "test-code") // 'test-code' is a placeholder for a valid authorization code, it will always fail
+	if err != nil {
```
mxyng, line 58: Is it expected to return valid if error is neither client ID or secret related? Are there other errors that could be returned? Might be a good idea to explicitly check invalid authorization code and return nil and error otherwise?

internal/server/authn/errors.go
```diff
@@ -0,0 +1,13 @@
+package authn
+
+import (
+	"fmt"
+
+	"github.com/infrahq/infra/internal"
+)
+
+var (
+	ErrInvalidProviderURL          = fmt.Errorf("%w: %s", internal.ErrBadRequest, "invalid provider url")
```
mxyng, line 10: I appreciate the consistent formatting but the second format string isn't needed

```suggestion
	ErrInvalidProviderURL          = fmt.Errorf("%w: invalid provider url", internal.ErrBadRequest)
```

internal/server/errors.go
mxyng, line 44: Would this output `Bad Request: Bad Request: ...` if `err` is an authn error?

internal/server/authn/errors.go
BruceMacD: Good suggestion, I changed to using a base `ErrValidation` in the authn package

internal/server/handlers.go
BruceMacD: This was confusing, updated to an empty redirect URL and a comment

internal/server/errors.go
BruceMacD: Removed this

internal/server/authn/oidc.go
BruceMacD: Added this strict error check

internal/server/authn/oidc.go
BruceMacD: The error message for an invalid auth code varies a lot by OIDC provider so I'm leaving this one on the permissive side for now

internal/server/errors.go
BruceMacD: Removed this

internal/server/authn/errors.go
BruceMacD: Made this fix

api/client.go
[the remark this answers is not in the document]
dnephin: I put together #1766 to try and address this.

dnephin (approved):
Nice! Looks great

Issue #1657, opened 2022-04-19
Consolidate user and machine identities into a single entity


mxyng:
User and machine identities are similar in many ways but differ in a few key areas, namely authentication.

Users:
- Requires an email (`local@domain`)
- Email + password
- No access key

Machines:
- Requires a name (`^[a-zA-Z0-9-_./]+$`)
- Access key
- No email + password

### Proposal

Both users and machines already exist as a single abstract resource called identity. Instead, identity should be the sole resource representing an identity without user or machine subtypes.

- Identifier can be either an email (`local@domain`) or a name (`^[a-zA-Z0-9-_./]+$`)
- May or may not have a password. No password set disables login with email + password
- Can create and use access keys

PR #1746, opened 2022-04-27
api: more tests for CreateIdentity and DeleteProvider

dnephin:
Extracted out of #1740, since that PR needs more work


BruceMacD (approved):
Some of this behavior is going to change soon with #1657, but I'll tweak this in that branch. This looks good.

PR #1750, opened 2022-04-27
fix: loading connector.Options

dnephin:
## Summary

Flag default values were overwriting the values from the config. Existing tests did not capture that case apparently.

A consequence of this change is that default values for flags won't do anything for the server or connector. We can probably address that in a follow up to make it an error to set a non-zero default value for flags, or handle it in some other way.

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -12,8 +12,10 @@ import (
 	"time"
 
 	"gotest.tools/v3/assert"
+	"gotest.tools/v3/fs"
 
 	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/connector"
 	"github.com/infrahq/infra/uid"
 )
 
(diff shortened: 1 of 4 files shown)


internal/cmd/cmd_test.go
```diff
@@ -229,3 +231,39 @@ func newTestClientConfig(srv *httptest.Server, identity api.Identity) ClientConf
...
+	var actual connector.Options
+	patchRunConnector(t, func(ctx context.Context, options connector.Options) error {
+		actual = options
+		return nil
+	})
+
+	content := `
+name: the-name
+accessKey: file:/var/run/secrets/key
+server: the-server
```
dnephin, line 245: I'll include more fields, and maybe fix in some flags as well in more tests cases, as a follow up.

mxyng:
Decided to revert for now. Details are in https://github.com/infrahq/infra/pull/1751

PR #1753, opened 2022-04-27
improve!: remove the /v1/introspect endpoint

dnephin:
The endpoint was removed in https://github.com/infrahq/infra/pull/1739.

BREAKING CHANGE: removes the GET /v1/introspect endpoint. Use /v1/identities/self


jmorganca (approved):

dnephin:
commit-lint is happy, so I will merge this

mxyng (approved):

PR #1754, opened 2022-04-27
maintain: run all tests in CI

dnephin:
## Summary

Removes the -short flag from CI, so that all tests are run. Previously we were using it to skip tests that required docker containers and kubernetes.

Also update the version of gotestsum, which should use color by default.


dnephin:
The test flake is unrelated. I've seen it before locally as well.  I'll see if I can figure out why this flakes before re-running the tests.

I'm not sure why `lastSeenAt` would be nil.

jmorganca (approved):

PR #1757, opened 2022-04-27
maintain: stateless signupEnabled

### LINKED ISSUE #1676: Signup check should be stateless
Currently, Infra stores a `Settings.SignupEnabled` boolean to in the database to [determine if signup is enabled](https://github.com/infrahq/infra/blob/main/internal/server/server.go#L594) but it could be checked in a stateless manner in the `GET /v1/signup` handler without storing state.

mxyng:
## Summary

Remove SignupEnable flag in Settings. Instead check state when necessary. Include deleted resources as well when looking for identities, providers, and grants

## Checklist

## Related Issues

Resolves #1676


internal/server/data/data.go
```diff
@@ -179,16 +179,16 @@ func deleteAll[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) error
...
+func Count[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (int64, error) {
 	db2 := db
 	for _, selector := range selectors {
 		db2 = selector(db2)
 	}
 
 	var count int64
-	if err := db.Model((*T)(nil)).Count(&count).Error; err != nil {
-		return nil, err
+	if err := db2.Model((*T)(nil)).Count(&count).Error; err != nil {
```
mxyng, line 189: db2 wasn't used before so selectors were not applied. Haven't run into this previously since most calls to `Count` don't use selectors

internal/server/handlers.go
```diff
@@ -479,6 +481,8 @@ func (a *API) SignupEnabled(c *gin.Context, _ *api.EmptyRequest) (*api.SignupEna
 }
 
 func (a *API) Signup(c *gin.Context, r *api.SignupRequest) (*api.Identity, error) {
+	c.Set("enableSignup", a.server.options.EnableSignup)
```
jmorganca, line 484: Could we quick-return false here instead of setting the context?

internal/server/handlers.go
```diff
@@ -468,6 +468,8 @@ func (a *API) DeleteGrant(c *gin.Context, r *api.Resource) error {
 }
 
 func (a *API) SignupEnabled(c *gin.Context, _ *api.EmptyRequest) (*api.SignupEnabledResponse, error) {
+	c.Set("enableSignup", a.server.options.EnableSignup)
```
jmorganca, line 471: Could we quick-return false here instead of setting the context?

internal/server/handlers.go
mxyng: That's fine too. I decided to take the current approach as it consolidates the checking into a single place

internal/access/signup.go
```diff
@@ -7,35 +7,64 @@ import (
...
-	settings, err := data.GetSettings(db)
+	enableSignupVar, ok := c.Get("enableSignup")
+	if !ok {
+		return false, nil
+	}
+
+	enableSignup, ok := enableSignupVar.(bool)
+	if !ok {
+		return false, nil
+	}
```
dnephin, line 31: I learned yesterday there were more ways of getting things out of the `gin.Context`. I think this could use:

```suggestion
	enableSignup := c.GetBool("enableSignup")
```

Although avoiding using the context for this might also be nice (as Jeff suggested in another comment).

internal/server/data/data.go
```diff
@@ -179,16 +179,16 @@ func deleteAll[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) error
 	return db2.Delete(new(T)).Error
 }
 
-func Count[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*int64, error) {
+func Count[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (int64, error) {
```
dnephin, line 182: Nice! No reason to return a pointer here

internal/server/handlers.go
dnephin: Not having to use context for this would be nice.  

Not blocking, but I think if the call to `access.SignupEnabled` happened here (instead of from `access.Signup`), we could either pass in the bool as an argument, or do the early return.

internal/server/models/settings.go
```diff
@@ -5,6 +5,4 @@ type Settings struct {
 
 	PrivateJWK EncryptedAtRestBytes
 	PublicJWK  []byte
-
-	SignupEnabled bool
```
dnephin, line 9: How does GORM deal with deleted fields? Does this need any kind of migration?

dnephin (commented):
LGTM!

Mostly minor style things. My main question is about how GORM deals with removing fields. I'm guessing it's fine, but wanted to make sure.

internal/server/data/data.go
mxyng: That was in the original version of selectors but if I recall there were some weird side effects which is why we're using a temporary var for the db here

internal/server/models/settings.go
mxyng: I expected it to drop the column but it looks like it doesn't so will add a migration instead

BruceMacD (approved):
Looks good. Makes sense to me.

dnephin (approved):
LGTM! We'll probably want some API tests for these cases at some point (not necessarily every permutation, but maybe one case for signing enabled being disallowed, and one when it's allowed.  
The tests of the access function give us some reasonable confidence in this for now.

PR #1760, opened 2022-04-28
improve: user facing errors and formatting

kimskimchi:
## Checklist

## Related Issues

Resolves #


internal/cmd/login.go
```diff
@@ -525,12 +600,16 @@ func promptSkipTLSVerify() error {
...
+			UserFacingMessage: "Non-interactive login requires the [SERVER] argument.",
+		}
 	}
 
 	config, err := readOrCreateClientConfig()
 	if err != nil {
-		return "", err
+		return "", UserFacingError{
+			Underlying: err,
+		}
```
dnephin, line 612: I think it would be nice to avoid having to wrap errors with `UserFacingError` if we aren't going to add a `UserFacingMessage`.

Instead of this wrapping , we could change `main()`. to use `Internal Error` instead of `Unhandled error`.

internal/cmd/errors_userfacing.go
```diff
@@ -0,0 +1,75 @@
...
+
+// UserFacingError wraps an error to provide better error messages to users.
+// Any non-UserFacingError will be printed as "unhandled errors".
+type UserFacingError struct {
+	// original error
+	Underlying error
+
+	// formatted outputs that users see when running the CLI commands
+	// should be a readable sentence rather than a stacktrace
+	UserFacingMessage string
```
dnephin, line 17: Godoc comments always start with the name of the symbol they are documenting, ex:

```
// UserFacingMessage ...
UserFacingMessage string
```

Maybe we can shorten this field name to `Message`, since the type is already called `UserFacingError` ?

```suggestion
	// Message is a human readable error message that is added to Underlying to
	// provider users with more guidance about how to solve the problem.
	Message string
```

dnephin (commented):
Overall I like this approach! An error type gives us good control over what the users sees, and makes it clear from reading the code the purpose of the error.

Just a couple suggestions for how we might make this type a bit easier to use.

main.go
```diff
@@ -13,9 +13,15 @@ import (
 
 func main() {
 	if err := cmd.Run(context.Background(), os.Args[1:]...); err != nil {
-		if !errors.Is(err, terminal.InterruptErr) {
-			logging.S.Error(err.Error())
+		var userErr cmd.UserFacingError
+		switch {
+		case errors.Is(err, terminal.InterruptErr):
+		case errors.As(err, &userErr):
```
BruceMacD, line 19:
```suggestion
		case errors.Is(err, terminal.InterruptErr), errors.As(err, &userErr):
```
I think you can do both of these cases in one line

dnephin: I think that changes the behaviour. Right now we're printing nothing for the interruptErr.

I guess a comment on that line would make it more obvious?

```go
case errors.Is(err, terminal.InterruptErr):
    // print nothing, the user initiated the exit
```

mxyng: Golang doesn't have the same semantics for switch-case as other languages in that fallthrough is not implied. You need to add `fallthrough` to `case errors.Is(err, terminal.InterruptErr):` to mimic the behaviour from other languages. In this case, Bruce's suggestion will change behaviour. Not sure if this was your original intention.

internal/cmd/errors_userfacing.go
mxyng: +1 on the field name though I think even the type can be shortened to `Error` since it exists in the cmd package

internal/cmd/errors_userfacing.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// show or hide Underlying error from outputting with the UserFacingMessage
+	ShowUnderlying bool
```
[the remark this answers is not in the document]
mxyng: +1 though it seems `Underlying` should not be nil since it's needed for `Unwrap`. Instead, if we need to print the raw error, just don't set a message

internal/cmd/errors_userfacing.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (u UserFacingError) Error() string {
```
mxyng, line 23: This seems unnecessarily complex. It should probably output the Message if it exists or the raw Error if it does not

internal/cmd/errors_userfacing.go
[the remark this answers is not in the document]
kimskimchi: I think I over engineered it for now. It was for anytime when we expect an error (eg. we wouldn't need to print 404 not found, just that 'something wasn't found' on the cli). 
but I agree as well that it's better if we just set underlying as nil and log (debug) it instead, since wherever ShowUnderlying:false was to be used, it's an expected and valid workflow. if user really wants to debug, then they still can.

internal/cmd/login.go
kimskimchi: will do!

api/client.go
```diff
@@ -52,7 +52,7 @@ func checkError(status int, body []byte) error {
 	}
 
 	if status >= 400 {
-		return errors.New(http.StatusText(status))
+		return fmt.Errorf("%w: %s: %s", ErrUnexpected, http.StatusText(status), apiError.Message)
```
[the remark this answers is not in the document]
kimskimchi: I asked Steven about this, and the expected ones will be translated in `api.Errors.go`, while rest is 'unexpected'. 
If we do >=500, what do you think is a better name for the rest of 400s? `ErrClient`?

internal/cmd/cmd.go
```diff
@@ -172,7 +172,11 @@ func defaultAPIClient() (*api.Client, error) {
 func apiClient(host string, accessKey string, skipTLSVerify bool) (*api.Client, error) {
 	u, err := urlx.Parse(host)
 	if err != nil {
-		return nil, err
+		return nil, UserFacingError{
+			UserFacingMessage: fmt.Sprintf("Could not reach %s. Please check server hostname.", host),
```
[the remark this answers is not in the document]
kimskimchi: I wanted a general statement saying "something is wrong with your hostname, whether it is during parsing or connecting" (and then print the actual error). Do you have any suggestions for the wording since the current one could be misleading? 

since to the user, it is a bad hostname, whether it be infra or the server rejecting it

internal/cmd/login.go
```diff
@@ -392,32 +458,41 @@ func newAPIClient(options loginCmdOptions) (*api.Client, error) {
...
+		return UserFacingError{
+			UserFacingMessage: fmt.Sprintf("Could not reach server [%s]. Please check server hostname.", host),
+			Underlying:        err,
+			ShowUnderlying:    true,
+		}
 	}
 	url.Scheme = "https"
 	urlString := url.String()
 
+	logging.S.Debug("creating request to verify tls")
```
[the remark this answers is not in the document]
kimskimchi: This was added to log before each possibility of an error (so we know where it failed)
I actually couldn't tell where the error was coming from (during testing) before these logs 😅

api/client.go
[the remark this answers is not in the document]
dnephin: We should talk some more about this. I think `api.checkError` is not the right place to do this kind of translation.  Other users of the `api.Client` will need to be able to check the status code.

I think `api.CheckError` should only build an `api.Error` and return it, and the caller should translate the error into whatever they want. That's what `formatAPIError` was supposed to do (but I see it is commented out in this branch now).

main.go
kimskimchi: original intention was to do nothing when its a terminal error, while do something for the other cases

kimskimchi: I was also confused at first, Golang has the exact opposite behavior of what I'm familiar with (which is fall through by default)

internal/cmd/errors_userfacing.go
kimskimchi: the reason I am attempting to make it a bit more complex, is so that user can have a general guideline of why that error is relevant to them, but still keep the original error.

First point (for context) is on why we need a user facing message at all (this was brought up a lot of times): 
- if I'm logging in, and it says connection refused, is this the infra server refusing connection or the provider? 
- if I'm adding a grant, why is there a random error saying "provider could not be added"?

Second point is then why do we need the original error:
- I initially suggested without it, that we either print `Internal error: stack trace...`, or a user facing message `This sis a sentence. To output a better sentence, add a provider. 'infra provider add PROVIDER'`
- however, after asking a couple of people, it seemed that a bunch of people would prefer to see the actual failure as well as a friendly error message

internal/cmd/errors_userfacing.go
mxyng: As in just call the type `Error`

e.g.

```go
if somethingBad {
  return Error{Message: "something bad happened"}
}
```

or 
```go
if err {
  return NewError(err, "something bad happened") # returns Error{Error: err, Message: "..."}
}

api/client.go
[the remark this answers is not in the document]
kimskimchi: yep, I didn't want to address api errors yet, so thank you for the pr! :D

kimskimchi:
Thanks for all your suggestions, will request prs updates now! :)

PR #1766, opened 2022-04-28
api: improve errors returned by api.Client

dnephin:
## Summary

This PR attempts to address two problems:
1. Right now the `api.Client` doesn't expose enough detail about the request error. The caller of an `api.Client` method should be able to access all of the structured fields in the response body, so it can format a message appropriately. `api.Client` should also expose the underlying error code in some way, so that the caller can make a choice based on the error code.
2. Both the `internal` package and the `api` package had a set of error variables that are very similar. This makes it easy to use the wrong variable. This PR fixes one instance where an `api.ErrBadRequest` was used in place of `internal.ErrBadRequest`, which would result in a 500 status code, instead of a 400 status code. The `api` error variables are removed, and replaced with `api.Error`.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -22,40 +22,35 @@ type Client struct {
 	Headers http.Header
 }
 
-func checkError(status int, body []byte) error {
-	var apiError Error
+// checkError checks the resp for an error code, and returns an api.Error with
+// details about the error. Returns nil if the status code is 2xx.
+//
+// 3xx codes are considered an error because redirects should have already
+// been followed before calling checkError.
+func checkError(resp *http.Response, body []byte) error {
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return nil
+	}
 
+	var apiError Error
 	err := json.Unmarshal(body, &apiError)
 	if err != nil {
+		// Use the full body as the message if we fail to decode a response.
 		apiError.Message = string(body)
-		apiError.Code = int32(status)
+		apiError.Code = int32(resp.StatusCode)
 	}
 
-	switch apiError.Code {
-	case http.StatusUnauthorized:
-		return ErrUnauthorized
-	case http.StatusForbidden:
-		return ErrForbidden
-	case http.StatusConflict:
-		return ErrDuplicate
-	case http.StatusNotFound:
-		return ErrNotFound
-	case http.StatusBadRequest:
-		return fmt.Errorf(apiError.Message)
-	case http.StatusBadGateway:
-		// this error should be displayed to the user so they can see its an external problem
-		return fmt.Errorf(apiError.Message)
-	case http.StatusInternalServerError:
-		return ErrInternal
-	case http.StatusGone:
-		return fmt.Errorf(apiError.Message)
-	}
+	return apiError
+}
 
-	if status >= 400 {
-		return errors.New(http.StatusText(status))
+// ErrorStatusCode returns the http status code from the error.
+// Returns 0 if the error is nil, or if the error is not of type Error.
+func ErrorStatusCode(err error) int32 {
+	var apiError Error
+	if errors.As(err, &apiError) {
+		return apiError.Code
 	}
-
-	return nil
+	return 0
 }
 
 func get[Res any](client Client, path string) (*Res, error) {
@@ -78,9 +73,9 @@ func get[Res any](client Client, path string) (*Res, error) {
 		return nil, fmt.Errorf("reading response: %w", err)
 	}
 
-	err = checkError(resp.StatusCode, body)
+	err = checkError(resp, body)
 	if err != nil {
-		return nil, fmt.Errorf("GET %q responded %d: %w", path, resp.StatusCode, err)
+		return nil, fmt.Errorf("GET %v failed: %w", path, err)
 	}
 
 	var res Res
@@ -113,9 +108,9 @@ func list[Res any](client Client, path string, query map[string][]string) ([]Res
 		return nil, fmt.Errorf("reading response: %w", err)
 	}
 
-	err = checkError(resp.StatusCode, body)
+	err = checkError(resp, body)
 	if err != nil {
-		return nil, fmt.Errorf("GET %q responded %d: %w", path, resp.StatusCode, err)
+		return nil, fmt.Errorf("GET %v failed: %w", path, err)
 	}
 
 	var res []Res
@@ -152,9 +147,9 @@ func request[Req, Res any](client Client, method string, path string, req *Req) 
 		return nil, fmt.Errorf("reading response: %w", err)
 	}
 
-	err = checkError(resp.StatusCode, body)
+	err = checkError(resp, body)
 	if err != nil {
-		return nil, fmt.Errorf("%s %q responded %d: %w", method, path, resp.StatusCode, err)
+		return nil, fmt.Errorf("%s %v failed: %w", method, path, err)
 	}
 
 	var res Res
@@ -192,9 +187,9 @@ func delete(client Client, path string) error {
 		return fmt.Errorf("reading response: %w", err)
 	}
 
-	err = checkError(resp.StatusCode, body)
+	err = checkError(resp, body)
 	if err != nil {
-		return fmt.Errorf("DELETE %q responded %d: %w", path, resp.StatusCode, err)
+		return fmt.Errorf("DELETE %v failed: %w", path, err)
 	}
 
 	return nil
--- a/api/errors.go
+++ b/api/errors.go
@@ -2,26 +2,28 @@ package api
 
 import (
 	"fmt"
+	"net/http"
+	"strings"
 )
 
-var (
-	// ErrUnauthorized refers to the http response code unauthorized, which really means not authenticated, despite its name. See https://stackoverflow.com/a/6937030/155585
-	ErrUnauthorized = fmt.Errorf("unauthorized")
-	// ErrForbidden means you don't have permissions to the requested resource
-	ErrForbidden = fmt.Errorf("forbidden")
-	// ErrBadGateway means an invalid response was received from an upstream server (probably an OIDC provider)
-	ErrBadGateway = fmt.Errorf("bad gateway")
-
-	ErrDuplicate  = fmt.Errorf("duplicate record")
-	ErrNotFound   = fmt.Errorf("record not found")
-	ErrBadRequest = fmt.Errorf("bad request")
-	ErrInternal   = fmt.Errorf("internal error")
-)
-
+// Error is used as the response body for failed HTTP requests. It is also
+// the error returned by api.Client methods when the request fails.
 type Error struct {
-	Code        int32        `json:"code"` // should be a repeat of the http response status code
-	Message     string       `json:"message"`
+	// Code is the HTTP status of the response.
+	Code int32 `json:"code"`
+	// Message contains the full text of the failure as a single string. The
+	// details of the failure may also be available in a structured representation
+	// from one of the other fields on the Error struct.
+	Message string `json:"message"`
+	// FieldErrors contains a structured representation of any validation errors.
 	FieldErrors []FieldError `json:"fieldErrors,omitempty"`
+}
+
+func (e Error) Error() string {
+	if e.Message == "" {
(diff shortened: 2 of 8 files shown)


internal/server/errors.go
```diff
@@ -77,6 +75,20 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
...
+	switch {
+	case base == full:
+		return ""
+	case strings.HasPrefix(full, base+": "):
+		return strings.TrimPrefix(full, base+": ")
+	case strings.HasSuffix(full, ": "+base):
+		return strings.TrimSuffix(full, ": "+base)
+	}
+	return full
+}
```
dnephin, line 90: This is not ideal for sure. I think if we used domain-specific errors we wouldn't have this problem. `validator.ValidationErrors` in `sendAPIError`  is a good example of this. We don't have to filter out a message for that error, because the error text is specific to what failed, not a generic mirror of HTTP status codes.

internal/server/errors.go
```diff
@@ -47,16 +48,13 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
...
 	case errors.Is(err, internal.ErrNotImplemented):
 		resp.Code = http.StatusNotImplemented
-		resp.Message = internal.ErrNotImplemented.Error()
+		resp.Message = trimErrorMsg(internal.ErrNotImplemented, err)
 	case errors.Is(err, internal.ErrBadGateway):
 		resp.Code = http.StatusBadGateway
-		resp.Message = err.Error()
-	case errors.Is(err, (*validator.InvalidValidationError)(nil)):
-		resp.Code = http.StatusBadRequest
-		resp.Message = err.Error()
```
dnephin, line 59: I tested this to confirm, but this is an incorrect use of `errors.Is`. I didn't realize this either until I was pairing with Elizabeth last week.

`errors.Is` only works with error variables. It's like the `==` operation, and `err` here isn't going to be `nil`.

This would need to use `errors.As` to do a type check. But I think even that isn't necessarily correct. `InvalidValidationError` is returned when an invalid argument is passed to `Validate`. If we pass  an invalid argument that's a programming error, and I think should probably be a 500 status code. So by removing this case we should end up with a 500 status code.

internal/server/errors.go
BruceMacD: This is trimming things like `bad request: ...`? It's not super clear to me.

api/client.go
```diff
@@ -22,40 +21,20 @@ type Client struct {
...
-func checkError(status int, body []byte) error {
-	var apiError Error
+func checkError(resp *http.Response, body []byte) error {
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return nil
+	}
 
+	var apiError Error
 	err := json.Unmarshal(body, &apiError)
 	if err != nil {
```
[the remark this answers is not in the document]
dnephin: Ah yes, good call! I messed that up. That should be done in `sendAPIError`, right? So that we never return it from the API endpoints.

api/client.go
```diff
@@ -78,9 +57,9 @@ func get[Res any](client Client, path string) (*Res, error) {
 		return nil, fmt.Errorf("reading response: %w", err)
 	}
 
-	err = checkError(resp.StatusCode, body)
+	err = checkError(resp, body)
 	if err != nil {
-		return nil, fmt.Errorf("GET %q responded %d: %w", path, resp.StatusCode, err)
+		return nil, fmt.Errorf("GET %v failed: %w", path, err)
```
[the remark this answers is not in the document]
dnephin: Hmm, I might not understand your comment. This is only returning an error to the caller, right? This is in the api client. I don't think we are logging anything here.

internal/server/errors.go
dnephin: Yup, exactly! I think the new test cases below demonstrate it, but definitely it's not super obvious from the code. I'll add a godoc string to clarify.

internal/cmd/login.go
```diff
@@ -154,7 +154,8 @@ func login(options loginCmdOptions) error {
 func loginToInfra(client *api.Client, loginReq *api.LoginRequest) error {
 	loginRes, err := client.Login(loginReq)
 	if err != nil {
-		if errors.Is(err, api.ErrUnauthorized) {
+		var apiError api.Error
+		if errors.As(err, &apiError) && apiError.Code == http.StatusUnauthorized {
```
dnephin, line 158: If we want to make this operation more convenient, I think we could add this function to the `api` package. `grpc` has a similar helper that makes it convenient to deal with their status codes.

```go
func ErrorStatusCode(err error) (int32, bool) {
    var apiError api.Error
    if errors.As(err, &apiError) {
        return apiError.Code, true
    }
    return 0, false
}
```

BruceMacD (approved):
My comment about log injection was wrong, this looks good to me now

dnephin:
I've been following the discussion on this [Go proposal](https://github.com/golang/go/issues/52607). I think a cleaner way of doing this would be to wrap errors with something like [errorWithType](https://github.com/juju/errors/blob/master/errortypes.go#L69-L85).

That would remove the need for `trimErrorMsg` because our sentinel errors would no longer be included in the error message. `errorWithType` should only print the text of the original error.

I think that approach is an extension of what is done in this PR, so I could do that as a follow up. Or I can try that first before merging this.

internal/server/errors.go
```diff
@@ -77,6 +75,24 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
...
+	switch {
+	case base == full:
+		return ""
+	case strings.HasPrefix(full, base+": "):
+		return strings.TrimPrefix(full, base+": ")
+	case strings.HasSuffix(full, ": "+base):
+		return strings.TrimSuffix(full, ": "+base)
+	}
+	return full
+}
```
dnephin, line 94: https://github.com/infrahq/infra/pull/1766#issuecomment-1114018693

```suggestion
```

internal/server/errors.go
ssoroka: the goal of the code here was to return StatusBadRequest when the request failed validation. We should try to accomplish that.

api/errors.go
```diff
@@ -2,28 +2,32 @@ package api
...
-	// ErrForbidden means you don't have permissions to the requested resource
-	ErrForbidden = fmt.Errorf("forbidden")
-	// ErrBadGateway means an invalid response was received from an upstream server (probably an OIDC provider)
-	ErrBadGateway = fmt.Errorf("bad gateway")
-
-	ErrDuplicate  = fmt.Errorf("duplicate record")
-	ErrNotFound   = fmt.Errorf("record not found")
-	ErrBadRequest = fmt.Errorf("bad request")
-	ErrInternal   = fmt.Errorf("internal error")
-)
```
ssoroka, line 19: this belongs to the API and not the client, and should not be removed

internal/server/errors.go
```diff
@@ -30,33 +31,30 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
 	switch {
 	case errors.Is(err, internal.ErrUnauthorized):
 		resp.Code = http.StatusUnauthorized
-		resp.Message = "unauthorized"
+		// no messages for this type of error, it may contain sensitive information
```
ssoroka, line 34: seems like a hard-coded message here was not harmful

internal/server/errors.go
```diff
@@ -77,6 +75,24 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
...
+// trimErrorMsg attempts to prevent duplicate messages in error responses by
+// trimming the errBase message from errFull. If the two errors contain identical
+// messages then an empty string is returned. The errBase text is only removed
+// if it appears as a prefix or suffix to errFull.
+func trimErrorMsg(errBase error, errFull error) string {
+	base, full := errBase.Error(), errFull.Error()
+
+	switch {
+	case base == full:
+		return ""
```
ssoroka, line 87: seems like it should still return a basic message

internal/server/errors.go
```diff
@@ -77,6 +75,24 @@ func (a *API) sendAPIError(c *gin.Context, err error) {
...
+	case strings.HasPrefix(full, base+": "):
+		return strings.TrimPrefix(full, base+": ")
+	case strings.HasSuffix(full, ": "+base):
+		return strings.TrimSuffix(full, ": "+base)
+	}
```
ssoroka, line 92: why do we want to trim the base error message out? Seems like it could only add to clarity. Keep in mind some bad clients don't show http status codes to users, but they do show the full message.

api/client.go
```diff
@@ -22,40 +21,20 @@ type Client struct {
 	Headers http.Header
 }
 
-func checkError(status int, body []byte) error {
-	var apiError Error
+func checkError(resp *http.Response, body []byte) error {
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
```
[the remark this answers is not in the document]
dnephin: True. Generally 3xx codes will be handled by the `http.Client` and never make it here.  If there are too many redirects [defaultCheckRedirect](https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/client.go;l=811-816) returns an error, which will be returned by `client.HTTP.Do`.

The only way to get here from a 3xx status code is a custom `http.Client.CheckRedirect`  function which doesn't error in the same way.

If we don't count 3xx codes as errors, then we'll return `nil` here and in the caller we will try to unmarshal a response, which will fail (because we didn't get a 2xx code, and the user will get a `parsing json response` error.

That seems like the wrong error to return. The user should be informed they ended up with a 3xx status code, so they can fix their proxy settings.

So I think `< 300` is correct, but I can add a comment about why.

internal/server/errors.go
dnephin: That is already handled in the case above `case errors.As(err, validationErrors)`. Validation errors return `validator.ValidationErrors`.

`validator.InvalidValidationError` is documented as :

> // InvalidValidationError describes an invalid argument passed to
> // `Struct`, `StructExcept`, StructPartial` or `Field`

That sounds like a programming error , and should be a 500. So I think removing this is correct.

internal/server/errors.go
dnephin: We should have some way of providing better error messages for these types. I think we should define an error type for forbidden and unauthorized. We would only create the type with a safe and sanitized message, which lets us include it here.  Any errors that are simply wrapped with a sentinel value will continue to have no response.

internal/server/errors.go
dnephin: I've reverted this part for now, and focused this PR on the client side changes.

dnephin:
This PR should be ready for review now. I reverted the api changes and removed `trimErrorMessage`. This should still allow the CLI to get access to the error details, without changing how we populate the error response.

internal/server/handlers.go
```diff
@@ -576,7 +576,7 @@ func (a *API) Login(c *gin.Context, r *api.LoginRequest) (*api.LoginResponse, er
 		return &api.LoginResponse{PolymorphicID: user.PolyID(), Name: user.Name, AccessKey: key, Expires: api.Time(expires)}, nil
 	}
 
-	return nil, api.ErrBadRequest
+	return nil, fmt.Errorf("%w: missing login credentials", internal.ErrBadRequest)
```
dnephin, line 579: This is the example of why having error variables in the `api` package causes problems. This API endpoint needed to return `internal.ErrBadRequest`, but returning an error from the `api` package feels natural here. 

By removing the second set of error variables we remove the problem.

internal/server/errors.go
ssoroka: Do we need any message at all for these? Usually we don't want to leak any info for authn/authz problems, and leave details in logs.

internal/server/handlers.go
ssoroka: yeah, definitely reasonable.

ssoroka (approved):
Very sharp and clean 👍🏼

internal/server/errors.go
dnephin: I'm not sure. I think there are times when an authenticated request could benefit from more details about the authorization failure, without it being a security risk. I guess we don't have the problem right now, so we don't have to worry about it.

PR #1768, opened 2022-04-28
improve: make canonicalPath more strict

dnephin:
## Summary

Previously the following could happen (assuming a value of `/home/user` for `$HOME`):

```
$HOMERUN/config -> /home/userRUN/config
${HOME}/config  -> /working/dir/${HOME}/config
~daniel/config  -> /home/userdaniel/config
```

With these changes both `$HOME` and `${HOME}` are supported substitutions.  As a side-effort other environment variables can also be interpolated into the string, but that shouldn't be a problem. We don't have to document that any env var can be used.
Only `~/` will be replaced with the home dir. Trying to use `~username` will be ignored and `~username` will be treated as a relative path.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -252,25 +252,18 @@ $ infra use development.kube-system`,
 	}
 }
 
-func canonicalPath(in string) (string, error) {
-	homeDir, err := os.UserHomeDir()
-	if err != nil {
-		return "", err
+func canonicalPath(path string) (string, error) {
+	path = os.ExpandEnv(path)
+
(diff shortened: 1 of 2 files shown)


dnephin:
I considered making a `pflag.Value` type for this logic, so that we don't need to call `canonicalPath` on the fields. `cliopts.Load` would take care of resolving the path to an absolute path. I couldn't decide if that was an improvement or made things less obvious, so I left it this way for now.

BruceMacD (approved):
I like this change more than adding a new flag value, nice and simple 👍

Issue #1394, opened 2022-03-30
cli: Refactor error messages for better readability


kimskimchi:
This is a tracking bug. We need to :

1. properly distinguishing stdout / stderr
2. having user-facing errors, that states the error itself as a sentence, then suggests next command to run. Similar to git
3. centralize all those user-facing errors in one place
4. distinguish between error, debug, etc.
5. distinguish between user-facing errors vs actual internal errors
6. utilize printError() in cmd.go properly to translate to user facing messages

Proposal: 
- #1760 

Features: 
- #1786 - implements CLI's own error struct 
- #1920 - updates output as single line
- #1936 - uses CLI error struct in main.go
- #2156 - migrates user facing errors to use Error for basic outputs/errors

PR #1786, opened 2022-04-29
feat: introduce cli errors

kimskimchi:
## Summary

Introduce cli errors. These will be used for outputting user facing messages in a consistent format, where as the original errors (`OriginalError`) should be used for debugging. 

See a draft of how this is to be used: #1760

## Checklist

## Related Issues

Related #1394

### DIFF
--- a/internal/cmd/error.go
+++ b/internal/cmd/error.go
@@ -25,7 +25,7 @@ func (e Error) Error() string {
 		if string(e.Message[len(e.Message)-1]) == "." {
 			e.Message = e.Message[:len(e.Message)-1]
 		}
-		return fmt.Sprintf("%v:\n       %v", e.Message, e.OriginalError)
+		return fmt.Sprintf("%v:\n%v", e.Message, e.OriginalError)
 	}
 
 	return e.Message
--- a/internal/cmd/error.go
+++ b/internal/cmd/error.go
@@ -5,13 +5,13 @@ import "fmt"
 // CLI Errors are user facing errors that are formatted.
 // Should be used for communication, rather than a stacktrace.
 type Error struct {
-	// Error that is caused by the system, used for debugging
-	// Only set this if you need it to be printed as part of the message
+	// OriginalError is an error that is caused by the system, used for logging/debugging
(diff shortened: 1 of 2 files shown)


internal/cmd/error.go
```diff
@@ -0,0 +1,36 @@
+package cmd
+
+import "fmt"
+
+// CLI Errors are user facing errors that is formatted.
```
BruceMacD, line 5: nit:
```suggestion
// CLI Errors are user facing errors that are formatted.
```

internal/cmd/error.go
```diff
@@ -0,0 +1,36 @@
...
+// Should be used for communication, rather than a stacktrace.
+type Error struct {
+	// Error that is caused by the system, used for debugging
+	// Only set this if you need it to be printed as part of the messsage
```
BruceMacD, line 9:
```suggestion
	// Only set this if you need it to be printed as part of the message
```

internal/cmd/error.go
```diff
@@ -0,0 +1,36 @@
...
+	if e.OriginalError != nil {
+		if len(e.Message) == 0 {
+			return fmt.Sprintf("Internal error:\n%v", e.OriginalError)
+		}
+
+		// Strip '.' at the end when message includes the original error
+		if string(e.Message[len(e.Message)-1]) == "." {
+			e.Message = e.Message[:len(e.Message)-1]
+		}
+		return fmt.Sprintf("%v:\n       %v", e.Message, e.OriginalError)
```
BruceMacD, line 28: Should this indent be a `\t`?

BruceMacD (approved):

dnephin (commented):
This looks good! I think the only thing that is missing is test cases for `Error` method. We should probably have 3 test cases to show each of the possible formats.

internal/cmd/error.go
kimskimchi: \t seems to be off by one space: 
image

kimskimchi:
Will push pr on Monday after we get a couple of test cases in.

PR #1789, opened 2022-04-29
fix: do not label signup user grant as created by system

### LINKED ISSUE #1788: Signup user loses admin grant on server restart
## Describe the bug

The user created by the signup workflow is granted Infra admin but on server restart, the user loses all access

## Expected behavior

Admin grant is persisted across restarts

mxyng:
## Summary

anything labelled "created by system" is pruned on startup by config
loading. since the admin user/grant is not part of the config, it will
be pruned.

## Checklist

## Related Issues

Resolves #1788

### DIFF
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -76,7 +76,7 @@ func Signup(c *gin.Context, name, password string) (*models.Identity, error) {
 		Subject:   uid.NewIdentityPolymorphicID(identity.ID),
 		Privilege: models.InfraAdminRole,
 		Resource:  "infra",
-		CreatedBy: models.CreatedBySystem,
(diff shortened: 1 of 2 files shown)


dnephin (commented):
Thanks for fixing this!

Should we make this granted `CreatedBy` the admin identity we created earlier in this function, so that we don't end up a 0 value for `CreatedBy` ?

How can we test this, to prevent a regression?

mxyng:
Added a test case. Since this particular bug requires a few components interacting with each other, it's tested at a higher layer. The test emulates a server restart by calling `loadConfig` explicitly. It's not exactly the same as a restart but it does catch this error

dnephin (approved):
Looks good!

PR #1792, opened 2022-04-29
fix: mac os uses symlink for canonical path

BruceMacD:
## Summary
Quick change. This test fails on mac, need to append the `/private` link to the path to pass.

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/cmd/cmd_test.go
```diff
@@ -35,6 +36,12 @@ func TestCanonicalPath(t *testing.T) {
 	t.Setenv("HOME", "/home/user")
 	t.Setenv("USERPROFILE", "/home/user")
 	wd := t.TempDir()
+
+	if runtime.GOOS == "darwin" {
```
BruceMacD, line 40: Not the most elegant solution but `filepath.Abs(path)` didn't add the `/private` to wd, and `os.Readlink` errors on the test temp dir.

dnephin (approved):
Thanks for fixing this!

This is not the first time that the MacOS tempDir-is-symlink has tricked me, and I suspect it won't be the last. One of these days I'll remember!

PR #1794, opened 2022-04-29
improve: consolidate user and machine identities into a single entity

### LINKED ISSUE #1657: Consolidate user and machine identities into a single entity

BruceMacD:
## Summary
Make identity the sole resource representing an identity without user or machine subtypes.
- Identifier can be either an email (local@domain) or a name (^[a-zA-Z0-9-_./]+$)
- May or may not have a password. No password set disables login with email + password
- Can create and use access keys

## Checklist

## Related Issues

Resolves #1657


internal/server/server_test.go
```diff
@@ -371,7 +371,7 @@ func TestServer_PersistSignupUser(t *testing.T) {
 	passwd := "supersecretpassword"
 
 	// run signup for "admin@email.com"
-	signupReq := api.SignupRequest{Email: email, Password: passwd}
+	signupReq := api.SignupRequest{Name: email, Password: passwd}
```
jmorganca, line 374: Is this a breaking change for CLI <> API?

jmorganca: nice!

jmorganca (approved):
Great cleanup

Makefile
```diff
@@ -10,6 +10,10 @@ test:
 test-all:
 	go test ./...
 
+# update the expected command output file
```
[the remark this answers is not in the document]
dnephin: I'm added support for using just the `-update` flag, but haven't tagged a release yet. Hopefully that will make it much easier to remember.

Also doesn't hurt to have it in the Makefile.

internal/server/config.go
```diff
@@ -678,6 +679,7 @@ func (Server) loadGrant(db *gorm.DB, input Grant) (*models.Grant, error) {
 
 		id = uid.NewGroupPolymorphicID(group.ID)
 
+	// TODO: this gotta go
```
dnephin, line 682: Is this TODO still relevant?

internal/server/handlers_test.go
```diff
@@ -307,24 +307,42 @@ func TestAPI_DeleteProvider(t *testing.T) {
...
 	opts.Grants = append(opts.Grants, Grant{
-		Machine:  "admin",
+		User:     "admin",
 		Role:     "admin",
 		Resource: "infra",
 	})
 }
 
-func TestAPI_CreateIdentity(t *testing.T) {
+func TestDeleteProvider_NoDeleteInternalProvider(t *testing.T) {
```
dnephin, line 323: I think this is covered by the `"infra provider can not be deleted"` case in the table test above, no?

internal/server/errors_test.go
```diff
@@ -60,36 +60,6 @@ func TestSendAPIError(t *testing.T) {
...
-				Code:    http.StatusBadRequest,
-				Message: `Email: failed the "email" check`,
-				FieldErrors: []api.FieldError{
-					{
-						FieldName: "Email",
-						Errors:    []string{`failed the "email" check`},
-					},
-				},
-			},
-		},
```
dnephin, line 92: These test cases a are still relevant, they were testing that validation errors return the right HTTP response code. I think we still want that, we just need a different field to use.

dnephin: I guess we don't need both, but at least one would be good.

PR #1796, opened 2022-04-29
fix: handle panics properly

dnephin:
## Summary

While working on some tests I ran into a situation where I was receiving a 200 response code unexpectedly. It turns out the handler had panic'ed, but there were no logs indicating that is what happened.

Without this change a panic in an HTTP handler will result in a 200 response code with no logs. For some reason the sentrygin middleware doesn't set the response code.

By having it re-panic, the `gin.Recovery` middleware does the right thing. It logs the panic to stder and returns a 500 status code.

I'm not sure if we are going to keep the sentry-gin middleware, but this fixes it for now.


mxyng (approved):

dnephin:
Replaced by #1801

PR #1801, opened 2022-05-01
improve: remove crash reporting

jmorganca:


internal/server/server.go
```diff
@@ -157,7 +149,7 @@ func New(options Options) (*Server, error) {
 		return nil, fmt.Errorf("loading certificate provider: %w", err)
 	}
 
-	settings, err := data.InitializeSettings(server.db)
+	_, err = data.InitializeSettings(server.db)
```
BruceMacD, line 152: we should update `InitializeSettings` to not return a settings struct. It isn't used anymore as of this change.

BruceMacD (commented):
Should we have crash reporting in our telemetry now or removing this completely?

dnephin (approved):
LGTM, we could change the `InitializeSettings` signature in a follow up.

Including panics in telemetry sounds like a good idea as well, but could be a follow up

PR #1802, opened 2022-05-01
improve: remove CreatedBy field and always replace records if provided via server options

jmorganca:
This removes the `CreatedBy` field which was previously used to track configuration loading. While helpful to know who created what, this is better handled by later storing event data. The hope is this would help remove tricky edge cases (e.g. a resource created by "the system") which could avoid issues similar to #1788 in the future.

Instead, the Infra server now disregards how records are created. If non-empty slices are provided for `providers:`, `grants:` or `identities:` in the server configuration, existing records (except default must-have records) are simply deleted (no matter where they originate from).

Note: as an important change, if any of the above slices are provided in the server's configuration, any existing records **of that kind** are deleted.

The user can then decide to:
1. "Lock" records (e.g. `grants`) to what is provided in the `server.yaml` file
2. Use the API or UI to manage these records ad-hoc

As a future change we can consider moving the server configuration on top of the API, allowing:
1. Custom semantics around how "declared" records are managed
2. Additional ways to configure Infra via code: terraform, etc
3. New CLI commands such as `infra import -f <infra.yaml>` or similar

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -523,57 +523,86 @@ func getKindFromUnstructured(data interface{}) string {
 	return ""
 }
 
-func (s Server) loadConfig(config Config) error {
-	if err := validator.New().Struct(config); err != nil {
-		return err
-	}
-
+func (s Server) loadConfig() error {
 	return s.db.Transaction(func(tx *gorm.DB) error {
-		// inject internal infra provider
-		config.Providers = append(config.Providers, Provider{
-			Name: models.InternalInfraProviderName,
-		})
+		// load default infra provider
+		provider, err := s.loadProvider(tx, Provider{Name: models.InternalInfraProviderName})
+		if err != nil {
+			return err
+		}
 
-		config.Identities = append(config.Identities, Identity{
-			Name: models.InternalInfraConnectorIdentityName,
-		})
+		// load default connector identity
+		connector, err := s.loadIdentity(tx, Identity{Name: models.InternalInfraConnectorIdentityName}, provider)
+		if err != nil {
+			return err
+		}
 
-		config.Grants = append(config.Grants, Grant{
+		// load default connector grant
+		connectorGrant, err := s.loadGrant(tx, Grant{
 			User:     models.InternalInfraConnectorIdentityName,
 			Role:     models.InfraConnectorRole,
 			Resource: "infra",
 		})
-
-		if err := s.loadProviders(tx, config.Providers); err != nil {
-			return fmt.Errorf("load providers: %w", err)
+		if err != nil {
+			return err
 		}
 
-		// extract identities from grants and add them to identities
-		for _, g := range config.Grants {
-			switch {
-			case g.User != "":
-				config.Identities = append(config.Identities, Identity{Name: g.User})
-			case g.Machine != "":
-				logging.S.Warn("please update 'machine' grant to 'user', the 'machine' grant type is deprecated and will be removed in a future release")
-				config.Identities = append(config.Identities, Identity{Name: g.Machine})
+		if err := validator.New().Struct(s.options.Config); err != nil {
+			return err
+		}
+
+		// load and replace providers if specified in the config
+		if len(s.options.Config.Providers) > 0 {
+			if err := s.loadProviders(tx, s.options.Config.Providers, []uid.ID{provider.ID}); err != nil {
+				return err
 			}
 		}
 
-		if err := s.loadIdentities(tx, config.Identities); err != nil {
-			return fmt.Errorf("load identities: %w", err)
+		var ids []uid.ID
+
+		// load and replace grants if specified in the config
+		if len(s.options.Config.Grants) > 0 {
+			for _, g := range s.options.Config.Grants {
+				if g.Group != "" {
+					continue
+				}
+
+				// create placeholder identities for users declared in grants
+				var id *models.Identity
+				var err error
+
+				switch {
+				case g.User != "":
+					id, err = s.loadIdentity(tx, Identity{Name: g.User}, provider)
+				}
+
+				if err != nil {
+					return err
+				}
+
+				if id != nil {
+					ids = append(ids, id.ID)
+				}
+			}
+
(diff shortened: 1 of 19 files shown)


internal/server/config.go
```diff
@@ -523,56 +523,88 @@ func getKindFromUnstructured(data interface{}) string {
...
 
-		// extract identities from grants and add them to identities
-		for _, g := range config.Grants {
-			switch {
-			case g.User != "":
-				config.Identities = append(config.Identities, Identity{Email: g.User})
-			case g.Machine != "":
-				config.Identities = append(config.Identities, Identity{Name: g.Machine})
+		// load and replace providers if specified in the config
+		if s.options.Config.Providers != nil {
```
jmorganca, line 555: This is where we decide wether to "overwrite" or not – if users don't specify `providers:` in the server configuration then this won't run since the slice will be `nil` (as opposed to an empty slice – in which case this _will_ run and delete all providers as one may expect)

BruceMacD (changes requested):
While testing I found that something in these changes caused my default admin user to not work when a config is applied.

Re-production steps:
1. Stand up a new Infra instance with no config
2. Create default admin user
3. Apply a config (for example, with an identity provider)
4. The created default admin user can no longer login and appears to have lost their grants

mxyng (changes requested):
The server `configmap` template will need to be updated otherwise it will template out an empty list instead of omitting `providers`, `grants`, and `identities` keys which, with this change, will delete any resources created by the user using the CLI

jmorganca:
Thanks for the catch: I've updated the helm chart since it was setting empty slices by default

internal/server/config.go
```diff
@@ -582,8 +614,7 @@ func (s Server) loadProviders(db *gorm.DB, providers []Provider) error {
 		keep = append(keep, provider.ID)
 	}
 
-	// remove any provider previously defined by config
-	if err := data.DeleteProviders(db, data.NotIDs(keep), data.CreatedBy(models.CreatedBySystem)); err != nil {
+	if err := data.DeleteProviders(db, data.NotName(models.InternalInfraProviderName), data.NotIDs(keep)); err != nil {
```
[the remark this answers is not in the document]
jmorganca: Thanks!

dnephin (commented):
LGTM! I think avoiding deletes when the input is empty is enough of a safeguard to prevent accidentally deleting everything.

We still need to make sure that we carefully document that using the config is mutually exclusive with using the API.  Ithink there needs to be a big loud warning in the config documentation/examples saying.

```
WARNING: adding any identities here will delete all existing identities.
```

I suspect that makes this config unusable for most people. But I guess that's an acceptable trade-off for now to keep things simple.

Mostly some questions and a suggestion for additional error context.

dnephin:
Any additional test cases we need for config loading?

internal/server/config.go
```diff
@@ -734,8 +752,7 @@ func (s Server) loadIdentities(db *gorm.DB, identities []Identity) error {
 		keep = append(keep, identity.ID)
 	}
 
-	// remove any provider previously defined by config
-	if err := data.DeleteIdentities(db, data.NotIDs(keep), data.CreatedBy(models.CreatedBySystem)); err != nil {
+	if err := data.DeleteIdentities(db, data.NotIDs(keep)); err != nil {
```
dnephin, line 755: Is this missing a filter for "from infra provider" ? Otherwise we lose all identities from other providers. That would be a good test case, if we don't have it already.

internal/server/config.go
```diff
@@ -523,57 +523,86 @@ func getKindFromUnstructured(data interface{}) string {
...
 			Resource: "infra",
 		})
+		if err != nil {
+			return err
+		}
 
-		if err := s.loadProviders(tx, config.Providers); err != nil {
-			return fmt.Errorf("load providers: %w", err)
+		if err := validator.New().Struct(s.options.Config); err != nil {
+			return err
```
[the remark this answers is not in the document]
jmorganca: It absolutely would. I'll move it there!

internal/server/config.go
jmorganca: In this case we'll still want to delete all – since using this config implies Infra will converge to these users existing in the system (vs leaving users lying around).

internal/server/config.go
```diff
@@ -523,57 +523,86 @@ func getKindFromUnstructured(data interface{}) string {
 	return ""
 }
 
-func (s Server) loadConfig(config Config) error {
-	if err := validator.New().Struct(config); err != nil {
-		return err
-	}
-
+func (s Server) loadConfig() error {
```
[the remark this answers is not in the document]
jmorganca: Great callout. I'll update this PR to reduce (vs increase) coupling between this function and `Server`

jmorganca:
@dnephin this is a good call, created https://github.com/infrahq/infra/issues/1885 to better document how to use this part

internal/server/config.go
```diff
@@ -523,57 +523,78 @@ func getKindFromUnstructured(data interface{}) string {
...
+		var ids []uid.ID
+
+		// load and replace grants if specified in the config
+		if len(config.Grants) > 0 {
+			for _, g := range config.Grants {
+				if g.Group != "" || g.User == "" {
+					continue
+				}
+
+				id, err := s.loadIdentity(tx, Identity{Name: g.User}, provider)
```
mxyng, line 570: The implication here is users/groups created through grants will never be removed, even if the grants are removed, since it's not calling `DeleteIdentities`

mxyng (approved):

mxyng (changes requested):
There's a complication where if user enables the connector in the helm chart, it will append an identity with the access key generated for that connector. In that case, the config loading will see a non-empty identities list, load the identity, then remove all non-config identities. This would be a bad experience for the user since they would not expect enabling the connector to remove their identities

jmorganca:
@mxyng this is a great catch! Going to close this for now while consider a future design that's based on the API

PR #1803, opened 2022-05-02
fix: grant user should only be created once

### LINKED ISSUE #1761: ui: should not be able to create a second grant for a user
It should replace the existing grant

Screen Shot 2022-04-28 at 12 55 14 PM

hoyyeva:
## Summary
when creating grant user, it should be created once. 

 `test@infrahq.com` - `view` is already created: 
- if user types `test@infrahq.com` and selects `view` then there should not trigger any call to the endpoints, 
- if user types `test@infrahq.com` and selects `admin` then it should be acting the same as role updates

## Related Issues
Resolves #1761

### DIFF
--- a/ui/components/grant.js
+++ b/ui/components/grant.js
@@ -7,18 +7,6 @@ import InputDropdown from '../components/input'
 import ErrorMessage from '../components/error-message'
 import InfoModal from './modals/info'
 
-function Grant ({ id }) {
-  if (!id) {
-    return null
-  }
-
-  const { data: user } = useSWR(`/v1/identities/${id.replace('i:', '')}`, { fallbackData: { name: '' } })
-
-  return (
-    <p>{user.name}</p>
-  )
-}
-
 export default function ({ id, modalOpen, handleCloseModal }) {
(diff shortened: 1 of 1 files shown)


ui/components/grant.js
```diff
@@ -30,6 +18,23 @@ export default function ({ id, modalOpen, handleCloseModal }) {
   const [role, setRole] = useState('view')
 
   const options = ['view', 'edit', 'admin', 'remove']
+  const emailList = [];
```
jmorganca, line 21: In this case we should show the most "powerful" role for the user (from `view` to `admin`) vs simply taking the first. Subsequently we should fix this in the API (I've created https://github.com/infrahq/infra/issues/1804 to track)

A good approach could be to store a map `{email: role}` (example below) to display in the UI so users only see the "most powerful" role a user has. Removing their role in the UI should remove _all_ roles for this user even if we're only showing one in the UI.

Example:
```
{
  'jeff@infrahq.com': 'admin', 'view'
}
```

hoyyeva:
close this for now. After discuss with @jmorganca  the best way to handle this is do it on the API. So we are going to keep what is it for now.

PR #1805, opened 2022-05-02
improve(docs): First draft of the API development guidelines

pdevine:
## Summary

First draft of API Development Guidelines

### DIFF
--- a/docs/api/guidelines.md
+++ b/docs/api/guidelines.md
@@ -13,11 +13,6 @@ into other software. Resource based APIs are defined by _endpoints_ (resources)
 which comprise the _nouns_ of the API, and _methods_ (such as GET, PUT, POST,
 etc.) which define the _verbs_.
 
-This guide only covers REST based APIs, however, there are times when it may be
-necessary to define a different type of API.  Although REST APIs can be easy to
-consume, they aren't necessarily performant. Consider using gRPC and Protobufs
-when you need more speed.
-
 ## Creating a new API
 
 Any good API requires a lot of thought and consideration. Good APIs stick
@@ -71,15 +66,15 @@ When you have determined what resources you will use, you will need to also
 determine the paths on which they get exposed. The paths for our current
 API tend to be very flat. Some examples include:
 
-| Path                     | Description                    |
-| :----------------------- | :----------------------------- |
-| /v1/destinations         | Kubernetes resources           |
-| /v1/identities           | Users and machines             |
-| /v1/identities/%s/grants | Grants for a specific identity |
-| /v1/identities/%s/groups | Groups for a specific identity |
-| /v1/grants               | Authorization grants           |
-| /v1/groups               | Groups and group membership    |
-| /v1/providers            | Identity providers             |
+| Path                | Description                    |
+| :------------------ | :----------------------------- |
+| /v1/destinations    | Kubernetes resources           |
+| /v1/users           | Users                          |
+| /v1/users/%s/grants | Grants for a specific identity |
+| /v1/users/%s/groups | Groups for a specific identity |
+| /v1/grants          | Authorization grants           |
+| /v1/groups          | Groups and group membership    |
+| /v1/providers       | Identity providers             |
 
 Try to keep your pathnames flat. Good APIs provide a simple list of base
 resources which do not have deep, nested hierarchies. As a rule of thumb,
@@ -271,17 +266,6 @@ We currently use the following status codes when the API returns:
 | 500    | Internal server error |
 | 502    | Bad gateway           |
 
-
-Currently unused status codes include:
-
-| Status | Type                  |
-| :----- | :---                  |
-| 409    | Conflict              |
-| 429    | Too many requests     |
-| 503    | Service unavilable    |
-| 504    | Gateway timeout       | 
-
-
 We use a standard format for returning an error:
  
 ```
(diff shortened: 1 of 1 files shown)


docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+  * cohesive;
+  * resource based; and
+  * internally consistent
+
+Good APIs can be easily understood by humans, and can be easily implemented
+into other software. Resource based APIs are defined by _endpoints_ (resources)
+which comprise the _nouns_ of the API, and _methods_ (such as GET, PUT, POST,
+etc.) which define the _verbs_.
+
+This guide only covers REST based APIs, however, there are times when it may be
```
BruceMacD, line 16: Should we be mentioning these things here our sticking to a concise description of our API practices?

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+## Endpoint Hierarchies
+
+When you have determined what resources you will use, you will need to also
+determine the paths on which they get exposed. The paths for our current
+API tend to be very flat. Some examples include:
+
+| Path                     | Description                    |
+| :----------------------- | :----------------------------- |
+| /v1/destinations         | Kubernetes resources           |
+| /v1/identities           | Users and machines             |
```
BruceMacD, line 77:
```suggestion
| /v1/identities           | Users                          |
```

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+
+If you are creating a compound field name, use lower camel case for the JSON field with the first character
+lowercase, and capitalize the first letter of each subsequent word (e.g. thisIsAField). Do not use characters
+other that ASCII a-z and A-Z inside of the field name.
+
+### Sequential IDs
+
+When returning a resource, avoid using a sequential ID in the ID field. Using sequential IDs allows
+callers to potentially guess a particular resource which can lead to leaking information. We
+have a Polymorphic ID in our `infrahq/infra/api` package which you should use instead.
```
BruceMacD, line 215:
```suggestion
use ["Snowflake IDs"](https://en.wikipedia.org/wiki/Snowflake_ID) in our `infrahq/infra/api` package which you should use instead.
```

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+| 403    | Forbidden             |
+| 404    | Not found             |
+| 409    | Duplicate             |
+| 500    | Internal server error |
+| 502    | Bad gateway           |
+
+
+Currently unused status codes include:
+
+| Status | Type                  |
```
BruceMacD, line 277: These unused status codes feel like they will change, and we probably won't realize to update this document.

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+   "fieldName": [ "<error1>"...<"errorN">]
+  }
+ ]
+}
+```
+
+### Returning a 404 or 403
+
+If a client isn't authenticated or doesn't have access to a particular resource, be careful about returning a
+```403 Forbidden``` error as it may leak information about a resource. Instead, consider returning a
```
BruceMacD, line 302: If a client isn't authenticated they will get a `401 unauthorized` because the first thing we will do is check their token if they're requesting a protected resource. If it's a public endpoint with protected resources that vary depending on the caller we should follow this format.

BruceMacD (approved):

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+| 409    | Duplicate             |
+| 500    | Internal server error |
+| 502    | Bad gateway           |
+
+
+Currently unused status codes include:
+
+| Status | Type                  |
+| :----- | :---                  |
+| 409    | Conflict              |
```
dnephin, line 279: We use 409, it's in the list above
```suggestion
```

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+| 429    | Too many requests     |
+| 503    | Service unavilable    |
+| 504    | Gateway timeout       |
```
dnephin, line 282: We use 504, for request timeouts
```suggestion
```

docs/api/guidelines.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Timestamps
+
+For timestamp fields, we follow [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) which overlaps
+with ISO 8601. Timestamps are always stored in in UTC form. The client is responsible for converting any
+timestamp from UTC to local time.
+
+Timestamp fields follow the form:
+
+```YYYY-MM-DDThh:mm:ss.SSSZ```
```
dnephin, line 233: Looks like this renders ok, but is confusing githubs syntax highlighting.

```suggestion
`YYYY-MM-DDThh:mm:ss.SSSZ`
```

Or maybe add newlines before/after the ```

dnephin (commented):
Looks good! Some minor suggestions for edits

docs/api/guidelines.md
pdevine: I removed this section for now. We can update it in the future.

docs/api/guidelines.md
pdevine: I'll switch this to single ticks.

PR #1833, opened 2022-05-02
improve: do not allow invalid roles for grants

kimskimchi:
## Summary

Introduce errors when user attempts to grant an invalid role for that type of destination, as our doc states. 
In the future, we should read from API responses (bad request - fields) and output the same error there, rather than 'failing early'. 

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -202,6 +202,10 @@ func addGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 		return err
 	}
 
+	if err := checkRole(cmdOptions.Role, cmdOptions.Destination); err != nil {
+		return err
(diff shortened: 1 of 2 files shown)


internal/cmd/grants.go
```diff
@@ -359,3 +363,31 @@ func addGrantIdentity(client *api.Client, name string, identityType identityType
...
 }
+
+func checkRole(role string, destination string) error {
+	if destination == "infra" {
+		switch role {
+		case "admin":
+			fallthrough
+		case "view":
+			fallthrough
+		case "connector":
```
ssoroka, line 374:
```suggestion
		case "admin", "view", "connector":
```

PR #1835, opened 2022-05-02
feat: update list apis to return a response object

### LINKED ISSUE #1834: api: response objects instead of slices
Response should be an object which includes the list response such that additional fields can be added without breaking the API, e.g. pagination.

e.g.

```
{
  "items": [
    {
      <response_object>
    },
    {
      <response_object>
    }
  ]
}
```

mxyng:
## Summary

List API endpoints should return an object such that additional fields can be added as needed without breaking the API, e.g. for pagination. This is a breaking change as it modifies the existing endpoints to return a format that will not be recognized by previous clients.

## Checklist

## Related Issues

Resolves #1834


internal/cmd/destinations.go
```diff
@@ -87,11 +87,11 @@ func newDestinationsRemoveCmd() *cobra.Command {
 				return err
 			}
 
-			if len(destinations) == 0 {
+			if len(destinations.Items) == 0 {
```
mxyng, line 90: This is a common pattern. The resposne object should probably have a `Count` field

hoyyeva (approved):
the ui part looks good to me

api/client.go
```diff
@@ -86,7 +86,7 @@ func get[Res any](client Client, path string) (*Res, error) {
 	return &res, nil
 }
 
-func list[Res any](client Client, path string, query map[string][]string) ([]Res, error) {
+func list[Res any](client Client, path string, query map[string][]string) (*Res, error) {
```
mxyng, line 89: This change makes `list` very similar to `get` with minor differences around query parameters. It can probably be merged into a single func at some point

internal/server/ui/404.html
```diff
@@ -1 +1 @@
...
..."/><link rel="preload" href="/_next/static/css/a7b25e39ac458b0f.css" as="style"/><link rel="stylesheet" href="/_next/static/css/a7b25e39ac458b0f.css" data-n-g=""/><noscript data-n-css=""></noscript><script defer="" nomodule="" src="/_next/static/chunks/polyfills-5cd94c89d3acac5f.js"></script><script src="/_next/static/chunks/webpack-d7b038a63b619762.js" defer=""></script><script src="/_next/static/chunks/framework-5f4595e5518b5600.js" defer=""></script><script src="/_next/static/chunks/main-164920eac96bc1c7.js" defer=""></script><script src="/_next/static/chunks/pages/_app-49d2a68a378223e0.js" defer=""></script><script src="/_next/static/chunks/pages/_error-1995526792b513b2.js" defer=""></script><script src="/_next/static/embed/_buildManifest.js" defer=""></script><script src="/_next/static/embed/_ssgManifest.js" defer=""></script><script src="/_next/static/embed/_middlewareManifest.js" defer=""></script></head><body><div id="__next" data-reactroot=""></div><script id="__NEXT_DATA__" type="application/json">{"props":{"pageProps":{"statusCode":404}},"page":"/_error","query":{},"buildId":"embed","nextExport":true,"isFallback":false,"gip":true,"scriptLoader":[]}</script></body></html>
```
ssoroka, line 1: out of curiousity, why did the html and js pages change?

ssoroka: ah, nm, it's because the frontend was updated.

ui/components/grant.js
```diff
@@ -46,7 +46,7 @@ export default function ({ id, modalOpen, handleCloseModal }) {
 
       setEmail('')
 
-      return [...(grants || []).filter(grant => grant?.subject !== id), data]
+      return [...(grants.items || []).filter(grant => grant?.subject !== id), data]
```
ssoroka, line 49: should these be `grants?.items`? I don't see error checking anywhere.

pdevine (approved):

internal/server/apimigrator.go
```diff
@@ -0,0 +1,157 @@
...
+			w.Flush()
+
+			if w.flushErr != nil {
+				panic(w.flushErr)
+			}
+		},
+	})
+}
+
+type responseWriter struct {
```
dnephin, line 62: Having to wrap the response writer seems like it's going to be the biggest maintenance cost of this implementation.  

I wonder if we could avoid this somehow. Our API endpoints currently return types. If the response migration could accept types and translate those types (before encoding to JSON) that might remove the need to wrap ResponseWriter? I don't know how we do that in a generic way, but it seems easy enough to do by wrapping the function with another one.

internal/server/version.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+)
+
+type Version struct {
+	Major uint8
+	Minor uint8
+	Patch uint8
+}
+
+func NewVersion(ver string) Version {
```
dnephin, line 14: Could we use a non-semantic version to make this easier to parse? We only need migrations for breaking changes anyway, so semantic versions don't seem like they would apply here. Maybe just an sequential int, or some form of encoded date (`20220405`) ?

dnephin (commented):
Had a quick look at the api migration, it seems cool pretty cool.

My main concerns would be that a header makes the version a little less explicit. We should still have some way to deprecate and remove old versions, so that we aren't supporting every version forever. Using a header (instead of the path) seems like it might make deprecation a bit harder to communicate.

Also the concern about having to wrap `ResponseWriter` (more about that below).

I don't know if there is any way we could split up the API changes in this PR from the api migration to make them easier to review independently. Maybe open a PR based on this one (target branch of `mxyng/list-responses`)  so that the changes can be viewed independently?

mxyng:
@dnephin split the api migrations into #1884

internal/server/apimigrator.go
ssoroka: could be doable, but sort of breaks the middleware approach. Let's look at this as an optimization later.

internal/server/version.go
ssoroka: we only publish semantic versions, so the client can get no better resolution than that.

ssoroka:
> a header makes the version a little less explicit. 

Not if you require it to be supplied, which I think is the plan.

> We should still have some way to deprecate and remove old versions, so that we aren't supporting every version forever. 

I think we actually want to support versions forever; or at least until they're not used anymore (cue analytics).

> Using a header (instead of the path) seems like it might make deprecation a bit harder to communicate.

Deprecation isn't really a concern. We're aiming for API stability here.

internal/server/apimigrator.go
```diff
@@ -0,0 +1,157 @@
...
+
+func addResponseRewrite[newResp any, oldResp any](method, path, version string, f func(newResp) oldResp) {
+	migrations = append(migrations, apiMigration{
+		method:  method,
+		path:    path,
+		version: version,
+		responseRewrite: func(c *gin.Context) {
+			reqVer := c.Request.Header.Get("X-Infra-Version")
+			if reqVer == "" || NewVersion(reqVer).GreaterThanStr(version) {
+				c.Next()
```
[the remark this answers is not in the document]
mxyng: I changed revert this to `default to earliest` after thinking about it some more. This case will be useful once most client are up-to-date and reporting a version header. In the current environment where clients do not have the version header, it's better to default to earliest.

PR #1836, opened 2022-05-02
api: remove extra grant endpoints

### LINKED ISSUE #1680: api: remove /v1/idenities/:id/grants and /v1/groups/:id/grants
`ListIdentityGrants`, and `ListGroupGrants` are both endpoints for returning a list of grants.

We could remove them, because it is already possible to list grants for identities and groups using `GET /v1/grants`. The only change we need to make is how the requests are authorized.  Currently `ListIdentityGrants` and `ListGroupGrants` have special cases for "allow listing grants for the authenticated user". 

We don't need separate endpoints for that. We could add those authorization rules to `/v1/grants`.

dnephin:
## Summary

Removes `ListIdentityGrants` and `ListGroupGrants` and moves those special authorization rules to `ListGrants`.

This PR keeps the old deprecated routes for now (as aliases of `ListGrants`), but maybe we should remove them.

Most of the change is the addition of API tests, best viewed by individual commit.

## Related Issues

Related to #1680, but maybe the group endpoint should be changed in the same way before we close that issue.

TODO:

* [x] update CLI to use `ListGrant` (will wait until #1865 merges to avoid conflicts)
* [x] deprecate or remove the `Client` methods
* [x] remove or ticket for removing the deprecated API routes - #1868

### DIFF
--- a/internal/access/grant.go
+++ b/internal/access/grant.go
@@ -35,7 +35,8 @@ func ListIdentityGrants(c *gin.Context, identityID uid.ID) ([]models.Grant, erro
 		return nil, err
 	}
 
-	return data.ListIdentityGrants(db, identityID)
+	polymorphicID := uid.NewIdentityPolymorphicID(identityID)
+	return data.ListGrants(db, data.BySubject(polymorphicID))
 }
 
 func ListGroupGrants(c *gin.Context, groupID uid.ID) ([]models.Grant, error) {
@@ -44,7 +45,8 @@ func ListGroupGrants(c *gin.Context, groupID uid.ID) ([]models.Grant, error) {
 		return nil, err
 	}
 
-	return data.ListGroupGrants(db, groupID)
+	polymorphicID := uid.NewGroupPolymorphicID(groupID)
+	return data.ListGrants(db, data.BySubject(polymorphicID))
 }
 
 func CreateGrant(c *gin.Context, grant *models.Grant) error {
--- a/internal/access/grant.go
+++ b/internal/access/grant.go
@@ -1,8 +1,12 @@
 package access
 
 import (
-	"github.com/gin-gonic/gin"
+	"errors"
 
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+
+	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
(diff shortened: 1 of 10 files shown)


internal/access/grant.go
```diff
@@ -18,33 +22,49 @@ func GetGrant(c *gin.Context, id uid.ID) (*models.Grant, error) {
...
+	if errors.Is(err, internal.ErrForbidden) {
+		// Allow an authenticated identity to view their own grants
+		db := getDB(c)
+		subjectID, _ := subject.ID()
+		identity := AuthenticatedIdentity(c)
+		switch {
+		case identity == nil:
+			return nil, err
+		case subject.IsIdentity() && identity.ID == subjectID:
+			selectors = append(selectors, data.ByOptionalSubject(subject))
```
BruceMacD, line 45: The subject should not be optional in this case should it? The caller without permissions should only be able to see their own resources.

internal/access/grant.go
```diff
@@ -18,33 +22,49 @@ func GetGrant(c *gin.Context, id uid.ID) (*models.Grant, error) {
...
+			return data.ListGrants(db, selectors...)
+		case subject.IsGroup() && userInGroup(db, identity.ID, subjectID):
+			selectors = append(selectors, data.ByOptionalSubject(subject))
```
BruceMacD, line 48: (same comment on subject not being optional)

internal/access/grant.go
```diff
@@ -18,33 +22,49 @@ func GetGrant(c *gin.Context, id uid.ID) (*models.Grant, error) {
...
+		}
 	}
 
-	return data.ListIdentityGrants(db, identityID)
+	return nil, err
 }
 
-func ListGroupGrants(c *gin.Context, groupID uid.ID) ([]models.Grant, error) {
-	db, err := hasAuthorization(c, groupID, isUserInGroup, models.InfraAdminRole, models.InfraViewRole)
+func userInGroup(db *gorm.DB, authnUserID uid.ID, groupID uid.ID) bool {
```
BruceMacD, line 56: I'd like to see something that can be passed to `hasAuthorization` if possible, this seems like it could be a common pattern. If that requires a lot of changes, then this is ok as is.

internal/server/handlers.go
```diff
@@ -446,6 +419,18 @@ func (a *API) ListGrants(c *gin.Context, r *api.ListGrantsRequest) ([]api.Grant,
 	return results, nil
 }
 
+// TODO: remove after deprecation period
```
BruceMacD, line 422: Let's create a ticket to track this or add it to an existing one so we don't forget.

ssoroka (changes requested):
Blocking for the time being because I don't agree with https://github.com/infrahq/infra/issues/1680. Will articulate more info in that issue and try to get this resolved.

internal/access/grant.go
dnephin: Yes, good catch! I fixed exactly this bug in an earlier commit, but then accidentally restored it when I went to add the other selectors to this query.  It should be `BySubject`.

I still wish we did not have `Optional` selectors, and just used an if statement.

internal/access/grant.go
dnephin: I think that would be nice. I would like to do that as well, but my thinking was that maybe it would be better as a follow up. This is still a contentious change, so I didn't want to go with a big refactor up front.

I'll explore what that might look like now.

internal/server/handlers.go
dnephin: Maybe I should remove them in this PR?  At the very least I should update the CLI to use the new endpoints, and mark the Client methods deprecated as well.

internal/access/grant.go
ssoroka: I'm good with this suggestion.

internal/server/grants_test.go
```diff
@@ -0,0 +1,314 @@
...
+		"not authenticated": {
+			urlPath: "/v1/grants?subject=i:" + idOther.String(),
+			setup: func(t *testing.T, req *http.Request) {
+				req.Header.Del("Authorization")
+			},
+			expected: func(t *testing.T, resp *httptest.ResponseRecorder) {
+				assert.Equal(t, resp.Code, http.StatusUnauthorized)
+			},
+		},
+		"not authorized, wrong identity": {
```
ssoroka, line 130:
```suggestion
		"authorized, wrong identity": {
```

ssoroka: actually... I think these are confusing authorized/authenticated. It's not clear to me what they should say

ssoroka (approved):
looks good

PR #1838, opened 2022-05-03
fix: add additional signup check

mxyng:
## Summary

When connector is enabled in the same helm install as server, it will
create an identity and access key to allow the connector to communicate
with the server. This identity is created with "connector" which is
filtered out in the signup check leading to signup to be enabled.

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -35,7 +35,12 @@ func SignupEnabled(c *gin.Context) (bool, error) {
 		return false, err
 	}
 
-	return identities+providers+grants == int64(0), nil
+	accessKeys, err := data.Count[models.AccessKey](db.Unscoped())
+	if err != nil {
+		return false, err
+	}
+
+	return identities+providers+grants+accessKeys == 0, nil
 }
 
 // Signup creates a user identity using the supplied name and password and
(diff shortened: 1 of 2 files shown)


internal/access/signup_test.go
```diff
@@ -64,6 +64,34 @@ func TestSignupEnabled(t *testing.T) {
 		assert.Assert(t, !enabled)
 	})
 
+	t.Run("DisabledByInClusterConnector", func(t *testing.T) {
```
jmorganca, line 67: Nice! thanks for adding a test

internal/access/signup.go
```diff
@@ -35,7 +35,12 @@ func SignupEnabled(c *gin.Context) (bool, error) {
 		return false, err
 	}
 
-	return identities+providers+grants == int64(0), nil
+	accessKeys, err := data.Count[models.AccessKey](db.Unscoped())
+	if err != nil {
+		return false, err
+	}
+
+	return identities+providers+grants+accessKeys == int64(0), nil
```
jmorganca, line 43: Is the `int64` conversion necessary?

jmorganca (approved):
small comment otherwise lgtm

internal/access/signup.go
mxyng: I thought it is because it was complaining previously but doesn't seem to be the case anymore

PR #1841, opened 2022-05-03
maintain: restore required email error test

### LINKED ISSUE #1657: Consolidate user and machine identities into a single entity

BruceMacD:
## Summary
A little more clean-up around the identity kind consolidation. Restoring an error test, removing a redundant test and updating a comment.

Restoring

## Checklist

## Related Issues

Resolves #1657

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -679,7 +679,7 @@ func (Server) loadGrant(db *gorm.DB, input Grant) (*models.Grant, error) {
 
 		id = uid.NewGroupPolymorphicID(group.ID)
 
-	// TODO: this gotta go
+	// TODO: remove this when deprecated machines in config are removed
 	case input.Machine != "":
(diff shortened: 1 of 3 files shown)


dnephin (approved):
LGTM

PR #1846, opened 2022-05-03
maintin: Change external "identity" references to "user"

### LINKED ISSUE #1829: Rename `identities` to `users`

BruceMacD:
## Summary
Users is more understandable concept than "identities". With the recent change of removing identity kinds we can now unify the "identity" terminology to "users".

This change leaves internal references (access layer, data, and models) to identity, but changes user-facing references to identities to be "users". This touches the following parts of Infra:

**API**:
- `/v1/identities` is deprecated, we now use `/v1/users`

**Config**:
- `identities:` is deprecated, we now use `users:`.

**CLI**:
- `infra identities ... ` has changed to `infra users ...`
- references to identity in help messages and keys have been updated

All deprecated commands, endpoints, and config will continue to work for now.

## Checklist

## Related Issues

Resolves #1829


internal/server/config.go
```diff
@@ -45,7 +45,8 @@ type Identity struct {
 type Config struct {
 	Providers  []Provider `mapstructure:"providers" validate:"dive"`
 	Grants     []Grant    `mapstructure:"grants" validate:"dive"`
-	Identities []Identity `mapstructure:"identities" validate:"dive"`
+	Identities []User     `mapstructure:"identities" validate:"dive"` // #1829: remove, migrate identities to users
```
ssoroka, line 48: can use config migration for this.

internal/server/handlers.go
```diff
@@ -367,7 +367,12 @@ func (a *API) CreateToken(c *gin.Context, r *api.EmptyRequest) (*api.CreateToken
 }
 
 func (a *API) ListAccessKeys(c *gin.Context, r *api.ListAccessKeysRequest) ([]api.AccessKey, error) {
-	accessKeys, err := access.ListAccessKeys(c, r.IdentityID, r.Name)
+	// #1829: remove, migrate identities to users
+	if r.IdentityID != 0 {
+		r.UserID = r.IdentityID
+	}
```
ssoroka, line 371: can be a request migration

internal/server/routes.go
```diff
@@ -48,13 +48,22 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) *gin.Engine
...
+	get(a, authn, "/v1/users/:id/groups", a.ListUserGroups)
+	get(a, authn, "/v1/users/:id/grants", a.ListUserGrants)
+
+	get(a, authn, "/v1/identities", a.ListUsers)
+	post(a, authn, "/v1/identities", a.CreateUser)
+	get(a, authn, "/v1/identities/:id", a.GetUser)
+	put(a, authn, "/v1/identities/:id", a.UpdateUser)
+	delete(a, authn, "/v1/identities/:id", a.DeleteUser)
+	get(a, authn, "/v1/identities/:id/groups", a.ListUserGroups)
+	get(a, authn, "/v1/identities/:id/grants", a.ListUserGrants)
```
ssoroka, line 66: I'm thinking we should do some kind of path redirect instead of this, which will duplicate the endpoints in the docs. Will discuss with you offline

helm/charts/infra/templates/server/configmap.yaml
```diff
@@ -23,6 +23,8 @@ data:
 {{- end }}
 {{- end }}
 
+    version: {{ .Values.server.config.version }}
```
mxyng, line 26: Instead of templating the config with possibly an older version, it's better to template it to the current version. In this case, merging `identities` and `users` into a single list. This approach is problematic because if anyone has previously defined `identities` through helm can potentially get an unusable system because 1) `users` contains possibly admin and connector credentials and 2) `identities` overwrite `users`

PR #1849, opened 2022-05-03
fix!: remove `kubernetes.` prefix from destinations

### LINKED ISSUE #1814: Replace destination dot notation (e.g. `kubernetes.`)
Feedback from users has been that the `kubernetes.` dot notation is confusing since Infra is designed to support Kubernetes first. Commands such as `infra grant` and `infra use` are especially hard to understand and in some places we don't have the syntax (e.g. the connector helm install)

* Remove `kubernetes.` notation in cluster names
* Add a `kind` field to `/v1/destinations`, starting with `cluster`
* Update `infra list` `infra grant` and `infra use` commands to use this new notation

jmorganca:
* Fixes #1814


ssoroka (approved):
probably alright but i would have liked to see us still prepending `kubernetes.` on the backend.

docs/getting-started/key-concepts.md
```diff
@@ -55,8 +55,8 @@ Groups are sets of identities (usually representing a team or organization) that
 
 Destinations are infrastructure resources that can be accessed via Infra. Some examples of destinations include:
 
-* Kubernetes cluster (e.g. `kubernetes.production`)
-* Kubernetes namespace (e.g. `kubernetes.production.web`)
+* Kubernetes cluster (e.g. `production`)
+* Kubernetes namespace (e.g. `production.web`)
 * Database (e.g. `postgres.production`)
 * Database Table (e.g. `postgres.production.orders`)
 * SSH Server (e.g. `ssh.staging-1`)
```
mxyng, line 62: These need to be updated as well. Probably remove for now?

internal/cmd/cmd.go
```diff
@@ -163,14 +163,13 @@ $ infra use development.kube-system`,
...
 			if parts[0] == "kubernetes" {
-				if len(parts) > 2 {
-					return kubernetesSetContext(parts[1], parts[2])
+				if len(parts) > 1 {
+					return kubernetesSetContext(parts[0], parts[1])
 				}
 
-				return kubernetesSetContext(parts[1], "")
+				return kubernetesSetContext(parts[0], "")
 			}
```
mxyng, line 171: This entire block is no longer necessary

jmorganca: Good catch, thanks!

docs/getting-started/key-concepts.md
jmorganca: Removing for now - good catch!

Issue #1820, opened 2022-05-02
Keep kubeconfig always up to date


jmorganca:
While logged in Kubeconfigs are only updated when users run `infra list` or `infra use`. Kube configs should always be kept up to date while logged in.

PR #1852, opened 2022-05-04
feat: use email name for authinfo inside of kube

### LINKED ISSUE #1851: cli: update kubeconfig to have proper `auth`
Currently `kubectl config get-contexts` prints out:
```
CURRENT   NAME                 CLUSTER              AUTHINFO              NAMESPACE
          docker-desktop       docker-desktop       docker-desktop
*         infra:docker-local   infra:docker-local   infra:docker-local
```

This makes it difficult to see at a glance what infra identity has access to a particular cluster. Instead, we should use the infra identity and share it across the various contexts.

pdevine:
## Summary

Changes AuthInfo inside of kube to use the infra email name for a user

## Checklist

## Related Issues

Resolves #1851

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/spf13/pflag"
 	"gotest.tools/v3/assert"
+	is "gotest.tools/v3/assert/cmp"
 	"gotest.tools/v3/env"
 	"gotest.tools/v3/fs"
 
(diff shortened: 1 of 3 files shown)


internal/cmd/list.go
```diff
@@ -138,5 +143,5 @@ func list(cli *CLI) error {
 		cli.Output("You have not been granted access to any active destinations")
 	}
 
-	return writeKubeconfig(destinations, grants)
+	return writeKubeconfig(identity, destinations, grants)
```
dnephin, line 146: Not directly related to your PR, but I am surprised we update the config as part of `infra list`.

internal/cmd/cmd_test.go
```diff
@@ -180,8 +192,9 @@ XlW7KilKI5YkcszGoPB4RePiHsH+7trf7l8IQq5r5kRq7SKsZ41BI6s1E1PQVW93
 
 		assert.Equal(t, len(kubeconfig.Clusters), 2)
 		assert.Equal(t, len(kubeconfig.Contexts), 2)
-		assert.Equal(t, len(kubeconfig.AuthInfos), 2)
+		assert.Equal(t, len(kubeconfig.AuthInfos), 1)
```
dnephin, line 195: We should update the example config to include an `AuthInfos` that we are not modifying, to show that we preserve other entries. 

I think that's why all the other cases here have 2 entries.

dnephin (commented):
Looks good, just one suggestion for the tests.

internal/cmd/list.go
BruceMacD: It's to make sure the local config reflects whatever a users most recent access is. It may change for #1820

internal/cmd/cmd_test.go
pdevine: oh good point. I'll make another issue for that.

PR #1853, opened 2022-05-04
improve: feedback on grant roles

kimskimchi:
## Summary

Gives user feedback if the grant is an invalid one for infra, or the proper links if it is kubernetes. 

- API returns bad request with invalid roles for infra
- CLI links user to infra documentation if role is a default expected one (admin, view, etc.)
- CLI warns user if it is not one of the defaults, and links to the kubernetes [article](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles) 

## Checklist

## Related Issues

Resolves #
Related to #1394

### DIFF
--- a/docs/guides/granting-access.md
+++ b/docs/guides/granting-access.md
@@ -2,7 +2,7 @@
 
 ## Roles
 
-Infra allows granting different levels of access via **roles**, such as `view`, `edit` or `admin`. Each connector has different roles that can be used:
+Infra allows granting different levels of access via **roles**, such as `view`, `connector` or `admin`. Each connector has different roles that can be used:
 
 - [Kubernetes Roles](../connectors/kubernetes.md#roles)
 
--- a/docs/guides/granting-access.md
+++ b/docs/guides/granting-access.md
@@ -2,7 +2,7 @@
 
 ## Roles
 
-Infra allows granting different levels of access via **roles**, such as `view`, `connector` or `admin`. Each connector has different roles that can be used:
+Infra allows granting different levels of access via the **roles** `view` or `admin`. Each connector has different roles that can be used:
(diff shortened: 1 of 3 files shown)


docs/guides/granting-access.md
```diff
@@ -2,7 +2,7 @@
 
 ## Roles
 
-Infra allows granting different levels of access via **roles**, such as `view`, `edit` or `admin`. Each connector has different roles that can be used:
+Infra allows granting different levels of access via **roles**, such as `view`, `connector` or `admin`. Each connector has different roles that can be used:
```
BruceMacD, line 5:
```suggestion
Infra allows granting different levels of access via the **roles** `view` or `admin`. Each connector has different roles that can be used:
```

Good catch on these docs, we don't need to document the connector role, it should be handled internally at the moment.

internal/cmd/grants.go
```diff
@@ -222,6 +222,15 @@ func addGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 
 	cli.Output("Access granted!")
 
+	// Warn user if destination is kubernetes that role is not validated
+	if cmdOptions.Destination != models.InternalInfraProviderName {
```
BruceMacD, line 226: Not sure if we should output this every time a user runs this command. Maybe we could throw the information about roles in the `grants add` help section for now?

kimskimchi: I was attempting to give feedback to the user when it wasn't one of the four mentioned in our docs - "hey, this is not one of the defaults we talked about, so make sure you meant this"

if I add it in grants add, then I believe message might look like:
`don't recognize this grant, for more information, see grants add --help"

Issue #1860, opened 2022-05-04
Fork snowflake bwmarrin/snowflake


dnephin:
We use this library for all the IDs in our API. The library is no longer maintained, and has a number of bugs that we need fixed.

If we delete all the parts we don't need, the code we actually use is under 200 lines. You can see that here: https://github.com/dnephin/snowid/blob/main/snowid.go. In this fork I've added a bunch of fuzz tests, and fixed a few bugs.

We could move this fork into the `infrahq/` org on GitHub , but I think for such a small dependency that is so central to our API, it would be better off as part of the infra project. 

I propose we move it into the `uid` package , which will allow us to remove most of the existing `snowflake.go` as well. 

Maybe in the future clients will want to generate their own IDs. If there's a need we can extract it into a library, but for now it's going to be easier to manage it inline. Putting it inline removes the need to setup separate CI, and makes it easier for us change.

PR #1865, opened 2022-05-04
feat(api): remove polymorphic ids from grant responses

### LINKED ISSUE #1859: api: remove polymorphic IDs
Polymorphic IDs is a micro-format which is difficult for users to understand, parse and serialize.

mxyng:
## Summary

Remove the the `PolymorphicID` field since it requires the user dereference IDs returned by the Grant API multiple ways, first to dereference the ID to either a group or user ID, then dereference the ID to an actual resource. Instead, return the ID in their respect fields, `Group` if the polymorphic ID points to a group resource and `Identity` if the polymorphic ID points to an identity resource.

API migration will be added in #1884

## Checklist

## Related Issues

Resolves #1859


uid/snowflake.go
```diff
@@ -29,6 +29,10 @@ func New() ID {
 }
 
 func (u ID) String() string {
+	if u == 0 {
```
mxyng, line 32: This should be superseded by #1860 which fixes a number of snowflake related issues

internal/cmd/grants.go
```diff
@@ -60,19 +60,35 @@ func newGrantsListCmd(cli *CLI) *cobra.Command {
 
 			type row struct {
 				Identity string `header:"IDENTITY"`
-				Access   string `header:"ACCESS"`
+				Access   string `header:"ROLE"`
```
BruceMacD, line 63: might wanna hold off on updating this one until we have a design in place for what grants should be exposed like, just to keep things in sync

internal/cmd/grants.go
```diff
@@ -105,10 +121,10 @@ $ infra grants remove janedoe@example.com kubernetes.docker-desktop
 # Remove all grants of a group in a destination
 $ infra grants remove group-a kubernetes.staging --group
 
-# Remove a specific grant 
+# Remove a specific grant
 $ infra grants remove janedoe@example.com kubernetes.staging --role viewer
 
-# Remove access to infra 
+# Remove access to infra
```
BruceMacD, line 127:
```suggestion
# Remove admin access to Infra
```

BruceMacD (approved):
Looks good to me, just comments around unrelated docs

dnephin (commented):
I like this approach. I think the additional structure of a `Subject` or `SubjectGrant` type (with the identity and group fields on that type) would make the API more obvious, but this works as well.

api/grant.go
```diff
@@ -11,19 +11,22 @@ type Grant struct {
...
+	Privilege string `json:"privilege" note:"a role or permission"`
+	Resource  string `json:"resource" note:"a resource name in Infra's Universal Resource Notation"`
 }
 
 type ListGrantsRequest struct {
-	Subject   uid.PolymorphicID `form:"subject"`
-	Resource  string            `form:"resource" example:"kubernetes.production"`
-	Privilege string            `form:"privilege" example:"view"`
+	Identity  uid.ID `json:"identity"`
+	Group     uid.ID `json:"group"`
```
dnephin, line 22: Oops, this comment didn't post for some reason.

I think we need the same `required_without` validation here. A query can't include both ofthese.

mxyng: This validation is slightly different since neither are required. I think the semantics you're looking for is that one of Identity or Group _can_ be specified but not both. From what I see in the validator docs, that's not an option so it will require custom validator. I'd suggest punting that to a separate issue

dnephin: Ahh yes, that's right, it should be "at most one of".

Does `excluded_if` not work for this?

mxyng: Ah yes that should work. I misread the description and thought it meant something else

PR #1866, opened 2022-05-04
fix: remove and update helm links

### LINKED ISSUE #1856: helm/charts/infra/values.yaml has outdated URLs in the comments

BruceMacD:
## Summary
Removing dead links from helm chart. Updating where possible. 

## Checklist

## Related Issues

Resolves #1856

### DIFF
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -50,8 +50,6 @@ global:
   #     name: secret-name
 
 # Default server configurations
-# Ref: https://github.com/infrahq/infra/blob/main/helm/charts/infra/charts/server/README.md
-
 server:
 
   ## Deploy the server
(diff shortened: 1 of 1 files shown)


helm/charts/infra/values.yaml
```diff
@@ -50,8 +50,6 @@ global:
   #     name: secret-name
 
 # Default server configurations
-# Ref: https://github.com/infrahq/infra/blob/main/helm/charts/infra/charts/server/README.md
```
BruceMacD, line 53: Not sure what was here, seems like it was empty

helm/charts/infra/values.yaml
```diff
@@ -384,7 +382,6 @@ server:
   affinity: {}
 
   ## Primary server configuration
-  ## Ref: https://github.com/infrahq/infra/blob/main/docs/cli.md#infra-server
```
BruceMacD, line 387: This CLI documentation is gone now that the command is hidden

mxyng (approved):

PR #1867, opened 2022-05-04
uid: inline snowflake ID library

### LINKED ISSUE #1860: Fork snowflake bwmarrin/snowflake

dnephin:
The library we were using is unmaintained. Once we remove all of the stuff we don't use the code is less than 200 lines.

This commit also includes a few bug fixes:

- leading 1s in the base58 encoded ID are no longer valid. Those characters would translate into leading zeros
- `ID(0)` now encodes to the empty string instead of `"1"`
- overflows are now errors
- negative numbers are now errors

The full history of the changes from the library can be seen here: https://github.com/dnephin/snowid/commits/main

Mostly I deleted stuff we don't use (other base encodings), and replaced the tests. I added better unit and fuzz tests, but left the benchmarks unmodified. We can revisit the benchmarks when it's important.  I'll leave a few more comments inline.

Resolves #1860


uid/snowid.go
```diff
@@ -0,0 +1,192 @@
...
+
+	n.mu.Unlock()
+	return r
+}
+
+// Bytes returns a byte array of the base58 encoded value for this ID.
+func (id ID) Bytes() []byte {
+	switch {
+	case id <= 0:
+		return nil
```
dnephin, line 118: This is new, any negative numbers or 0 now encode to the empty string.

uid/snowid.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// Parse parses a base58 encoded value into an ID.
+func Parse(b []byte) (ID, error) {
+	switch {
+	case bytes.HasPrefix(b, []byte("1")):
+		return -1, fmt.Errorf("invalid base58: not in canonical form")
+	case len(b) > 11:
+		return -1, fmt.Errorf("invalid base58: too long")
+	}
```
dnephin, line 149: This is new. We had the length check in our wrapper around `ID`. Leading `"1"` characters are invalid because they decode into `0`, and generally leading `0` are omitted.

uid/snowid.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		shifted, ok := multiplyCheckOverflow(id, 58)
+		if !ok {
+			return -1, fmt.Errorf("invalid base58: value too large")
+		}
+		id = shifted + int64(decodeBase58Map[b[i]])
+		if id <= 0 {
+			return -1, fmt.Errorf("invalid base58: value too large")
+		}
```
dnephin, line 164: These two overflow checks are new. It took me a while to figure out the multiply overflow because it often doesn't result in a negative number. Eventually I found enough examples using the fuzz test to understand what was happening.

uid/snowid_test.go
```diff
@@ -0,0 +1,342 @@
...
+
+	b.ReportAllocs()
+
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		_ = node.Generate()
+	}
+}
+
+func TestParse(t *testing.T) {
```
dnephin, line 184: This is mostly the test that we already had for our wrapper around `ID`.

uid/snowid_test.go
```diff
@@ -0,0 +1,342 @@
...
+	for i := 0; i < 1000000; i++ {
+		y = node.Generate()
+		if x == y {
+			t.Errorf("x(%d) & y(%d) are the same", x, y)
+		}
+		x = y
+	}
+}
+
+func TestNode_Generate_ConcurrencyAndUniqueness(t *testing.T) {
```
dnephin, line 40: This is a new property test, to show that IDs are unique, and to catch any concurrency bugs in `NewNode`.

uid/snowid_test.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.base58, func(t *testing.T) {
+			run(t, tc)
+		})
+	}
+}
+
+func FuzzParse_NoPanic(f *testing.F) {
```
dnephin, line 244: The 3 fuzz tests below here are new, and helped find the bugs that were fixed.

In CI it runs the seed cases. As a follow up I'd like to add a nightly job that runs the fuzzer to find new interesting cases.

internal/server/data/migrations.go
```diff
@@ -94,7 +94,7 @@ func migrate(db *gorm.DB) error {
 					key.IssuedFor = strings.TrimPrefix(key.IssuedFor, "u:")
 					key.IssuedFor = strings.TrimPrefix(key.IssuedFor, "m:")
 
-					iss, err := uid.ParseString(key.IssuedFor)
+					iss, err := uid.Parse([]byte(key.IssuedFor))
```
BruceMacD, line 97: What's the advantage of getting rid of `ParseString`?  Is converting to the byte slice more standard?

BruceMacD (commented):
This is looking good to me so far, just a couple questions

uid/snowid.go
```diff
@@ -0,0 +1,192 @@
+package uid
```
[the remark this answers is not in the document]
dnephin: Ya, good question. Everything here should be from the upstream repo. The changes I made (that I haven't noted yet) are:

* unexported `epoch`, `nodeBits`, `stepBits`, so they can't be modified outside this package
* `epoch` is set to the value we were using (instead of the twitter value which was previously the default)
* `Bytes` and `String` now returned the base58 encoded ID. In the upstream they would return the int64 value, and our wrapper defined a different `String`. So now the `String` matches what our wrapper did (return the base58 encoded value).
* method receiver was renamed from `f` to `id`
* `MarshalJSON` and `UnmarshalJSON` were replaced by `MarshalText` and `UnmarshalText` (which we had in our wrapper). The JSON decode/encoder knows to call those, and removes the need for us to handle the `""`.

That should be it. Everything else is noted below, or was a change to the tests. There were also lots of things deleted, but we don't use those things.

uid/snowid.go
dnephin: The generation should not have this problem. It should never generate an `int64` that is out of range. We don't have a proper fuzz test for that yet. I think the `if < 0` check in `TestNode_Generate_ConcurrencyAndUniqueness` should confirm that generation always generates valid IDs that don't overflow. We could add a fuzz test for it as well.

internal/server/data/migrations.go
dnephin: My main motivation for removing this is that I always try to avoid functions that do almost the same thing. It becomes hard to use a package if there are a bunch of similar methods with only vary slight differences in args. If we expect `ParseString` to be called in a lot of places, I might be worth it to add.

I don't expect us to have to parse IDs in many places. Most of the places that I changed were in tests. Most of the time I'd expect JSON encode/decode to call parse for us, and it's pretty easy to cast a string to `[]byte` when we do need to call it.

I'm not against restoring `ParseString` if we think it'll be used more often.

BruceMacD (approved):

pdevine (approved):
looks good.

PR #1872, opened 2022-05-04
feat: add resources/roles to destinations api

### LINKED ISSUE #1818: Detect and verify roles for clusters
Infra should detect (or automatically create) roles for a cluster and publish them to users. Today users need to guess (i.e. know which roles ahead of time can be assigned) which creates room for errors if users select the wrong names.

mxyng:
## Summary

Update the destinations API to allow the connector to report its capabilities (e.g. resources and roles)

## Checklist

## Related Issues

Resolves #1818

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -246,6 +247,10 @@ func addGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 		}
 	}
 
+	if err := checkResourcesPrivileges(client, cmdOptions.Destination, cmdOptions.Role); err != nil {
+		return err
+	}
+
 	_, err = client.CreateGrant(&api.CreateGrantRequest{
 		User:      user,
 		Group:     group,
@@ -261,7 +266,55 @@ func addGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 	return nil
 }
 
-// userOrGroupByName gets the ID of the identity or group to be associated with the grant
+func checkResourcesPrivileges(client *api.Client, resource, privilege string) error {
+	parts := strings.SplitN(resource, ".", 2)
+	destination := parts[0]
+	subresource := ""
+
+	if len(parts) > 1 {
+		subresource = parts[1]
+	}
+
+	supportedResources := make(map[string]struct{})
+	supportedRoles := make(map[string]struct{})
+
+	if destination != "infra" {
+		destinations, err := client.ListDestinations(api.ListDestinationsRequest{Name: destination})
+		if err != nil {
+			return err
+		}
+
+		if destinations.Count == 0 {
+			return fmt.Errorf("unknown destination %q", destination)
+		}
+
+		for _, d := range destinations.Items {
+			for _, r := range d.Resources {
+				supportedResources[r] = struct{}{}
+			}
+
+			for _, r := range d.Roles {
+				supportedRoles[r] = struct{}{}
+			}
+		}
+
+		if subresource != "" {
+			if _, ok := supportedResources[subresource]; !ok {
+				return fmt.Errorf("unknown resource %q for %q", subresource, destination)
+			}
+		}
+
+		if privilege != "connect" {
+			if _, ok := supportedRoles[privilege]; !ok {
+				return fmt.Errorf("unknown role %q for %q", privilege, destination)
+			}
+		}
+	}
+
+	return nil
+}
+
+// identityOrGroupByName gets the ID of the identity or group to be associated with the grant
 func userOrGroupByName(client *api.Client, subject string, isGroup bool) (uid.ID, uid.ID, error) {
 	if isGroup {
 		group, err := GetGroupByName(client, subject)
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -481,31 +481,48 @@ func Run(ctx context.Context, options Options) error {
 		endpoint := fmt.Sprintf("%s:%d", host, port)
 		logging.S.Debugf("connector serving on %s", endpoint)
 
-		if destination.ID == 0 {
-			destination.Connection.CA = string(caCertPEM)
-			destination.Connection.URL = endpoint
+		namespaces, err := k8s.Namespaces()
+		if err != nil {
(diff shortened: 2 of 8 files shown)


internal/kubernetes/kubernetes.go
```diff
@@ -589,3 +589,26 @@ func (k *Kubernetes) IsServiceTypeClusterIP() (bool, error) {
...
+	}
+
+	clusterRoles, err := clientset.RbacV1().ClusterRoles().List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	results := make([]string, 0, len(clusterRoles.Items))
+	for _, n := range clusterRoles.Items {
+		if strings.HasPrefix(n.Name, "system:") {
```
jmorganca, line 606: Love the filtering – wondering if we should do more – i.e. perhaps anything with a colon ? or perhaps for now even an "allow" list that includes `view`, `edit`, `admin`, `cluster-admin` so we can control the experience. Alternatively we can filter on the "front-end" (i.e. ui, cli, api handlers)

internal/connector/connector.go
```diff
@@ -619,8 +624,12 @@ func Run(ctx context.Context, options Options) error {
 	return tlsServer.ListenAndServeTLS("", "")
 }
 
-// createDestination creates a destination in the infra server if it does not exist
+// createDestination creates a destination in the infra server if it does not exist and updates it if it does
 func createDestination(client *api.Client, local *api.Destination) error {
```
jmorganca, line 628: nit: rename to `createOrUpdate` or something more upsert-y?

internal/kubernetes/kubernetes.go
dnephin: Does kubernetes set anything in `ClusterRole.ObjectMeta.Annotations` or `ClusterRole.ObjectMeta.Labels` that would identify this as a system role?

dnephin (commented):
Nice, I like this approach! Some thoughts on how we might encode a list in the db

internal/server/models/destination.go
```diff
@@ -12,6 +14,9 @@ type Destination struct {
 
 	ConnectionURL string
 	ConnectionCA  string
+
+	Resources string
+	Roles     string
```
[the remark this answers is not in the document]
mxyng: I like the suggestion. Agreed we shouldn't transform the data if possible. I'll take a look at your suggestions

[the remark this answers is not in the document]
mxyng: @dnephin pointed out we've implemented a type `CommandSeparatedStrings` for exactly this use case. I think it'll be sufficient for the time being

internal/kubernetes/kubernetes.go
pdevine: We definitely don't want to show those cluster roles normally, but is there a use case for admins wanting to set those for a different reason?

internal/server/models/destination.go
```diff
@@ -12,6 +12,9 @@ type Destination struct {
 
 	ConnectionURL string
 	ConnectionCA  string
+
+	Resources CommaSeparatedStrings
+	Roles     CommaSeparatedStrings
```
jmorganca, line 17: I think @dnephin had a comment that was resolved around this, but why not use the native `json` type to store the slice? Less code for us to maintain and we may be able to query those fields now.

ssoroka:
I'd like to take a different approach vs this. Instead we populate key-value lists of things (eg namespaces) along with what parent(s) they belong to (eg which cluster), the UI then just asks the user to select from each list, and then we know the unique reference to the resource. We're going to need this to support resource-based access controls, and I'd prefer to do it sooner rather than later.

ssoroka (approved):
seems like resources should maybe have its own endpoint. likely this list can get huge.

internal/cmd/grants.go
```diff
@@ -230,6 +231,10 @@ func addGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 		}
 	}
 
+	if err := checkResourcesPrivileges(client, cmdOptions.Destination, cmdOptions.Role); err != nil {
```
mxyng, line 234: This is a nice place to add a `force` flag to ignore nonexistent destinations, resources, or roles

internal/connector/connector.go
```diff
@@ -481,31 +481,42 @@ func Run(ctx context.Context, options Options) error {
...
 
-			err = createDestination(client, destination)
-			if err != nil {
-				logging.S.Errorf("initializing destination: %v", err)
-				return
-			}
-		} else if destination.Connection.URL != endpoint || destination.Connection.CA != string(caCertPEM) {
+			fallthrough
+
+		case !slicesEqual(destination.Resources, namespaces), !slicesEqual(destination.Roles, clusterRoles):
```
jmorganca, line 509: This `switch` control flow is confusing: could we just use regular `if` statements ?

E.g.
```
if destination.ID == 0 {
  // create a destination
} else {
 // update the destination if fields have changed
}
```

Not urgent: but this goroutine is accessing an unprotected variable `destination`. While in practice it shouldn't be an issue we should move it in this goroutine

internal/cmd/grants.go
```diff
@@ -245,6 +250,54 @@ func addGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
...
+			return err
+		}
+
+		if destinations.Count == 0 {
+			return fmt.Errorf("unknown destination %q", destination)
+		}
+
+		for _, d := range destinations.Items {
+			for _, r := range d.Resources {
+				supportedResources[r] = struct{}{}
```
jmorganca, line 277: If it's just one resource and one role, we don't need to build a map, you can just use a single `bool` variable for each

internal/connector/connector.go
mxyng: The switch allows more concise code. It's not straight forward since multiple things are checked and updated before calling the API. `destination` is only after modified in the goroutine after being initialized. It's important its available for all goroutines as it caches the destination object which is used to determine when to call create or update

internal/cmd/grants.go
mxyng: The map is for multiple destinations that may match the name query. Since the returned value is a list of destinations, it's better to build and use this set

internal/connector/connector.go
mxyng: Created #1942

jmorganca (approved):

PR #1880, opened 2022-05-05
maintain: refactor printTable to use CLI

mxyng:
## Summary

CLI provides an interface for CLI outputs and is an appropriate place to put table printing


mxyng:
Not going to do this right now.

PR #1883, opened 2022-05-05
api: remove unused method receiver and unused method

dnephin:
Following up on #1874

I noticed that now `sendAPIError` doesn't need to be a method anymore (it doesn't use `API`).

Also noticed the `Telemetry.Group` method was not being used. Should it be, or is it safe to remove for now?


ssoroka (approved):
seems fine. it was only there for telemetry iirc.

ssoroka:
`Telemetry.Group` can be removed if we never want to track groups. I'm not the right person to answer that.

dnephin:
Sounds good. I'll restore it from git if we end up needing it

PR #1884, opened 2022-05-05
feat: add api migration

mxyng:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/apimigrator.go
+++ b/internal/server/apimigrator.go
@@ -38,7 +38,7 @@ func addRequestRewrite[oldReq any, newReq any](a *API, method, path, version str
 		path:    path,
 		version: version,
 		requestRewrite: func(c *gin.Context) {
-			reqVer := NewVersion(c.Request.Header.Get("VERSION"))
+			reqVer := NewVersion(c.Request.Header.Get("Infra-Version"))
 			if reqVer.GreaterThanStr(version) {
 				c.Next()
 				return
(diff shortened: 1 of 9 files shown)


ssoroka (approved):

internal/server/apimigrator.go
```diff
@@ -0,0 +1,264 @@
...
+			if reqVer.GreaterThanStr(version) {
+				c.Next()
+				return
+			}
+
+			oldReqObj := new(oldReq)
+
+			err := bind(c, oldReqObj)
+			if err != nil {
+				sendAPIError(c, err)
```
BruceMacD, line 51: We should probably add a trace to this error or it will cause a lot of pain to troubleshoot

internal/server/apimigrator.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+		}
+		if fieldname, ok := t.Field(i).Tag.Lookup("json"); ok {
+			body[fieldname] = f.Interface()
+		}
+	}
+	c.Request.URL.RawQuery = query.Encode()
+
+	if c.Request.Method != http.MethodGet {
+		bodyJSON, err := json.Marshal(body)
```
BruceMacD, line 107: We shouldn't need to get the request body for delete requests either

BruceMacD (approved):

internal/server/apimigrator.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func addRequestRewrite[oldReq any, newReq any](a *API, method, path, version string, f func(oldReq) newReq) {
+	a.migrations = append(a.migrations, apiMigration{
+		method:  method,
+		path:    path,
+		version: version,
+		requestRewrite: func(c *gin.Context) {
+			reqVer := NewVersion(c.Request.Header.Get("VERSION"))
+			if reqVer.GreaterThanStr(version) {
```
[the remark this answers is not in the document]
ssoroka: no, not if we're setting the migration version to the last version that needed the patch.

PR #1888, opened 2022-05-05
maintain: remove unused BindGroupIdentities

dnephin:
## Summary

While working on `ListGroups` I came across `data.BindGroupIdentities` which was only used in tests. I suspect this function was replaced by `data.AssignIdentityToGroups` at some point.

This PR removes `data.BindGroupIdentities` and the tests for it. The one test of other functionality that used this function was changed to use `data.AssignIdentityToGroups` to better mirror production code paths.

### DIFF
--- a/internal/access/access_test.go
+++ b/internal/access/access_test.go
@@ -94,14 +94,29 @@ func TestBasicGrant(t *testing.T) {
 }
 
 func TestUsersGroupGrant(t *testing.T) {
+	models.SkipSymmetricKey = true
+	t.Cleanup(func() {
+		models.SkipSymmetricKey = false
+	})
 	db := setupDB(t)
-	err := data.CreateIdentity(db, tom)
+
+	tom = &models.Identity{Name: "tom@infrahq.com"}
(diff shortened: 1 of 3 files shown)


internal/access/access_test.go
```diff
@@ -94,14 +94,29 @@ func TestBasicGrant(t *testing.T) {
...
 func TestUsersGroupGrant(t *testing.T) {
+	models.SkipSymmetricKey = true
+	t.Cleanup(func() {
+		models.SkipSymmetricKey = false
+	})
 	db := setupDB(t)
-	err := data.CreateIdentity(db, tom)
+
+	tom = &models.Identity{Name: "tom@infrahq.com"}
+	tomsGroup = &models.Group{Name: "tom's group"}
```
dnephin, line 104: These were defined at the package level previously. I think it's pretty risky to re-use these in tests because they are pointers that get updated by each test. It's also very little extra work to copy them into the test that uses them, so I did that for this one test I was changing. There are a few other tests that would probably benefit from the same cleanup.

mxyng (approved):

BruceMacD (approved):

PR #1892, opened 2022-05-06
improve: small improvements to uid tests

dnephin:
## Summary
Some minor cleanup and improvements to the tests in the `uid` package. Follow up to  #1867

### DIFF
--- a/uid/snowid.go
+++ b/uid/snowid.go
@@ -11,7 +11,7 @@ import (
 
 var (
 	// epoch is the baseline date used for generating IDs.
-	epoch = time.Date(2022, 1, 1, 0, 0, 0, 0, time.UTC).UnixMilli()
+	epoch = time.Date(2022, 1, 1, 0, 0, 0, 0, time.UTC)
 
 	// nodeBits holds the number of bits to use for Node
 	// Remember, you have a total 22 bits to share between Node/Step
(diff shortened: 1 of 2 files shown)


uid/snowid.go
```diff
@@ -74,7 +74,7 @@ func NewNode(node int64) (*Node, error) {
 
 	var curTime = time.Now()
 	// add time.Duration to curTime to make sure we use the monotonic clock if available
-	n.epoch = curTime.Add(time.Unix(epoch/1000, (epoch%1000)*1000000).Sub(curTime))
+	n.epoch = curTime.Add(epoch.Sub(curTime))
```
dnephin, line 77: Previously the epoch was an int. We used a `time.Date` to define it (which is much nicer). So now we can use the date directly instead of converting it to an int and back to a date.

uid/snowid_test.go
```diff
@@ -46,12 +47,15 @@ func TestNode_Generate_ConcurrencyAndUniqueness(t *testing.T) {
...
-	nodeMax := -1 ^ (-1 << nodeBits)
+	nodeMax := -1 ^ (-1 << nodeBits) // modeMax=1024
 	nodeIDs := make([]int64, nodeMax)
 	for n := 0; n < nodeMax; n++ {
 		nodeIDs[n] = int64(n)
 	}
-	rand.Shuffle(nodeMax, func(i, j int) {
+	seed := time.Now().UnixNano()
+	t.Log("seed", seed)
+	r := rand.New(rand.NewSource(seed))
```
dnephin, line 57: Use a different seed each time so that we actually get a random shuffle. Without this it is likely the random would always produce the same results because it gets seeded to the same value (`1`).

Also print the seed to the test log. If this test fails, that gives us an easy way to reproduce (by  hardcoding that seed value).

jmorganca (approved):

PR #1893, opened 2022-05-06
fix: `created_at` no longer zeros on save

### LINKED ISSUE #1847: api: new destinations have `created: null`
After connecting a new cluster, the `created` field is set to null

```
[{"id":"6ZZpojwtko","uniqueID":"4a5f5c496514a778cd70eb14cf54569ed712c1d395a006333ed351fbc0a26578","name":"testing","created":null,"updated":"2022-05-03T23:25:06Z" ... }}]
```

jmorganca:
Fixes #1847

### DIFF
--- a/internal/server/models/model.go
+++ b/internal/server/models/model.go
@@ -19,7 +19,8 @@ type Model struct {
 	ID uid.ID
 	// CreatedAt is set by GORM to time.Now when a record is first created.
 	// See https://gorm.io/docs/conventions.html#Timestamp-Tracking
-	CreatedAt time.Time
+	// gorm:"<-:create" allows read and create, but not updating
+	CreatedAt time.Time `gorm:"<-:create"`
 	// UpdatedAt is set by GORM to time.Now() when a record is updated.
 	// See https://gorm.io/docs/conventions.html#Timestamp-Tracking
 	UpdatedAt time.Time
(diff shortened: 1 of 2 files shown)


internal/server/models/model.go
```diff
@@ -19,7 +19,8 @@ type Model struct {
 	ID uid.ID
 	// CreatedAt is set by GORM to time.Now when a record is first created.
 	// See https://gorm.io/docs/conventions.html#Timestamp-Tracking
-	CreatedAt time.Time
+	// gorm:"<-:create" allows read and create, but not updating
+	CreatedAt time.Time `gorm:"<-:create"`
```
mxyng, line 23: This is a side effect of using `Save` as opposed to `Update`. `Save` saves all fields regardless if they've been initialized while `Update` will only save the initialized fields. It's fine to make this change for `CreatedAt` since it only needs to be set once but ideally this is fixed at the source to avoid running into the same problem else where

dnephin (commented):
Looks good! A test case that reproduces the problem would be awesome. I guess we don't need to test every model, but a test for one would be great.

internal/server/models/model.go
jmorganca: Fair, although our PUT semantics are to replace every field, and this change lines up nicely with the server being the authoritative source for `CreatedAt`. `Updates()` will only replace non-zero fields

jmorganca:
@dnephin great call. Added a test case for Destination (the resource where we found this)

BruceMacD (approved):

ssoroka (approved):

PR #1894, opened 2022-05-06
fix: polymorphic ID api migration

### LINKED ISSUE #1859: api: remove polymorphic IDs

mxyng:
## Summary

Mutate the APIs requesting/responding with polymorphic IDs for backwards compat

## Checklist

## Related Issues

Resolves #1859

### DIFF
--- a/api/login.go
+++ b/api/login.go
@@ -21,7 +21,7 @@ type LoginRequest struct {
 }
 
 type LoginResponse struct {
-	ID                     uid.ID `json:"id"`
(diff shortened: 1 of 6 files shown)


ssoroka (changes requested):
Maybe not directly related to this migration, but I'm hoping we can change the new login response

ssoroka (approved):

BruceMacD (approved):

PR #1909, opened 2022-05-09
fix: golint 1.6.0 errors

BruceMacD:
## Summary
The lint runner is running the new golint version (1.6.0) and it is detecting some new warnings. This PR cleans up these problems.

## Checklist

### DIFF
--- a/internal/cmd/about.go
+++ b/internal/cmd/about.go
@@ -40,15 +40,12 @@ func newTriangle(a, b, c *point3D, ch rune) *triangle {
 }
 
 func (t *triangle) Draw(surf *sprite.Surface) {
-	err := surf.Triangle(
+	// nolint
+	surf.Triangle(
 		t.A.ScreenX(), t.A.ScreenY(),
 		t.B.ScreenX(), t.B.ScreenY(),
 		t.C.ScreenX(), t.C.ScreenY(),
 		t.Color, true)
-	//nolint
-	if err != nil {
-		// don't do anything
-	}
 }
 
 type point3D struct {
--- a/internal/cmd/about.go
+++ b/internal/cmd/about.go
@@ -45,6 +45,7 @@ func (t *triangle) Draw(surf *sprite.Surface) {
 		t.B.ScreenX(), t.B.ScreenY(),
 		t.C.ScreenX(), t.C.ScreenY(),
 		t.Color, true)
+	//nolint
 	if err != nil {
 		// don't do anything
 	}
@@ -137,8 +138,10 @@ type scrollText struct {
 }
 
 func newScrollText(width, height int) *scrollText {
-	s := &scrollText{BaseSprite: sprite.BaseSprite{
-		Visible: true},
+	s := &scrollText{
+		BaseSprite: sprite.BaseSprite{
+			Visible: true,
+		},
 		TimeOut: 2,
 	}
 	f := sprite.NewJRSMFont()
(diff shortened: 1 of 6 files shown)


internal/cmd/about.go
```diff
@@ -45,6 +45,7 @@ func (t *triangle) Draw(surf *sprite.Surface) {
 		t.B.ScreenX(), t.B.ScreenY(),
 		t.C.ScreenX(), t.C.ScreenY(),
 		t.Color, true)
+	//nolint
```
BruceMacD, line 48: This is an intentionally empty branch

internal/cmd/keys.go
```diff
@@ -157,7 +157,7 @@ func newKeysListCmd(cli *CLI) *cobra.Command {
 				return err
 			}
 
-			keys := &api.ListResponse[api.AccessKey]{}
+			keys := &api.ListResponse[api.AccessKey]{} // nolint
```
BruceMacD, line 160: False positive: was being detected as unused but `keys.Items` is used

internal/cmd/local_server.go
```diff
@@ -106,7 +109,7 @@ func formatURLMsg(msg string, fallback string) string {
 	}
 
 	formatted := strings.ReplaceAll(msg, "_", " ")
-	formatted = strings.Title(strings.ToLower(formatted))
+	formatted = cases.Title(language.English).String(formatted)
```
BruceMacD, line 112: Go 1.18 deprecated `strings.Title`

jmorganca:
Thanks for fixing this!

internal/cmd/about.go
BruceMacD: That worked, updated

mxyng (approved):

internal/cmd/about.go
```diff
@@ -40,14 +40,12 @@ func newTriangle(a, b, c *point3D, ch rune) *triangle {
 }
 
 func (t *triangle) Draw(surf *sprite.Surface) {
-	err := surf.Triangle(
+	// nolint
+	surf.Triangle(
```
dnephin, line 44: I think this should work to indicate "ignore error":
```suggestion
	_ = surf.Triangle(
```

internal/cmd/keys.go
dnephin: I don't think this is a false positive. The value of `keys` is never used here, it's being overwritten.

This should change to just a variable declaration:
```suggestion
			var keys *api.ListResponse[api.AccessKey]
```

dnephin (commented):
Thanks for fixing these! A couple suggestions for the fixes

internal/cmd/about.go
BruceMacD: This worked 👍

internal/cmd/about.go
```diff
@@ -40,14 +40,12 @@ func newTriangle(a, b, c *point3D, ch rune) *triangle {
 }
 
 func (t *triangle) Draw(surf *sprite.Surface) {
-	err := surf.Triangle(
+	// nolint
```
dnephin, line 43:
```suggestion
```

Do we still need this?

dnephin (approved):

internal/cmd/about.go
dnephin: Not a big deal, it's fine to leave it

PR #1910, opened 2022-05-09
improve: confirm password mismatch returns to first password prompt

kimskimchi:
## Summary

If password confirm does not match the first one, right now it prompts only the second one. 
This change remprompts both: 

**Old:**
```
  Password: abc
  Confirm password: abcd
Passwords do not match. 
  Confirm password: abcd
```

**Now:**
```
  Password: abc
  Confirm password: abcd
Passwords do not match. Please try again.
  Password: ...
  Confirm password: ...
```

**Reason**
During admin signup, if user mistypes first password, they have no clear way to fix it. 

**Additional small changes**
- merges `promptUpdatePassword` and `promptConfirmPassword` because it only included a message `fmt.println`. Moved that msg out to the respective functions that call it, since the message is different each time.

### DIFF
--- a/internal/cmd/identities.go
+++ b/internal/cmd/identities.go
@@ -36,10 +36,10 @@ func newIdentitiesCmd(cli *CLI) *cobra.Command {
 func newIdentitiesAddCmd(cli *CLI) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "add IDENTITY",
-		Short: "Create an identity.",
-		Long: `Create an identity.
+		Short: "Create an identity",
+		Long: `Create an identity
 
-A new user identity must change their one time password before further usage.`,
+Note: A new user identity must change their one time password before further usage.`,
 		Args: ExactArgs(1),
 		Example: `# Create a user
 $ infra identities add johndoe@example.com`,
@@ -228,7 +228,8 @@ func UpdateIdentity(name string, cmdOptions editIdentityCmdOptions) error {
 	}
 
 	if cmdOptions.Password {
-		req.Password, err = promptUpdatePassword("")
+		fmt.Fprintf(os.Stderr, "  Enter a new one time password (min length 8):\n")
+		req.Password, err = promptSetPassword("")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


internal/cmd/identities.go
```diff
@@ -289,14 +280,19 @@ func promptPasswordConfirm(oldPassword string) (string, error) {
...
-			Validate: checkConfirmPassword(&passwordConfirm.Password),
+			Validate: checkConfirmPassword(),
 		},
 	}
 
 	if err := survey.Ask(prompts, &passwordConfirm, survey.WithStdio(os.Stdin, os.Stderr, os.Stderr)); err != nil {
 		return "", err
 	}
 
+	if passwordConfirm.Password != passwordConfirm.Confirm {
```
kimskimchi, line 291: @ssoroka moved to here

internal/cmd/identities.go
```diff
@@ -289,14 +280,19 @@ func promptPasswordConfirm(oldPassword string) (string, error) {
 		{
 			Name:     "Confirm",
 			Prompt:   &survey.Password{Message: "Confirm Password:"},
-			Validate: checkConfirmPassword(&passwordConfirm.Password),
+			Validate: checkConfirmPassword(),
```
BruceMacD, line 283:
```suggestion
```

Looks like this validation should be removed from the survey input completely if it happens after the input

internal/cmd/identities.go
```diff
@@ -319,17 +315,12 @@ func checkPasswordRequirements(oldPassword string) survey.Validator {
 	}
 }
 
-func checkConfirmPassword(password *string) survey.Validator {
+func checkConfirmPassword() survey.Validator {
```
BruceMacD, line 318: I think this function should be removed entirely

BruceMacD (commented):
Good catch on this bad behavior, I think the validator function to confirm that the passwords match here that can be cleaned up. In the case that the validator function test could be removed too.

internal/cmd/identities.go
kimskimchi: remove the type check?

BruceMacD: Yeah, anything the survey library lets them enter should be a valid string (it doesn't allow nil response)

internal/cmd/identities.go
kimskimchi: done

BruceMacD (approved):

Issue #1868, opened 2022-05-04
Remove deprecated grant/group list API endpoints


dnephin:
#1836 and #1914 deprecates three endpoints. This ticket is to remove them after a couple versions. Also remove the `api.Client` methods.

dnephin:
Done in #2604

PR #1914, opened 2022-05-09
api: remove extra group endpoint

### LINKED ISSUE #1680: api: remove /v1/idenities/:id/grants and /v1/groups/:id/grants

dnephin:
## Summary

Deprecates the `ListIdentityGroups` endpoint and replaces it with an extra query parameter to `ListGroups`. Both of these endpoints returned the same data, but they were authorized in different ways. Now the single endpoint handles both. 

Resolves #1680


internal/server/data/group.go
```diff
@@ -19,12 +19,19 @@ func ListGroups(db *gorm.DB, selectors ...SelectorFunc) ([]models.Group, error)
...
+	return func(db *gorm.DB) *gorm.DB {
+		return db.
+			Joins("JOIN identities_groups ON groups.id = identities_groups.group_id").
+			Where("identities_groups.identity_id = ?", id)
 	}
+}
 
-	return result, nil
+// TODO: replace calls with ListGroups(..., WhereGroupIncludesUser(id))
+func ListIdentityGroups(db *gorm.DB, identityID uid.ID) ([]models.Group, error) {
```
dnephin, line 31: This query is used from other endpoints. It seems like it is equivalent to `ListGroups` with `WhereGroupIncludesUser`. I'm thinking we only want one way to do this.

internal/server/data/group_test.go
```diff
@@ -52,7 +58,7 @@ func TestCreateGroupDuplicate(t *testing.T) {
 	)
 
 	db := setup(t)
-	createGroups(t, db, everyone, engineers, product)
+	createGroups(t, db, &everyone, &engineers, &product)
```
dnephin, line 58: I changed these to pointers so that we could get access to the ID which is generated on save. The ID is often needed for other queries.

api/group.go
```diff
@@ -12,7 +12,10 @@ type Group struct {
 }
 
 type ListGroupsRequest struct {
+	// Name filters the results to only the group matching this name.
 	Name string `form:"name"`
+	// UserID filters the results to only groups where this user is a member.
+	UserID uid.ID `form:"user-id"`
```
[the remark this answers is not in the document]
mxyng: A common format seems to be camelCase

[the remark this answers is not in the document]
dnephin: I saw that for the json request/response body, but I'm not sure about query parameters.

We could also use `user` if `ID` is generally implied.

api/client.go
```diff
@@ -229,6 +229,7 @@ func (c Client) ListIdentityGrants(id uid.ID) (*ListResponse[Grant], error) {
 	return list[ListResponse[Grant]](c, fmt.Sprintf("/v1/identities/%s/grants", id), nil)
 }
 
+// Deprecated: use ListGroups
```
BruceMacD, line 232: Should we create an issue to track removing this?

internal/access/group.go
```diff
@@ -30,13 +33,29 @@ func isUserInGroup(c *gin.Context, requestedResourceID uid.ID) (bool, error) {
...
+	}
+	if userID != 0 {
+		selectors = append(selectors, data.WhereGroupIncludesUser(userID))
+	}
+
 	db, err := RequireInfraRole(c, models.InfraAdminRole, models.InfraViewRole, models.InfraConnectorRole)
-	if err != nil {
-		return nil, err
+	if err == nil {
+		return data.ListGroups(db, selectors...)
```
BruceMacD, line 47: I like that you defaulted to no access here 👍

BruceMacD (approved):

api/client.go
dnephin: It is added to https://github.com/infrahq/infra/issues/1868, I think we can use a single issue to track all of it.

api/group.go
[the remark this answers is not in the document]
dnephin: Camcel case in query params is pretty strange. I wasn't able to find any existing multi-word query parameters. Are these some I missed or is this the first instance?

internal/access/group.go
```diff
@@ -30,13 +33,29 @@ func isUserInGroup(c *gin.Context, requestedResourceID uid.ID) (bool, error) {
 	return false, nil
 }
 
-func ListGroups(c *gin.Context, name string) ([]models.Group, error) {
+func ListGroups(c *gin.Context, name string, userID uid.ID) ([]models.Group, error) {
+	var selectors []data.SelectorFunc
+	if name != "" {
+		selectors = append(selectors, data.ByName(name))
+	}
```
ssoroka, line 40: this is what data.ByOptionalName is for. move this logic into the selector.

internal/access/group.go
```diff
@@ -30,13 +33,29 @@ func isUserInGroup(c *gin.Context, requestedResourceID uid.ID) (bool, error) {
...
+	if userID != 0 {
+		selectors = append(selectors, data.WhereGroupIncludesUser(userID))
+	}
```
ssoroka, line 43: `data.ByOptionalGroupMember()` ?

internal/access/group.go
```diff
@@ -30,13 +33,29 @@ func isUserInGroup(c *gin.Context, requestedResourceID uid.ID) (bool, error) {
...
+	if err == nil {
+		return data.ListGroups(db, selectors...)
+	}
+
+	if errors.Is(err, internal.ErrForbidden) {
+		db, err := hasAuthorization(c, userID, isIdentitySelf)
+		if err != nil {
+			return nil, err
+		}
+		return data.ListGroups(db, selectors...)
```
ssoroka, line 55: We need to add a selector to filter this by the user's groups. they don't automatically get access to do whatever query they are asking for

ssoroka (changes requested):

internal/server/data/group.go
ssoroka: would just go ahead and remove this.

dnephin (commented):
Considering the confusion around the API endpoint implementation I am thinking it would be better to change it to the approach I used in #1836

api/client.go
dnephin: We already effectively removed it in the rename from identity to user, so I went ahead and removed it. No reason to deprecate a method that was never in a release.

internal/server/handlers.go
```diff
@@ -125,21 +125,13 @@ func (a *API) DeleteUser(c *gin.Context, r *api.Resource) error {
 	return access.DeleteIdentity(c, r.ID)
 }
 
+// TODO: remove after deprecation period
```
ssoroka, line 128: we can probably come up with a rewrite rule that would work for these rather than a deprecation period.

internal/server/data/group_test.go
```diff
@@ -83,17 +86,57 @@ func TestListGroups(t *testing.T) {
...
+	secondUser := models.Identity{
+		Name:   "secondarly",
+		Groups: []models.Group{everyone, product},
+	}
+	createIdentities(t, db, &firstUser, &secondUser)
 
-	groups, err = ListGroups(db, ByName(engineers.Name))
-	assert.NilError(t, err)
-	assert.Equal(t, 1, len(groups))
+	t.Run("all", func(t *testing.T) {
```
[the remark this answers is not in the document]
dnephin: Could you elaborate? This is the data layer so no authorization here. 

Don't the other two test cases show that only matching results are returned (because there are groups that do not match the query) ?

api/group.go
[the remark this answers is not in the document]
dnephin: Both the stripe API and the github API seem to use snake_case consistently for all fields regardless of where they appear in the request.

I'm not going to propose we change that here, so CamcelCase it is.

internal/server/routes.go
```diff
@@ -103,6 +102,7 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) *gin.Engine
 	// TODO: remove after a couple versions
 	get(a, authn, "/v1/users/:id/grants", a.ListUserGrants)
 	get(a, authn, "/v1/groups/:id/grants", a.ListGroupGrants)
+	get(a, authn, "/v1/users/:id/groups", a.ListUserGroups)
```
dnephin, line 105: Oops, these should have been added as regular gin routes that don't add to the openAPI doc. I missed that. I'll fix that here.

internal/access/group.go
ssoroka: Not sure I agree with that. It cleans up the callers quite a bit, and the caller needs to explicitly think about whether or not the parameter could be optional or not, as it introduces security problems if a required argument is treated as optional.

docs/api/openapi3.json
```diff
@@ -3558,24 +3555,37 @@
...
+        "summary": "Version",
         "tags": [
-          "Groups",
-          "Users"
+          "Misc"
         ]
       }
     },
-    "/api/version": {
+    "/v1/groups/{id}/grants": {
```
ssoroka, line 3571: shouldn't be in the api docs anymore. We could add support for redirects that change a path id to a query parameter to make supporting these easy.

internal/server/data/group_test.go
```diff
@@ -83,17 +86,57 @@ func TestListGroups(t *testing.T) {
...
+		expected := []models.Group{
+			{Name: "Engineering"},
+		}
+		assert.DeepEqual(t, actual, expected, cmpGroupShallow)
+	})
+
+	t.Run("filter by identity membership", func(t *testing.T) {
+		fetched, err := GetIdentity(db, ByID(firstUser.ID))
+		assert.NilError(t, err)
+		fmt.Println(fetched.Groups)
```
ssoroka, line 124: leaving the println in?

internal/server/routes.go
ssoroka: yeah I was saying above redirect rewrites could be made to support this case

internal/server/apimigrator.go
```diff
@@ -26,24 +26,48 @@ type apiMigration struct {
...
+			fmt.Println("Updating ", a.migrations[i].path, " to ", newMigration.redirect)
+			a.migrations[i].path = newMigration.redirect
 		}
-		if len(mig.redirect) > 0 && mig.redirect == path {
-			a.migrations[i].redirect = newPath
+		// Update any existing redirect migrations, so that the target of the redirect
+		// is the new canonical path.
+		// TODO: but what if there are intermediate paths that are necessary?
+		if mig.redirect != "" && mig.redirect == newMigration.path {
+			a.migrations[i].redirect = newMigration.redirect
```
dnephin, line 68: I think this update logic (previously it was in `addRedirect`) is not quite right. It seems to work when there is only one migration for the endpoint, but adding another one in this PR causes some strange behaviour.

The tests pass, but if you look at the debugging output I added "Applying ...", it applies a bunch of request rewrites incorrectly. 

It doesn't seem correct to update all of these migrations because it changes the meaning of them.

internal/server/handlers.go
dnephin: I spent 2 days on it, but it requires adding even more complexity to the API migration system to be able to support "merge" migrations (where two endpoints are implemented by a single one). We'd have to filter the migrations by version and apply them. I think the approach in this PR gives users the same benefits (no breaking change, no deprecated routes in the OpenAPI doc), without us having to do that extra work.

internal/server/handlers.go
```diff
@@ -125,21 +125,13 @@ func (a *API) DeleteUser(c *gin.Context, r *api.Resource) error {
...
 func (a *API) ListUserGroups(c *gin.Context, r *api.Resource) (*api.ListResponse[api.Group], error) {
```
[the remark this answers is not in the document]
dnephin: renamed to `deprecatedListUserGroups` to indicate this route is deprecated.

internal/server/apimigrator.go
dnephin: Reverted this part. The problem is more involved than simply this update.

internal/server/data/group.go
```diff
@@ -16,15 +16,16 @@ func GetGroup(db *gorm.DB, selectors ...SelectorFunc) (*models.Group, error) {
 }
 
 func ListGroups(db *gorm.DB, selectors ...SelectorFunc) ([]models.Group, error) {
+	db = db.Order("name ASC")
```
dnephin, line 19: Noticed some test flakes because of a lack of sort order, so added this, which matches the order we use for `ListIdentities`

ssoroka (approved):

Issue #1234, opened 2022-03-18
helm installing infra-connector conflict 


mchiang0610:
Testing this in docker desktop:
```
mchiang@Michaels-MacBook-Pro ~ % helm install infra infrahq/infra --set server.enabled=false --set engine.config.name=docker-desktop --set engine.config.accessKey=sample-keyDeI1wsKhOG --set engine.config.server=localhost --set engine.config.skipTLSVerify=true
Error: INSTALLATION FAILED: cannot re-use a name that is still in use
```

The output should install 
```
helm install infra-connector infrahq/infra --set server.enabled=false --set engine.config.name=docker-desktop --set engine.config.accessKey=sample-keyDeI1wsKhOG --set engine.config.server=localhost --set engine.config.skipTLSVerify=true
```

version: 
```
mchiang@Michaels-MacBook-Pro ~ % helm show chart infrahq/infra
apiVersion: v2
appVersion: 0.5.12
description: A Helm chart for Infra
name: infra
type: application
version: 0.5.12

mchiang@Michaels-MacBook-Pro ~ % helm list
NAME           	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART       	APP VERSION
infra          	default  	1       	2022-03-18 00:55:49.947953 -0400 EDT	deployed	infra-0.5.12	0.5.12     
infra-connector	default  	1       	2022-03-18 00:59:17.881725 -0400 EDT	deployed	infra-0.5.12	0.5.12     
```

Issue #1858, opened 2022-05-04
cli: revisit output & error feedback


kimskimchi:
Tracking bug. Working in progress along with #1394. We need to: 
- make sure no command succeeds or fails silently. 
- log state changes as they happen on debug level 
- confirm errors are outputted (however, the actual formatting of the output should be linked in #1394)
- warnings are outputted if state change is unknown, or partial

Proposal: 
- #1833 

Pull requests:
- #1853

kimskimchi:
Will close this, as the big ones were addressed.

Going forward, these will be individual issues, eg: 
https://github.com/infrahq/infra/issues/2096

Issue #1400, opened 2022-03-30
Refactor client to not be so modifiable within package `cmd`


kimskimchi:
Current we pass in someFunctionToCallAPI(client, ...)

We should be discouraging the modification of the client, since it really shouldn't be touched anywhere except when the state of config changes (eg. login, logout, ...)

We can have a client.go , and set a var there of that client, for lazy initialization.

ssoroka:
not sure I understand what this issue is trying to say.

dnephin:
Ya, using a package-level variable is probably not what we want.

A great way to indicate that a function doesn't modify something, without switching to using the struct value, is to accept an interface. For example:

```go
type usersClient interface {
    ListUsers(req ListUsersRequest) (*ListResponse[User], error)
    ...
}

func someFunctionToCallAPI(client usersClient) error {
    ...
}
```

`api.Client` already implements` usersClient`. This way the function declares that it's only going to use a few methods, and the reader can pretty quickly see that the methods are likely not going to modify anything in the client.

PR #1920, opened 2022-05-09
improve: feedback and logging for dst remove and id remove

kimskimchi:
## Summary

**Major**
- introduces force flag to `destinations remove` and `identities remove`
   ```
   $ infra dst rm abc
   No destinations named [abc].
   $ infra echo $?
   1

   $ infra dst rm abc --force
   $ echo $?
   0
   ```
- `destinations remove` and `identities remove` starts using CLI Error 

**Minor**
- updates logging of CLI Error to not be two lines, but rather one long one 
  ```
  old: 
    Internal error: 
    some error..
  now: 
    Internal error: some error...
  ```
- adds logging for `destinations [list|remove]`, and `identities remove`
- updates docs for `--force` flag for `destinations remove` and `identities remove`
- fixes tests for delete (api get and delete are now working properly)

## Checklist

- [x] Wrote appropriate unit tests --> **[updated tests]**
- [x] Updated associated docs where necessary  --> **[make docs]**
- [x] Is change backwards compatible? --> **[yes, will return 0 on not found if `--force` is specified]**
- [x] Nothing sensitive logged --> **[logs input, no sensitive info]**

## Related Issues

<!--
Link any related issues. Each issue should be on
its own line. For example:

Resolves #1234
Resolves #4321
--
Related #1858


internal/cmd/destinations.go
```diff
@@ -40,6 +41,7 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
 				return err
 			}
 
+			logging.S.Debug("call server: list destinations")
```
BruceMacD, line 44: Any info about the current runtime we can add to these debug logs that might help give even more context? Not needed, just something I thought of.

main.go
```diff
@@ -3,19 +3,28 @@ package main
...
 
 func main() {
+
 	if err := cmd.Run(context.Background(), os.Args[1:]...); err != nil {
-		if !errors.Is(err, terminal.InterruptErr) {
-			logging.S.Error(err.Error())
+		var userErr cmd.Error
+		switch {
+		case errors.Is(err, terminal.InterruptErr):
+			// print nothing, the user initiated the exit
```
BruceMacD, line 20: should we debug log this too?

internal/cmd/identities.go
```diff
@@ -175,6 +189,8 @@ $ infra identities remove janedoe@example.com`,
 		},
 	}
 
+	cmd.Flags().BoolVar(&force, "force", false, "Exit successfully destination not found")
```
BruceMacD, line 192:
```suggestion
	cmd.Flags().BoolVar(&force, "force", false, "Exit successfully when destination not found")
```

internal/cmd/destinations.go
kimskimchi: I have been struggling a bit attempting to error handle / log  'apiclient'/contexts, etc, since there is #1400 that needs to be addressed first

internal/cmd/destinations.go
```diff
@@ -82,16 +87,24 @@ func newDestinationsRemoveCmd() *cobra.Command {
...
 			if err != nil {
 				return err
 			}
 
 			if destinations.Count == 0 {
-				return fmt.Errorf("no destinations named %s.", args[0])
+				if force {
+					return nil
+				}
+				return Error{
```
pdevine, line 100: We haven't switched all of the other errors over to using this style of Error{} object. Do you want to create this one in the old fmt.Errorf() format and then convert everything over at once? That way we won't have skew between the other styles of errors.

main.go
```diff
@@ -12,10 +13,19 @@ import (
 )
 
 func main() {
+
 	if err := cmd.Run(context.Background(), os.Args[1:]...); err != nil {
-		if !errors.Is(err, terminal.InterruptErr) {
-			logging.S.Error(err.Error())
+		var userErr cmd.Error
```
pdevine, line 18: I think this can go in as a separate PR

pdevine (changes requested):
Overall looks fine. It'd be nice to split out the errors part from this PR and put them in separately though just to keep local consistency.

main.go
kimskimchi: done~

internal/access/identity_test.go
```diff
@@ -1,100 +0,0 @@
-package access
```
ssoroka, line 1: not sure why this is here.

internal/cmd/destinations.go
```diff
@@ -58,10 +60,11 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
 				})
 			}
 
+			logging.S.Debug("print destinations")
```
ssoroka, line 63: doesn't seem useful

internal/cmd/destinations.go
ssoroka, line 100: is this a user-facing error? I think perhaps we should change the name of the type to something more descriptive? at the moment it seems like `Error` is appropriate to use for all errors, but I'm assuming it's only appropriate for errors where you want to customize the text to be something the user sees and acts on. This probably differs from other errors in that these shouldn't be wrapped, prefixed, or displayed with stack traces, etc. thoughts?

internal/cmd/error.go
```diff
@@ -18,14 +18,14 @@ type Error struct {
 func (e Error) Error() string {
 	if e.OriginalError != nil {
 		if len(e.Message) == 0 {
-			return fmt.Sprintf("Internal error:\n%v", e.OriginalError)
+			return fmt.Sprintf("Internal error: %v", e.OriginalError)
```
ssoroka, line 21: `Internal error` doesn't seem accurate here. The Error type is for user-facing errors is it not?

ssoroka (changes requested):
The `logging.S.Debug` statements don't seem that useful for users debugging what's going wrong with a command; instead they seem to be of the breakpoint style that a developer would use while writing the program. Users instead want to see hidden server responses, warnings, etc.

internal/cmd/destinations.go
kimskimchi: removed

internal/cmd/destinations.go
kimskimchi: We initially named it a "userfacing error" or some sort, but then after some discussion, a few of us were okay with naming it "error", since if we named it "UserFacingError" we don't need "Error" right now anyways (in the future we could rename if it becomes more robust and we need to split them)

internal/cmd/error.go
kimskimchi: The purpose of the Error is to allow for user-facing output, so any errors (internal or developer specified) can all be formatted consistently. Basically to wrap any errors (or create any) that can be printed on the CLI nicely

internal/cmd/users.go
```diff
@@ -156,12 +158,24 @@ $ infra users remove janedoe@example.com`,
...
+			if users.Count == 0 {
+				if force {
+					return nil
+				}
+				return Error{
+					Message: fmt.Sprintf("No users named [%s].", name),
+				}
+			}
+
+			logging.S.Debug("deleting %d users named [%s]...", users.Count, name)
```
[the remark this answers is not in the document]
kimskimchi: I actually agree, but this was the behavior that was requested (by I believe Jeff). I initially proposed that we error out and provide a message, to panic or output "multiple users found", but the behavior that was requested was to just ignore and assume API is always doing the correct thing.

internal/cmd/users_test.go
```diff
@@ -142,10 +156,27 @@ func TestUsersCmd(t *testing.T) {
 	t.Run("removes only the specified user", func(t *testing.T) {
 		modifiedUsers := setup(t)
 		ctx := context.Background()
-		err := Run(ctx, "users", "remove", "to-delete-user@example.com")
+		err := Run(ctx, "users", "add", "to-delete-user@example.com")
```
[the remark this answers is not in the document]
kimskimchi: done

internal/access/identity_test.go
kimskimchi: done

mxyng:
The force remove flag has been merged in #2042

PR #1921, opened 2022-05-09
fix: make dev using wrong version

mxyng:
## Summary

`make dev` makes some assumptions about the code base, one of which is to set infra version. This has side effects since the introduction of API migrations which causes the connector to report its API version incorrectly.

1. Docker image will always tag `dev` docker images when build locally with `make dev`
2. `IMAGE_TAG` is no longer supported to change the docker image tag
3. The version is injected into the Docker image at build time with contents of `git describe --tag --abbrev=0`
4. Server and connector on start up with report their full version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-tag := $(patsubst v%,%,$(shell git describe --tags))
+tag := $(patsubst v%,%,$(shell git describe --tags --abbrev=0))
 version := $(tag:v%=%)
 
 generate:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-tag := $(patsubst v%,%,$(shell git describe --tags))
+tag := $(patsubst v%,%,$(shell git describe --tags --abbrev=0))
 version := $(tag:v%=%)
```
BruceMacD, line 2: Is this version variable still used?

ssoroka: seems to be

Makefile
```diff
@@ -39,12 +39,8 @@ goreleaser:
...
 	goreleaser build --snapshot --rm-dist
 
-export IMAGE_TAG=0.0.0-development
-
-build/docker:
-	docker build --build-arg TELEMETRY_WRITE_KEY=${TELEMETRY_WRITE_KEY} . -t infrahq/infra:$(IMAGE_TAG)
-
-export OKTA_SECRET=infra-okta
+dev/docker:
+	docker build --build-arg BUILDVERSION=$(version) --build-arg TELEMETRY_WRITE_KEY=${TELEMETRY_WRITE_KEY} . -t infrahq/infra:dev
```
ssoroka, line 43: is BUILDVERSION expecting the leading `v` character or no?

ssoroka (approved):

pdevine (approved):
Seems to be working well.

Makefile
mxyng: The tag is set to the git tag which in our case has the leading v. Version is the stripped version of the tag. It's used in a few places to set various versions

PR #1922, opened 2022-05-09
ci: run data package tests against all supported databases

dnephin:
## Summary

The `data` package is our abstraction for interacting with the database. We should run the tests in this package against all supported database (sqlite and postgres) to make sure that queries work in both scenarios.

This PR setups up CI config and adds supports for running the tests against postgres locally as well. Postgres is required in CI, but optional for local runs.

Tests in the `data` package should receive a db from `runDBTests`, to run against all databases.  I've only added this to a couple tests to demonstrate the pattern. I'll apply it to all the tests once we're happy with the approach.

Since there are plenty of whitespace changes, the diff is best viewed with [hide whitespace](https://github.com/infrahq/infra/pull/1922/files?diff=split&w=1).

TODO:
* [x] use `runDBTests` in all the other `data` tests

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -6,6 +6,18 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:14-alpine
+        env:
+          POSTGRES_PASSWORD: password123
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports: ["5432:5432"]
+
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
@@ -13,7 +25,12 @@ jobs:
           go-version: '1.18'
       - run: go install gotest.tools/gotestsum@v1.8.0
       - run: go mod tidy
-      - run: ~/go/bin/gotestsum -ftestname -- -race ./...
+
+      - name: go test
+        run: ~/go/bin/gotestsum -ftestname -- -race ./...
+        env:
+          POSTGRESQL_CONNECTION: "host=localhost port=5432 user=postgres dbname=postgres password=password123"
+
       - name: Check that tests leave a clean git checkout
         run: |
           # show and check changes to committed files
--- a/internal/server/data/data_test.go
+++ b/internal/server/data/data_test.go
@@ -15,12 +15,21 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
+// TODO: replace all calls to setup with runDBTests, or setupDB
 func setup(t *testing.T) *gorm.DB {
+	t.Helper()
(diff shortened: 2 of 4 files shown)


internal/server/data/data_test.go
```diff
@@ -51,24 +60,69 @@ func setupLogging(t *testing.T) {
...
+			assert.NilError(t, db.Exec("DROP SCHEMA testing CASCADE").Error)
+		})
+		assert.NilError(t, db.Exec("CREATE SCHEMA testing").Error)
+
+		pgsql, err = NewPostgresDriver(pgConn + " search_path=testing")
+		assert.NilError(t, err, "postgresql driver")
+
+		drivers = append(drivers, pgsql)
+	case os.Getenv("CI") != "":
+		t.Fatalf("CI must test all drivers, set POSTGRESQL_CONNECTION")
```
dnephin, line 92: This is to ensure we always run everything in CI.

Example failure when CI is not setup properly: https://github.com/infrahq/infra/runs/6360878148?check_suite_focus=true

.github/workflows/pull_request.yml
```diff
@@ -6,14 +6,29 @@ on:
...
+    services:
+      postgres:
+        image: postgres:14-alpine
+        env:
+          POSTGRES_PASSWORD: password123
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
dnephin, line 18: Copied this from the github actions docs: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-in-containers

internal/server/data/data_test.go
```diff
@@ -15,13 +15,22 @@ import (
...
+	return setupDB(t, driver)
+}
 
-	db, err := NewDB(driver)
+func setupDB(t *testing.T, driver gorm.Dialector) *gorm.DB {
+	t.Helper()
+	db, err := NewRawDB(driver)
 	assert.NilError(t, err)
 
+	// TODO: mirror production
```
BruceMacD, line 31: Does this `TODO` mean you're going to pre-populate this with some production data?

BruceMacD (approved):
This is nice 👍

pdevine (approved):

ssoroka:
It might be easier to have the tests take a DSN for the database connection from an environment variable, and just run the test suite twice, once for each db to test against.

internal/server/data/data_test.go
dnephin: Ah, no this TODO is about making sure the test code paths match production.

We used to have a `NewDB` function that setup the database for production code paths. Then we introduced `NewRawDB` which production uses, but all of our tests use `NewDB`. This divergence means our tests can lie to us because they are not testing the right state of the db.

This TODO is about fixing that problem. I'll word it better.

ssoroka (approved):

dnephin:
I opened #1989 which applies this new helper to all the functions.

PR #1924, opened 2022-05-09
improve: enable Sqlite WAL

ssoroka:
## Summary

I noticed about a 3-4x performance improvement from enabling sqlite's WAL, which oddly isn't enabled by default. This enables it.

In the course of testing the old database locked messages, I also noticed we were hiding some db errors which didn't seem correct, so I've fixed that too.

## Checklist

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -3,6 +3,7 @@ package data
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/server/data/data.go
```diff
@@ -64,6 +65,14 @@ func NewSQLiteDriver(connection string) (gorm.Dialector, error) {
 			return nil, err
 		}
 	}
+	uri, err := url.Parse(connection)
+	if err != nil {
+		return nil, err
+	}
+	query := uri.Query()
+	query.Add("_journal_mode", "WAL")
```
BruceMacD, line 73: Should we be checking if the uri already contains this query?

BruceMacD (approved):

internal/server/data/data.go
ssoroka: only if someone would ever want to disable WAL. Not sure that we should give them the choice.

PR #1926, opened 2022-05-10
fix: logout should always succeed

### LINKED ISSUE #1863: cli: logout fails if host cert is not trusted
## Describe the bug

``` 
$ dev logout --all --log-level=debug
DEBUG  logging out of all servers

DEBUG  requested but not logged in to server [localhost:8443]
ERROR  POST "/v1/logout": Post "https://localhost/v1/logout": x509: “Infra” certificate is not trusted
```

## Expected behavior

logout should happen whether or not the cert is trusted 

## Environment

```
$ infra version
```

```
$ kubectl version
```

ssoroka:
## Summary

If the /logout api call fails for some reason, we should just delete the access key and continue; we don't really care why it failed, and the key will time out anyway.

## Checklist

## Related Issues

Resolves #1863

### DIFF
--- a/internal/cmd/logout.go
+++ b/internal/cmd/logout.go
@@ -70,6 +70,7 @@ func logoutOfServer(hostConfig *ClientHostConfig) (success bool) {
 
 	client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
 	if err != nil {
+		logging.S.Debugf("err: %s", err)
 		return false
 	}
 
@@ -80,8 +81,10 @@ func logoutOfServer(hostConfig *ClientHostConfig) (success bool) {
 	err = client.Logout()
 	switch {
 	case api.ErrorStatusCode(err) == http.StatusUnauthorized:
+		logging.S.Debugf("err: %s", err)
 		return false
 	case err != nil:
+		logging.S.Debugf("err: %s", err)
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


internal/cmd/logout.go
```diff
@@ -62,15 +62,15 @@ $ infra logout --all --clear`,
...
 	if !hostConfig.isLoggedIn() {
 		logging.S.Debugf("requested but not logged in to server [%s]", hostConfig.Host)
-		return false, nil
+		return false
 	}
 
 	client, err := apiClient(hostConfig.Host, hostConfig.AccessKey, hostConfig.SkipTLSVerify)
 	if err != nil {
-		return false, err
+		return false
```
BruceMacD, line 73: we should add a debug log for this error just in case

internal/cmd/logout.go
```diff
@@ -80,14 +80,13 @@ func logoutOfServer(hostConfig *ClientHostConfig) (success bool, err error) {
 	err = client.Logout()
 	switch {
 	case api.ErrorStatusCode(err) == http.StatusUnauthorized:
-		logging.S.Warn(err.Error())
-		return false, nil
+		return false
```
BruceMacD, line 83: we should add a debug log for this error too, could be hard to troubleshoot otherwise

BruceMacD, line 83: we should add a debug log for this error too, could be hard to troubleshoot otherwise

BruceMacD (commented):
I agree that we should just clear the session on an error response, just some comments around supportability

jmorganca (approved):

pdevine (approved):

PR #1927, opened 2022-05-10
format grants output for separate user/group tables

### LINKED ISSUE #1862: infra grants list does not describe what kind of identity the grant is issued to
## Describe the bug

It's possible to create a grant for both a user and a group sharing the same name. This grant will show up in the list outputs with the same content so it's impossible to differentiate between the two.

```
$ infra grants list
  IDENTITY (4)     ACCESS     DESTINATION
  tmp              connect    infra
  tmp              connect    infra
```

## Expected behavior

It should be obvious from the list outputs whether a grant is issued to a user or a group

ssoroka:
## Summary

New output:
```sh
$ infra grants list
  USER (7)                 ACCESS         DESTINATION                     
  steven@example.com       admin          infra                           
  steven+test@example.com  cluster-admin  kubernetes.example-cluster      
  steven+test@example.com  edit           kubernetes.example-cluster.web  
  connector                connector      infra                           
  admin                    admin          infra                           
  admin                    connect        infra                           
  admin                    connect        kubernetes.example-cluster      

  GROUP     ACCESS  DESTINATION                 
  Everyone  view    kubernetes.example-cluster  

```

## Checklist

## Related Issues

Resolves #1862


mxyng (approved):
Nice! I like the approach of splitting the tables. Better than adding another column

jmorganca (approved):

internal/cmd/grants.go
```diff
@@ -58,46 +58,62 @@ func newGrantsListCmd(cli *CLI) *cobra.Command {
 				return err
 			}
 
-			type row struct {
-				Identity string `header:"IDENTITY"`
+			type userTable struct {
```
pdevine, line 61: These should probably be userRow{} and groupRow{} given that you're creating a slice of both (not a slice of tables)

pdevine (approved):
overall looks good, just the one small nit. I do prefer the tables to be split out.

PR #1928, opened 2022-05-10
fix: resolves issue with stacked response migrations

ssoroka:
## Summary

stacked migrations were interacting oddly with the flush behavior, where flush would flush layers it didn't own. This was code adopted from gin that didn't make sense in this context. added tests and resolved.

## Checklist

### DIFF
--- a/internal/server/apimigrator_test.go
+++ b/internal/server/apimigrator_test.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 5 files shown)


internal/server/apimigrator_test.go
```diff
@@ -33,21 +34,51 @@ func TestAddRequestRewrite(t *testing.T) {
...
+	})
+
+	addRequestRewrite(a, "get", "/test", "0.1.1", func(old upgradedTestRequest) upgradedTestRequest {
+		return upgradedTestRequest{
+			VegetableCount: old.VegetableCount * 2,
+		}
+	})
+
+	get(a, router.Group("/"), "/test", func(c *gin.Context, req *upgradedTestRequest) (*api.EmptyResponse, error) {
+		assert.Equal(t, req.VegetableCount, 24)
```
mxyng, line 67: might be clearer if this was something like `(5 + 7) * 2` so I don't need to do math

mxyng (approved):

PR #1929, opened 2022-05-10
fix: pin UI API version

mxyng:
## Summary

API migration assume without a version that all migrations are required. This has some side effects. The main side effect is that any request that does not require any migration must set the newest version. Since the UI is bundled with the API server, setting an valid version is difficult since the version won't be known until after release. For now, pin the UI version to the latest stable which means the API that the UI uses will lag behind.

## Checklist

## Related Issues

Resolves #


ui/pages/_app.js
```diff
@@ -8,7 +8,14 @@ import '../lib/dayjs'
 import '../styles/globals.css'
 
 const fetcher = async (resource, init) => {
-  const res = await fetch(resource, init)
+  const res = await fetch(resource, {
+    ...init,
+    ...{
```
jmorganca, line 13:
```
{
  ..init,
  headers: {
    "Infra-Version": "0.12.2",
  },
}
```

jmorganca (approved):
small comment worth fixing, otherwise LGTM

mxyng:
Created #1930 to track using the latest API version in the UI

PR #1936, opened 2022-05-11
improve: support cli error struct 

kimskimchi:
## Summary

Last bit to support the new Error CLI struct. Part 2 of #1786 - main.go will now handle the new Error, as well as our previous errors. 

## Related Issues

Related #1786


main.go
```diff
@@ -13,9 +14,17 @@ import (
 
 func main() {
 	if err := cmd.Run(context.Background(), os.Args[1:]...); err != nil {
-		if !errors.Is(err, terminal.InterruptErr) {
-			logging.S.Error(err.Error())
+		var userErr cmd.Error
+		switch {
+		case errors.Is(err, terminal.InterruptErr):
+			logging.S.Debug("user interrupted (kill, stop) the process")
```
ssoroka, line 20:
```suggestion
			logging.S.Debug("user interrupted the process")
```

main.go
```diff
@@ -13,9 +14,17 @@ import (
...
+		case errors.As(err, &userErr):
+			fmt.Fprintln(os.Stderr, userErr.Error())
+		default:
+			userErr.OriginalError = err
```
ssoroka, line 24: why are we setting userErr here?

dnephin: I think the goal here is to get the formatting provided by `Error.Error()`. I think there are a few ways we could accomplish that goal.

Since we know that in this case there is no `Error.Message` we'll always end up in the branch that prefixes this error with `Internal error:\n%v`. We could extract that to a function called `formatInternalError(err) string`, and call that function both here, and from `Error.Error.

Although, maybe using `cmd.Error` for that case is unnecessary. I guess we only ever want to use `cmd.Error` if we are provided a human readable error message. So maybe we could remove that branch from `Error.Error`, and do the suggestion in the other comment.

main.go
```diff
@@ -13,9 +14,17 @@ import (
...
+			fmt.Fprintln(os.Stderr, userErr.Error())
```
dnephin, line 25:
```suggestion
			fmt.Fprintf(os.Stderr, "Internal error:\n%v\n", err)
```

(or however we want to format these)

kimskimchi: done

main.go
```diff
@@ -13,9 +14,16 @@ import (
 
 func main() {
 	if err := cmd.Run(context.Background(), os.Args[1:]...); err != nil {
-		if !errors.Is(err, terminal.InterruptErr) {
-			logging.S.Error(err.Error())
+		var userErr cmd.Error
+		switch {
+		case errors.Is(err, terminal.InterruptErr):
+			logging.S.Debug("user interrupted the process")
```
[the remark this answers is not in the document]
kimskimchi: done

BruceMacD (approved):

PR #1937, opened 2022-05-11
improve: generate UI in CI

### LINKED ISSUE #1882: ui: remove generated `internal/server/ui` code from the repo
Having generated code in the repo has little value and adds a maintenance overhead in making sure it's up-to-date. Instead, the code should be generated as needed, in most cases right before the Docker image build.

Requirements:

- `internal/server/ui` is removed from the repo
- UI is generated by make before a docker image build
- UI is not generated if the directory is already present
  - user should run `make clean` to clean up the directory if necessary to regenerate the UI
- UI is available for CI docker image build

mxyng:
## Checklist

## Related Issues

Resolves #1882


BruceMacD (approved):
These changes look good to me. We will have to merge this one with failing tests right?

dnephin (changes requested):
No, we should not merge this with failing tests. This prevents everyone from building the infra binary.

We need to solve these problems. I outlined them here:  https://github.com/infrahq/infra/issues/1882#issuecomment-1118797979

These aren't test failures, they are build failures because `go:embed` requires the files to exist.

.github/workflows/release_pull_request.yml
```diff
@@ -18,7 +18,22 @@ jobs:
       - run: go build github.com/$GITHUB_REPOSITORY
       - uses: github/codeql-action/analyze@v2
 
+  generate-ui:
+    if: startsWith(github.head_ref, 'release-please-') == true
```
dnephin, line 22: It would be great to build the UI on every PR. If there are problems with the build, I think we'd like to learn about them as soon as possible.

Could we have this run without the `if` ?

dnephin (commented):
I'd be happy to pair on some of the challenges with `go:embed` and testing if you run into any problems.

mxyng:
This was my bad. Should've marked as Draft

.github/workflows/release_pull_request.yml
mxyng: This definitely needs to be generated for all pull requests. I initially thought the UI is only necessary for the image but your points in #1882 and the failing tests prove that to not be true. This should be changed to a prereq for test jobs in both CI and local dev i.e. the Makefile

dnephin: We should not require someone to build the UI to be able to build the Go binary, or run the Go tests.

I think we can make that work, but generating the UI in every PR is nice anyway so that we verify the build still works in advance of the release.

mxyng:
Superseded by #2016

PR #1938, opened 2022-05-11
fix: revert ui's api version to 0.12.2

hoyyeva:
## Summary
ui is pointing to an older version of the api, so we don't need to update the ui every time when we have a newer version of api unless we are bumping the api version on ui


ssoroka (approved):

ui/components/grant.js
```diff
@@ -65,27 +65,27 @@ export default function ({ id, modalOpen, handleCloseModal }) {
...
+        if (!res.ok) {
           throw data
         }
 
-        if (data.length === 0) {
-          res = await fetch('/v1/users', {
-                  method: 'POST',
-                  body: JSON.stringify({ name: email })
-                })
+        if (data.count === 0) {
```
mxyng, line 75: this should be `length`. count was introduced after 0.12.2

ui/components/grant.js
```diff
@@ -65,27 +65,27 @@ export default function ({ id, modalOpen, handleCloseModal }) {
...
+            body: JSON.stringify({ name: email })
+          })
           const user = await res.json()
 
           await grantPrivilege('i:' + user.id)
           setEmail('')
           setRole('view')
         } else {
-          grantPrivilege('i:' + data[0].id)
+          grantPrivilege('i:' + data.items[0].id)
```
mxyng, line 86: similar to `count`, `items` is introduced after 0.12.2

internal/server/migrations.go
```diff
@@ -89,13 +89,13 @@ func (a *API) addResponseRewrites() {
 	addResponseRewrite(a, "get", "/v1/access-keys", "0.12.2", func(newResponse *api.ListResponse[api.AccessKey]) []api.AccessKey {
 		return newResponse.Items
 	})
-	addResponseRewrite(a, "get", "/v1/identities", "0.12.2", func(newResponse *api.ListResponse[api.User]) []api.User {
+	addResponseRewrite(a, "get", "/v1/users", "0.12.2", func(newResponse *api.ListResponse[api.User]) []api.User {
```
mxyng, line 92: This is necessary because the path has been redirected from `/v1/identities` to `/v1/users`

BruceMacD (commented):
Thanks for fixing this. As Mike points out, we will want to leave the API `migrations.go` file with `identities` conversion but the UI changes are good to go.

internal/server/migrations.go
BruceMacD: Yeah, you'll want to leave this file unchanged.

mxyng: I mean this change is necessary _to_ have the correct migration behaviour. Otherwise the redirect modifies the path from `/v1/identities` to `/v1/users` causing it to no longer match this path

mxyng: This change can be omitted once #1943 is merged

jmorganca (approved):

Issue #1939, opened 2022-05-11
api: remove or replace package-level static cache of InfraProvider and ConnectorIdentity


dnephin:
`data.infraConnectorCache` and `data.infraProviderCache` are package-level (aka static) variables that cache the ID of these models. Static variables like this make testing more difficult, because every test that uses them has to know to call `data.InvalidateCache`. 

Some tests are doing that correctly now, but some do not. It's not obvious to a developer that they need to call clear in tests. By not calling `InvalidateCache` we risk that our tests are not behaving correctly and hiding bugs.

I think there are other options to solving this problem:
1. Cache these identifiers on a struct. It would be great to have a strongly typed struct that we pass to API handlers to replace the use of the untyped and mutable `gin.Context`. Or we could have a struct for the `data` layer, which could also work for `models.SymmetricKey`.
2. We could hardcode the ID used for these. Since our ID generation will always produce IDs above some minimum value, we could define IDs in the range of 1000-2000 as reserved for system resources. We would always initialize these models with the hard coded ID, so we wouldn't need to cache the ID. A database migration can take care of any already existing entities with the old ID.

I think my preference would be option 2.

Related to #1609, which is a similar issue for the static package-level logging vars.

dnephin:
Still relevant, and becomes even more important to replace with the introduction of orgs, as it makes caching more difficult and less effective.

dnephin:
This was done in 7a769470d8d349143098d779f3053d04396ca9a3. Thank you @ssoroka !

PR #1941, opened 2022-05-11
api: replace static openapi document with a field on API

dnephin:
## Summary

Package-level variables make testing more challenging and can make code harder to reason about. They introduce [side effects](https://en.wikipedia.org/wiki/Side_effect_(computer_science)) that could be avoided.

This commit replaces the boolean to skip generation in tests (which likely makes the tests less like production) and replaces it with an `openapi3.T` document on the `API` struct. The document is now returned from `GenerateRoutes`, which also removes the risk of something referencing the static variable before it is fully populated. 

With this change, it's not a lot more obvious that `GenerateRoutes` is responsible for constructing the document, and it makes it a lot more difficult for tests to make changes to the real production document (because that is only saved explicitly by one test).

Also has two commits which:
* add some guard rails to CI by checking for files that are modified, to prevent similar problems in the future
* fixes a problem with other package-level variables that I noticed while investigating this issue (opened #1939 to track alternatives to those)

### DIFF
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -20,7 +20,6 @@ import (
 )
 
 var (
-	openAPISchema             = openapi3.T{}
 	pathIDReplacer            = regexp.MustCompile(`:\w+`)
 	funcPartialNameToTagNames = map[string]string{
 		"Grant":       "Grants",
@@ -35,10 +34,8 @@ var (
 	}
 )
 
+// register an API endpoint that has both a request and response value.
 func register[Req, Res any](a *API, method, path string, handler ReqResHandlerFunc[Req, Res]) {
-	if a.disableOpenAPIGeneration {
-		return
-	}
 	funcName := getFuncName(handler)
 
 	//nolint:gocritic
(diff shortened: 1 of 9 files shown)


jmorganca (approved):

internal/server/openapi.go
```diff
@@ -146,17 +143,17 @@ func orderedTagNames() []string {
 	return tagNames
 }
 
-func createComponent(rst reflect.Type) *openapi3.SchemaRef {
-	if openAPISchema.Components.Schemas == nil {
-		openAPISchema.Components.Schemas = openapi3.Schemas{}
+func createComponent(openAPIDoc openapi3.T, rst reflect.Type) *openapi3.SchemaRef {
```
ssoroka, line 146: are you sure this works? openapi3.T is a struct and will be copied when passed by value

ssoroka (approved):
bit concerned about the openapi3.T being passed by value and losing some of the values set, but otherwise no concerns.

internal/server/openapi.go
dnephin: Great question! I missed this when I was adding the args.

This worked because we only ever modified the schemas map (which is always pass by reference), but it is still really surprising.

I pushed another commit which changes the first arg to a few functions to be just the `openapi.Schemas`, which is a `map[string]*SchemaRef`. That should prevent any surprises in the future because we only pass the mutable part to these functions.

internal/server/openapi.go
```diff
@@ -142,17 +142,15 @@ func orderedTagNames() []string {
 	return tagNames
 }
 
-func createComponent(openAPIDoc openapi3.T, rst reflect.Type) *openapi3.SchemaRef {
-	if openAPIDoc.Components.Schemas == nil {
-		openAPIDoc.Components.Schemas = openapi3.Schemas{}
-	}
```
dnephin, line 151: This gets initialized in `register`

dnephin:
I pushed another commit with some small changes to the args. I'm happy to make any changes to that in a follow up. I'm going to open another PR with some small refactors to the `register` call.

PR #1943, opened 2022-05-11
fix: support redirecting previous path rewrites

ssoroka:
## Summary

to avoid requiring devs to edit old rewrites, support redirects that update old rewrites themselves

## Checklist

### DIFF
--- a/internal/server/apimigrator_test.go
+++ b/internal/server/apimigrator_test.go
@@ -10,6 +10,7 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/uid"
 )
 
 type legacyTestRequest struct {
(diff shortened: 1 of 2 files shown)


internal/server/apimigrator_test.go
```diff
@@ -95,13 +96,121 @@ func TestRedirect(t *testing.T) {
 	assert.Assert(t, resp.Result().StatusCode == 200)
 }
 
+func TestRedirectOfRequestAndResponseRewrite(t *testing.T) {
```
mxyng, line 99: I don't think this test has the right behaviour. It's setting a redirect from `oldtest` to `test` and also running migrations on `test`. Instead, it should run migrations on `oldtest` since that's the path being modified

ssoroka: There's a later test that does it in the order you're expecting. Both should be valid.

mxyng: cool wasn't obvious what it was testing

mxyng (approved):

PR #1947, opened 2022-05-12
Docs: Add animated gifs to the quickstart

technovangelist:
Signed-off-by: Matthew Williams <m@technovangelist.com>

## Summary

Added animated gifs to each step that shows what the user will be doing

## Checklist

- [x ] Wrote appropriate unit tests
- [x ] Considered security implications of the change
- [x ] Updated associated docs where necessary
- [x ] Updated associated configuration where necessary
- [x ] Change is backwards compatible if it needs to be (user can upgrade without manual steps?)
- [x ] Nothing sensitive logged
- [x ] Considered data migrations for smooth upgrades

## Related Issues

Resolves #


BruceMacD:
This is looking good, the destination names are out of date in the gifs at the moment though.

One suggestion (not required) the speed of the gifs and the fact they loop made me want to pause them. Maybe we should embed these as videos instead so that they can have media controls.

To fix the `commit-lint` error you can update the commit message to `improve: add animated gifs to the quickstart`. We only accept a pretty narrow format for the commit message at the moment:
https://github.com/infrahq/infra/blob/main/.github/commitlint.config.js

mchiang0610:
closing since we have a quickstart video

Issue #1965, opened 2022-05-12
CLI client requests should timeout


ssoroka:
client commands hang forever if the server goes away

PR #1966, opened 2022-05-12
fix: client requests should time out

ssoroka:
Resolves https://github.com/infrahq/infra/issues/1965

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -15,6 +16,8 @@ import (
 )
 
 var clientVersion = "0.13.0"
+
+var ErrTimeout = errors.New("client timed out waiting for response from server")
 
 type Client struct {
 	URL       string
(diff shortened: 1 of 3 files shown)


pdevine (commented):
This seems OK, but the error is probably not going to be great for the user. We should trap the error and return a reasonable response saying that the client timed out waiting for the infra server.

ssoroka:
I think there's multiple types of networking errors that can happen, and it's possibly somewhat os-dependent. it'd be [hard to catch them all](https://stackoverflow.com/questions/22761562/portable-way-to-detect-different-kinds-of-network-error) and have a reasonable message that doesn't assume too much.

dnephin:
We don't need to catch them all. I think there's only two places where this can timeout:
1. in the call to `http.Client.Do`
2. reading the `response.Body`

`http.Client.Do` says:

> // Any returned error will be of type *url.Error. The url.Error
> // value's Timeout method will report true if the request timed out.

That makes it pretty easy to check for a timeout on that call.

The error from reading `response.Body` is likely less easy, but also likely less common. A check for `io.EOF` or `errors.Is(err, context.DeadlineExceeded)` would probably be reliable enough for this.

dnephin (approved):
LGTM

pdevine (approved):

PR #1967, opened 2022-05-13
improve(docs): Add guide for installing Infra with k0s

pdevine:
## Summary

Create a guide which explains how to install Infra with k0s

## Checklist

### DIFF
--- a/docs/install-infra-on-k0s.md
+++ b/docs/install-infra-on-k0s.md
@@ -1,0 +1,168 @@
+# Installing k0s with Infra
+
+Getting k0s to work with Infra requires a few extra steps to get working.
+Out of the box, k0s does not include a load balancer or persistent storage
+which are required to run Infra.
+
+We'll install:
+  * [Metallb](https://docs.k0sproject.io/head/examples/metallb-loadbalancer/)
+  * [OpenEBS](https://docs.k0sproject.io/head/storage/)
+
+This guide assumes that you've already installed the Infra CLI. You
+can check out the [Quickstart guide](https://infrahq.com/docs/getting-started/quickstart)
+which contains directions on how to install the CLI.
+
+### 1. Download k0s
+
+`curl -sSLf https://get.k0s.sh | sudo sh`
+
+1. Create a k0s config file
+
+As the root user, run:
+
+`k0s config create > /etc/k0s/k0s.yaml`
+
+This will dump out a k0s config file which you will need to edit.
+
+### 2. Edit /etc/k0s/k0s.yaml to include OpenEBS
+
+Modify `/etc/k0s/k0s.yaml` and include the section:
+
+```
+spec:
+  extensions:
+    storage:
+      type: openebs_local_storage
+```
+
+### 3. Install k0s
+
+Run the command:
+
+`sudo k0s install controller --single -c /etc/k0s/k0s.yaml`
+
+If this is a multinode system, you can omit `--single` flag when you're doing the install, otherwise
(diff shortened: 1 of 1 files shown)


docs/install-infra-on-k0s.md
```diff
@@ -0,0 +1,168 @@
+# Installing k0s with Infra
+
+Getting k0s to work with Infra requires a few extra steps to get working.
```
BruceMacD, line 3:
```suggestion
Getting k0s to work with Infra requires a few extra steps.
```

docs/install-infra-on-k0s.md
```diff
@@ -0,0 +1,168 @@
...
+Out of the box, k0s does not include a load balancer or persistent storage
+which are required to run Infra.
+
+We'll install:
+  * [Metallb](https://docs.k0sproject.io/head/examples/metallb-loadbalancer/)
+  * [OpenEBS](https://docs.k0sproject.io/head/storage/)
+
+This guide assumes that you've already installed the Infra CLI. You
+can check out the [Quickstart guide](https://infrahq.com/docs/getting-started/quickstart)
```
BruceMacD, line 12:
```suggestion
can check out [this guide](https://github.com/infrahq/infra/blob/main/docs/install/install-infra-cli.md)
```

docs/install-infra-on-k0s.md
```diff
@@ -0,0 +1,168 @@
...
+which contains directions on how to install the CLI.
+
+### 1. Download k0s
```
BruceMacD, line 15: We should just link to the k0s guide here rather than re-iterating their install steps, that way we don't get out of sync:
https://docs.k0sproject.io/v1.21.2+k0s.1/install/

docs/install-infra-on-k0s.md
```diff
@@ -0,0 +1,168 @@
...
+    address-pools:
+    - name: default
+      protocol: layer2
+      addresses:
+      - <ip-address-range-start>-<ip-address-range-stop>
+```
+
+You can use private IP addresses in the IP Address range fields, but you'll need to make certain that any IP addresses are routed correctly.
+For example, if your network is using `192.168.1.1` to `192.168.1.254`, you can pick a small range of IPs inside of that network
+(make certain that your router isn't handing those IP addresses out with DHCP).
```
BruceMacD, line 103: How can I make check this?

docs/install-infra-on-k0s.md
```diff
@@ -0,0 +1,168 @@
...
+We'll call this file `openebs-persistence.yaml`.
+
+```
+---
+server:
+  persistence:
+    className: openebs-hostpath
+```
+
+### 9. Deploy Infra
```
BruceMacD, line 121: Should we make this a separate section? Just an idea.

Ex: 
```
# Configuring k0s
## Install k0s
...
# Deploying Infra
## Install Infra
```

BruceMacD (commented):
This should be in a sub-directory of `docs` rather than the root. Maybe `docs/guides`.

docs/install-infra-on-k0s.md
pdevine: The problem is there are some non-quickstart steps that need to be done first unfortunately. Since we'll post this in the discussion forum for now, I think it will probably be fine.

docs/install-infra-on-k0s.md
pdevine: Added a note about `netstat -rn` and `route -n`.

pdevine:
Closing this in favour of posting it into the discussion forum for now.

PR #1971, opened 2022-05-13
fix: set correct version for release and release candidate builds

### LINKED ISSUE #1952: Release branch image tag sets an invalid internal.Version
## Describe the bug

Release branch image is sets the tag to `next` which is useful as the Docker image tag but is harmful when set as the client version. Instead, this should be separated into two values. In reality, the values may be the same but for the release branch, it will be different; `next` for the docker image tag and `<version>-next` for the client version

mxyng:
## Summary

- Unset `Prerelease` and `Metadata` fields when building release
- Set `Prerelease` to `rc` and unset `Metadata` when building release candidate
- Unset `Prerelease` and set `Metadata` to `dev` when building local docker image
- Update version string for server to `internal.FullVersion()` and client to `api.GetClientVersion()`
- Update `api.GetVersion()` to `api.GetServerVersion()`
- Split `BUILDVERSION` and `IMAGEVERSION` since they may be different
- Remove codeql-analysis since the value add is minimal at this point in time

## Checklist

## Related Issues

Resolves #1952

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,9 +40,10 @@ jobs:
         with:
           path: /root/.cache/
           key: ${{ runner.os }}-go-build
+      # TODO: use goreleaser github action: https://github.com/goreleaser/goreleaser-action
       - run: |
           go install github.com/goreleaser/goreleaser@v1.7.0
-          make release tag=${{ needs.release-please.outputs.release_name }}
+          make release
         env:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
           GEMFURY_TOKEN: ${{ secrets.GORELEASER_GEMFURY_TOKEN }}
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -40,9 +40,10 @@ jobs:
         with:
           path: /root/.cache/
           key: ${{ runner.os }}-go-build
+      # TODO: use goreleaser github action: https://github.com/goreleaser/goreleaser-action
       - run: |
           go install github.com/goreleaser/goreleaser@v1.7.0
-          make release tag=${{ needs.release-please.outputs.release_name }}
```
mxyng, line 45: Removing this since goreleaser doesn't use the version passed in through Makefile

ssoroka (approved):
LGTM

internal/cmd/version.go
```diff
@@ -28,7 +28,7 @@ func version(cli *CLI) error {
 	defer w.Flush()
 
 	fmt.Fprintln(w)
-	fmt.Fprintln(w, "Client:\t", strings.TrimPrefix(internal.Version, "v"))
+	fmt.Fprintln(w, "Client:\t", strings.TrimPrefix(api.GetClientVersion(), "v"))
```
dnephin, line 31: I think we want to keep using `internal.Version` here. 

I think the user is more interested in "what version did this Go binary come from", not "what version of the API is it using".

The API version is misleading to a user because the CLI can use endpoints that did not exist in that version.

mxyng: You're completely right. This will be reverted in #1972

PR #1972, opened 2022-05-13
fix: revert client to report binary version, not api version

mxyng:
## Summary

This reverts part of 3662a6fab75ca77570fb19175dca27e950daf3a9.

The command line should report its binary version. The API version is used to describe the target API version so it's not appropriate to report as the version of the CLI


jmorganca (approved):

dnephin (approved):
Thanks for the quick fix!

ssoroka (approved):

PR #1973, opened 2022-05-13
ci: add daily schedule job to run fuzzing

dnephin:
This PR adds a github workflow for running all the fuzz tests. The workflow runs daily (at 3am UTC), and can also be triggered on a PR by adding the `action/fuzz` label (also added in this PR).

The `go test -fuzz` flag can only be used to run a single Fuzz test at a time, so the fuzz workflow does this:
* start by finding all the fuzz tests
* uses the list of fuzz tests to create a `matrix` for the actual fuzz job
* runs a separate job for every fuzz test it finds
* the fuzz corpus should be cached, so each run should resume from where it stopped last time

The fuzzing will run for 30s in the nightly run, but only 5 minutes for the PR runs. We can change those times if we want.

### DIFF
--- a/.github/workflows/fuzz.yaml
+++ b/.github/workflows/fuzz.yaml
@@ -1,14 +1,14 @@
 name: Fuzzing
 
 on:
-  # TODO: remove push, replace with pull_request label
-  push:
-    branches: [dnephin/ci-fuzz]
+  pull_request:
+    types: [labeled]
   schedule:
     - cron: "14 3 * * *"
 
 jobs:
   fuzz-matrix:
+    if: ${{ github.event.label.name == 'action/fuzz' }}
     runs-on: ubuntu-latest
     outputs:
       fuzz-names: ${{ steps.list-fuzz.outputs.list }}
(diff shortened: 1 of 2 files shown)


.github/workflows/fuzz.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: '1.18'
+
+      - uses: actions/cache@v3
+        with:
+          key: go-fuzz-corpus-${{ matrix.name }}-${{ github.run_number }}
+          restore-keys: go-fuzz-corpus-${{ matrix.name }}-
+          path: /tmp/go/cache/fuzz
```
dnephin, line 46: Looking at https://github.com/infrahq/infra/runs/6430042316 (the most recent run on this PR), it seems like the cache is being saved, but the restore isn't working for some reason. This `restore-keys` is a prefix match, I'm not sure why it's not matching.

dnephin: I read over https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key again, and it seems like this should be correct.

I'm not sure if I'm missing something, or if this config needs to be in main first.

mxyng: This looks right to me. The run even says it cached successfully

mxyng (approved):

.github/workflows/fuzz.yaml
ssoroka: Perhaps the run number is causing problems. wouldn't that make each of the caches unique, so you could never restore it?

dnephin: Ya, the `run_number` is unique for each run.  `restore-keys` will do a prefix match though, which should still be a match on previous runs. The docs for that are here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

I'm guessing it's because this config isn't in `main` yet, so something about the [restrictions](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) is preventing a cache hit.

dnephin:
Let's see if the cache works better when this runs on `main`.

dnephin:
The cache did start working after this landed in `main`, so it must be an issue that these docs don't describe properly: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

PR #1974, opened 2022-05-13
fix: openapi gen should not set -dev version in release branches

mxyng:
## Summary

If the check is run from the release branch pull request, unset `Metadata` so the version is set correctly

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -52,9 +52,11 @@ jobs:
           go-version: '1.18'
       - name: Check generated docs are updated
         run: |
+          # if checking a release candidate pull request, unset `internal.Metadat`
+          if [[ "$GITHUB_REF_NAME" =~ ^release-please-.*$ ]]; then LDFLAGS='-X github.com/infrahq/infra/internal.Metadata='; fi
           # fake a terminal to get the right defaults for non-interactive
-          script -e -q -c "go run ./internal/docgen"
-          go run ./internal/openapigen docs/api/openapi3.json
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -52,9 +52,11 @@ jobs:
           go-version: '1.18'
       - name: Check generated docs are updated
         run: |
+          # if checking a release candidate pull request, unset `internal.Metadat`
+          if [[ "$GITHUB_REF_NAME" =~ ^release-please-.*$ ]]; then LDFLAGS='-X github.com/infrahq/infra/internal.Metadata='; fi
           # fake a terminal to get the right defaults for non-interactive
-          script -e -q -c "go run ./internal/docgen"
-          go run ./internal/openapigen docs/api/openapi3.json
+          script -e -q -c "go run ${LDFLAGS:+-ldflags \"$LDFLAGS\"} ./internal/docgen"
+          go run ${LDFLAGS:+-ldflags "$LDFLAGS"} ./internal/openapigen docs/api/openapi3.json
```
dnephin, line 59: today I learned what `:+` does!

dnephin (approved):
Looking at the failure in the currently opened PR [here](https://github.com/infrahq/infra/runs/6428778824?check_suite_focus=true#step:4:241), it seemed like unsetting metdata wouldn't fix the full problem. They would still mismatch (`0.13.1`, vs `0.13.2`).

But I'm guessing that the `.2` is coming from https://github.com/infrahq/infra/blob/main/internal/version.go#L21-L23

Which increments the patch when there is metadata. The behaviour of `FullVersion` is bit surprising, but I guess it works.

Assuming my understanding is correct, LGTM!

PR #1975, opened 2022-05-14
maintain: simpler contributor guide and Makefile

jmorganca:
* Remove the need for `envsubst` and `docker-desktop.yaml.in`. Instead users can provide a custom `values.yaml` to `make dev` via `make dev flags="-f values.yaml"`
* Remove some `.gitignore` entries that aren't needed anymore
* Remove nonessential sections from `contributor.md` (e.g. disabling telemetry since it's always disabled anyways in development)


dnephin (commented):
I like the simplification! Makefiles are good for encoding simple tasks, and for new contributors to a project to get started quickly. Many of us will have different workflows that work well for us, but we don't need to encode all of them into the project Makefile. Trying to put all the complex workflows into a Makefile often makes it pretty hard to maintain and can make it harder to use.

Keeping it to a few common things like `test`, `lint`, `docs`, `dev`, `clean`, and `helm` looks great.

I wonder if we could even remove `release` from the Makefile (doesn't have to be in this PR). The bulk of the release logic is in `.github/workflows/release.yml`. Calling `make` from that workflow can sometimes make it harder to trace (have to jump between two files), and likely doesn't benefit us much. In the unlikely scenario where we have to do a manual release when github actions are down we'll have to refer to the workflow config anyway (to find all the steps), so copying a few one line commands out of that config is probably not a big deal.

mxyng:
I'm in favour of simplifying it further. The current Makefile has evolved organically so it's probably time to prune it. For example, the `envsubst` was necessary at one point but no longer so removing it is appropriate.

The targets @dnephin mentioned is a great place to start. There are currently multiple ways for achieving very similar results, e.g. `make docker`, `make release/docker`. Instead these should be a single target that gets configured different based on the situation

docs/contributing.md
```diff
@@ -16,91 +15,80 @@
...
-      role: admin
-      resource: infra
+    grants:
+      - user: admin@local.dev
+        role: admin
+        resource: infra
 
 connector:
   enabled: true
+  config:
```
BruceMacD, line 76: Will this work without an access key?

BruceMacD (approved):

Makefile
```diff
@@ -16,61 +13,25 @@ test/update:
...
-
-dev: $(VALUES) dev/docker
-	# docker desktop setup for the dev environment
-	# create a token and get the token secret from:
-	# https://dev-02708987-admin.okta.com/admin/access/api/tokens
-	# get client secret from:
-	# https://dev-02708987-admin.okta.com/admin/app/oidc_client/instance/0oapn0qwiQPiMIyR35d6/#tab-general
-	# create the required secret with:
-	# kubectl $(NS) create secret generic $(OKTA_SECRET) --from-literal=clientSecret=$$OKTA_CLIENT_SECRET
+	rm -r helm/*.tgz
```
mxyng, line 23: Should use `$(RM)` instead. It's equivalent to `rm -f`

docs/contributing.md
mxyng: It should. If the connector is enabled in the same helm deployment as the server, and no access key is provided, one will be generated automatically

mxyng (approved):

PR #1984, opened 2022-05-16
fix: build for release candidates

mxyng:
## Summary

- `BUILDVERSION` gets unset since the Makefile sets it to an empty string. Instead remove it as it's set statically in the Dockerfile
- Use `GITHUB_HEAD_REF` instead of `GITHUB_REF_NAME`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 v ?= $(shell git describe --tags --abbrev=0)
-BUILDVERSION ?= $(v:v%=%)
 IMAGEVERSION ?= $(v:v%=%)
 
 generate:
@@ -42,7 +41,6 @@ build: goreleaser
 
 dev/docker:
 	docker build \
-		--build-arg BUILDVERSION=$(BUILDVERSION) \
 		--build-arg BUILDVERSION_METADATA=dev \
 		--build-arg TELEMETRY_WRITE_KEY=$(TELEMETRY_WRITE_KEY) \
 		-t infrahq/infra:dev \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,5 +1,4 @@
 v ?= $(shell git describe --tags --abbrev=0)
-BUILDVERSION ?= $(v:v%=%)
```
dnephin, line 2: I had to look it up for myself, so thought I'd mention it here. The `BUILDVERSION` is set in the `Dockerfile`. That seems good to me. We can avoid passing it from here.

internal/version.go
```diff
@@ -5,8 +5,9 @@ import "github.com/Masterminds/semver/v3"
 var (
 	Branch = "main"
 	// {x-release-please-start-version}
-	Version = "0.13.0"
+	realVersion = "0.13.0"
```
mxyng, line 8: `realVersion` should only be modified by release automation and hold valid semantic versions

internal/version.go
```diff
@@ -5,8 +5,9 @@ import "github.com/Masterminds/semver/v3"
...
 	// {x-release-please-end}
+	Version    = ""
```
dnephin, line 10: We still reference this directly in a few places (CLI user agent, telemetry, and metrics), so we probably don't want to set it to `""`.

internal/version.go
```diff
@@ -17,10 +18,15 @@ var (
 // This is because release-please keeps this at the released version, and not the upcoming next version.
 // While the next version may not match the patch release, it causes the right behavior for semver version comparisons.
 func FullVersion() string {
-	v, _ := semver.NewVersion(Version)
+	v, err := semver.NewVersion(Version)
+	if err != nil {
+		v, _ = semver.NewVersion(realVersion)
+	}
```
dnephin, line 24: Instead of real version, what do you think about?


```suggestion
	v, err := semver.NewVersion(Version)
	if err != nil {
		panic(fmt.Sprintf("invalid version %v: %v", Version, err))
	}
```

Still a panic, but should make it easier to debug.

internal/version.go
BruceMacD: ^ I think we should add this comment in the `version.go` file

dnephin (approved):
LGTM!

PR #1986, opened 2022-05-16
api: small refactor to register openAPI routes

dnephin:
Follow up to #1941

I was wondering if we could avoid the need for 3 similarly named functions for registering the routes in the OpenAPI doc.  This is one approach. It is less type-safe locally, but the types are checked by the caller before calling `newDefinition`.

### DIFF
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -34,33 +34,43 @@ var (
 	}
 )
 
-// register an API endpoint that has both a request and response value.
-func register[Req, Res any](a *API, method, path string, handler ReqResHandlerFunc[Req, Res]) {
-	funcName := getFuncName(handler)
-
-	//nolint:gocritic
-	rqt := reflect.TypeOf(*new(Req))
-	//nolint:gocritic
-	rst := reflect.TypeOf(*new(Res))
-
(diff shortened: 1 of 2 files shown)


internal/server/openapi.go
```diff
@@ -34,33 +34,37 @@ var (
...
+	if fn.NumOut() == 2 { // the last out is always type error
+		resp = fn.Out(0)
+	}
 
-	a.register(method, path, funcName, rqt, rst)
+	return routeDefinition{
+		Method:   method,
+		Path:     path,
+		Name:     getFuncName(handler),
+		Request:  fn.In(1),
```
BruceMacD, line 60: what is `1` here?

dnephin: I should comment this. `1` is the second argument to the request function, the first argument is always `gin.Context`

ssoroka (approved):

dnephin:
I think I've found a better way to solve this. Instead of supporting different handler signatures I think we make a lot of things easier by having the delete endpoints return `api.EmptyResponse`. That change would remove the need for this one, so I'll close this PR for now and open one with the other approach.

Sorry for the noise.

PR #1989, opened 2022-05-16
data: use runDBTests in all tests

dnephin:
Follow up to #1922

Applies the new helper to all tests.

Best viewed with [hide whitespace](https://github.com/infrahq/infra/pull/1989/files?diff=split&w=1).

### DIFF
--- a/internal/server/data/data_test.go
+++ b/internal/server/data/data_test.go
@@ -15,14 +15,6 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-// TODO: replace all calls to setup with runDBTests, or setupDB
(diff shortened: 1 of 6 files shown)


dnephin:
I missed the migration tests, I'll do those as well in this PR.

BruceMacD (approved):

dnephin:
Ahh, the migration tests don't work with postgres because we need different sql files. The ones we have are specific to sqlite.  So I'm going to defer that change to a later PR and keep this one to only the non-migration tests.  If we want to test migrations on postgres as well, we'll have to make it easy to generate the sql dumps for both.

PR #1990, opened 2022-05-16
data: unify testing and prod paths for NewDB

dnephin:
TODO: write a better description before making this not a draft

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -22,22 +22,33 @@ import (
 )
 
 // NewDB creates a new database connection and runs any required database migrations
-// before returning the connection.
-func NewDB(connection gorm.Dialector) (*gorm.DB, error) {
-	db, err := NewRawDB(connection)
+// before returning the connection. The loadDBKey function is called after
+// initializing the schema, but before any migrations.
+func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*gorm.DB, error) {
+	db, err := newRawDB(connection)
 	if err != nil {
 		return nil, fmt.Errorf("db conn: %w", err)
 	}
 
-	if err = Migrate(db); err != nil {
+	if err := preMigrate(db); err != nil {
+		return nil, err
+	}
+
+	if loadDBKey != nil {
+		if err := loadDBKey(db); err != nil {
+			return nil, fmt.Errorf("load DB key failed: %w", err)
+		}
+	}
+
+	if err = migrate(db); err != nil {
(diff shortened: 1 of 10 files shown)


internal/server/data/data.go
```diff
@@ -22,22 +22,33 @@ import (
 )
 
 // NewDB creates a new database connection and runs any required database migrations
-// before returning the connection.
-func NewDB(connection gorm.Dialector) (*gorm.DB, error) {
-	db, err := NewRawDB(connection)
+// before returning the connection. The loadDBKey function is called after
+// initializing the schema, but before any migrations.
+func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*gorm.DB, error) {
```
dnephin, line 27: This is the primary change. Instead of having tests use `NewDB`, and production code using `NewRawDB` this allows both tests and production to use the same `NewDB` function.

It would be nice to have tests also use the same `loadDBKey` function, but that is a larger refactor, so I left it for another time. 

This change allows us to unexport `migrate` and `preMigrate`, so that we expose a smaller interface to other packages.

internal/testing/patch/patch.go
```diff
@@ -0,0 +1,37 @@
...
+type TestingT interface {
+	assert.TestingT
+	Helper()
+	TempDir() string
+	Cleanup(func())
+}
+
+// ModelsSymmetricKey sets model.ModelsSymmetricKey to a random key for the lifetime of the test.
+// This function modifies global state, it must not be used with t.Parallel.
+func ModelsSymmetricKey(t TestingT) {
```
dnephin, line 24: We were doing this same patching in 8 different places. Sometimes I'd be fine copying a test helper like this, but in this case most of the implementations did not use the right temp dir (`os.TempDir` is not automatically cleaned up), and most were not calling `t.Cleanup` to reset the value.

I've replaced these 8 places with calls to this function.

internal/server/data/migrations.go
```diff
@@ -466,19 +466,22 @@ func Migrate(db *gorm.DB) error {
 		// next one here
 	})
 
-	m.InitSchema(PreMigrate)
+	// TODO: why? isn't this already called by NewDB?
```
BruceMacD, line 469: We should capture this follow-up in an issue

BruceMacD (approved):
Nice clean-up

internal/server/data/migrations.go
dnephin: Created https://github.com/infrahq/infra/issues/2066 to track this

PR #1991, opened 2022-05-17
fix: optional grant role

mxyng:
## Summary

Grant role is optional. In the case a grant does not have a role, Infra will grant it `connect` privileges, allowing the user to authenticate with Infra but does not configure any authorization for the destination infrastructure


pdevine (approved):
Looks good.

If the role *is* specified and it's not a valid role, it should still error out.

ssoroka (approved):

PR #1994, opened 2022-05-17
api: remove deprecated API routes from the OpenAPI reference

dnephin:
## Summary

These two list grants routes were deprecated in a previous version, but I missed that they were still being added to the OpenAPI document used to generate our API reference documentation.

This PR refactors the API handler registration a bit to make it possible to register routes that are not added to the OpenAPI document. 

Changes include:
* change `DELETE` handlers to return an `EmptyResponse`, so that they have the same function signature as all the other handlers. This removes a few special cases and makes it possible to have a single function for all the handler registration.
* parameterize the handler registration using the `route` type, and `add` function. The `get`, `post`, `put`, and `delete` functions still exist for convenience. Special cases (like deprecated routes) can now use `add` directly without having to duplicate the common parts (binding request parameters to struct, sending api errors, and sending the response).

I will be building on this to complete the work in #1914


internal/server/apimigrator.go
```diff
@@ -23,26 +23,28 @@ type apiMigration struct {
...
+		// update request/response rewrite migrations to the new canonical path
+		if existing.rename == "" && existing.path == oldPath {
 			a.migrations[i].path = newPath
 		}
-		if len(mig.redirect) > 0 && mig.redirect == path {
-			a.migrations[i].redirect = newPath
+		// update redirect migrations to redirect to the new canonical path
+		if existing.rename != "" && existing.rename == oldPath {
+			a.migrations[i].rename = newPath
 		}
```
dnephin, line 41: A couple things changed here, so calling them out.

The main change is to compare strings using `== ""` and `!= ""` instead of length checks. That should be a little more obvious that these are strings.

The other changes are renaming of variables: `path` -> `oldPath`, `redirect` -> `rename`.

internal/server/routes.go
```diff
@@ -251,17 +225,15 @@ func redirectsFor(a *API, method, path string) []apiMigration {
 		if strings.ToUpper(migration.method) != method {
 			continue
 		}
-		if migration.redirect != path {
-			continue
-		}
-		if len(migration.redirect) > 0 {
+		// TODO: also check version
+		if migration.rename == path {
```
dnephin, line 229: A small simplification here. We can check redirect once for equality. `path` has to be a non-empty string, so this will never compare to non-rename migrations.

dnephin (commented):
I pushed one more commit with a few other small improvements.

internal/server/apimigrator.go
```diff
@@ -23,26 +23,28 @@ type apiMigration struct {
...
+	rename          string
 	requestRewrite  func(c *gin.Context)
 	responseRewrite func(c *gin.Context)
 }
 
-func addRedirect(a *API, method, path, newPath, version string) {
-	// update any existing migrations with the legacy path
-	for i, mig := range a.migrations {
-		if len(mig.redirect) == 0 && mig.path == path {
+// TODO: rename to addRename
```
BruceMacD, line 31: Is there something blocking renaming this function in this PR?

internal/server/routes.go
```diff
@@ -108,8 +109,18 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
...
+		path:         "/v1/users/:id/grants",
+		handler:      a.ListUserGrants,
+		omitFromDocs: true,
+	})
+	add(a, authn, route[api.Resource, *api.ListResponse[api.Grant]]{
+		method:       http.MethodGet,
+		path:         "/v1/groups/:id/grants",
+		handler:      a.ListGroupGrants,
+		omitFromDocs: true,
+	})
```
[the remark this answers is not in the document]
BruceMacD: Is your plan that we update all routes to the `add` func rather than `get, post, ...`?

BruceMacD (approved):
This all looks solid to me

internal/server/apimigrator.go
dnephin: Nothing blocking. It is going to increase the size of the diff quite a bit, so I wanted to avoid it initially to keep the focus on the other changes. It should be an easy "refactor -> rename" in GoLand in a follow up

internal/server/routes.go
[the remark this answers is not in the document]
dnephin: Good question! I was not planning on doing that. The reason these are special cases is because they use `omitFromDocs: true`. We could hide some of this with a `deprecated` helper (similar to `get`, `post`, etc) that accepts the method, path, and handler.

I think the `get`, `post`, etc helpers are nice and keep the route definitions concise and readable.

One thing I might look at would be some way to include the api migrations / previous versions along with the new canonical route, so that we have separate lists per method+path, instead of a single list for everything. I'm not sure what that would look like yet, but hopefully those could be passed along to the `get`, `post`, etc functions.

Issue #1441, opened 2022-04-01
Temporary password should only allow for updating password


BruceMacD:
## Is your feature request related to a problem? Please describe.
If a user logs in with a temporary password, they must set a new password. If they for some reason exit out of this flow they will have to request a new temporary password.

## Describe the solution you'd like
The temporary password should be able to set the user's password. Then the actual username/password are used for login after that is set.

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

PR #1999, opened 2022-05-17
fix: update password before saving session configs

mxyng:
## Summary

Login will prompt user to change their password if server requires a new password. This should be set and updated before saving the session to disk. Otherwise the user can interrupt the password update and still end up with a valid session.

Update login errors to not inform users of their previous session. It's not possible to know definitively whether the session is valid from the client alone so to avoid misleading the user, omit this message completely.

### DIFF
--- a/internal/cmd/errors.go
+++ b/internal/cmd/errors.go
@@ -3,6 +3,11 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"strings"
+
+	"github.com/muesli/termenv"
+
+	"github.com/infrahq/infra/internal/logging"
 )
 
 // internal errors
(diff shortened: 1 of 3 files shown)


internal/cmd/errors.go
```diff
@@ -29,31 +35,27 @@ var (
 	ErrTLSNotVerified = errors.New(`The authenticity of the host can't be established.`)
 )
 
-// User facing variable errors
-type FailedLoginError struct {
-	LoggedInIdentity string
-	LoginMethod      loginMethod
```
mxyng, line 35: Neither of these fields add much value.

`LoggedInIdentity` could be useful to pass in context without needing a config lookup but `Error()` is doing a lookup anyway by comparing the passed in value with the config value. 

`LoginMethod` is used to determine the error message to output but it's set by the parent who already has this and more info about the error

internal/cmd/login.go
```diff
@@ -146,29 +146,44 @@ func login(options loginCmdOptions) error {
...
 		}
-
 	}
 
 	return loginToInfra(client, loginReq)
 }
 
 func loginToInfra(client *api.Client, loginReq *api.LoginRequest) error {
 	loginRes, err := client.Login(loginReq)
-	if api.ErrorStatusCode(err) == http.StatusUnauthorized {
```
mxyng, line 157: Just checking `StatusUnauthorized` isn't sufficient as there may be other reasons why login will fail. One example is OIDC login errors which never actually returns unauthorized. Instead, it may return 404 Not Found if the provider was removed between starting and completing login.

internal/cmd/login.go
```diff
@@ -146,29 +146,44 @@ func login(options loginCmdOptions) error {
...
+			}
 		}
-	}
-	if err != nil {
+
 		return err
 	}
 
-	fmt.Fprintf(os.Stderr, "  Logged in as %s\n", termenv.String(loginRes.Name).Bold().String())
+	if loginRes.PasswordUpdateRequired {
```
mxyng, line 172: As mentioned in the PR description, the password update request needs to be handled before writing configs to disk. Otherwise the user can gets a session without completing the workflow

internal/cmd/login.go
dnephin: We should be normalizing all of those errors on the server to always return unauthorized.

By removing this check we will be misleading users. Ex: if the error is actually a connection error we should not say "your access key may be invalid". That's not going to help them debug the problem.

I think this needs to be `if api.ErrorStatusCode(err) == http.StatusUnauthorized`, and we should audit the `Login` endpoint to make sure it's returning proper status codes in all cases.

If we need to, we can add other `if api.ErrorStatusCode(err) == http.StatusNotFound`, or we could add other fields to the response error to provide more fine grained information about the error.

internal/cmd/login.go
dnephin: Not for this PR, but I wonder if we should make changes to the API so that the one-time-password session only allows them to change their password. That way they can continue to attempt to update it without being locked out.

With this change, if they ctrl-c at this prompt, they'll be locked out until an admin resets their one-time-password, is that right?

dnephin (commented):
Nice! One concern about removing the http status code check, but generally looks great!

internal/cmd/login.go
mxyng: You're correct that the OTP will be expended as soon as they login even if they don't change their password. The behaviour you're asking about is captured in #1441 but supporting that behaviour requires substantial changes to the backend service so it's out of scope of this PR

internal/cmd/login.go
mxyng: You're probably right. I took this approach since it was the path of least resistance. In hindsight, there are other scenarios I didn't consider. Will revert to check status code explicitly with additional checks for the cases I mentioned

dnephin (approved):
I don't know this code super well, but this LGTM.

I'd be great to have some kind of test coverage for this, but I'm guessing that requires the ability to test interactive things, and something like #1889, so is a lot of initial setup.

I'm going to prioritize #1889 first thing next week to try and get at least one test with that pattern, which should hopefully get us closer to being able to test this in CI.

mxyng:
I've tried adding tests but without interactive support, it's difficult to add useful cases. Once #1889 implemented, I'll add more coverage

PR #2000, opened 2022-05-17
maintain: update infra users

mxyng:
## Summary

Remove `--non-interactive`. This flag has no value. The only editable field is `password` which is explicitly not supported with `non-interactive`. If/when other editable fields are introduced, `non-interactive` may make a return.

Inform users of an updated password even if it's updating their own password

### DIFF
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -26,7 +26,7 @@ func newUsersCmd(cli *CLI) *cobra.Command {
 	}
 
 	cmd.AddCommand(newUsersAddCmd(cli))
-	cmd.AddCommand(newUsersEditCmd())
+	cmd.AddCommand(newUsersEditCmd(cli))
 	cmd.AddCommand(newUsersListCmd(cli))
 	cmd.AddCommand(newUsersRemoveCmd(cli))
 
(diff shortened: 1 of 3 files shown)


internal/cmd/users.go
```diff
@@ -246,7 +229,9 @@ func UpdateUser(name string, cmdOptions editUserCmdOptions) error {
 
 	if !isSelf {
 		// Todo otp: update term to temporary password (https://github.com/infrahq/infra/issues/1441)
-		fmt.Fprintf(os.Stderr, "  Updated one time password for user.\n")
+		fmt.Fprintf(os.Stderr, "  Updated password for %q.\n", name)
+	} else {
+		fmt.Fprintf(os.Stderr, "  Updated password.\n")
```
dnephin, line 234: In other PRs and above in this PR I notice you're using `cli.Output` for this kind of message, but here it's going to stderr.  What's the reason for that?

I don't have a strong preference for this, I think either could be fine as long as we are consistent.

internal/cmd/users.go
```diff
@@ -198,7 +183,7 @@ func CreateUser(req *api.CreateUserRequest) (*api.CreateUserResponse, error) {
 	return resp, nil
 }
 
-func UpdateUser(name string, cmdOptions editUserCmdOptions) error {
```
dnephin, line 201: Should this be called `updateUserPassword` ? Seems like it can only update that one field now.

dnephin (commented):
Looks good! Just a couple questions

internal/cmd/users.go
mxyng: No reason for this one. Will update it to use `cli.Output` instead

internal/cmd/users.go
mxyng: Will update this in a follow up PR. There already exists a `updateUserPassword` which #1999 is removing

dnephin (approved):
I guess testing this has the same challenges as https://github.com/infrahq/infra/pull/1999#pullrequestreview-977445332

Changes LGTM once the build/lint failures are fixed

PR #2001, opened 2022-05-17
maintain: only output actual removals

mxyng:
## Summary

infra keys rm: Inform users of removal only after successful removal
infra dst rm: Add message to inform users destination was removed

### DIFF
--- a/internal/cmd/destinations.go
+++ b/internal/cmd/destinations.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 
 	"github.com/infrahq/infra/api"
@@ -23,7 +21,7 @@ func newDestinationsCmd(cli *CLI) *cobra.Command {
 	}
 
 	cmd.AddCommand(newDestinationsListCmd(cli))
-	cmd.AddCommand(newDestinationsRemoveCmd())
+	cmd.AddCommand(newDestinationsRemoveCmd(cli))
 
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


internal/cmd/destinations.go
```diff
@@ -87,15 +85,13 @@ func newDestinationsRemoveCmd() *cobra.Command {
 				return err
 			}
 
-			if destinations.Count == 0 {
```
BruceMacD, line 90: I still think this is a nice thing to be notified of

internal/cmd/keys.go
```diff
@@ -112,28 +110,20 @@ func newKeysRemoveCmd() *cobra.Command {
...
+				err = client.DeleteAccessKey(key.ID)
+				if err != nil {
+					return err
+				}
 
-			err = client.DeleteAccessKey(key.ID)
-			if err != nil {
-				return err
-			}
+				issuedFor := key.IssuedForName
```
BruceMacD, line 119: What is this check for?

internal/cmd/destinations.go
mxyng: This will be in a follow up where it will error if no matches are found unless a flag is supplied. In the meantime, this is bringing the destinations remove command in line with other commands

internal/cmd/keys.go
mxyng: This is to set `issuedFor` in case `issuedForName` is empty. I believe it's for compatibility between the CLI and older version of the API which does not set `issuedForName`

BruceMacD (approved):

PR #2006, opened 2022-05-18
Use version handlers as alternative to apimigrator

dnephin:
This approach is functionally a superset to apimigrator. It supports every possible migration, because everything is simply a gin handler. It continues to use the existing version header. With this approach we can delete ~400 lines of code, and replace it with about 20 lines to look up the right handler for a version.

I think there are many advantages to this approach:
1. there is no extra framework for migrations. Everything is a regular handler that accepts an API type and returns an API response. We wrap those handlers with `wrapRoute` to turn them into `gin.HandlerFunc`.
2. the code is much easier to trace because functions call other functions directly. There's no map of functions that gets called based on lookups.
3. we no longer need to marshal and unmarshal the struct for each migration. This can be very expensive in the future. Instead we only ever unmarshal and marshal once.
4. it's a lot less code to maintain. ~20 lines instead of the existing ~400
5. it should be a lot more obvious how to implement a new version
6. it supports all possible migrations. `apimigrator` only works for some cases (renames and request/response changes), but doesn't work for less predictable changes (ex: changes to authn #3599).

The latest commit in this PR shows an example of adding a versioned handler, and a test to cover the behaviour.

TODO:
* [x] ensure the versions are in the correct order, either via a test or by sorting them before returning the routes


internal/server/routes.go
```diff
@@ -157,33 +155,85 @@ func add[Req, Res any](a *API, r *gin.RouterGroup, route route[Req, Res]) {
 		a.register(openAPIRouteDefinition(route))
 	}
 
-	handlers := includeRewritesFor(a, route.method, route.path)
-	handlers = append(handlers, func(c *gin.Context) {
+	handler := func(c *gin.Context) {
```
dnephin, line 158: This is the important part of the change. These 4 functions (below) and 2 types handle the version header, and allow us to replace all of apimigrator.

internal/server/routes.go
```diff
@@ -297,6 +304,33 @@ type isBlockingRequest interface {
...
+
+	// TODO: iterate in reverse order to benefit newer versions over older versions
+	for _, v := range versions {
+		if reqVer.GreaterThan(v.version) {
+			continue
+		}
+		return v.handler
+	}
+	return nil
+}
```
dnephin, line 332: This function is really the only new code in this PR. These 13 lines replace ~400 lines of migration code.

BruceMacD: This looks good, but as far as I can see there's nowhere that grantees the `versions` slice is in the correct order (oldest version to newest version). Should we sort it before applying the migrations?

dnephin: Yes, great point! That is something I forgot to add. 

We could do that at runtime before returning the routes, but I'm tempted to do it via a test (which will fail if any routes are in the wrong order). That way the routes should always be listed in the correct order in the source code.

dnephin:
I've added that test, and rebased to resolve the merge conflict.

internal/server/handlers.go
```diff
@@ -22,13 +24,62 @@ import (
...
+	versions   map[routeIdentifier][]routeVersion
+}
+
+type routeVersion struct {
+	version *semver.Version
+	handler func(c *gin.Context)
+}
+
+// nolint // will be used in the future
+func addVersionHandler[Req, Res any](a *API, method, path, version string, routeDef route[Req, Res]) {
```
BruceMacD, line 43: What is this going to be used for?

dnephin: Oops, this function is already used in this PR! I used it for the one api migration we have today (just below here). I forgot to remove this comment. I've done that now in the latest commit.

BruceMacD (approved):
lets get this one in 👍

jmorganca (approved):

internal/server/handlers.go
```diff
@@ -22,13 +24,61 @@ import (
...
+	version *semver.Version
+	handler func(c *gin.Context)
+}
+
+func addVersionHandler[Req, Res any](a *API, method, path, version string, routeDef route[Req, Res]) {
+	if a.versions == nil {
+		a.versions = make(map[routeIdentifier][]routeVersion)
+	}
+
+	routeDef.routeSettings.omitFromDocs = true
```
pdevine, line 47: I'm assuming this is because we would end up with conflicting docs for the route. Ideally we would be generating docs for every version that we release, so I _think_ this makes sense.

internal/server/routes.go
```diff
@@ -402,3 +438,21 @@ func (a *API) notFoundHandler(c *gin.Context) {
...
+	// CLI clients before v0.14.4 rely on sign-up being false to continue with login
+	type SignupEnabledResponse struct {
+		Enabled bool `json:"enabled"`
+	}
+
+	add(a, noAuthnNoOrg, http.MethodGet, "/api/signup", route[api.EmptyRequest, *SignupEnabledResponse]{
+		handler: func(c *gin.Context, _ *api.EmptyRequest) (*SignupEnabledResponse, error) {
+			return &SignupEnabledResponse{Enabled: false}, nil
+		},
+		routeSettings: routeSettings{
```
pdevine, line 452: Do we need a `defaultDeprecatedRouteSettingsGet` similar to `defaultRouteSettingsGet`? I would imagine as we add deprecated routes here we're going to pile up routeSettings.

Side note, the way we add routeSettings to the route feels weird given that we pass in part of the arguments through the settings here, but the `route.authenticationOptional` and `route.organiaztionOptional` settings get set through the route group (and have different names for the settings)

internal/server/routes.go
```diff
@@ -297,6 +304,33 @@ type isBlockingRequest interface {
...
+	return reqVer, nil
+}
+
+func handlerForVersion(versions []routeVersion, reqVer *semver.Version) func(c *gin.Context) {
+	if reqVer == nil {
+		return nil
+	}
+
+	// TODO: iterate in reverse order to benefit newer versions over older versions
+	for _, v := range versions {
```
pdevine, line 325: I may have missed it, but I didn't see anywhere where the versions of the route are being sorted. This means someone someone could add a deprecated route in the wrong order in `addPreviousVersionHandlers()` where an older route could match first instead of the correct route.

e.g. You have handlers for:
* A = (0.0.0->0.15.0)
* B = [0.15.0->0.16.0)
* C = [0.16.0, latest)

There is an entry in `addPreviousVersionHandlers()` for the change from 0.15.0 -> 0.16.0 and one for 0.16.0 -> latest.

If the person writing the second handler accidentally put the 0.0.0 -> 0.15.0 handler in first, and a client called requesting 0.16.5, handler A could match instead of handler B since reqVer is greater. This isn't an issue now, but I would worry about maintainability in the future.

internal/server/handlers.go
```diff
@@ -22,13 +24,61 @@ import (
...
+		Name        string `form:"name"`
+		ShowExpired bool   `form:"show_expired"`
+		api.PaginationRequest
+	}
+	addVersionHandler(a,
+		http.MethodGet, "/api/access-keys", "0.16.1",
+		route[listAccessKeysRequest_v0_16_1, *api.ListResponse[api.AccessKey]]{
+			routeSettings: defaultRouteSettingsGet,
+			handler: func(c *gin.Context, reqOld *listAccessKeysRequest_v0_16_1) (*api.ListResponse[api.AccessKey], error) {
+				req := &api.ListAccessKeysRequest{
```
pdevine, line 73: I _think_ you can just do `req := reqOld` here since the structs are the same.

pdevine (commented):
Overall I really like this approach. I think it's a nice simplification of the api migrator. The current way that we're doing it with reflection in the api migrator seems brittle (and I'm not super thrilled to have added the recursive function to pull apart nested API requests).

I'm assuming that the way deprecated routes will work is that we remove the route from the main routes, and then include the old route in `addPreviousVersionHandlers()`, which seems pretty straightforward.

dnephin:
> I'm assuming that the way deprecated routes will work is that we remove the route from the main routes, and then include the old route in addPreviousVersionHandlers(), which seems pretty straightforward.

We have [`API.deprecatedRoutes`](https://github.com/infrahq/infra/blob/main/internal/server/migrations.go#L45) today, which uses basically this same pattern. We don't actually set an error if someone is using the deprecated route with a newer API version, but we do hide it from the docs.  With versioned handlers we could add a handlers for the newer version that returns a HTTP `410 Gone` response code.

internal/server/handlers.go
dnephin: Ya, we don't want previous versions in the latest docs, but we do want to public previous versions of the docs.

internal/server/routes.go
dnephin: Bruce asked this as well. I added a test `TestAPI_AddPreviousVersionHandlers_Order` in this PR to make sure the order is correct.

I think a test is nicer than sorting than because it ensures the order is correct in the source code. That way a reader can read it in the correct order, instead of people adding routes in bad orders and forcing every reader to sort it in their head.

internal/server/handlers.go
dnephin: I think we'd need to type convert it as well. I decided against that approach because it will break as soon as we add another field to `api.ListAcessKeysRequest`. Someone adding a new field to that struct shouldn't need to come fix this code.

internal/server/routes.go
dnephin: > Do we need a `defaultDeprecatedRouteSettingsGet` similar to `defaultRouteSettingsGet`?

Maybe, I guess we'll see how many deprecated routes we end up with.

----

> the way we add routeSettings to the route feels weird given that we pass in part of the arguments through the settings here, but the route.authenticationOptional and route.organiaztionOptional settings get set through the route group (and have different names for the settings)

Fixing the names is easy enough.  The `routeGroup` struct has a TODO on it to be removed. I think we should get rid of route groups and set everything on package level `var` for each route. There are a couple routes that already do that.  That's going to be necessary for #3562.

I think that would fix the problem because all settings would come from that `var` . The `routeGroup` would no longer exist.

internal/server/handlers.go
```diff
@@ -22,13 +24,146 @@ import (
...
+		IssuedForName     string   `json:"issuedForName"`
+		IssuedFor         uid.ID   `json:"issuedFor"`
+		ProviderID        uid.ID   `json:"providerID"`
+		Expires           api.Time `json:"expires"`
+		ExtensionDeadline api.Time `json:"extensionDeadline"`
+	}
+	addVersionHandler(a,
+		http.MethodGet, "/api/access-keys", "0.18.0",
+		route[api.ListAccessKeysRequest, *api.ListResponse[accessKeyV0_18_0]]{
+			handler: func(c *gin.Context, req *api.ListAccessKeysRequest) (*api.ListResponse[accessKeyV0_18_0], error) {
```
dnephin, line 140: I think the only downside to this approach is that when we change an api request or response struct (in this case `api.ListAccessKeysRequest`) we'll need to check any existing version handlers, and replace them as well.

For example, if in 0.19.0 we add a migration for `ListAccessKeysRequest` struct, then this line here will also need to change to `listAccessKeysRequestV0_19_0` since that's the version of the request that was relevant to this version.

I think we can mitigate this by changing how we store these handlers for previous versions. Instead of a single function for everything, we can store them by entity. That way when something changes it should be a shorter list of migrations to check.

dnephin: Ah, it turns out this is already a problem now. There's a migration for this endpoint from 0.16.0 that needs this change. I'll push it to this branch.

internal/server/routes.go
dnephin: I fixed the names of the `routeGroup` fields in this PR.

internal/server/routes.go
pdevine: Could you put in a comment here which explains the ordering? I agree the tests should also cover this, but it should be clear to future authors that the order is important.

pdevine (approved):

internal/server/routes.go
```diff
@@ -173,8 +176,8 @@ type routeIdentifier struct {
 // constructed from the get, post, put, del helper functions.
 type routeGroup struct {
 	*gin.RouterGroup
-	noAuthentication bool
-	noOrgRequired    bool
+	authenticationOptional bool
```
jmorganca, line 179: Nice rename! Negative booleans names are always confusing

jmorganca:
Overall looks good!

internal/server/routes.go
```diff
@@ -300,6 +309,33 @@ type isBlockingRequest interface {
...
+	}
+	return reqVer, nil
+}
+
+func handlerForVersion(versions []routeVersion, reqVer *semver.Version) func(c *gin.Context) {
+	if reqVer == nil {
+		return nil
+	}
+
+	// TODO: iterate in reverse order to benefit newer versions over older versions
```
dnephin, line 329:
```suggestion
```

internal/server/access_keys.go
```diff
@@ -58,3 +60,120 @@ func (a *API) CreateAccessKey(c *gin.Context, r *api.CreateAccessKeyRequest) (*a
...
+		oldResp := &api.ListResponse[accessKeyV0_18_0]{
+			PaginationResponse: latest.PaginationResponse,
+			Count:              latest.Count,
+			LastUpdateIndex:    latest.LastUpdateIndex,
+			Items:              make([]accessKeyV0_18_0, len(latest.Items)),
+		}
+		for i, item := range latest.Items {
+			oldResp.Items[i] = newAccessKeyV0_18_0FromLatest(item)
+		}
+		return oldResp
```
mxyng, line 109: This can use `api.NewListResponse`

```suggestion
		return api.NewListResponse(latest.Items, latest.PaginationResponse, 	func(newItem api.AccessKey, oldItem accessKeyV0_18_0) {
    			return newAccessKeyV0_18_0FromLatest(newItem)
		})
```

internal/server/routes.go
mxyng: Nit: `optionalAuthentication` and `optionalOrganization`?

internal/server/access_keys.go
dnephin: It could, but using `NewListResponse` is actually what almost caused me to introduce a bug in the list grants endpoint. That handler doesn't set `LastUpdateIndex`.

Maybe we should have a function to copy `ListResponse` and change the type of the items.

internal/server/routes.go
dnephin: I think `optionalOrganization`  would make sense for a field that identifies the organization by ID or name. This is a boolean about if the organization is optional or not. I think putting optional second reads a bit better because it's short for "organization is optional" -> `organizationOptional`

internal/server/access_keys.go
dnephin: I created `CopyListResponse` which calls `NewListResponse` and sets the extra fields. I think this works well for now.

I have been finding `api.ListResponse` a bit difficult to work with. It forced us to put other metadata onto this struct, instead of being able to embed it where necessary.

I think we might want to consider having separate api structs for each list response, and have each of them embed a `ListMetadata` struct that provides the `Count` and `PaginationResponse` fields. That would let us embed other structs without overloading `ListResponse`. I created #3993 to track that change.

api/types.go
```diff
@@ -141,6 +141,12 @@ type ListResponse[T any] struct {
 	LastUpdateIndex `json:"-"`
 }
 
+func (r ListResponse[T]) SetHeaders(h http.Header) {
+	if r.LastUpdateIndex.Index > 0 {
+		h.Set("Last-Update-Index", strconv.FormatInt(r.LastUpdateIndex.Index, 10))
+	}
+}
```
dnephin, line 148: I forgot to comment on this. I moved this over from `ListGrantsResponse` because we need this header set for previous versions of the `ListGrants` endpoint as well.

I think #3993 would allow us to move this to the `BlockingRequest` type, that we can embed in the list responses.

PR #2007, opened 2022-05-18
improve: refactor api migrator to work nice with redirects

ssoroka:
## Summary

The general idea is that we build the list of redirect/rewrite middleware backwards from the defined migrations, and this lets us do the right thing when we encounter redirects etc. Redirects now basically fork the path and end up with multiple paths, and all of them get bound to gin. Before there was no way to make each path have its own middleware, but now each is selected so it will exactly have the correct versioning middleware for that particular path/version

## Checklist

### DIFF
--- a/internal/server/apimigrator.go
+++ b/internal/server/apimigrator.go
@@ -224,30 +224,31 @@ func bindRoute(a *API, method, path string, handler gin.HandlerFunc, methodFunc 
 		if strings.ToUpper(migration.method) != method {
 			continue
 		}
-		route, ok := routes[migration.path]
-		// nolint:gocritic
-		if ok {
+
+		// check if the migration path matches the route we're defining
+		if route, ok := routes[migration.path]; ok {
 			if migration.requestRewrite != nil {
 				route = append([]gin.HandlerFunc{migration.requestRewrite}, route...)
 			}
 			if migration.responseRewrite != nil {
 				route = append([]gin.HandlerFunc{migration.responseRewrite}, route...)
 			}
-		} else if len(migration.redirect) > 0 {
-			// Redirects end up duplicating/splitting into a new path without destroying the old one
-			route, ok = routes[migration.redirect]
-			if !ok {
-				panic(fmt.Sprintf("invalid migration: there is no http %s route named %q defined to redirect to", migration.method, migration.redirect))
-			}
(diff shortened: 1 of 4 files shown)


internal/server/apimigrator.go
```diff
@@ -206,6 +208,55 @@ func (m *apiMigration) RedirectHandler() gin.HandlerFunc {
...
+			continue
+		}
+
+		// check if the migration redirects to the route we're defining
+		if len(migration.redirect) > 0 {
+			// Redirects end up duplicating/splitting into a new path without destroying the old one
+			if route, ok := routes[migration.redirect]; ok {
+				route = append([]gin.HandlerFunc{migration.RedirectHandler()}, route...)
+				if migration.redirectHandler != nil {
+					// if the migration has a custom redirect handler, prepend it.
```
BruceMacD, line 246: What would be a custom redirect handler?

internal/server/apimigrator.go
```diff
@@ -26,23 +27,22 @@ type apiMigration struct {
...
-		if len(mig.redirect) == 0 && mig.path == path {
-			a.migrations[i].path = newPath
-		}
-		if len(mig.redirect) > 0 && mig.redirect == path {
-			a.migrations[i].redirect = newPath
-		}
+func addRedirect(a *API, method, path, newPath, version string, optMiddleware ...gin.HandlerFunc) {
+	var optRedirectMiddleware gin.HandlerFunc
+	if len(optMiddleware) > 0 {
+		optRedirectMiddleware = optMiddleware[0]
```
BruceMacD, line 37: Setting the redirect middleware to the first optRedirectMiddleware means it will go through the whole chain of middleware if there are many?

BruceMacD (commented):
Looks good to me, just a couple questions

internal/server/apimigrator.go
ssoroka: this just allows addRedirect to have a single optional custom middleware handler. There should only ever be one, but most redirects don't have it. There's an example of this in the tests:
```go
	addRedirect(a, "get", "/oldtest/:id", "/awesometest", "0.1.3", func(c *gin.Context) {
		id := c.Param("id")
		q := c.Request.URL.Query()
		q.Add("vegetableCount", id)
		c.Request.URL.RawQuery = q.Encode()
		c.Next()
	})
```

BruceMacD (approved):

PR #2013, opened 2022-05-20
feat: agent to keep kube config up to date

### LINKED ISSUE #1820: Keep kubeconfig always up to date

BruceMacD:
## Summary
In order to keep the user's local state up to date we need an agent that runs continuously in the background.  This change adds a hidden `infra agent` command which is executed on login and continues to run. The only task in the agent right now is to sync the Kubernetes configuration automatically. If the agent encounters an error it will stop running, this means that when a user logs out that the agent also quits.

The agent does not currently run on start-up. This requires different installation steps for different operating systems, and it will be a separate change.

## Checklist

## Related Issues

Resolves #1820

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,6 +61,7 @@ require (
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
 	github.com/tklauser/numcpus v0.4.0 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 )
 
 require (
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
 	github.com/tklauser/numcpus v0.4.0 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
-	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 )
 
 require (
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,15 @@ require (
 )
 
 require (
+	github.com/go-ole/go-ole v1.2.6 // indirect
+	github.com/lufia/plan9stats v0.0.0-20211012122336-39d0f177ccd0 // indirect
+	github.com/power-devops/perfstat v0.0.0-20210106213030-5aafc221ea8c // indirect
+	github.com/tklauser/go-sysconf v0.3.10 // indirect
+	github.com/tklauser/numcpus v0.4.0 // indirect
+	github.com/yusufpapurcu/wmi v1.2.2 // indirect
+)
+
+require (
 	github.com/PuerkitoBio/purell v1.1.1 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
 	github.com/armon/go-metrics v0.3.10 // indirect
@@ -136,6 +145,7 @@ require (
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/ryanuber/go-glob v1.0.0 // indirect
+	github.com/shirou/gopsutil/v3 v3.22.4
 	github.com/ugorji/go/codec v1.2.6 // indirect
 	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c // indirect
 	go.uber.org/atomic v1.9.0 // indirect
--- a/internal/cmd/agent.go
+++ b/internal/cmd/agent.go
@@ -1,0 +1,139 @@
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -137,6 +146,7 @@ require (
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/ryanuber/go-glob v1.0.0 // indirect
+	github.com/shirou/gopsutil/v3 v3.22.4
```
BruceMacD, line 149: The changes to the dependencies come from this library I use to check if the agent process is running. It is better to pull in this dependency than check manually in my opinion due to difference in operating systems.

There may come a point where we run the agent on start-up and we won't need to worry about if it is running. When we get there we can remove this library.

internal/cmd/list.go
```diff
@@ -145,3 +101,56 @@ func list(cli *CLI) error {
 
 	return writeKubeconfig(user, destinations.Items, grants.Items)
 }
+
+func getUserDestinationGrants() (*api.User, *api.ListResponse[api.Destination], *api.ListResponse[api.Grant], error) {
```
BruceMacD, line 105: Pulled this out so I can call it from the agent without executing `infra list`. No logical changes.

internal/cmd/agent.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/internal/repeat"
+)
+
+// AgentConfig stores details about the agent in config separate from the CLI config file
+// this config file must be separate to avoid concurrent writes with the CLI config
+type AgentConfig struct {
+	ProccessID int `json:"pid"` // used to manage the agent lifecycle
```
mxyng, line 24: Does this need to be structured? It's usually enough to write the pid to a file as plaintext, e.g.

```
$ cat agent.pid
12345
```

internal/cmd/login.go
```diff
@@ -200,6 +200,21 @@ func loginToInfra(client *api.Client, loginReq *api.LoginRequest) error {
...
+	if err != nil {
+		// do not block login, just proceed, potentially without the agent
+		logging.S.Errorf("unable to check background agent: %v", err)
+	}
+
+	if !backgroundAgentRunning {
+		// the agent is started in a separate command so that it continues after the login command has finished
+		if err := execAgent(); err != nil {
+			// user still has a valid session, so do not fail
+			logging.S.Errorf("exec login agent: %w", err)
```
mxyng, line 213: does this need both `logging` and `fmt.Fprintf`?

mxyng:
Looks good overall. A few minor comments

internal/cmd/agent.go
BruceMacD: Made this change in [c92fe61](https://github.com/infrahq/infra/pull/2013/commits/c92fe6171674757cade345501c89e3a432c2ba53)

internal/cmd/login.go
BruceMacD: I was writing the actual error in the logger and a human friendly warning in the output. Is there a different convention I should follow here?

internal/cmd/agent.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+// configAgentRunning checks if the agent process stored in config is still running
+func configAgentRunning() (bool, error) {
+	config, err := readAgentConfig()
+	if err != nil {
+		if os.IsNotExist(err) {
+			// this is the first time the agent is running, suppress the error and continue
+			logging.S.Debug(err)
```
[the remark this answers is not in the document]
BruceMacD: This is called during `infra login` so it gets output to the actual interactive prompt 👍

internal/cmd/agent.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return running, nil
+}
+
+// execAgent executes the agent in the background and stores its process ID in configuration
+func execAgent() error {
+	cmd := exec.Command("infra", "agent")
```
mxyng, line 75: This will fail if `infra` is not in `PATH` or if the `infra` in `PATH` is actually a different binary

internal/cmd/login.go
mxyng: Not sure for this case since it's technically an error but I don't know if it should error out

internal/cmd/agent.go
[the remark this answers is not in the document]
mxyng: Sorry, I didn't mean this case specifically but agent logs in general, e.g. [line 130](https://github.com/infrahq/infra/pull/2013/files#diff-fa09e72c0519d3e6e84a9652bdfbf0a167f326170a8b520cc81b0d90161841afR130)

internal/cmd/login.go
BruceMacD: Switched to just error logging

internal/cmd/agent.go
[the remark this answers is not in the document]
BruceMacD: Ah, I see now. Added a file logger here: [90b2890](https://github.com/infrahq/infra/pull/2013/commits/90b289006a5978f3f9fa37a8837dbf051a5b686f)

mxyng (approved):

PR #2016, opened 2022-05-21
improve: remove `ui` from `go generate`

jmorganca:
This removes UI from go generate since it generates too many little files that change too often.

Instead with this change we:

1. Build the UI static files at release under `internal/server/ui/static`
2. Replace tests that relies on built UI code with a mock filesystem

In passing this replaces a dependency we don't need anymore `github.com/spf13/afero`


internal/server/ui/.gitignore
```diff
@@ -0,0 +1 @@
+static
```
jmorganca, line 1: This is a bit of a trick for `go:embed`, so that it always works (even with the ui static files directory `ui/static` being empty). Consequently all generated ui files are under `ui/static` now

internal/server/server.go
```diff
@@ -80,6 +79,7 @@ type ListenerOptions struct {
 type UIOptions struct {
 	Enabled  bool
 	ProxyURL types.URL `mapstructure:"proxyURL"`
+	FS       fs.FS     `mapstructure:"-"`
```
jmorganca, line 82: This shouldn't really need to exist but was added for injecting a fake `fs.FS` in tests – open to a better alternative if one exists!

internal/server/static_fs_test.go
```diff
@@ -1,68 +1,77 @@
...
+		},
+	}
 
 	sfs := &StaticFileSystem{
-		base: fs,
+		base: http.FS(fs),
 	}
 
-	exists := sfs.Exists("/", "/dashboard/foo")
+	exists := sfs.Exists("/", "/foo.html")
```
dnephin, line 75: Doesn't this change the meaning of the test? The test name suggests we are trying to cover when `/foo` is queried that it looks for a file called `foo.html`.

Does this still work?
```suggestion
	exists := sfs.Exists("/", "/foo")
```

internal/server/routes.go
```diff
@@ -282,9 +283,14 @@ func (a *API) notFoundHandler(c *gin.Context) {
...
-	if err != nil {
-		logging.S.Error(err)
+
+	buf := []byte("404 not found")
+	var err error
+	if a.server.options.UI.FS != nil {
+		buf, err = fs.ReadFile(a.server.options.UI.FS, "ui/static/404.html")
+		if err != nil {
+			logging.S.Error(err)
+		}
```
dnephin, line 293: This only works in tests, right? It will no longer serve the 404 page from the static files because we never set the FS on `server.options.UI.FS`.

internal/server/server.go
dnephin: A shim like this is a good option, but I think in this case we might only be testing this shim.

I wonder if we can use `uiFS` and patch that to a different value in tests, to bring the tests closer to production.

dnephin (commented):
Looks good! I think this is a good approach, but I think we might have lost the nice 404 page from the static files. I think a couple small changes can bring that back.

internal/server/ui/.gitignore
mxyng: It looks like `go:embed` needs the directory to not be empty. Instead of including a `.gitignore`, you could include a file `.empty` or something and generate into `ui/`. In the top level `.gitignore`, you can add an entry `ui/*` to the contents after adding `aaa/.empty` to the repo

internal/server/server.go
mxyng: What's the behaviour if the requested file isn't found, i.e. if someone neglects to generate the UI for dev?

internal/server/ui/.gitignore
jmorganca: Yeah great point. That was my initial stab at this – although this makes `.gitignore` the `.empty` file, but also one that's handy to remove `static` from being checked in.

internal/server/server.go
jmorganca: It returns a 404 with page body `404 not found`

internal/server/static_fs_test.go
dnephin: I applied this change in a commit, and the test still passes.

internal/server/routes.go
```diff
@@ -282,12 +283,19 @@ func (a *API) notFoundHandler(c *gin.Context) {
...
-	buf, err := assetFS.ReadFile("ui/404.html")
-	if err != nil {
-		logging.S.Error(err)
+
+	buf := []byte("404 not found")
+
+	const filePath404 = "ui/static/404.html"
+	uiFS := a.server.options.UI.FS
+	if _, err := fs.Stat(uiFS, filePath404); err == nil {
+		buf, err = fs.ReadFile(uiFS, filePath404)
```
dnephin, line 292: I changed this if from `!= nil`, to stat the file. If the UI static files exist then it'll use the page from the static assets. Otherwise it will use the string above.

internal/server/server.go
```diff
@@ -101,6 +102,7 @@ type Addrs struct {
 
 // newServer creates a Server with base dependencies initialized to zero values.
 func newServer(options Options) *Server {
+	options.UI.FS = uiFS
```
dnephin, line 105: This is the important bit that fixes the problem with the `notFoundHandler` not being able to access the static fs.

dnephin (approved):
LGTM! Thanks for making this change

internal/server/server.go
dnephin: Using a package-level variable as the shim didn't work because the variable needs to be an `embed.FS` for `go:embed` to work, and we can't create an `embed.FS` with actual files in it. 

So instead I changed everything to read `options.UI.FS` and ignore the package level variable.

dnephin:
I rebased with `main`, and tests appear to be running again

PR #2029, opened 2022-05-24
improve: update ui to use api 0.13.0

jmorganca:


ui/components/admin.js
```diff
@@ -57,43 +57,43 @@ function Grant ({ id, subject, grants }) {
 }
 
 export default function () {
-  const { data: grants } = useSWR(() => '/v1/grants?resource=infra&privilege=admin', { fallbackData: [] })
+  const { data: { items: grants } = {} } = useSWR(() => '/api/grants?resource=infra&privilege=admin', { fallbackData: [] })
```
jmorganca, line 60: This needs an update to `{ items : [] }`

ui/components/grant.js
```diff
@@ -65,25 +65,25 @@ export default function ({ id }) {
...
+        let res = await fetch(`/api/users?name=${email}`)
         const data = await res.json()
 
         if (!res.ok) {
           throw data
         }
 
-        if (data.length === 0) {
-          res = await fetch('/v1/identities', {
+        if (data?.items?.length === 0) {
```
hoyyeva, line 75: i think we can do `data?.count`

ui/components/grant.js
```diff
@@ -137,16 +137,16 @@ export default function ({ id }) {
       {error && <ErrorMessage message={error} />}
       {list?.length > 0 &&
```
hoyyeva, line 138: this should be `list?.count`?

hoyyeva: or maybe on the line 24 
``const { data: { items: list } = {} } = useSWR(() => `/api/grants?resource=${destination.name}`)``

it should be 
``const { data: { items: list, count } = {} } = useSWR(() => `/api/grants?resource=${destination.name}`)``

hoyyeva (approved):

PR #2037, opened 2022-05-24
fix: allow password and accesskey on config user

### LINKED ISSUE #1958: Identities configuration should allow setting both Password and AccessKey
## Describe the bug

It should be possible to configure both a password and an access key. Right now these fields are mutually exclusive.

This doesn't work:

```
  - user: alice@email.com
    password: password
    accessKey: aaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbb
```

This would work:

```
  - user: alice@email.com
    password: password
  - user: alice@email.com
    accessKey: aaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbb
```

BruceMacD:
## Summary
It is now valid for a user identity to be assigned both a password and an access key in config. Updating to allow that and adding some more tests around config users.

## Checklist

## Related Issues

Resolves #1958

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -36,10 +36,10 @@ type Grant struct {
 
 type User struct {
 	Name      string `mapstructure:"name" validate:"excluded_with=Email"`
-	AccessKey string `mapstructure:"accessKey" validate:"excluded_with=Password"`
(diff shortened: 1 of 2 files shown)


mxyng (approved):
:shipit:

PR #2041, opened 2022-05-24
improve: require Infra-Version header

ssoroka:
## Summary

The Infra-Version header works best if clients are required to supply it. This PR makes the header required and errors with a helpful message if it is not supplied.

## Checklist

### DIFF
--- a/docs/api/guidelines.md
+++ b/docs/api/guidelines.md
@@ -81,6 +81,15 @@ resources which do not have deep, nested hierarchies. As a rule of thumb,
 if modifying a child resource also requires modifying the parent resource,
 you should probably make your resource higher up in the hierarchy.
 
+## Version Control
+
+The Infra API is versioned. Requests to the API must contain a header named "Infra-Version".
+The best practice is to set this to the version matching the API docs reference you're using, or the version of the server you're using.
+Once you set this value you can forget about it until you want to use features from newer API versions.
+A valid version header looks like this:
+
+    Infra-Version: 0.13.0
+
 ## Method Calls
 
 Method calls are the 'verbs' which determine how to interact with resources. Our API uses standard CRUD (Create, Read,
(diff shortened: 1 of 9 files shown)


docs/api/openapi3.json
```diff
@@ -689,6 +689,17 @@
         "description": "ListAccessKeys",
         "operationId": "ListAccessKeys",
         "parameters": [
+          {
+            "in": "header",
+            "name": "Infra-Version",
+            "required": true,
+            "schema": {
+              "description": "Version of the API being requested",
+              "example": "0.13.1+dev",
```
dnephin, line 698: The examples should probably not include the `+dev`. Does `+dev` get get removed before publish?

dnephin (approved):
LGTM, just a question about the version in the examples.

docs/api/openapi3.json
ssoroka: it gets generated on release with the release version, so the published one won't have dev in it

BruceMacD (approved):
Looks good to me too, we will have to make it easy to find previous version API docs but that is another issue.

mxyng (approved):

docs/api/guidelines.md
```diff
@@ -81,6 +81,15 @@ resources which do not have deep, nested hierarchies. As a rule of thumb,
 if modifying a child resource also requires modifying the parent resource,
 you should probably make your resource higher up in the hierarchy.
 
+## Version Control
```
pdevine, line 84: This part feels a little bit weird here, as the sections are more about how to design/name endpoints and not how to call the API specifically. I added the changes lower in the document which explain how versioning works.

dnephin: It seems like we are missing a section in our API reference doc to communicate this kind of stuff (versioning, auth, common headers like `Accept`). I think that's pretty common for API reference docs. Maybe we could use https://swagger.io/specification/#info-object for this?

Otherwise we could make a separate doc and use https://swagger.io/specification/#external-documentation-object

PR #2046, opened 2022-05-25
fix: fail fast on duplicate config access key

### LINKED ISSUE #1983: Specifying an existing access key in config for a different user should fail-fast
## Describe the bug
When an existing access key is specified for a different or new user in config that user fails to be created (which is good because it prevents the access key from changing the identity it was issued for in this case).

Steps to reproduce:
1. Create a config with a user/key:
```
server:
  config:
    users:
      - name: test@example.com
        accessKey: aaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbb
```
2. Apply the config.
`helm upgrade --install -f infra.yaml`
3. Observe that you can successfully login.
4. Update the config to specify a new user, but with the same access key previously used.
```
server:
  config:
    users:
      - name: different@example.com
        accessKey: aaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbb
```
5. Apply the config.
`helm upgrade --install -f infra.yaml`
6. Attempt to login with the access key. It will fail.

## Expected behavior
Infra fails to apply the config and error logs that the access key specified for a user was already in use.

BruceMacD:
## Summary
When an access key that is specified for a user in config is already specified for another user, fail fast. Access keys can only be linked to one user so this is an invalid configuration.

Ex:
```
users:
      - name: test
        accessKey: aaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbb
      - name: anotherTest
        accessKey: aaaaaaaaaa.bbbbbbbbbbbbbbbbbbbbbbbb
```
```
{"level":"error","ts":1653485595.127271,"caller":"infra/main.go:17","msg":"creating server: configs: load users: access key assigned to \"anotherTest\" is already assigned to another user, a user's access key must have a unique ID"}
```

## Checklist

## Related Issues

Resolves #1983

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -859,6 +859,10 @@ func (s Server) loadAccessKey(db *gorm.DB, identity *models.Identity, key string
 		return nil
 	}
 
+	if accessKey.IssuedFor != identity.ID {
+		return fmt.Errorf("access key assigned to %q is already assigned to another user, a user's access key must have a unique ID", identity.Name)
+	}
+
 	accessKey.Secret = secret
 
 	if err := data.SaveAccessKey(db, accessKey); err != nil {
--- a/internal/server/config_test.go
+++ b/internal/server/config_test.go
@@ -800,7 +800,7 @@ func TestLoadAccessKey(t *testing.T) {
 		assert.NilError(t, err)
 
 		err = s.loadAccessKey(s.db, alice, testAccessKey)
-		assert.Error(t, err, "access key assigned to alice is already assigned to another user, a user's access key must have a unique ID")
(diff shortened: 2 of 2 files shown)


mxyng:
To clarify, the first access key will be loaded but the second will error and since config loading is in a transaction, the entire config will be rolled back right?

BruceMacD:
> To clarify, the first access key will be loaded but the second will error and since config loading is in a transaction, the entire config will be rolled back right?

Yes, the config will be rolled back and the pod will fail to stand-up.

mxyng (approved):

PR #2049, opened 2022-05-25
fix: dashboard loading and redirect states

jmorganca:
Some of the loading states were flickering due to empty arrays `[]` being a truthy value in Javascript. Also removes the loading spinner since this looked like flickering as well on a fast connection

### DIFF
--- a/ui/pages/users/index.js
+++ b/ui/pages/users/index.js
@@ -8,7 +8,6 @@ import { useAdmin } from '../../lib/admin'
 
 import EmptyTable from '../../components/empty-table'
 import PageHeader from '../../components/page-header'
-import Loader from '../../components/loader'
 import Table from '../../components/table'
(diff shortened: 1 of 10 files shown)


ui/pages/users/index.js
```diff
@@ -119,45 +119,43 @@ export default function Users () {
...
-              <Sidebar
-                handleClose={() => setSelectedUser(null)}
-                title={selectedUser.name}
-                profileIcon={selectedUser.name[0]}
-              >
-                <SidebarContent selectedUser={selectedUser} admin={admin} setSelectedUser={setSelectedUser} />
-              </Sidebar>}
-          </div>
-          )}
+      {!loading && (
```
hoyyeva, line 122: we can remove the import Loader from this file then since we are not using it anymore

jmorganca: great catch!

Issue #2014, opened 2022-05-20
Remove unused JWT claim


BruceMacD:
## Describe the bug
We are adding a nonce value to the JWT but it is not validated or checked anywhere. A nonce is used to prevent replay attacks. This is useful for an implicit OIDC flow, but not needed for our JWTs that will be sent and used multiple times.

## Expected behavior
No unnecessary claims on the JWT.

## Environment

```
$ infra version
 Client: 0.13.1
 Server: 0.13.1
```

## Additional context
This claim is from a very old version of Infra and was just not cleaned up.

BruceMacD:
Re-opening. Needed to do a transitional change (#2086) for backwards-compatibility. We can remove this completely in a future version.

PR #2086, opened 2022-05-26
fix: keep jwt nonce for version compatibility

### LINKED ISSUE #2081: Removing the JWT nonce is not backwards compatible
## Describe the bug
The nonce on the JWT is marked as a required field on claims in v0.13.0 and earlier.

## Expected behavior
The nonce is not needed, it should not be added to the JWT. For a smooth transition we should update the claims to not require a nonce. Then eventually remove it.

## Environment

```
$ infra version
```

```
$ kubectl version
```

BruceMacD:
## Summary
This JWT nonce is not needed, but it is marked as required on the claims struct so the struct will fail validation on previous versions of infra without this field. This is a transitional change that makes the `nonce` not required, we can remove it fully in a few versions when 0.13.0 connectors have been phased out.

## Checklist

## Related Issues

Resolves #2081
Re-opened #2014 to track removing this completely.

### DIFF
--- a/internal/server/data/token.go
+++ b/internal/server/data/token.go
@@ -9,6 +9,7 @@ import (
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal/claims"
+	"github.com/infrahq/infra/internal/generate"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
 )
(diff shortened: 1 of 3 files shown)


internal/server/data/token.go
```diff
@@ -51,6 +52,7 @@ func createJWT(db *gorm.DB, identity *models.Identity, groups []string, expires
 	custom := claims.Custom{
 		Name:   identity.Name,
 		Groups: groups,
+		Nonce:  generate.MathRandom(10),
```
BruceMacD, line 55: This is a math random rather than a crypto random to not drain the entropy on a field that is not used.

mxyng (approved):

PR #2090, opened 2022-05-26
feat: enable server UI by default

mxyng:
## Summary

Enable server UI by default


jmorganca (approved):
lgtm once tests pass

internal/cmd/server.go
```diff
@@ -73,7 +73,7 @@ func newServerCmd() *cobra.Command {
 	cmd.Flags().String("db-encryption-key", "", "Database encryption key")
 	cmd.Flags().String("db-encryption-key-provider", "", "Database encryption key provider")
 	cmd.Flags().Bool("enable-telemetry", false, "Enable telemetry")
-	cmd.Flags().BoolVar(&options.UI.Enabled, "enable-ui", false, "Enable Infra server UI")
```
mxyng, line 76: Keeping this option and hiding it isn't necessary since it's not something that should be used currently

Issue #2082, opened 2022-05-26
ui: logging out and logging in with a OTP skips password change dialog


jmorganca:
Steps to reproduce:

1. Log in as a user with a one time password
2. Create another user
3. Log in with second one time password

hoyyeva:
this is another use case:
If user gets created with OTP, logs in, updates password, and is removed, recreating the user generates a new OTP but does not prompt to change password after using new OTP. This CLI handles this correctly by forcing the user to change their password.

PR #2091, opened 2022-05-26
fix: fix auth ui redirects for new otp users

jmorganca:
Fixes https://github.com/infrahq/infra/issues/2082

Also refactors ui auth code to be easier to reason about

### DIFF
--- a/ui/components/layouts/dashboard.js
+++ b/ui/components/layouts/dashboard.js
@@ -4,8 +4,9 @@ import useSWR, { useSWRConfig } from 'swr'
 
 import { useAdmin } from '../../lib/admin'
 import ProfileIcon from '../profile-icon'
+import AuthRequired from '../auth-required'
 
-export default function ({ children }) {
+function Dashboard ({ children }) {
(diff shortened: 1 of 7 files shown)


ui/components/layouts/dashboard.js
```diff
@@ -106,3 +107,13 @@ export default function ({ children }) {
     </div>
   )
 }
+
+export default function ({ children }) {
+  return (
+    <AuthRequired>
```
hoyyeva, line 113: should we move the `AuthRequired` to `_app.js` so we don't need to wrap it in inside the dashboard and fullscreen components?

jmorganca: Great question – I thought about this, but `_app.js` is a component that's rendered for every page, whereas `AuthRequired` should only be wrapping a subset of pages – the idea here that all `Dashboard` and `Fullscreen` layouts need auth – but you could do it on an individual page basis too

hoyyeva (approved):

PR #2098, opened 2022-05-27
cli: add test for login

### LINKED ISSUE #1889: testing: add a CLI integration test of signup with a real API server

dnephin:
Adds a test for `infra login` that does an initial setup of an admin user. The test uses a real server in-process, and uses `go-expect` to navigate the interactive prompts. I found a recommendation for testing survey here: https://github.com/AlecAivazis/survey#testing, which worked nicely and did not require any modification to survey.

Resolves #1889


internal/cmd/login_test.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+
+		exp := expector{console: console}
+		exp.ExpectString(t, "Username:")
+		exp.Send(t, "admin\n")
+		exp.ExpectString(t, "Password")
+		exp.Send(t, "password\n")
+		exp.ExpectString(t, "Confirm")
+		exp.Send(t, "password\n")
+		exp.ExpectString(t, "Logged in as")
```
dnephin, line 68: The test was failing here in CI. I've increased the timeout from 500ms to 2s to see if that helps.

internal/certs/certs.go
```diff
@@ -142,3 +124,29 @@ func SelfSignedOrLetsEncryptCert(manager *autocert.Manager, serverName string) f
 		return &keypair, nil
 	}
 }
+
+func newCA() (*x509.Certificate, crypto.PrivateKey, error) {
```
dnephin, line 128: I moved this out into a function to make it easier to generate the certs that I checked into git.

I used:

```go
func TestCreateCerts(t *testing.T) {
	ca, pk, err := newCA()
	assert.NilError(t, err)

	certBytes, keyBytes, err := GenerateCertificate([]string{"127.0.0.1", "::1"}, ca, pk)

	ioutil.WriteFile("localhost.crt", certBytes, 0600)
	ioutil.WriteFile("localhost.key", keyBytes, 0600)
}
```

We could check this in somewhere as a way to re-generate them. They only last a year, so we'll have to do it again in 11 months.

internal/cmd/server.go
```diff
@@ -74,18 +81,19 @@ func newServerCmd() *cobra.Command {
...
 	return cmd
 }
 
-func defaultServerOptions() server.Options {
+func defaultServerOptions(infraDir string) server.Options {
 	return server.Options{
-		TLSCache:                "$HOME/.infra/cache",
-		DBFile:                  "$HOME/.infra/sqlite3.db",
-		DBEncryptionKey:         "$HOME/.infra/sqlite3.db.key",
+		Version:                 0.2, // update this as the config version changes
```
dnephin, line 93: Previously this was being set in the test function, so I think in the last release we weren't setting the version. I guess that's ok because we haven't modified any keys, it's a separate key in the config.

internal/cmd/login_test.go
dnephin: Ok, changing to 2s seems to fix it. I guess CI is quite a bit slower because we use `-race`, and the CPU/disk is probably shared with other processes.

dnephin: I spoke too soon. It's flaked again with the 2 second timeout!  

I'm not sure why it's so much slower in CI, but I'll try increasing this further to see if it can be reliable. It's possible there is something else that is causing the flake.

dnephin: I increased the timeout to 20 seconds in CI, but kept it to 1 second for local dev. Hopefully that works.

BruceMacD (approved):
Looks good, nice to be able to test the interactive stuff

internal/certs/certs.go
dnephin: It would probably be fine to use a longer expiry. We hardcode to 1 year right now in the production code, so I just used the same value.

PR #2120, opened 2022-05-27
fix: ui bugs

### LINKED ISSUE #2074: ui: only allow granting access to users or destinations that exist
Currently the UI operates like the old version of the CLI: it doesn't verify if a user or destination exists before allowing access. Infra should provide an error when trying to grant access (vs creating placeholder resources)

### LINKED ISSUE #2107: ui: wording bugs
**setting - revoke admin modal**
* “This action cannot be undone.” seems unnecessary. It can be undone if I re-add the email as admin 
* delete button - should rename to revoke

**sidebar**
* Meta should probably be Metadata
* infrastructure access section - when there is no user access - The * is unnecessary. Slightly grey and italics is sufficient

**Infrastructure page**
* It’s weird to have “Infrastructure” on the UI when we only support kube clusters. - we should rename it to clusters, + cluster

### LINKED ISSUE #2109: ui: login page OIDC
OIDC should show name as well as URL to match with what's on CLI

hoyyeva:
## Summary
ui bugs that occurred during bug bash on May 26

## Related Issues
Resolves #2074
Resolves #2107
Resolves #2109

### DIFF
--- a/ui/components/admin.js
+++ b/ui/components/admin.js
@@ -36,7 +36,7 @@ function Grant ({ id, userID, grants }) {
         <DeleteModal
           open={open}
           setOpen={setOpen}
-          actionBtnTitle='Revoke'
+          primaryButtonText='Revoke'
           onSubmit={() => {
             mutate('/api/grants?resource=infra&privilege=admin', async ({ items: grants } = { items: [] }) => {
               await fetch(`/api/grants/${id}`, { method: 'DELETE' })
--- a/ui/components/admin.js
+++ b/ui/components/admin.js
@@ -36,6 +36,7 @@ function Grant ({ id, userID, grants }) {
         <DeleteModal
           open={open}
           setOpen={setOpen}
+          actionBtnTitle='Revoke'
           onSubmit={() => {
             mutate('/api/grants?resource=infra&privilege=admin', async ({ items: grants } = { items: [] }) => {
               await fetch(`/api/grants/${id}`, { method: 'DELETE' })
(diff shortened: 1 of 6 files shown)


ui/components/grant.js
```diff
@@ -63,23 +64,15 @@ export default function ({ id }) {
...
-        res = await fetch('/api/users', {
-          method: 'POST',
-          body: JSON.stringify({ name })
-        })
-        const user = await res.json()
-
-        await grantPrivilege(user.id)
-        setName('')
-        setRole('view')
+        setError('User does not exist')
```
BruceMacD, line 75: We should warn that the user does not exist, but allow the grant to be created anyway. The user not being known to Infra will be a common case as long as we are doing identities "just in time" from identity providers.

hoyyeva: but i think for the current workflow, we will not be able to get the one time password if the user create a user on the grant view. I think what we should do is maybe come up with a make sense workflow for that and before that we should prevent user to create a user on the grant

ui/components/admin.js
```diff
@@ -36,6 +36,7 @@ function Grant ({ id, userID, grants }) {
         <DeleteModal
           open={open}
           setOpen={setOpen}
+          actionBtnTitle='Revoke'
```
jmorganca, line 39: nit: a better title could be `primaryButtonText` (i.e. primary vs secondary buttons)

ui/pages/destinations/index.js
```diff
@@ -135,11 +135,11 @@ export default function Destinations () {
                   />
                   {destinations?.length === 0 &&
                     <EmptyTable
-                      title='There is no infrastructure'
-                      subtitle='There is currently no infrastructure connected to Infra'
+                      title='There is no cluster'
```
jmorganca, line 138: `There are no clusters` would be more consistent with other pages

ui/components/grant.js
jmorganca: That's okay: for now we'll restrict this, and handle JIT user capabilities in a future iteration (e.g. a prompt, checkbox or similar - similar to the CLI's `-f` flag). For now users will have to log in to get granted access

jmorganca (approved):

PR #2121, opened 2022-05-27
fix: release build should not contain dev

mxyng:
## Summary

Goreleaser configs should also inject `Prerelease` and `Metadata` vars. Use env vars `BUILDVERSION_PRERELEASE` and and `BUILDVERSION_METADATA` but set reasonable defaults (empty strings)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -4,16 +4,19 @@ env:
   # TODO (https://github.com/infrahq/infra/issues/65)
   - CGO_ENABLED=0
   - GO111MODULE=on
+  - BUILDVERSION_PRERELEASE={{ if index .Env "BUILDVERSION_PRERELEASE" }}{{ .Env.BUILDVERSION_PRERELEASE }}{{ end }}
+  - BUILDVERSION_METADATA={{ if index .Env "BUILDVERSION_METADATA" }}{{ .Env.BUILDVERSION_METADATA }}{{ end }}
 project_name: infra
 builds:
   - id: infra
     ldflags:
       - -s -w
         -X github.com/infrahq/infra/internal.Branch={{ .Branch }}
-        -X github.com/infrahq/infra/internal.Version={{ .Version }}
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -4,13 +4,15 @@ env:
...
+  - BUILDVERSION_PRERELEASE={{ if index .Env "BUILDVERSION_PRERELEASE" }}{{ .Env.BUILDVERSION_PRERELEASE }}{{ end }}
+  - BUILDVERSION_METADATA={{ if index .Env "BUILDVERSION_METADATA" }}{{ .Env.BUILDVERSION_METADATA }}{{ end }}
 project_name: infra
 builds:
   - id: infra
     ldflags:
       - -s -w
         -X github.com/infrahq/infra/internal.Branch={{ .Branch }}
-        -X github.com/infrahq/infra/internal.Version={{ .Version }}
+        -X github.com/infrahq/infra/internal.Version={{ .RawVersion }}
```
mxyng, line 15: Use `RawVersion` instead which doesn't contain the Prerelease part as determined by goreleaser

ssoroka (approved):

.goreleaser.yml
```diff
@@ -4,16 +4,19 @@ env:
   # TODO (https://github.com/infrahq/infra/issues/65)
   - CGO_ENABLED=0
   - GO111MODULE=on
+  - BUILDVERSION_PRERELEASE={{ if index .Env "BUILDVERSION_PRERELEASE" }}{{ .Env.BUILDVERSION_PRERELEASE }}{{ end }}
+  - BUILDVERSION_METADATA={{ if index .Env "BUILDVERSION_METADATA" }}{{ .Env.BUILDVERSION_METADATA }}{{ end }}
```
dnephin, line 8: For my own understanding, why is this necessary? It's not possible to reference the env directly below unless we set it here?

mxyng: This is setting a default. goreleaser will error if the var isn't set. The unspoken part of this if is the else which is an empty string

PR #2124, opened 2022-05-27
improve: strip whitespace of off incoming request string fields

### LINKED ISSUE #1763: api: stricter field validation
Quite a few API fields don't trim or validate against leading & trailing whitespace:

* `name` in all resources
* `url`, `clientID` and `clientSecret` in `/v1/providers`

This results in api objects being created with fields with leading whitespace. We should trim whitespace for these fields in our API (or at least in the client)

ssoroka:
## Summary

a simple solution to closing #1763

## Checklist

## Related Issues

Resolves #1763

### DIFF
--- a/internal/server/handlers_test.go
+++ b/internal/server/handlers_test.go
@@ -710,7 +710,7 @@ func TestAPI_ListGrantsV0_12_2(t *testing.T) {
 	routes := srv.GenerateRoutes(prometheus.NewRegistry())
 
 	resp := httptest.NewRecorder()
-	req, err := http.NewRequest(http.MethodGet, "/v1/grants?privilege=admin", nil)
+	req, err := http.NewRequest(http.MethodGet, "/v1/grants?privilege=%20admin", nil)
 	assert.NilError(t, err)
 	req.Header.Add("Authorization", "Bearer "+adminAccessKey(srv))
 	req.Header.Add("Infra-Version", "0.12.2")
--- a/internal/server/handlers_test.go
+++ b/internal/server/handlers_test.go
@@ -710,7 +710,7 @@ func TestAPI_ListGrantsV0_12_2(t *testing.T) {
 	routes := srv.GenerateRoutes(prometheus.NewRegistry())
 
 	resp := httptest.NewRecorder()
-	req, err := http.NewRequest(http.MethodGet, "/v1/grants?privilege=%20admin", nil)
+	req, err := http.NewRequest(http.MethodGet, "/v1/grants?privilege=admin", nil)
 	assert.NilError(t, err)
 	req.Header.Add("Authorization", "Bearer "+adminAccessKey(srv))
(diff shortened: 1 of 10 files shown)


internal/server/handlers_test.go
```diff
@@ -710,7 +710,7 @@ func TestAPI_ListGrantsV0_12_2(t *testing.T) {
 	routes := srv.GenerateRoutes(prometheus.NewRegistry())
 
 	resp := httptest.NewRecorder()
-	req, err := http.NewRequest(http.MethodGet, "/v1/grants?privilege=admin", nil)
+	req, err := http.NewRequest(http.MethodGet, "/v1/grants?privilege=%20admin", nil)
```
dnephin, line 713: Test coverage for this is great! To be a bit more explicit about what is being tested, maybe we could have a separate test case that includes whitespace (newline, tab, space, vertical space) in both query parameters and request body? 

That test case would also probably be a great place to include some negative tests, to show that a `[]byte` with leading or trailing whitespace is preserved.

internal/server/routes.go
```diff
@@ -186,6 +189,21 @@ func add[Req, Res any](a *API, r *gin.RouterGroup, route route[Req, Res]) {
 	bindRoute(a, r, route.method, route.path, wrappedHandler)
 }
 
+func trimWhitespace(req interface{}) {
+	v := reflect.ValueOf(req)
+	for v.Type().Kind() == reflect.Pointer && !v.IsNil() {
+		v = v.Elem()
+	}
```
dnephin, line 196: Is this equivalent to https://pkg.go.dev/reflect#Indirect ? It looks similar, but `Indirect` doesn't have the nil check.

internal/server/routes.go
```diff
@@ -186,6 +189,21 @@ func add[Req, Res any](a *API, r *gin.RouterGroup, route route[Req, Res]) {
...
+	if v.Kind() == reflect.Struct {
+		for i := 0; i < v.NumField(); i++ {
+			f := v.Field(i)
+			if f.Kind() == reflect.String {
```
dnephin, line 200: minor: I think this will also match types like:

```go
type Custom string
```

If we want to avoid those I think this might work

```go
strType := reflect.TypeOf("")
...
if f.Type() == strType {
```

Not a blocker if there's a reason to prefer the comparison to kind.

dnephin (commented):
Looks good! I think a little more test coverage would be great, but otherwise no concerns.

It's a small performance hit for a little more reflection, but our requests are small enough right now that it shouldn't be an issue for the foreseeable future.

The only place that comes to mind as needing trailing whitespace is `api.Desination.Connection.CA`. I believe certs are supposed to have a single trailing newline in them. We might want to change that field to a `[]byte`.

dnephin (approved):
LGTM

PR #2125, opened 2022-05-27
cli: add a test for `infra list`

dnephin:
Branched from #2098 which introduces support for testing against an in-process api server.

This PR is a follow up to #2097. It fixes one test, and adds another. Both changes would have caught that bug.

This PR adds a test for the `infra list` command. The test uses an in-process API server, and tests the list output is correct for a user with no destinations, and a user with two destinations.


dnephin:
This PR is now ready for review. The PR it was based on has merge.

jmorganca (approved):

PR #2143, opened 2022-05-30
handle data duplicate errors

### LINKED ISSUE #1686: api: errors are not human readable
Examples:
* `409`: `{"code":409,"message":"duplicate record: UNIQUE constraint failed: providers.name"}` – should be `a provider with this name already exists`

mxyng:
## Summary

Inspect data errors to detect duplication and provide more details to user when unique constraints are violated.

Update `CreateUser` to _only_ create users and credentials. Previously it will accept the request if the user already exists and update their credentials. This use case should be fulfilled by the UpdateUser endpoint.

## TODO

- [x] Unit tests for sqlite3 and postgres unique constraint violations

## Checklist

## Related Issues

Resolves #1686

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -6,8 +6,8 @@ import (
 	"net/url"
 	"os"
 	"path"
-	"regexp"
 	"strings"
+	"unicode"
 
 	"github.com/go-playground/validator/v10"
 	"github.com/jackc/pgconn"
@@ -173,11 +173,14 @@ func handleError(err error) error {
 			//                            key_desc) : 0,
 			//                 errtableconstraint(heapRel,
 			//                                    RelationGetRelationName(rel))));
-			re := regexp.MustCompile(`Key \((?P<columnName>[[:print:]]+)\)=\([[:print:]]+\) already exists.`)
-			matches := re.FindStringSubmatch(pgErr.Detail)
-			if matches != nil {
-				columnNameIndex := re.SubexpIndex("columnName")
-				return UniqueConstraintError{Table: pgErr.TableName, Column: matches[columnNameIndex]}
+
+			fields := strings.FieldsFunc(pgErr.Detail, func(r rune) bool {
+				return unicode.IsSpace(r) || r == '(' || r == ')'
+			})
+
+			// fields = [Key, <column>, =, <value>, already, exists.]
+			if len(fields) == 6 {
+				return UniqueConstraintError{Table: pgErr.TableName, Column: fields[1]}
 			}
 		}
 	}
@@ -186,12 +189,15 @@ func handleError(err error) error {
 	// pRtree->base.zErrMsg = sqlite3_mprintf(
 	//     "UNIQUE constraint failed: %s.%s", pRtree->zName, zCol
 	// );
-	re := regexp.MustCompile(`UNIQUE constraint failed: (?P<tableName>[[:print:]]+)\.(?P<columnName>[[:print:]]+)`)
-	matches := re.FindStringSubmatch(err.Error())
-	if matches != nil {
-		tableNameIndex := re.SubexpIndex("tableName")
-		columnNameIndex := re.SubexpIndex("columnName")
-		return UniqueConstraintError{Table: matches[tableNameIndex], Column: matches[columnNameIndex]}
+	if strings.HasPrefix(err.Error(), "UNIQUE constraint failed:") {
+		fields := strings.FieldsFunc(err.Error(), func(r rune) bool {
+			return unicode.IsSpace(r) || r == '.'
+		})
+
+		// fields = [UNIQUE, constraint, failed:, <table>, column>]
+		if len(fields) == 5 {
+			return UniqueConstraintError{Table: fields[3], Column: fields[4]}
+		}
 	}
 
 	return err
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -149,7 +149,7 @@ type UniqueConstraintError struct {
 }
 
 func (e UniqueConstraintError) Error() string {
-	return fmt.Sprintf("value for %v already exist in %v", e.Column, e.Table)
+	return fmt.Sprintf("value for %v already exists in %v", e.Column, e.Table)
 }
 
 // handleError looks for well known DB errors. If the error is recognized it
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -149,7 +149,12 @@ type UniqueConstraintError struct {
 }
 
 func (e UniqueConstraintError) Error() string {
-	return fmt.Sprintf("value for %v already exists in %v", e.Column, e.Table)
+	if e.Table == "" {
+		return "value already exists"
+	} else if e.Column == "" {
+		return fmt.Sprintf("value already exists for %v", e.Table)
+	}
+	return fmt.Sprintf("value for %v already exists for %v", e.Column, e.Table)
 }
 
 // handleError looks for well known DB errors. If the error is recognized it
@@ -164,24 +169,18 @@ func handleError(err error) error {
 	if errors.As(err, &pgErr) {
 		switch pgErr.Code {
 		case pgerrcode.UniqueViolation:
-			// https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/access/nbtree/nbtinsert.c;h=f6f4af8bfe3f0876e860944e2233328f8cc3303a;hb=HEAD#l668
-			// ereport(ERROR,
-			//         (errcode(ERRCODE_UNIQUE_VIOLATION),
-			//          errmsg("duplicate key value violates unique constraint \"%s\"",
-			//                 RelationGetRelationName(rel)),
-			//                 key_desc ? errdetail("Key %s already exists.",
-			//                            key_desc) : 0,
-			//                 errtableconstraint(heapRel,
-			//                                    RelationGetRelationName(rel))));
-
-			fields := strings.FieldsFunc(pgErr.Detail, func(r rune) bool {
-				return unicode.IsSpace(r) || r == '(' || r == ')'
-			})
-
-			// fields = [Key, <column>, =, <value>, already, exists.]
-			if len(fields) == 6 {
-				return UniqueConstraintError{Table: pgErr.TableName, Column: fields[1]}
+			constraintFields := map[string]string{
+				"idx_identities_name":         "name",
+				"idx_groups_name":             "name",
+				"idx_providers_name":          "name",
+				"idx_access_keys_name":        "name",
+				"idx_destinations_unique_id":  "unique_id",
+				"idx_access_keys_key_id":      "key_id",
+				"idx_credentials_identity_id": "identity_id",
 			}
+
+			columnName := constraintFields[pgErr.ConstraintName]
+			return UniqueConstraintError{Table: pgErr.TableName, Column: columnName}
 		}
 	}
 
@@ -198,6 +197,8 @@ func handleError(err error) error {
 		if len(fields) == 5 {
 			return UniqueConstraintError{Table: fields[3], Column: fields[4]}
 		}
+
+		return UniqueConstraintError{}
 	}
 
 	return err
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -6,12 +6,12 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/go-playground/validator/v10"
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgerrcode"
-	"github.com/mattn/go-sqlite3"
 	"gorm.io/driver/postgres"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
@@ -144,15 +144,12 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
 }
 
 type UniqueConstraintError struct {
-	Field string
+	Table  string
+	Column string
 }
 
-func (d UniqueConstraintError) Error() string {
-	return fmt.Sprintf("value for field already exists: %v", d.Field)
-}
-
-func (d UniqueConstraintError) Is(target error) bool {
-	return d == target || target == internal.ErrDuplicate
+func (e UniqueConstraintError) Error() string {
+	return fmt.Sprintf("value for %v already exist in %v", e.Column, e.Table)
 }
 
 // handleError looks for well known DB errors. If the error is recognized it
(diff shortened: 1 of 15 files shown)


internal/server/handlers.go
```diff
@@ -60,48 +60,27 @@ func (a *API) GetUser(c *gin.Context, r *api.GetUserRequest) (*api.User, error)
 func (a *API) CreateUser(c *gin.Context, r *api.CreateUserRequest) (*api.CreateUserResponse, error) {
 	user := &models.Identity{Name: r.Name}
 
-	setOTP := r.SetOneTimePassword
-
-	// infra identity creation should be attempted even if an identity is already known
```
mxyng, line 65: Removing this because the behaviour is not in line with other POST endpoints. Instead, the user should use the UpdateUser endpoint to create or update the password. A side effect of this change is that admins will no longer be able to (re)set an one-time password one the user exists

internal/server/data/data.go
```diff
@@ -152,28 +139,62 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
+		}
 	}
 
-	if strings.HasPrefix(err.Error(), "UNIQUE constraint failed:") {
-		return true
+	// https://sqlite.org/src/file?name=ext/rtree/rtree.c:
+	// pRtree->base.zErrMsg = sqlite3_mprintf(
+	//     "UNIQUE constraint failed: %s.%s", pRtree->zName, zCol
+	// );
+	re := regexp.MustCompile(`UNIQUE constraint failed: (?P<tableName>\w+).(?P<columnName>\w+)`)
```
mxyng, line 189: Not checking for `sqlite3.Error` specifically here because doing so requires cross compiling

internal/server/data/data.go
```diff
@@ -152,28 +139,62 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
-		if isUniqueConstraintViolation(err) {
-			return fmt.Errorf("%w: %s", internal.ErrDuplicate, err)
-		}
+	err := db.Create(model).Error
+	return handleError(err)
+}
 
-		return err
-	}
+type UniqueConstraintError struct {
```
mxyng, line 146: This should ideally store table, column, and value but some implementations (sqlite3) does not report the value. Even in cases where it does report the value, the value may be an ID which wouldn't be very useful to the end user

internal/server/data/data.go
```diff
@@ -152,28 +139,62 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
+			// ereport(ERROR,
+			//         (errcode(ERRCODE_UNIQUE_VIOLATION),
+			//          errmsg("duplicate key value violates unique constraint \"%s\"",
+			//                 RelationGetRelationName(rel)),
+			//                 key_desc ? errdetail("Key %s already exists.",
+			//                            key_desc) : 0,
+			//                 errtableconstraint(heapRel,
+			//                                    RelationGetRelationName(rel))));
+			re := regexp.MustCompile(`Key \((?P<columnName>\w+)\)=\(\w+\) already exists.`)
+			matches := re.FindStringSubmatch(pgErr.Detail)
```
dnephin, line 177: Aren't the fields already available from `pgErr` ? I thought we could use `pgErr.TableName`, and `pgErr.ColumnName`.

internal/server/data/data.go
```diff
@@ -152,28 +139,62 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
+			if matches != nil {
+				columnNameIndex := re.SubexpIndex("columnName")
+				return UniqueConstraintError{Table: pgErr.TableName, Column: matches[columnNameIndex]}
```
mxyng, line 180: While there is a `ColumnName` field in the error struct, it's no populated for this particular error

internal/server/handlers.go
dnephin: Would it be possible to do this in a separate PR? 

Is the change to `AssignIdentityToGroups` related to this?

dnephin (commented):
Nice!

internal/server/data/data.go
dnephin: Oh wow, that's surprising. Is it the `ConstraintName` that is set? Which maybe is not that helpful for us.

internal/server/handlers.go
mxyng: I'll split into a separate PR. It's something I noticed while testing the changes. When creating duplicate users, the error return complains about duplicate credentials rather than duplicate users which was surprising

internal/server/data/data.go
mxyng: I was surprised as well. This is the full error from psql
```
postgres      | 2022-05-30 18:15:11.405 UTC [35] ERROR:  duplicate key value violates unique constraint "idx_credentials_identity_id"
postgres      | 2022-05-30 18:15:11.405 UTC [35] DETAIL:  Key (identity_id)=(54270808554807296) already exists.
postgres      | 2022-05-30 18:15:11.405 UTC [35] STATEMENT:  INSERT INTO "credentials" ("created_at","updated_at","deleted_at","identity_id","password_hash","one_time_password","one_time_password_used","id") VALUES ($1,$2,$3,$4,$5,$6,$7,$8) RETURNING "id"
```
and seen from infra
```
infra-server  | {"level":"info","ts":1653934511.4067185,"caller":"data/data.go:165","msg":"&pgconn.PgError{Severity:\"ERROR\", Code:\"23505\", Message:\"duplicate key value violates unique constraint \\\"idx_credentials_identity_id\\\"\", Detail:\"Key (identity_id)=(54270808554807296) already exists.\", Hint:\"\", Position:0, InternalPosition:0, InternalQuery:\"\", Where:\"\", SchemaName:\"public\", TableName:\"credentials\", ColumnName:\"\", DataTypeName:\"\", ConstraintName:\"idx_credentials_identity_id\", File:\"nbtinsert.c\", Line:663, Routine:\"_bt_check_unique\"}"}
```

`ConstraintName` is populated but we'd still need to parse the constraint string

internal/server/data/identity.go
```diff
@@ -70,11 +70,16 @@ func AssignIdentityToGroups(db *gorm.DB, user *models.Identity, provider *models
 			groupID = group.ID
 		}
 
-		// add user to group
-		err = db.Exec("insert into identities_groups (identity_id, group_id) values (?, ?)", user.ID, groupID).Error
-		if err != nil {
-			if !isUniqueConstraintViolation(err) {
-				return fmt.Errorf("insert: %w", err)
+		var ids []uid.ID
+		if err := db.Raw("SELECT identity_id FROM identities_groups WHERE identity_id = ? AND group_id = ?", user.ID, groupID).Scan(&ids).Error; err != nil {
```
mxyng, line 74: This is necessary because the previous behaviour executed the insert regardless if the record already exists and discarded the error on unique constraint violations. This checks before insert and only inserts the record if it doesn't already exist

internal/server/data/data.go
```diff
@@ -152,28 +139,62 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
+			// https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/access/nbtree/nbtinsert.c;h=f6f4af8bfe3f0876e860944e2233328f8cc3303a;hb=HEAD#l668
+			// ereport(ERROR,
+			//         (errcode(ERRCODE_UNIQUE_VIOLATION),
+			//          errmsg("duplicate key value violates unique constraint \"%s\"",
+			//                 RelationGetRelationName(rel)),
+			//                 key_desc ? errdetail("Key %s already exists.",
+			//                            key_desc) : 0,
+			//                 errtableconstraint(heapRel,
+			//                                    RelationGetRelationName(rel))));
+			re := regexp.MustCompile(`Key \((?P<columnName>[[:print:]]+)\)=\([[:print:]]+\) already exists.`)
```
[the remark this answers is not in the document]
mxyng: I don't believe there's any guarantee but looking through postgresql code, the error string hasn't changed in a long time. Last change was in 2017 and it was indentation. Before that was 2015, also indentation. I imagine sqlite to be the similar

internal/server/data/provider_test.go
```diff
@@ -42,7 +41,7 @@ func TestCreateProviderDuplicate(t *testing.T) {
 		createProviders(t, db, providerDevelop, providerProduction)
 
 		err := CreateProvider(db, &providerDevelop)
-		assert.ErrorIs(t, err, internal.ErrDuplicate)
+		assert.ErrorContains(t, err, "value for name already exists in providers")
```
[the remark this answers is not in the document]
mxyng: `err` is no longer a value so `assert.ErrorIs` is not appropriate. Looking through the doc, I did find `assert.ErrorType` but I don't know if its any better than `assert.ErrorContains`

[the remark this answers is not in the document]
dnephin: Definitely agree that our tests should be checking the type of the error, since that's an important part of the interface we are exposing from the `data` package.

To work with typed errors like this we'll need to use `errors.As`. I'd do something like this:

```go
var uerr UniqueConstraintError
assert.Assert(t, err, errors.As(err, &uerr), "error is wrong type %T", err)
expected := UniqueConstraintError{Column: "name", Table: "providers"}
assert.DeepEqual(t, uerr, expected)
```

internal/server/errors_test.go
```diff
@@ -46,17 +46,6 @@ func TestSendAPIError(t *testing.T) {
 			err:    fmt.Errorf("hide this: %w", internal.ErrForbidden),
 			result: api.Error{Code: http.StatusForbidden, Message: "forbidden"},
 		},
-		{
-			err:    internal.ErrDuplicate,
-			result: api.Error{Code: http.StatusConflict, Message: "duplicate record"},
-		},
```
dnephin, line 52: We should replace these test cases  with ones that use `data.UniqueConstraintError`, to show it uses the right status code.

internal/server/errors.go
```diff
@@ -35,36 +37,45 @@ func sendAPIError(c *gin.Context, err error) {
...
 		// hide the error text, it may contain sensitive information
 		resp.Message = "unauthorized"
+
 	case errors.Is(err, internal.ErrForbidden):
 		resp.Code = http.StatusForbidden
 		// hide the error text, it may contain sensitive information
 		resp.Message = "forbidden"
-	case errors.Is(err, internal.ErrDuplicate):
+
+	case errors.As(err, &uniqueConstraintError):
```
dnephin, line 46: Should we also populate the `FieldErrors`, since the constraint is likely related to a field?  We'll have to make sure we can easily translate the column name to a field name.  If that's a challenge it should be fine to leave that part for a future PR.

dnephin (commented):
Left a few more suggestions for testing, and hopefully an approach that will allow us to remove the need for regex to parse strings.

internal/server/data/data.go
```diff
@@ -152,28 +139,62 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
-	if strings.HasPrefix(err.Error(), "UNIQUE constraint failed:") {
-		return true
+	// https://sqlite.org/src/file?name=ext/rtree/rtree.c:
+	// pRtree->base.zErrMsg = sqlite3_mprintf(
+	//     "UNIQUE constraint failed: %s.%s", pRtree->zName, zCol
+	// );
+	re := regexp.MustCompile(`UNIQUE constraint failed: (?P<tableName>[[:print:]]+)\.(?P<columnName>[[:print:]]+)`)
+	matches := re.FindStringSubmatch(err.Error())
+	if matches != nil {
+		tableNameIndex := re.SubexpIndex("tableName")
```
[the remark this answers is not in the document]
mxyng: I feel the same way about regex so I'm happy to look for an alternative. In fact, the first iterations used split/cut which works fine for sqlite but gets messy for postgres. Ideally, both variants use the same type of code for consistency and simplicity. 

`handleError` can take the table as an input since the caller has the model but this isn't really the problem. The table name is readily available for postgres and parsable for sqlite, which we need to do anyways. Additionally, I don't believe we're using any implicit associations so the model passed in should mirror the table name set by the error.

internal/server/errors.go
mxyng: Field errors should probably be omitted for now. If we need to expose the field in the future, we can add it then

internal/server/data/data.go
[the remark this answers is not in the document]
mxyng: I'm trying `strings.FieldsFunc` which is nicer is some areas but more brittle in others. Let me know what you think

internal/server/data/data.go
```diff
@@ -152,28 +139,68 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
-		return err
-	}
+type UniqueConstraintError struct {
+	Table  string
+	Column string
+}
 
-	return nil
+func (e UniqueConstraintError) Error() string {
+	return fmt.Sprintf("value for %v already exists in %v", e.Column, e.Table)
```
dnephin, line 152: We should probably have a check here for `if e.Column == ""` in the case where we failed to parse the message, so that we can provide a fallback message that reads better than "value for already exists in"

dnephin (commented):
Looks good! I like the `FieldsFunc` approach.  

~I think all we're missing is test coverage in the `data` package, to show it working for both databases.~

Edit: I missed it! there is test coverage. I think just a fallback message for when parsing fails, and this should be all set

Also wondering what you think of https://github.com/infrahq/infra/pull/2143#discussion_r887134246 as an option.

internal/server/errors.go
dnephin: Sounds good. I guess we'll see what the CLI needs from the API, and we can also add it at that time if we need it

internal/server/data/data.go
mxyng: That's a good idea. I'll update it to use the constraint name instead. This would only be applicable to postgres. AFAIK, sqlite still requires parsing

internal/server/data/data.go
```diff
@@ -152,28 +139,69 @@ func add[T models.Modelable](db *gorm.DB, model *T) error {
...
+}
 
-	return nil
+func (e UniqueConstraintError) Error() string {
+	if e.Table == "" {
+		return "value already exists"
+	} else if e.Column == "" {
+		return fmt.Sprintf("value already exists in %v", e.Table)
+	}
+	return fmt.Sprintf("value for %v already exists in %v", e.Column, e.Table)
```
dnephin, line 157: Very minor, not a blocker, but thinking about how the user will see this.

I wonder if "already exists for x" (ex: already exists for grant) would read better than "in".

Maybe not a concern if we're going to change the error later at the API or CLI layers.

dnephin (approved):
LGTM once CI is green

PR #2148, opened 2022-05-30
fix: infra default role is no role

### LINKED ISSUE #2141: ui: infra role display bugs
Today the UI has two small bugs when it comes to displaying roles:

1. If the user has no `infra` role, infra shows `View`, it should show nothing or `-`

Screen Shot 2022-05-30 at 11 48 24 AM

2. Infra roles should be lowercase (e.g. `view`, `admin`) since they are values that need to work on the cli as well (e.g. `view` or `admin` but not "View")

hoyyeva:
## Summary
infra default role is no role so it should be displayed as `-` in the UI

## Related Issues

Resolves #2141


jmorganca (approved):
Good stuff!

PR #2156, opened 2022-05-31
improve: migrates existing user facing errors to use Error, and formats them

### LINKED ISSUE #1394: cli: Refactor error messages for better readability

kimskimchi:
## Summary

Finishes up using the new Error struct in cli. Removes periods for all error messages (and uses `;` instead), including user facing. 

## Checklist

- [x] Wrote and fixed appropriate unit tests

## Related Issues

Resolves #1394

### DIFF
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -60,7 +60,7 @@ $ infra keys add connector
 
 			if options.Name != "" {
 				if strings.Contains(options.Name, " ") {
-					return Error{Message: "key name cannot contain spaces"}
+					return Error{Message: "Key name cannot contain spaces"}
 				}
 			}
 


internal/cmd/keys.go
```diff
@@ -56,7 +57,7 @@ $ infra keys add example-key identity@example.com --ttl=12h
 
 			if options.Name != "" {
 				if strings.Contains(options.Name, " ") {
-					return fmt.Errorf("key name cannot contain spaces")
+					return Error{Message: "key name cannot contain spaces"}
```
BruceMacD, line 60: Should this be capitalized to match the formatting in other messages?

internal/cmd/login.go
```diff
@@ -229,7 +236,7 @@ func updateInfraConfig(client *api.Client, loginReq *api.LoginRequest, loginRes
 
 	t, ok := client.HTTP.Transport.(*http.Transport)
 	if !ok {
-		return fmt.Errorf("Could not update config due to an internal error")
+		return fmt.Errorf("could not update config")
```
BruceMacD, line 239: nit: which config?

dnephin (approved):
Looks good! I like Bruce's suggestions as well

PR #2157, opened 2022-05-31
feat: add more default cluster roles

### LINKED ISSUE #90: Include default roles for Kubernetes
Currently, Infra requires users to have Kubernetes `roles` set up ahead of time. Infra simply takes care of mapping.

However, many users don't have Kubernetes roles set up ahead of time, and they have asked us to include "good default" roles as part of Infra:

* `readonly` Read (almost) all resources
* `edit` Edit (almost) all resources
* `port-forward` Allow port-forwarding
* `exec` Allow executing shell commands in pods
* `admin` Full access

Others:
* `rolling-restart` Allow deploying and restarting services
* `secrets` Allow secret access

mxyng:
## Summary

Filter ClusterRoles based on labels
- kubernetes.io/bootstrapping=rbac-defaults
- app.infrahq.com/include-role=true

## Checklist

## Related Issues

Resolves #90

### DIFF
--- a/helm/charts/infra/templates/connector/_helpers.tpl
+++ b/helm/charts/infra/templates/connector/_helpers.tpl
@@ -177,3 +177,11 @@ Infer whether Infra connector should be deployed based on connector.enabled, con
 {{- define "connector.enabled" -}}
 {{- or .Values.connector.enabled (not (empty .Values.connector.config.server)) (not (empty .Values.connector.config.accessKey)) }}
 {{- end }}
+
+{{/*
+ClusterRole labels
+*/}}
+{{- define "connector.clusterRoleLabels" }}
+{{- include "connector.labels" . }}
+app.infrahq.com/include-role: "true"
+{{- end }}
--- a/helm/charts/infra/templates/connector/clusterrole.yaml
+++ b/helm/charts/infra/templates/connector/clusterrole.yaml
@@ -36,4 +36,121 @@ rules:
       - watch
       - bind
       - escalate
+{{- if not (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "logs") }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: logs
+  labels:
+{{- include "connector.clusterRoleLabels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources:
+      - pods
+    verbs:
+      - get
+      - list
+  - apiGroups:
+      - batch
+    resources:
+      - jobs
+    verbs:
+      - get
+  - apiGroups:
+      - apps
+    resources:
+      - deployments
+      - replicasets
+      - statefulsets
+      - daemonsets
+    verbs:
+      - get
+  - apiGroups: [""]
+    resources:
+      - pods/log
+    verbs:
+      - get
+      - watch
+{{- end }}
+{{- if not (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "exec") }}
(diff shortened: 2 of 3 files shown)


helm/charts/infra/templates/connector/clusterrole.yaml
```diff
@@ -36,4 +36,61 @@ rules:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-logs
+  labels:
+{{- include "connector.clusterRoleLabels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources:
+      - pods
```
mxyng, line 49: This, and both cluster roles below, are pretty restrictive. `kubectl logs` also takes deployment, replica set, stateful set, daemon set, as arguments which this cluster role does not permit. To allow those use cases, we'll need to add `deployments`, `replicasets`, etc. to the cluster role resources

internal/kubernetes/kubernetes.go
```diff
@@ -596,13 +596,22 @@ func (k *Kubernetes) ClusterRoles() ([]string, error) {
 		return nil, err
 	}
 
-	clusterRoles, err := clientset.RbacV1().ClusterRoles().List(context.Background(), metav1.ListOptions{LabelSelector: "kubernetes.io/bootstrapping=rbac-defaults"})
+	rbacDefaults, err := clientset.RbacV1().ClusterRoles().List(context.Background(), metav1.ListOptions{
```
jmorganca, line 599: Is there a way to combine these into one query?

helm/charts/infra/templates/connector/clusterrole.yaml
jmorganca: This is super cool. To enable users have `kubectl logs [deployment|rs|pod]` under one `logs` role, let's just do that here. Is there a downside to that?

helm/charts/infra/templates/connector/clusterrole.yaml
```diff
@@ -36,4 +36,61 @@ rules:
       - watch
       - bind
       - escalate
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-logs
```
jmorganca, line 43: Re below comment: we can call this `logs`

internal/kubernetes/kubernetes.go
mxyng: Not that I'm aware. The value is taken as a union and I didn't see an OR operator

helm/charts/infra/templates/connector/clusterrole.yaml
mxyng: Not really. It gives more permissions but covers a broader set of use cases

helm/charts/infra/templates/connector/_helpers.tpl
```diff
@@ -177,3 +177,11 @@ Infer whether Infra connector should be deployed based on connector.enabled, con
 {{- define "connector.enabled" -}}
 {{- or .Values.connector.enabled (not (empty .Values.connector.config.server)) (not (empty .Values.connector.config.accessKey)) }}
 {{- end }}
+
+{{/*
+ClusteRole labels
```
BruceMacD, line 182:
```suggestion
ClusterRole labels
```

BruceMacD: Should we include some more explanation here, like:
> Default ClusterRole labels created by Infra

BruceMacD (approved):

PR #2158, opened 2022-05-31
api: add a library for declarative request validation

dnephin:
This is the first step in replacing our API request validation.

This PR adds a library with validation rules that we already use today. It does not integrate it into our API handler or openapi doc generation yet.

Follow up items:
* replace validate struct tags with this validation
* at the same time, integrate with openapi doc generation

Follow up PRs: #2501, #2527

Related to #1763

### DIFF
--- a/internal/validate/email.go
+++ b/internal/validate/email.go
@@ -3,7 +3,6 @@ package validate
 import (
 	"fmt"
 	"net/mail"
-	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
 )
@@ -24,7 +23,7 @@ func (e email) Validate() *Failure {
 	}
 	addr, err := mail.ParseAddress(e.value)
 	if err != nil {
-		return fail(e.name, "invalid email address: "+strings.TrimPrefix(err.Error(), "mail: "))
+		return fail(e.name, "invalid email address")
 	}
 	if addr.Name != "" {
 		return fail(e.name, fmt.Sprintf("email address must not contain display name %q", addr.Name))
--- a/internal/validate/email.go
+++ b/internal/validate/email.go
@@ -1,6 +1,7 @@
 package validate
 
 import (
+	"fmt"
 	"net/mail"
 	"strings"
 
@@ -26,7 +27,7 @@ func (e email) Validate() *Failure {
 		return fail(e.name, "invalid email address: "+strings.TrimPrefix(err.Error(), "mail: "))
 	}
 	if addr.Name != "" {
-		return fail(e.name, "email address must not contain a name")
+		return fail(e.name, fmt.Sprintf("email address must not contain display name %q", addr.Name))
 	}
 	return nil
 }
--- a/internal/validate/email.go
+++ b/internal/validate/email.go
@@ -1,0 +1,37 @@
+package validate
+
+import (
+	"net/mail"
+	"strings"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+// Email validates a field that should contain an email address.
+func Email(name string, value string) ValidationRule {
+	return email{name: name, value: value}
+}
+
+type email struct {
+	name  string
+	value string
+}
+
+func (e email) Validate() *Failure {
+	if e.value == "" {
+		return nil
+	}
+	addr, err := mail.ParseAddress(e.value)
+	if err != nil {
+		return fail(e.name, "invalid email address: "+strings.TrimPrefix(err.Error(), "mail: "))
+	}
+	if addr.Name != "" {
+		return fail(e.name, "email address must not contain a name")
+	}
+	return nil
+}
+
+func (e email) DescribeSchema(parent *openapi3.Schema) {
+	schema := schemaForProperty(parent, e.name)
+	schema.Format = "email"
+}
--- a/internal/validate/string_test.go
+++ b/internal/validate/string_test.go
@@ -9,26 +9,41 @@ import (
 	"gotest.tools/v3/assert"
 )
 
+type StringExample struct {
+	Field string
+}
+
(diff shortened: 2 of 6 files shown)


internal/server/openapi.go
```diff
@@ -178,31 +185,59 @@ func createComponent(schemas openapi3.Schemas, rst reflect.Type) *openapi3.Schem
...
-		s.Items = buildProperty(f, t.Elem(), parent, parentSchema)
+		s.Items = buildProperty(f, t.Elem(), parent)
 	}
 
 	if s.Type == "object" {
 		s.Properties = openapi3.Schemas{}
 
+		// TODO: cleanup
+		reqV := reflect.New(t)
+		req, ok := reqV.Interface().(ivalidate.Request)
```
dnephin, line 219: All the API request structs should implement this `validate.Request` interface, so we'll be able to remove the type check and use the passed in struct.

internal/validate/struct.go
```diff
@@ -0,0 +1,146 @@
...
+	for _, rule := range req.ValidationRules() {
+		fv := rule.fieldValue()
+		if fv.Kind() != reflect.Ptr {
+			return fmt.Errorf("field value must be a pointer, not %v", fv.Kind())
+		}
+		ft := findStructField(value, fv)
+		if ft == nil {
+			return fmt.Errorf("could not locate field for %v", fv.Type())
+		}
+		jsonName := jsonNameFromStructField(ft)
```
dnephin, line 32: I think this is the bit I'm least thrilled about. This approach was inspired by `ozzo-validate`, but it only works when the `ValidationRules` method has a pointer receiver.

I'm thinking it would be easier to have the validation rules accept the name of the json field instead of trying to look up the name by matching pointer addresses.

If we take that approach we can delete `findStructField`, `jsonNameFromStructField`, `StructRule`, and simplify this function and `RulesToMap`. The other rules would take an extra argument of `jsonName string`, and the field value would not need to be a pointer.

api/access_key.go
```diff
@@ -21,10 +22,18 @@ type ListAccessKeysRequest struct {
 }
 
 type CreateAccessKeyRequest struct {
-	UserID            uid.ID   `json:"userID" validate:"required"`
+	UserID            uid.ID   `json:"userID"`
 	Name              string   `json:"name" validate:"excludes= "`
```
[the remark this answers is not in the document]
dnephin: Ya, I think we should do something with the `StringRule` to restrict this to only alpha-numeric charcters plus a few limited symbols.

internal/validate/validate_test.go
```diff
@@ -0,0 +1,151 @@
...
+		e = OneOfExample{Second: true}
+		assert.NilError(t, Validate(e))
+
+		e = OneOfExample{Third: 123}
+		assert.NilError(t, Validate(e))
+	})
+	t.Run("with failure", func(t *testing.T) {
+		e := OneOfExample{}
+		err := Validate(e)
+		assert.Error(t, err, "validation failed: one of (first, second, third) is required")
```
dnephin, line 149: Thoughts on this error message? Any suggestions for how we can improve it?

internal/validate/validate_test.go
```diff
@@ -0,0 +1,151 @@
...
+		assert.Error(t, err, "validation failed: only one of (first, second) can be set")
+	})
+	t.Run("with failure three set", func(t *testing.T) {
+		e := MutualExample{
+			First:  "value",
+			Second: true,
+			Third:  123,
+		}
+		err := Validate(e)
+		assert.Error(t, err, "validation failed: only one of (first, second, third) can be set")
```
dnephin, line 116: Thoughts on this error message? Any suggestions for how we can improve it?

internal/validate/string_test.go
```diff
@@ -0,0 +1,69 @@
...
+func TestStringRule_Validate(t *testing.T) {
+	t.Run("min length", func(t *testing.T) {
+		r := &ExampleRequest{RequiredString: "a"}
+		err := Validate(r)
+		assert.ErrorContains(t, err, "length (1) must be at least 2")
+	})
+	t.Run("max length", func(t *testing.T) {
+		r := &ExampleRequest{RequiredString: "abcdefghijklm"}
+		err := Validate(r)
+		assert.ErrorContains(t, err, "length (13) must be no more than 10")
```
dnephin, line 21: Thoughts on this error message? Any suggestions for how we can improve it?

BruceMacD: `length of string (x) exceeds maximum length of y`

maybe

internal/validate/email.go
```diff
@@ -0,0 +1,37 @@
...
+func (e email) Validate() *Failure {
+	if e.value == "" {
+		return nil
+	}
+	addr, err := mail.ParseAddress(e.value)
+	if err != nil {
+		return fail(e.name, "invalid email address: "+strings.TrimPrefix(err.Error(), "mail: "))
+	}
+	if addr.Name != "" {
+		return fail(e.name, "email address must not contain a name")
```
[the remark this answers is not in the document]
dnephin: Ya, not sure what to call this. The grammar in https://www.rfc-editor.org/rfc/rfc5322.html#section-3.4 calls it `display-name`. It's this part:

> "Display Name" <someone@example.com>

I could call it display name, but I'm not sure if that's any more clear. 

I guess including the part that is invalid would help.

```go
fmt.Sprintf("email address must not contain display name %q", addr.Name)
```

I'll try that out

internal/validate/string_test.go
dnephin: Adding "of string" sounds great! 

"exceeds maximum length" vs "must be no more than" is interesting. I guess we want a consistent feel for all of these messages. I'll try to compile them all into one place and play around with some of these options.

PR #2167, opened 2022-06-01
maintain: extended infra sessions

### LINKED ISSUE #1857: cli: longer infra sessions
Right now sessions within Infra last for 23 hours. This means users need to run `infra login` once a day outside of their normal kubernetes workflow. Infra sessions should be less in the way of user experience.

We should explore ways to keep the Infra session valid behind the scenes through access key lifetime extensions.

## Describe alternatives you've considered
PKI authentication.

BruceMacD:
- move login logic to its own package
- give infra sessions a long lifetime with an extension deadline
- improve test coverage of authentication mechanisms

## Summary
This change makes a simple behavior change that requires users to login to Infra less often, and moves authentication logic into its own package.

The first change is to set an extension deadline (a time-frame within which an access key must be used within to remain valid up to its session expiry) to access keys issued on login. By default I give access keys issued on login a maximum lifetime of 30 days, but they must be used once every 3 days to remain valid. These session values are configurable.

The other change is to move all authentication logic out of the login handler and access packages since they don't require access checks. Authentication mechanisms now all conform to a `LoginMethod` interface.

## Checklist

## Related Issues

Resolves #1857


internal/access/credential.go
```diff
@@ -75,31 +73,3 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
 
 	return nil
 }
-
-func LoginWithPasswordCredential(c *gin.Context, name, password string, expiry time.Time) (string, *models.Identity, bool, error) {
```
BruceMacD, line 79: Moved to the authn package

internal/access/identity.go
```diff
@@ -108,24 +109,88 @@ func ListIdentities(c *gin.Context, name string, ids []uid.ID) ([]models.Identit
...
 
-	logging.S.Debugf("%s user authenticated with %q groups", provider.Name, groups)
+	return provider, providerUser.RedirectURL, nil
+}
 
-	if err := data.AssignIdentityToGroups(db, user, provider, groups); err != nil {
-		return fmt.Errorf("assign identity to groups: %w", err)
+// UpdateIdentityInfoFromProvider calls the identity provider used to authenticate this user session to update their current information
+func UpdateIdentityInfoFromProvider(c *gin.Context, oidc authn.OIDC) error {
+	// added by the authentication middleware
```
BruceMacD, line 139: Moved this logic from the `handlers.go` file

internal/access/provider.go
```diff
@@ -70,99 +64,8 @@ func DeleteProvider(c *gin.Context, id uid.ID) error {
 	return data.DeleteProviders(db, data.ByID(id))
 }
 
-// RetrieveUserProviderTokens gets the providerUser for the current session token was created for
-func RetrieveUserProviderTokens(c *gin.Context) (*models.ProviderUser, error) {
```
BruceMacD, line 74: Moved to the authn package

internal/access/signup_test.go
```diff
@@ -96,12 +97,17 @@ func TestSignupEnabled(t *testing.T) {
...
 
 	t.Run("SignupUser", func(t *testing.T) {
-		c, _ := setup(t)
+		c, db := setup(t)
 
 		enabled, err := SignupEnabled(c)
 		assert.NilError(t, err)
 		assert.Equal(t, enabled, true)
 
+		provider := models.Provider{Name: models.InternalInfraProviderName}
```
BruceMacD, line 106: This was relying on the provider being created in another test, added the explicit creation here

internal/server/authn/key_exchange.go
```diff
@@ -0,0 +1,51 @@
...
+
+	identity, err := data.GetIdentity(db, data.ByID(validatedRequestKey.IssuedFor))
+	if err != nil {
+		return nil, nil, fmt.Errorf("user is not valid: %w", err) // the user was probably deleted
+	}
+
+	return identity, data.InfraProvider(db), nil
+}
+
+func (a *keyExchangeAuthn) Name() string {
```
BruceMacD, line 45: These names are just used in telemetry

internal/server/authn/oidc.go
```diff
@@ -202,3 +291,22 @@ func (o *oidcImplementation) GetUserInfo(providerUser *models.ProviderUser) (*In
...
+// UpdateUserInfoFromProvider calls the user info endpoint of an external identity provider to see a user's current attributes
+func UpdateUserInfoFromProvider(db *gorm.DB, info *InfoClaims, user *models.Identity, provider *models.Provider) error {
+	// add user to groups they are currently in
+	var groups []string
+
+	for i := range info.Groups {
+		name := info.Groups[i]
+		groups = append(groups, name)
+	}
+
```
ssoroka, line 304: why not just reference info.Groups? these slices are identical

internal/server/authn/password_credentials.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Infra users can have only one username/password combo, look it up
+	userCredential, err := data.GetCredential(db, data.ByIdentityID(identity.ID))
+	if err != nil {
+		return nil, nil, fmt.Errorf("validate creds get user: %w", err)
+	}
+
+	// check if this is a single use password that was already used
+	if userCredential.OneTimePassword && userCredential.OneTimePasswordUsed {
+		return nil, nil, fmt.Errorf("one time password cannot be used more than once")
```
ssoroka, line 40: add some guidance? "have an administrator reset your password" ?

internal/server/authn/password_credentials_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"gotest.tools/v3/assert"
+
+	"github.com/infrahq/infra/internal/server/data"
+	"github.com/infrahq/infra/internal/server/models"
+)
+
+func TestPasswordCredentialAuthentication(t *testing.T) {
+	db := setupDB(t)
+
+	cases := map[string]map[string]interface{}{
```
ssoroka, line 17: using a struct for setup/verify would be typesafe.

internal/access/credential_test.go
```diff
@@ -2,167 +2,59 @@ package access
...
+
+	username := "bruce@example.com"
+	user := &models.Identity{Name: username}
+	err := data.CreateIdentity(db, user)
+	assert.NilError(t, err)
+
+	_, err = CreateCredential(c, *user)
+	assert.NilError(t, err)
+
+	t.Run("Update user credentials IS single use password", func(t *testing.T) {
```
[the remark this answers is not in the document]
BruceMacD: This scenario is tested in `authn/password_credentials_test.go`

internal/server/data/credential.go
```diff
@@ -26,27 +25,12 @@ func DeleteCredential(db *gorm.DB, id uid.ID) error {
 	return delete[models.Credential](db, id)
 }
 
-func ValidateCredential(db *gorm.DB, user *models.Identity, password string) (bool, error) {
+// IdentityCredentialMustBeUpdated checks if the associated identity has a one time password and that password has been used
+func IdentityCredentialMustBeUpdated(db *gorm.DB, user *models.Identity) (bool, error) {
```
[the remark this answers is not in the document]
BruceMacD: Updated to match this suggestion

internal/server/authn/authn_method.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/infrahq/infra/internal/server/models"
+)
+
+type LoginMethod interface {
+	Authenticate(db *gorm.DB) (*models.Identity, *models.Provider, error)
+	Name() string                             // Name returns the name of the authentication method used
+	RequiresUpdate(db *gorm.DB) (bool, error) // Temporary way to check for one time password re-use, remove with #1441
+}
+
+func Login[T LoginMethod](db *gorm.DB, loginMethod T, keyExpiresAt time.Time, keyExtension time.Duration) (*models.AccessKey, string, error) {
```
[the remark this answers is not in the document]
BruceMacD: Switched to the interface

internal/server/authn/oidc.go
BruceMacD: Good spot, I think this is leftover from something in the past. Cleaned it up.

internal/server/authn/password_credentials.go
BruceMacD: This error doesn't reach the user in the client, so more details dont really help here. We can revisit this in a separate change.

internal/server/authn/password_credentials_test.go
BruceMacD: I looked at this, but I liked keeping the setup/verify functions together for each case in the same map, so I'll leave it for now

internal/server/authn/authn_method.go
```diff
@@ -16,9 +16,14 @@ type LoginMethod interface {
 	RequiresUpdate(db *gorm.DB) (bool, error) // Temporary way to check for one time password re-use, remove with #1441
 }
 
-func Login[T LoginMethod](db *gorm.DB, loginMethod T, keyExpiresAt time.Time, keyExtension time.Duration) (*models.AccessKey, string, error) {
+func Login(db *gorm.DB, loginMethod interface{}, keyExpiresAt time.Time, keyExtension time.Duration) (*models.AccessKey, string, error) {
```
ssoroka, line 19: suggest replacing `loginMethod interface{}` with `loginMethod LoginMethod`, because you can reference the exact interface you need to call, then you don't need type assertion below.

BruceMacD: That makes way more sense, fixed

PR #2174, opened 2022-06-02
maintain: remove some unused server options

dnephin:
This PR removes some unused `loadCertificate` logic and options from the server. We may restore this in the future when it is being used. For now I think it's better to remove it so that contributors are not distracted by it. Removing it also avoids doing unnecessary work on server startup.

This PR also moves one test file into the `pki` package. It was only testing things in that package.  The `pki` package is also not used, but at least it's well contained and is not called by any production code.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -3,11 +3,7 @@ package server
 import (
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"embed"
-	"encoding/base64"
-	"encoding/json"
-	"encoding/pem"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -35,7 +31,6 @@ import (
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/metrics"
-	"github.com/infrahq/infra/pki"
 )
 
 type Options struct {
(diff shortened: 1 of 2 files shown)


BruceMacD:
Any way we can capture this commit in an issue for loading certificates in Infra? It seems like an option we will want soon and I'd like an easy way to find this existing work.

ssoroka (changes requested):
It wasn't too long ago that I resolved the issue that asked for the ability to load user-specified certs. Seems like a step backwards. cc @jmorganca to decide how to move forward.

dnephin:
Good call to create an issue about this. I was thinking of adding support for that next. I've created #2176 to track it.

I'm not really sure how this code being removed relates to users being able to specify their own cert and key. Supporting that should be really simple, just creating a `tls.Config` out of the config they provided. I suspect we won't use any of what is being removed here to support #2176.

ssoroka:
Ah, sorry, this is the certificate provider stuff. This is related to the plan that we will manage certificate rotation for the user. Can't say I'm excited to see it go away.

dnephin:
Git makes it easy to restore it back later. Is there an issue related to that feature? I can put a note into that issue to link back to this PR, so we can find it easily.

ssoroka (approved):
discussed concerns. good to go.

dnephin:
Thank you!

Issue #2176, opened 2022-06-02
server: allow user to provide their own TLS cert and key


dnephin:
Currently we use `acme/autocert` to accept either a cert from https://letsencrypt.org/, or generate one on the first request if a let's encrypt cert is not found.

Many orgs already have a chain of trust setup, and will likely want to provide a cert they've already generated to the `infra server`. Technically they could add the cert and key to the directory specified by `tlsCache`, but that's an undocumented interface, and probably not the one we want to support as a public interface.

We should have a way for the user to specify their own CA, cert, and private key.  Adding something like this to the server options should work to allow the user to specify their own certs:

```go
type TLSConfig struct {
    CA string
    Certificate string
    PrivateKey string
}
```

Where each field is the path to a file.  We might not need `CA` initially, and might be fine with just cert and key. The `CA` would allow us to do mTLS if the client or connector also passed a certificate signed by the same CA.

From these options we can construct a `tls.Config`, and use that `tls.Config` in the `http.Server`.

dnephin:
Done in #2398, follow up work will be part of #2362

PR #2177, opened 2022-06-02
cli: prompt to accept the server TLS certificate for manual verification

### LINKED ISSUE #1541: Allow users to manually trust the server certificate on `infra login`

Current `infra login` requires `--skip-tls-verify`, or the user must accept the prompt for "The authenticity of the host can't be established.".

Similar to how `ssh` works when you connect to a new host, we should be able to prompt the user to trust the certificate used by the server.

When the server starts we should print the certificate to the logs and/or to a file on the ephemeral filesystem. This will allow the admin to view the generated certificate using a `kubectl` command. I think we may already do this in some cases (`~/.infra/cache/` seems to have certs).

In the CLI we would need to make 2 changes:

1. When we get a `x509.UnknownAuthorityError`, print the details of the certificate (they are available as `x509.UnknownAuthorityError.Certificate`). Then prompt the user to see if they would like the trust the certificate. We can also suggest that they either ask their admin to confirm the details, or if they are an admin they can use `kubectl exec ...` to print the expected cert. Once trusted we can save the certificate to the local filesystem.
2. When we construct the API client in the CLI we can use a `tls.Config` that populates the `RootCAs` from the certs saved on disk. That way future calls to the API will continue to trust the certificate that was previously trusted by that user.

dnephin:
## Summary

This PR adds a new prompt to `infra login` when the servers TLS certificate is not already trusted. The prompt gives the user information about the server certificate and allows them to manually trust it.

Once trusted, the user will be able to login again to the same server without being prompted again.

This is a significant enhancement over `--skip-tls-verify` because even if the user doesn't properly inspect the fingerprint, the only opportunity for a man-in-the-middle attack is on that first request. Every subsequent request is secure.

In a follow up PR I will:
* add a command line flag to specify a trusted certificate
* add the same flag to the connector
* update the quickstart to remove `--skip-tls-verify`

## Related Issues

Resolves #1541

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -104,7 +104,7 @@ func httpTransportForHostConfig(config *ClientHostConfig) *http.Transport {
 	if config.TrustedCertificate != "" {
 		ok := pool.AppendCertsFromPEM([]byte(config.TrustedCertificate))
 		if !ok {
-			logging.S.Warnf("Failed to read trusted certificates for server: %v", err)
+			logging.S.Warnf("Failed to read trusted certificates for server")
 		}
 	}
 
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -153,7 +153,7 @@ $ infra use development.kube-system`,
 				return err
 			}
 
-			err = updateKubeconfig(client, id)
+			err = updateKubeConfig(client, id)
 			if err != nil {
 				return err
 			}
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
+	"github.com/infrahq/infra/internal/certs"
 	"github.com/infrahq/infra/internal/cmd/cliopts"
 	"github.com/infrahq/infra/internal/connector"
 	"github.com/infrahq/infra/internal/logging"
@@ -97,15 +98,17 @@ func apiClient(host string, accessKey string, transport *http.Transport) *api.Cl
 func httpTransportForHostConfig(config *ClientHostConfig) *http.Transport {
 	pool, err := x509.SystemCertPool()
 	if err != nil {
-		// TODO: print warning about this case
+		logging.S.Warnf("Failed to load trusted certificates from system: %v", err)
 		pool = x509.NewCertPool()
 	}
 
 	if len(config.TrustedCertificate) > 0 {
-		// TODO: log or return the error
-		// TODO: read this in PEM format
-		cert, _ := x509.ParseCertificate(config.TrustedCertificate)
-		pool.AddCert(cert)
+		cert, err := x509.ParseCertificate(certs.PEMDecode(config.TrustedCertificate))
+		if err != nil {
+			logging.S.Warnf("Failed to read trusted certificates for server: %v", err)
+		} else {
+			pool.AddCert(cert)
+		}
 	}
 
 	return &http.Transport{
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -77,23 +77,27 @@ func defaultAPIClient() (*api.Client, error) {
 		return nil, err
 	}
 
-	return apiClient(config.Host, config.AccessKey, config.SkipTLSVerify), nil
+	return apiClient(config.Host, config.AccessKey, defaultHTTPTransport(config.SkipTLSVerify)), nil
 }
 
-func apiClient(host string, accessKey string, skipTLSVerify bool) *api.Client {
+func apiClient(host string, accessKey string, transport *http.Transport) *api.Client {
 	return &api.Client{
 		Name:      "cli",
 		Version:   internal.Version,
 		URL:       "https://" + host,
 		AccessKey: accessKey,
 		HTTP: http.Client{
-			Timeout: 60 * time.Second,
-			Transport: &http.Transport{
-				TLSClientConfig: &tls.Config{
-					//nolint:gosec // We may purposely set insecureskipverify via a flag
-					InsecureSkipVerify: skipTLSVerify,
-				},
-			},
+			Timeout:   60 * time.Second,
+			Transport: transport,
+		},
+	}
+}
+
+func defaultHTTPTransport(skipTLSVerify bool) *http.Transport {
+	return &http.Transport{
+		TLSClientConfig: &tls.Config{
+			//nolint:gosec // We may purposely set insecureskipverify via a flag
+			InsecureSkipVerify: skipTLSVerify,
 		},
 	}
 }
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
-	"github.com/infrahq/infra/internal/certs"
 	"github.com/infrahq/infra/internal/cmd/cliopts"
 	"github.com/infrahq/infra/internal/connector"
 	"github.com/infrahq/infra/internal/logging"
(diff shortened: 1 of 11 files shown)


internal/cmd/cmd.go
```diff
@@ -77,23 +78,41 @@ func defaultAPIClient() (*api.Client, error) {
...
+	if config.TrustedCertificate != "" {
+		ok := pool.AppendCertsFromPEM([]byte(config.TrustedCertificate))
+		if !ok {
+			logging.S.Warnf("Failed to read trusted certificates for server: %v", err)
+		}
+	}
+
+	return &http.Transport{
+		TLSClientConfig: &tls.Config{
+			//nolint:gosec // We may purposely set insecureskipverify via a flag
```
BruceMacD, line 113: Just verifying, this TLS verify skip is left in for the connector, which you're updating in a different PR?

internal/cmd/login.go
```diff
@@ -81,16 +85,32 @@ $ infra login --key 1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy`,
...
+		}
+
+		options.Server, err = promptServer(cli, config)
 		if err != nil {
 			return err
 		}
 	}
 
-	client, err := newAPIClient(cli, options)
+	if len(options.TrustedCertificate) == 0 {
```
BruceMacD, line 104: seems like this will always be true in the current implementation

internal/cmd/login.go
```diff
@@ -176,25 +179,19 @@ func loginToInfra(cli *CLI, client *api.Client, loginReq *api.LoginRequest) erro
...
-	if err := updateInfraConfig(client, loginReq, loginRes); err != nil {
+	if err := updateInfraConfig(lc, loginReq, loginRes); err != nil {
 		return err
 	}
 
-	// Client needs to be refreshed from here onwards, based on the newly saved infra configuration.
-	client, err = defaultAPIClient()
-	if err != nil {
-		return err
-	}
```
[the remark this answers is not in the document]
BruceMacD: I think I remember this being done because the call to get the user's grants in `updateKubeConfig` was not getting the new access key that was issued on login. It might have been fixed as a side-effect somewhere else.

BruceMacD (commented):
I'm getting an error while attempting to test this on `localhost` any ideas? 

```
$ infra login localhost
Error: Get "https://localhost": x509: “Infra” certificate is not trusted
```

This shouldn't block login like this completely I imagine.

internal/cmd/login.go
dnephin: Ya, we need to add a `--tls-trusted-ca` or `--tls-trusted-cert` flag to the CLI. I was thinking of doing that as part of #2176.

There are some TODOs in this PR that mention it as well.

internal/cmd/cmd.go
dnephin: I wasn't planning on removing the flag entirely, just giving us other options so that we don't have to use it in the quickstart or error messages.

internal/cmd/login.go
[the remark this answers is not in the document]
dnephin: I was very confused by this, because the existing tests were passing. I found an incorrect `return nil` in `updateKubeConfig` which was masking the problem. With that fixed the existing tests fail.

I fixed this by setting the access key on the API client once login is successful.

dnephin:
@BruceMacD this should be working on MacOS now. We have to repeat the request without the system cert pool to get the right error type.

internal/cmd/login.go
dnephin: I opened #2288 which resolves those TODOs and adds the new flag.

internal/cmd/login.go
```diff
@@ -526,34 +597,71 @@ func promptLoginOptions(cli *CLI, client *api.Client) (loginMethod loginMethod,
...
+	// TODO: improve this message
+	// TODO: use color/bold to highlight important parts
+	// TODO: test format with golden
+	fmt.Fprintf(cli.Stderr, `
+The certificate presented by the server is not trusted by your operating system. It
+could not be automatically verified.
+
+Certificate
+
+Subject: %[1]s
```
jmorganca, line 610: This is awesome. I wonder if we can trim this down a bit – e.g. do we need validity, SANs, etc?

internal/cmd/login.go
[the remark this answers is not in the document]
jmorganca: Great catch

internal/cmd/login.go
```diff
@@ -373,69 +400,113 @@ func runSignupForLogin(cli *CLI, client *api.Client) (*api.LoginRequestPasswordC
...
+		Timeout: 60 * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{RootCAs: pool, MinVersion: tls.VersionTLS12},
+		},
+	}
+	res, err = httpClient.Do(req)
+	urlErr := &url.Error{}
+	switch {
+	case err == nil:
+		res.Body.Close()
```
BruceMacD, line 502: Should we be closing the response body in any case?

internal/cmd/login.go
```diff
@@ -373,69 +400,113 @@ func runSignupForLogin(cli *CLI, client *api.Client) (*api.LoginRequestPasswordC
...
+	httpClient := http.Client{Timeout: 60 * time.Second}
+	res, err := httpClient.Do(req)
+	if err == nil {
+		res.Body.Close()
+		return nil
 	}
 
-	logging.S.Debugf("call server: test tls for %q", host)
-	res, err := http.DefaultClient.Do(req)
+	// Second attempt with an empty cert pool. This is necessary because at least
```
BruceMacD, line 479: Any open issue for this in the Go project?

BruceMacD (approved):
This works great for me now, I agree with Jeff's comment about cutting down on the size of the output for trusting the cert, but I think we should leave the expiry in the output.

internal/cmd/login.go
dnephin: The godoc for `Client.Do` (https://pkg.go.dev/net/http#Client.Do) says this:

> On error, any Response can be ignored. A non-nil Response with a non-nil error only occurs when CheckRedirect fails, and even then the returned Response.Body is already closed.

So we can only close the response body when err is nil.

internal/cmd/login.go
dnephin: I couldn't find one, so I created https://github.com/golang/go/issues/53401.

I'll add a link to the comment.

internal/cmd/login.go
```diff
@@ -525,34 +596,71 @@ func promptLoginOptions(cli *CLI, client *api.Client) (loginMethod loginMethod,
...
+
+Validity
+  Not Before: %[3]v
+  Not After: %[4]v
+
+Subject Alternative Names:
+  DNS Names: %[5]s
+  IP Addresses: %[6]v
+
+SHA-256 Fingerprint
```
jmorganca, line 620: Does this get printed in the server logs?

dnephin: It does, but unfortunately only when the cert is first created. 

As part of #2176 I'd like to change cert generation to happen at startup, instead of on first request. If we make that change we can have the fingerprint printed on every startup, instead of only when it is first created.  That will ensure that the value is always in the logs for the pod. I believe right now it'll be hard to find if the server pod was recreated after the cert is generated.

The admin can also generate the the value using:

```
openssl x509 -noout -fingerprint -sha256 -inform pem -in
```

As long as they know where to find the certificate.

internal/cmd/login.go
dnephin: Ok, I removed the SAN fields. We definitely don't need those since they will be verified automatically by the client.

I left the validity times, but I added a human readable version first which hopefully makes it read better.

jmorganca (approved):

internal/cmd/login.go
jmorganca: Makes sense, we can address that in a follow up change that moves cert generation to server start

PR #2180, opened 2022-06-03
feat: added pagination + sorting for list 

### LINKED ISSUE #1341: API sorting
Results from `handlers.go`'s `ListProviders` should return alphabetically. This helps with the visualization on the user-facing areas.

j-sneh:
## Checklist

## Related Issues

Resolves #1341 

So far, this is a rough draft with a few broken/missing features:
- MaxPage and Next field in pagination do not work yet
- Pagination works for ListUsers only
- CLI needs changes
- Sorting options need to be more robust and easy to use
- List tests fail, as they're not properly updated

### DIFF
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -5,9 +5,9 @@ import (
 )
 
 type PaginationRequest struct {
-	Page  int    `form:"page"`
-	Limit int    `form:"limit"`
-	Sort  string `form:"sort" validate:"oneof='name ASC' 'name DESC' 'id ASC' 'id DESC' ''"`
+	Page  int    `form:"page" validate:"min=0"`
+	Limit int    `form:"limit" validate:"min=0,max=250"`
+	Sort  string `form:"sort" validate:"oneof='name_ASC' 'name_DESC' 'id_ASC' 'id_DESC' ''"`
 }
 
 type PaginationResponse struct {
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -1,0 +1,25 @@
+package api
+
+import (
+	"math"
+)
+
+type PaginationRequest struct {
+	Page  int    `form:"page"`
+	Limit int    `form:"limit"`
+	Sort  string `form:"sort" validate:"oneof='name ASC' 'name DESC' 'id ASC' 'id DESC' ''"`
+}
+
+type PaginationResponse struct {
+	Page  int    `json:"page"`
+	Limit int    `json:"limit"`
+	Sort  string `json:"sort"`
+
+	MaxPage int `json:"max_page"`
+	Next    int `json:"next,omitempty"`
+	Prev    int `json:"prev,omitempty"`
+}
+
+func GetMaxPage[T any](p *PaginationResponse, lr *ListResponse[T]) int {
+	return int(math.Ceil(float64(lr.Count) / float64(p.Limit)))
+}
--- a/internal/server/data/selectors.go
+++ b/internal/server/data/selectors.go
@@ -1,8 +1,6 @@
 package data
 
 import (
-	"math"
-
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal/server/models"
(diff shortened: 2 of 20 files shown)


api/pagination.go
```diff
@@ -0,0 +1,25 @@
+package api
+
+import (
+	"math"
+)
+
+type PaginationRequest struct {
+	Page  int    `form:"page"`
+	Limit int    `form:"limit"`
+	Sort  string `form:"sort" validate:"oneof='name ASC' 'name DESC' 'id ASC' 'id DESC' ''"`
```
BruceMacD, line 10:
```suggestion
	Sort  string `form:"sort" validate:"oneof='name_ASC' 'name_DESC' 'id_ASC' 'id_DESC' ''"`
```

Not having spaces in the query params is safer for use in a URL, for other two word parameters we use the underscore.

internal/server/data/selectors.go
```diff
@@ -118,6 +119,22 @@ func ByUserID(userID uid.ID) SelectorFunc {
...
+	if pg == nil {
+		defaultpg := models.RequestToPagination(nil)
+		pg = &defaultpg
+	}
+
+	return func(db *gorm.DB) *gorm.DB {
+		// var count int64
+		// db.Count(&count)
+
+		pg.MaxPage = -999999 //TODO: get maxpage working
```
BruceMacD, line 132: This seems like it will set after the db query, so maybe applied to the result in `data.list` if pagination is used.

internal/server/data/selectors.go
```diff
@@ -118,6 +119,22 @@ func ByUserID(userID uid.ID) SelectorFunc {
...
+		// I don't think this is the right place to do this necessarily
+		return db.Offset(pg.Limit * (pg.Page - 1)).Limit(pg.Limit).Order(pg.Sort)
```
BruceMacD, line 134: This seems ok here to me, other place could be in the `data.List` function as well. 

What does `(pg.Page - 1)` do? Could this result in unexpected results if the request specifies a negative page number?

internal/server/migrations.go
```diff
@@ -111,9 +111,10 @@ func (a *API) addResponseRewrites() {
 		return newResponse.Items
 	})
 	addResponseRewrite(a, http.MethodGet, "/v1/grants/:id", "0.12.2", migrateUserGrantToIdentity)
-	addResponseRewrite(a, http.MethodGet, "/v1/grants", "0.12.2", func(newResponse *api.ListResponse[api.Grant]) []identityGrant {
-		return api.NewListResponse(newResponse.Items, migrateUserGrantToIdentity).Items
-	})
+	// addResponseRewrite(a, http.MethodGet, "/v1/grants", "0.12.2", func(newResponse *api.ListResponse[api.Grant]) []identityGrant {
+	// 	return api.NewListResponse(newResponse.Items, migrateUserGrantToIdentity).Items
+	// })
+	//TODO: Fix List Migrations (need pagination or to search through all)
```
BruceMacD, line 117: Setting pagination to nil is what I would expect where pagination wasn't in this previous version. Might need a response re-write also that removes the pagination struct.

internal/server/models/pagination.go
```diff
@@ -0,0 +1,70 @@
...
+	Page    int
+	Limit   int
+	Sort    string
+	MaxPage int
+	Next    int
+	Prev    int
+}
+
+// function to convert API --> this object
+// method to convert Pagination to response
```
BruceMacD, line 16: There's a standard way to comment functions in go: https://go.dev/doc/effective_go#commentary

Pretty much you start the comment with the name of the function, for example:
```
// RequestToPagination converts an API request object to an internal pagination model
```

It's kinda redundant, but makes for consistency.

BruceMacD (commented):
Looking really good, this fits really nicely with the list response.

internal/server/data/selectors.go
j-sneh: No, because any number ≤ 0 gets set to page 1. Would it make more sense to send an error if someone passes a negative value or a value that's too large?

BruceMacD: I think an error makes sense 👍

api/pagination.go
mxyng: I suggest splitting this into multiple fields, one to communicate the sort by field and another to communicate the order. It's useful for scenarios where user wants to sort by more than just name or id, e.g. by created timestamp, without adding more combinations to this field.

```go
type PaginationRequest struct {
    ...
    SortBy  string `form:"sortBy"`
    OrderBy string `from:"orderBy"`
}
```

j-sneh: @mxyng What would be the benefit of that vs just having the OrderBy field? I don't think we're going to keep the oneof validation.

mxyng: It removes the need for a formatted input. Values can be validated easily while also being easier for the user to understand and use. The user no longer needs to know which format is supported and we no longer need to worry about spaces or other delimiters

PR #2181, opened 2022-06-03
fix: require user to have an email

### LINKED ISSUE #1878: Admin signup should be an email
Currently `/api/signup` supports a name or email. To keep it consistent, we should validate that users provide an email for the initial admin user.

BruceMacD:
## Summary
After the unification of user/machine identities it was no longer required for users to have an email as their identifier. We discussed this a bit internally and it seems more future-proof to require users created in Infra to have an email. This change requires users created through the Infra API to have an email, it also updates the UI and CLI to prompt for an email (and validate it).

Backwards compatibility:
- Users created through config will not require an email yet, just log a warning.
- Users previously created without an email will still work, you just can't create more.

## Checklist

## Related Issues

Resolves #1878

### DIFF
--- a/api/signup.go
+++ b/api/signup.go
@@ -5,7 +5,7 @@ type SignupEnabledResponse struct {
 }
 
 type SignupRequest struct {
-	Name     string `json:"name" validate:"required_without=Email"`
-	Email    string `json:"email" validate:"required_without=Name"` // #1825: remove, this is for migration
+	Name     string `json:"name" validate:"omitempty,email,required_without=Email"`
+	Email    string `json:"email" validate:"omitempty,email,required_without=Name"` // #1825: remove, this is for migration
 	Password string `json:"password" validate:"required"`
 }
--- a/api/user.go
+++ b/api/user.go
@@ -22,7 +22,7 @@ type ListUsersRequest struct {
 }
 
 type CreateUserRequest struct {
-	Name               string `json:"name" validate:"required"`
+	Name               string `json:"name" validate:"email,required"`
 	SetOneTimePassword bool   `json:"setOneTimePassword"`
 }
 
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -3,6 +3,7 @@ package server
 import (
 	"errors"
 	"fmt"
+	"net/mail"
 	"os"
 	"strings"
 	"time"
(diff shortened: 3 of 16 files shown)


api/signup.go
```diff
@@ -5,7 +5,7 @@ type SignupEnabledResponse struct {
 }
 
 type SignupRequest struct {
-	Name     string `json:"name" validate:"required_without=Email"`
-	Email    string `json:"email" validate:"required_without=Name"` // #1825: remove, this is for migration
+	Name     string `json:"name" validate:"omitempty,email,required_without=Email"`
```
mxyng, line 8: Is there a point to keeping `Name` if it needs to be an email?

api/user.go
```diff
@@ -22,7 +22,7 @@ type ListUsersRequest struct {
 }
 
 type CreateUserRequest struct {
-	Name               string `json:"name" validate:"required"`
+	Name               string `json:"name" validate:"email,required"`
```
mxyng, line 25: So this means machine users need to be have an email identifier?

internal/server/config.go
```diff
@@ -746,6 +747,11 @@ func (s Server) loadUser(db *gorm.DB, input User) (*models.Identity, error) {
 		}
 	}
 
+	_, err := mail.ParseAddress(name)
+	if err != nil {
+		logging.S.Warnf("user name %q in server configuration is not a valid email, please update this name to a valid email", name)
```
mxyng, line 752: This will always log for the connector user which has name `connector`

ui/lib/email.js
```diff
@@ -0,0 +1,7 @@
+export const validateEmail = email => {
+  return String(email)
+    .toLowerCase()
+    .match(
+      /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
mxyng, line 5: This will reject something like `name@domain` which is valid according to `mail.ParseAddress`

hoyyeva (approved):
the UI part looks good to me!

ui/lib/email.js
BruceMacD: Update the regex to make `.*` optional

PR #2184, opened 2022-06-03
fix: ui scrolling

### LINKED ISSUE #2085: ui: scrolling behavior for lists and sidebars
1. Longer lists cause headers to go off screen (optional)

Screen Shot 2022-05-26 at 2 55 50 PM

2. Sidebars scrolls off-screen even when clicking on an item further in the list

Screen Shot 2022-05-26 at 2 40 20 PM

jmorganca:
Fixes #2085

### DIFF
--- a/ui/pages/destinations/index.js
+++ b/ui/pages/destinations/index.js
@@ -119,12 +119,12 @@ export default function Destinations () {
       </Head>
       {!loading && (
         <div className='flex-1 flex h-full'>
-          <main className='flex-1 flex flex-col space-y-4'>
+          <div className='flex-1 flex flex-col space-y-4'>
             <PageHeader header='Clusters' buttonHref={admin && '/destinations/add'} buttonLabel='Cluster' />
             {error?.status
               ? <div className='my-20 text-center font-light text-gray-300 text-sm'>{error?.info?.message}</div>
               : (
-                <div>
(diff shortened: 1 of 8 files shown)


ui/pages/destinations/index.js
```diff
@@ -119,12 +119,12 @@ export default function Destinations () {
...
       {!loading && (
         <div className='flex-1 flex h-full'>
-          <main className='flex-1 flex flex-col space-y-4'>
+          <div className='flex-1 flex flex-col space-y-4'>
             <PageHeader header='Clusters' buttonHref={admin && '/destinations/add'} buttonLabel='Cluster' />
             {error?.status
               ? <div className='my-20 text-center font-light text-gray-300 text-sm'>{error?.info?.message}</div>
               : (
-                <div>
+                <div className='px-6 flex-1 min-h-0 overflow-y-scroll px-6'>
```
hoyyeva, line 127: small suggestion - i think we should move this into the `table.js` component.

jmorganca: I like this suggestion to avoid code duplication, but since this also effects the `EmptyTable` component below and needs to be a parent for the table element we may want to keep it here.

That said, this is a great suggestion since we should consider pulling out this table logic into it's own component – there's quite a bit of repeated code between `destinations.js` `users.js` and `providers.js`

hoyyeva (approved):

PR #2189, opened 2022-06-03
fix: loading of dbUsername field for server config

dnephin:
In #1755 we changed from using `mapstructure` tags to using `config` tags, but I forgot to remove or update all the existing struct tags.

Every other field continued to work because the struct field name matched the convention. Unfortunately `dbUsername` was the one case where it did not match convention and required the struct tag. A workaround was to use `dbUser` or `INFRA_SERVER_DB_USER`.

This PR fixes the bug by renaming the field to `DBUsername`.

It also adds test coverage for every field in `server.Options` to show this is the only field with the problem. 

I also removed the struct tags, since they no longer do anything and keeping them would be confusing. I'll leave some inline comments for where each of the fields is tested.

### DIFF
--- a/internal/cmd/server_test.go
+++ b/internal/cmd/server_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"net/url"
 	"path/filepath"
 	"reflect"
 	"testing"
@@ -50,32 +51,6 @@ func TestServerCmd_LoadOptions(t *testing.T) {
 	}
 
 	testCases := []testCase{
-		{
-			name: "secret providers",
-			setup: func(t *testing.T, cmd *cobra.Command) {
-				content := `
-                    secrets:
-                      - kind: env
-                        name: base64env
-                        config:
-                          base64: true`
-
-				dir := fs.NewDir(t, t.Name(),
-					fs.WithFile("cfg.yaml", content))
-				cmd.SetArgs([]string{"--config-file", dir.Join("cfg.yaml")})
-			},
-			expected: func(t *testing.T) server.Options {
-				expected := defaultServerOptions(filepath.Join(dir, ".infra"))
-				expected.Secrets = []server.SecretProvider{
-					{
-						Kind:   "env",
-						Name:   "base64env",
-						Config: server.GenericConfig{Base64: true},
-					},
-				}
-				return expected
-			},
-		},
 		{
 			name: "config filename specified as env var",
 			setup: func(t *testing.T, cmd *cobra.Command) {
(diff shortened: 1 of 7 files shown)


internal/cmd/server_test.go
```diff
@@ -50,32 +51,6 @@ func TestServerCmd_LoadOptions(t *testing.T) {
 	}
 
 	testCases := []testCase{
-		{
-			name: "secret providers",
```
dnephin, line 54: This test case is now part of "all options from config" below.

internal/connector/connector.go
```diff
@@ -38,12 +38,12 @@ import (
 )
 
 type Options struct {
-	Server        string `mapstructure:"server"`
-	Name          string `mapstructure:"name"`
-	AccessKey     string `mapstructure:"accessKey"`
-	CACert        string `mapstructure:"caCert"`
-	CAKey         string `mapstructure:"caKey"`
-	SkipTLSVerify bool   `mapstructure:"skipTLSVerify"`
```
dnephin, line 46: This change is covered by an existing test: `cmd.TestConnectorCmd`

internal/server/config.go
```diff
@@ -20,87 +20,87 @@ import (
...
-	Grants    []Grant    `mapstructure:"grants" validate:"dive"`
-	Users     []User     `mapstructure:"users" validate:"dive"`
+	Providers []Provider `validate:"dive"`
+	Grants    []Grant    `validate:"dive"`
+	Users     []User     `validate:"dive"`
 }
 
 type KeyProvider struct {
-	Kind   string      `mapstructure:"kind" validate:"required"`
+	Kind   string      `validate:"required"`
```
dnephin, line 52: The key provider and secret provider loading is covered by:
* `server.TestKeyProvider_PrepareForDecode_IntegrationWithDecode_FullConfig`
* `server.TestSecretProvider_PrepareForDecode_IntegrationWithDecode_FullConfig`

internal/server/options_migrations.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 type OptionsDiffV0dot1 struct {
-	Identities []User `mapstructure:"identities" validate:"dive"`
+	Identities []User `validate:"dive"`
```
dnephin, line 11: Covered by `cmd.TestApplyVersion0ConfigToLatest`

internal/server/server.go
```diff
@@ -34,26 +34,26 @@ import (
...
+	EnableSignup    bool
+	SessionDuration time.Duration
+
+	DBFile                  string
+	DBEncryptionKey         string
+	DBEncryptionKeyProvider string
+	DBHost                  string
+	DBPort                  int
+	DBName                  string
+	DBUsername              string
```
dnephin, line 49: The primary fix is this one line

ssoroka (approved):

internal/cmd/server_test.go
```diff
@@ -141,6 +116,152 @@ func TestServerCmd_LoadOptions(t *testing.T) {
...
+  - group: group1
+    resource: production
+    role: special
+
+users:
+  - name: username
+    accessKey: access-key
+    password: the-password
+
+# TODO: config providers/grants/users
```
BruceMacD, line 179:
```suggestion
```

Looks like this is actually in the test config

mxyng (approved):

internal/cmd/server_test.go
dnephin: Good catch! I forgot to remove the command after adding them

PR #2191, opened 2022-06-03
improve: name the connector sync function

ssoroka:
## Summary

for clarity, name the connector sync function. first step in cleaning it up.


BruceMacD (approved):
This seems easier to read to me

PR #2193, opened 2022-06-03
feat: add list users in group api + groups list cli

pdevine:
## Summary

List users for a given group

## Checklist

### DIFF
--- a/internal/cmd/groups.go
+++ b/internal/cmd/groups.go
@@ -40,7 +40,7 @@ func newGroupsCmd(cli *CLI) *cobra.Command {
 	cmd := &cobra.Command{
 		Hidden:  true,
 		Use:     "groups",
-		Short:   "Manage group identities",
+		Short:   "Manage groups of identities",
 		Aliases: []string{"group"},
 		Group:   "Management commands:",
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
--- a/internal/cmd/groups.go
+++ b/internal/cmd/groups.go
@@ -2,6 +2,9 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
 
 	"github.com/infrahq/infra/api"
 )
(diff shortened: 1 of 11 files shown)


internal/server/data/group_test.go
```diff
@@ -126,6 +126,41 @@ func TestListGroups(t *testing.T) {
 			}
 			assert.DeepEqual(t, actual, expected, cmpGroupShallow)
 		})
```
pdevine, line 128: There are also tests for `internal/server/handlers_test.go` however I realized that there isn't a way to add users to groups yet through the API, so those will have to wait until I add that part.

internal/server/data/identity.go
```diff
@@ -97,6 +97,23 @@ func ListIdentities(db *gorm.DB, selectors ...SelectorFunc) ([]models.Identity,
 	return list[models.Identity](db, selectors...)
 }
 
+func ListIdentitiesByGroup(db *gorm.DB, groupID uid.ID, selectors ...SelectorFunc) ([]models.Identity, error) {
+	group, err := GetGroup(db.Preload("Identities", func(db *gorm.DB) *gorm.DB {
+		for _, selector := range selectors {
```
pdevine, line 102: This adds the `name=?` and `ids=?` selectors to the query, and then sorts the results alphabetically to more closely resemble `ListIdentities()`

internal/cmd/groups.go
```diff
@@ -32,3 +35,75 @@ func createGroup(client *api.Client, name string) (*api.Group, error) {
 
 	return group, nil
 }
+
+func newGroupsCmd(cli *CLI) *cobra.Command {
+	cmd := &cobra.Command{
+		Hidden:  true,
```
pdevine, line 41: The command will stay hidden for now until the other subcommands are ready.

internal/cmd/groups.go
```diff
@@ -32,3 +35,75 @@ func createGroup(client *api.Client, name string) (*api.Group, error) {
...
+		Use:     "groups",
+		Short:   "Manage group identities",
```
BruceMacD, line 43:
```suggestion
		Short:   "Manage groups of identities",
```

internal/server/data/group_test.go
```diff
@@ -126,6 +126,41 @@ func TestListGroups(t *testing.T) {
...
+
+		t.Run("filter identities by group", func(t *testing.T) {
```
BruceMacD, line 130: This should be in `data/identity_test.go`

internal/cmd/groups.go
pdevine: changed.

internal/server/data/group_test.go
pdevine: moved.

BruceMacD (approved):

PR #2202, opened 2022-06-06
fix: deadlock before ingress ready

### LINKED ISSUE #2078: helm: infra connector will sometimes restart a few times before running
When deploying the Infra connector via `helm`, the pod will sometimes restart due to its liveness checks:

```
kubectl get pods -w
NAME                               READY   STATUS    RESTARTS     AGE
infra-connector-6486c8476f-5gc8n   0/1     Running   2 (2s ago)   83s
```

BruceMacD:
## Summary
Before an ingress is ready the connector should still respond to liveness checks. There was a deadlock in the certCache cert retrieval logic before the external host is resolvable. This caused the connector to restart until the external host was set.

The deadlock was that `CertCache.Certificate()` was locking the certCache mutex, but the `CertCache.AddHost()` function also uses this mutex, so it waits forever. Removed this lock from the reading logic, as it is only needed for writes (adding the host).

## Checklist

## Related Issues

Resolves #2078


internal/connector/connector.go
```diff
@@ -343,10 +343,9 @@ func (c *CertCache) AddHost(host string) error {
 }
 
 func (c *CertCache) Certificate() (*tls.Certificate, error) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
```
ssoroka, line 348: the lock protects c.cert, so you can't check it's nil without the lock.

BruceMacD: In the case multiple requests find it to be nil it will be ok due to locks and checks in the AddHost function.

For example:
1. Request A comes in and finds that the cert is nil.
2. At the same time request B comes in and finds that the cert is nil.
3. They both call c.AddHost("")
4. The addHost function has a lock that one of the requests gets.
5. The request adds the cert and releases the lock.
6. When the other request attemps to re-add the cert for the same host it finds that the host already was added and returns without re-generating.

ssoroka: multiple goroutines can't even call c.cert without being within a lock. it's not a matter of races.

BruceMacD: I see, thanks, I'll put a different lock here

ssoroka (changes requested):
still a no go. There's sync.RWMutex for this, but it still will cause you problems. Just do the nil check inside a lock and then unlock it immediately. You can separate the nil check from the conditional.

dnephin (commented):
Nice work tracking this down!

As Steven mentioned using two locks is probably not the right fix. We would have to acquire both locks in `AddHost` to make it work.

A `sync.RWLock` could work, but is probably more involved because we'd have to re-check `cert == nil` after acquiring the write lock.

Only acquiring the lock for the `nil` check is generally racy, but in this case probably works ok because `AddHost` checks the lists of hosts before doing any work. So even if two calls to `AddHost("")` happen,  the second one should be a no-op and avoids unnecessary work.

ssoroka (approved):

PR #2203, opened 2022-06-06
fix: create kubeconfig directory

### LINKED ISSUE #2201: Login fails if `KUBECONFIG` directory does not exist
## Describe the bug

On login, infra creates a temporary file in the `KUBECONFIG` directory. This may not exist if user hasn't set up kubectl yet. If the directory doesn't exist, login will fail

```
$ infra login host.docker.internal:8443 --skip-tls-verify
? Select a login method: Login with username and password
? Username: admin
? Password: ********
Error: cannot create temporary config file: open /root/.kube/infra-kube-config-2179185096: no such file or directory
exit status 1
```

## Expected behavior

Login is successful and `KUBECONFIG` is created

## Workaround

Create the directory. For most scenarios, this would be `mkdir -p $HOME/.kube`

mxyng:
## Summary

Infra CLI writes to a temporary file in the same directory as `KUBECONFIG`. Create this directory since it's possible it does not exist

## Checklist

## Related Issues

Resolves #2201


jmorganca (approved):
Nice fix!

PR #2206, opened 2022-06-07
fix: allow infra prefix for `infra use`

### LINKED ISSUE #2095: `infra use infra:<cluster name>` fails
Since we name the Kubernetes context `infra:<cluster name>`, we should support users who run `infra use infra:<cluster name>` vs just `infra use <cluster name>`

Today, Infra converts the `:` to a `.` and it fails:

```
$ infra use infra:digitalocean
ERROR  context not found: infra.digitalocean
```

j-sneh:
## Summary

Since a cluster is "officially" named infra:<cluster-name> in Kubernetes, this change allows a user to both write `infra use infra:name` and `infra use name`
## Checklist

~~Updated associated configuration where necessary~~
~~Considered data migrations for smooth upgrades~~

## Related Issues

Resolves #2095

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -119,9 +119,7 @@ func newUseCmd(_ *CLI) *cobra.Command {
 $ infra use development
 
 # Use a Kubernetes namespace context
-$ infra use development.kube-system
-
-# Note: infra:<cluster-name> is equivalent to <cluster-name>`,
+$ infra use development.kube-system`,
 		Args:  ExactArgs(1),
 		Group: "Core commands:",
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -119,7 +119,9 @@ func newUseCmd(_ *CLI) *cobra.Command {
 $ infra use development
 
 # Use a Kubernetes namespace context
-$ infra use development.kube-system`,
+$ infra use development.kube-system
+
+# Note: infra:<cluster-name> is equivalent to <cluster-name>`,
 		Args:  ExactArgs(1),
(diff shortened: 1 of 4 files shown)


internal/cmd/cmd.go
```diff
@@ -119,7 +119,9 @@ func newUseCmd(_ *CLI) *cobra.Command {
 $ infra use development
 
 # Use a Kubernetes namespace context
-$ infra use development.kube-system`,
+$ infra use development.kube-system
+
+# Note: infra:<cluster-name> is equivalent to <cluster-name>`,
```
jmorganca, line 124: We can remove this help text, since handling `infra:` is more of an exception / user error we're catching here

internal/cmd/cmd_test.go
```diff
@@ -196,6 +196,24 @@ func TestUse(t *testing.T) {
 		assert.Assert(t, is.Contains(kubeconfig.AuthInfos, "testuser@example.com"))
 	})
 
+	t.Run("InfraUse", func(t *testing.T) {
```
jmorganca, line 199: Nice work with this test!

internal/cmd/kubernetes.go
```diff
@@ -33,10 +33,11 @@ func kubernetesSetContext(cluster, namespace string) error {
 		return err
 	}
 
-	name := cluster
+	name := strings.TrimPrefix(cluster, "infra:")
+	fmt.Println(cluster, name)
```
jmorganca, line 37: Remove me

jmorganca (changes requested):
Nice work! A few comments

internal/cmd/kubernetes.go
j-sneh: haha oops

jmorganca (approved):

PR #2208, opened 2022-06-07
feat: introduce format flag, support json output

### LINKED ISSUE #972: cli: Support YAML for `--format` flag

- [x] `format=JSON` #2208
- [ ] `format=YAML`

```
$ infra list --format=text
```

Example, YAML format:

```
$ infra list --format=yaml
resources:
    access: cluster-admin
```

Example, JSON format:

```
$ infra list --format=json
{
  "resources": [
    {
      "access": "cluster-admin"
    }
  ]
}
```

kimskimchi:
## Summary

Introduces the `format` flag. Eg. `--format=json`. 
Introduces the json output for the existing table outputs. It is exactly 1-1 with the API. 

## Checklist

- [ ] Wrote appropriate unit tests - **IN PROGRESS**

## Related Issues

Implements #972

### DIFF
--- a/internal/cmd/destinations.go
+++ b/internal/cmd/destinations.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/spf13/cobra"
@@ -30,7 +31,8 @@ func newDestinationsCmd(cli *CLI) *cobra.Command {
 }
 
 func newDestinationsListCmd(cli *CLI) *cobra.Command {
-	return &cobra.Command{
+	var format string
+	cmd := &cobra.Command{
 		Use:     "list",
 		Aliases: []string{"ls"},
 		Short:   "List connected destinations",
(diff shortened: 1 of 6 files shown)


ssoroka (approved):
looks good. are we going to support this for other commands?

dnephin (commented):
Looks good! My only question is about the naming.

I generally think of text output as something like this:

```
Field1: value1
Field2: value2
```

For example, the output of the `infra info` command.

Where as table output is what we have here, output formatted as a table with columns and rows.

Are we sure we want to call this output `text` and not `table` ?

internal/cmd/users.go
```diff
@@ -115,22 +118,34 @@ func newUsersListCmd(cli *CLI) *cobra.Command {
...
 			}
 
-			for _, user := range users.Items {
-				rows = append(rows, row{
-					Name:       user.Name,
-					LastSeenAt: user.LastSeenAt.Relative("never"),
-				})
-			}
+			switch format {
+			case "text":
```
jmorganca, line 122: Since this is the default, it doesn't need it a name. Let's move this to `default` below

internal/cmd/destinations.go
```diff
@@ -47,28 +49,39 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
...
-
-			var rows []row
-			for _, d := range destinations.Items {
-				rows = append(rows, row{
-					Name: d.Name,
-					URL:  d.Connection.URL,
-				})
-			}
+			switch format {
+			case "text":
```
jmorganca, line 53: See below comment: make this `default:`

PR #2209, opened 2022-06-07
fix: migrate destination request CA from string

### LINKED ISSUE #2200: connector: error when using 0.13.1 with `main`
When using Connector on `main` with Infra Server on `0.13.1`:

```
{"level":"error","ts":1654277753.9561458,"caller":"connector/connector.go:525","msg":"initializing destination: error creating destination: bad request: illegal base64 data at input byte 0"}
```

BruceMacD:
## Summary
When using a connector on versions `0.13.1` and earlier they were unable to parse the CA response on destination requests as they were expecting a string. This PR adds migrations to convert the CA on destination request/responses for `0.13.1` and below.

## Checklist

## Related Issues

Resolves #2200

### DIFF
--- a/internal/server/migrations.go
+++ b/internal/server/migrations.go
@@ -79,6 +79,47 @@ func (a *API) addRequestRewrites() {
 			Privilege: oldRequest.Privilege,
 		}
 	})
+	type createDestinationRequestV0_13_1 struct {
+		UniqueID   string                       `json:"uniqueID"`
+		Name       string                       `json:"name" validate:"required"`
+		Connection destinationConnectionV0_13_1 `json:"connection"`
+
+		Resources []string `json:"resources"`
+		Roles     []string `json:"roles"`
+	}
+	type updateDestinationRequestV0_13_1 struct {
+		ID         uid.ID                       `uri:"id" json:"-" validate:"required"`
+		Name       string                       `json:"name" validate:"required"`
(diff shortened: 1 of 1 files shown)


dnephin (commented):
Could you add a test case that fails without this migration? I guess we have a way to reproduce it, right?

ssoroka (approved):
Ok I understand what's going on. relevant playground: https://go.dev/play/p/I_zKy4KKZeQ
[]byte and string aren't json encoded the same way. []byte is base64 encoded, and this migration fixes it by converting the encoding type properly.

dnephin (changes requested):
We should at the very least have a test for this before merging.

But I think given what we've learned, we should also reconsider using `[]byte` for this. A pem encoded cert is already base64 encoded.

I think another approach would be use to a `type PEMstring string` and make it so that the automatic trimming does not apply to custom types.

We can add a `UnmarshalText` to `PEMstring` to do the correct normalization.

dnephin:
I put together an example of that here: #2217, but it still needs more tests

ssoroka:
I've forgotten why I changed it to []byte already, but perhaps we should consider changing the API back to a string field for this.

dnephin:
I believe PEM encoded strings are supposed to have a single trailing newline, so we wanted to avoid removing it.

BruceMacD:
Closed in favor of #2217

PR #2212, opened 2022-06-07
fix: modify set-password help text

### LINKED ISSUE #2146: Incorrect `infra users edit --password` help text
`infra users edit` incorrectly mentions that `--password` sets the users' one time password, when it sets their permanent password.

```
# Set a user's password
$ infra users edit janedoe@example.com --password

Flags:
      --password   Set password
```

Related:
* #2145

hemachandarv:
## Summary

`infra users edit` incorrectly mentions that `--password` sets the users' "one-time password" when it sets their permanent password. "one-time" has been removed from the help text to resolve the ambiguity.

## Related Issues

Resolves #2146

### DIFF
--- a/docs/reference/cli-reference.md
+++ b/docs/reference/cli-reference.md
@@ -338,14 +338,14 @@ infra users edit USER [flags]
 ### Examples
 
 ```
-# Set a new one time password for a user
+# Set a new password for a user
 $ infra users edit janedoe@example.com --password
 ```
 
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

hemachandarv:
@jmorganca Great! Thank you so much for accepting it! 

I'll also check if there is an issue that I can address.

PR #2213, opened 2022-06-07
feat: updates cli forbidden to more user friendly messages

### LINKED ISSUE #2096: cli: reword `forbidden` errors to include context about which operation failed
After #2032, if an API call fails (especially one for authorization) the error message is just `forbidden` . I think we should include more detail in the error. Ex:

```
failed to list groups for user <uid>
```

We should be able to translate the http method and http path in the error into a human readable message.

kimskimchi:
## Summary

Refactors forbidden errors. Before : `Error: forbidden`

Example of one of them now:
```
You do not have privileges to disconnect providers from infra; contact your admin

Run `infra info` for more information about your session
```

## Checklist

- [ ] Wrote appropriate unit tests - **In progress**

## Related Issues

Resolves #2096


internal/cmd/destinations.go
```diff
@@ -81,6 +81,7 @@ func newDestinationsRemoveCmd(cli *CLI) *cobra.Command {
 		Example: "$ infra destinations remove docker-desktop",
 		Args:    ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			forbiddenMsg := "You do not have privileges to disconnect destinations from infra; contact your admin\n\nRun `infra info` for more information about your session"
```
mxyng, line 84: This seems unnecessarily repetitive. Why not handle it once in `main.go`?

mxyng: Also suggest to avoid abbreviations. The latter part of the message also seems unnecessary

```
You do not have privileges for this action. Contact your administrator.
```

ssoroka: i have a different problem with this. the message says "disconnect providers", but it's used everywhere for all kinds of endpoints. we probably want a unique message per endpoint

ssoroka (changes requested):
forbiddenMsg reuse is causing bad client messages. don't use variables for this.

internal/cmd/destinations.go
kimskimchi: updated

internal/cmd/destinations.go
```diff
@@ -90,6 +90,11 @@ func newDestinationsRemoveCmd(cli *CLI) *cobra.Command {
 			logging.S.Debugf("call server: list destinations named %q", name)
 			destinations, err := client.ListDestinations(api.ListDestinationsRequest{Name: name})
 			if err != nil {
+				if api.ErrorStatusCode(err) == 403 {
+					return Error{
+						Message: "Cannot disconnect destination: missing privileges for ListDestinations",
+					}
+				}
```
ssoroka, line 97: if you're going to replace the existing error like this, you need to debug log the original error.

dnephin: Isn't the original error always just `Forbidden` , since we hide everything in `sendAPIError` ?

internal/cmd/grants.go
```diff
@@ -174,6 +180,12 @@ func removeGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 
 	user, group, err := checkUserGroup(client, cmdOptions.Name, cmdOptions.IsGroup)
 	if err != nil {
+		var cliError Error
+		if errors.As(err, &cliError) {
+			return Error{
+				Message: fmt.Sprintf("Cannot revoke grants: %s", cliError.Message),
```
dnephin, line 186: I don't think I understand why we only wrap the error with additional context if it's already a `cmd.Error`. Wouldn't we want to add this context regardless of the underlying error?

dnephin (commented):
These errors look great, but I wonder why we wouldn't provide these nice errors from the API.

It seems like it would be easier to maintain from the API since the permission check is there, so we could change the errors along with the permission check.  Not something we need to address in this PR, but maybe something for us to consider later.

internal/cmd/destinations.go
kimskimchi: done

kimskimchi:
> I created #2312 to discuss possibly having the API return better errors for authorization failures.

Thanks. If API is agreed upon to return more information, we can easily update these lines with 
`You do not have enough privileges: <api-error-msg>; contact your admin`

ssoroka (approved):

PR #2215, opened 2022-06-07
fix: make testuserscmd more resilient

pdevine:
## Summary

We were pulling the user name out of the URL path with RawQuery which can change if `ListUsers()` ever gets modified (which it will, and it's really hard to debug this when it does). This change just pulls the name directly out of the `URL.Query()`

## Checklist


kimskimchi (approved):
Ahh, nice. Thanks!

PR #2217, opened 2022-06-07
fix: destination CA field encoding

dnephin:
This is an example of the proposal described here: https://github.com/infrahq/infra/pull/2209#pullrequestreview-998683828

Needs more tests

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -119,3 +119,8 @@ func NewListResponse[T, M any](items []M, fn func(item M) T) *ListResponse[T] {
 
 	return result
 }
+
+// PEM is a base64 encoded string, commonly used to store certificates and
+// private keys. PEM values will be normalized to remove any leading whitespace
+// and all but a single trailing newline.
+type PEM string
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -335,6 +335,7 @@ func setTypeInfo(t reflect.Type, schema *openapi3.Schema) {
 	case reflect.Slice:
 		if t.Elem().Kind() == reflect.Uint8 {
 			schema.Type = "string" // []byte
+			schema.Format = "base64"
 			return
 		}
 		schema.Type = "array"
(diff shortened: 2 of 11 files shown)


internal/server/openapi.go
```diff
@@ -335,6 +335,7 @@ func setTypeInfo(t reflect.Type, schema *openapi3.Schema) {
 	case reflect.Slice:
 		if t.Elem().Kind() == reflect.Uint8 {
 			schema.Type = "string" // []byte
```
BruceMacD, line 337: Is this needed now that we have the PEM instead of the byte array?

BruceMacD (approved):
This seems like a preferable solution to me

api/types.go
```diff
@@ -119,3 +119,15 @@ func NewListResponse[T, M any](items []M, fn func(item M) T) *ListResponse[T] {
...
+// private keys. PEM values will be normalized to remove any leading whitespace
+// and all but a single trailing newline.
+type PEM string
+
+// TODO: tests
+func (p *PEM) UnmarshalText(raw []byte) error {
+	v := string(raw)
+	*p = PEM(strings.TrimSpace(v) + "\n")
+	return nil
+}
```
dnephin, line 133: We could remove this for now, and introduce it later. It's not required to fix the problem we have now. 

If we did that I think the only test we'd be missing is one for the create destination endpoint to show that it doesn't trim whitespace for the `CA` field.

internal/server/openapi.go
dnephin: Not needed anymore, but given that this is how `encoding/json` deals with `[]byte` it seems like it doesn't hurt too much to keep it this way in case we add other `[]byte` fields in the future.

dnephin:
I removed the `UnmarshalText` method for the `api.PEM` type, and added a test for `CreateDestination`.

Making this PR as ready-for-review now.

BruceMacD:
This doesn't seem to handle the case of a leading space in the certificate, but otherwise it looks good:
```
" -----BEGIN CERTIFICATE-----\nok\n-----END CERTIFICATE-----\n"
```

PR #2218, opened 2022-06-07
improve: added pagination model and selector for future use

j-sneh:
## Summary

Created the PaginationResponse, PaginationRequest, and internal Pagination model as an incremental change toward fully-integrated pagination.

## Checklist

~~- [ ] Wrote appropriate unit tests~~
~~- [ ] Updated associated docs where necessary~~
~~- [ ] Updated associated configuration where necessary~~


internal/server/models/pagination.go
```diff
@@ -0,0 +1,83 @@
...
+	"strings"
+
+	"github.com/infrahq/infra/api"
+)
+
+// Internal Pagination Data
+type Pagination struct {
+	Page    int
+	Limit   int
+	Sort    string
```
jmorganca, line 14: Should this change include `Sort` ?

internal/server/models/pagination.go
```diff
@@ -0,0 +1,83 @@
...
+func (p *Pagination) SetCount(count int64) {
+	p.MaxPage = int(math.Ceil(float64(count) / float64(p.Limit)))
+	p.Total = int(count)
+
+	if p.Next > p.MaxPage {
+		p.Next = 0
+	}
+}
+
+func (p *Pagination) SetDefaultSort(sort string) {
```
jmorganca, line 79: Similar to above: do we want to include for this change, or in a future change?

internal/server/models/pagination.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/infrahq/infra/api"
+)
+
+// Internal Pagination Data
+type Pagination struct {
+	Page    int
+	Limit   int
+	Sort    string
+	Total   int
+	MaxPage int
```
jmorganca, line 16: Is this the total # of pages? We could use `Pages` instead

internal/server/models/pagination.go
j-sneh: Why push off sorting to a future change?

BruceMacD (commented):
Good step 👍

internal/server/data/selectors.go
```diff
@@ -118,6 +119,18 @@ func ByUserID(userID uid.ID) SelectorFunc {
 	}
 }
 
+func ByPagination(pg models.Pagination) SelectorFunc {
```
jmorganca, line 122: Would it be possible to add a few tests for this selector?

internal/server/models/pagination.go
```diff
@@ -0,0 +1,84 @@
...
+		Page:  page,
+		Limit: limit,
+		Sort:  sort,
+		Prev:  prevPage,
+		Next:  page + 1,
+	}
+}
+
+// PaginationToResponse converts the internal Pagination model to a response sent to the user
+func (p *Pagination) PaginationToResponse() api.PaginationResponse {
```
jmorganca, line 55: Similar to above comment: consider this in a follow up PR

internal/server/models/pagination.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Internal Pagination Data
+type Pagination struct {
+	Page  int
+	Limit int
+	Sort  string
+	Total int
+	Pages int
+	Next  int
+	Prev  int
```
mxyng, line 18: What are the use cases for these fields?

j-sneh: If there is no next or previous page then the field would be empty, other than that it's just a ±1. I guess they're not very necessary

internal/server/models/pagination.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+
+	"github.com/infrahq/infra/api"
+)
+
+// Internal Pagination Data
+type Pagination struct {
+	Page  int
+	Limit int
+	Sort  string
```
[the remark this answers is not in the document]
j-sneh: I think we should stick with the sorting. It's already working in the data layer, it's relevant to pagination and quite simple (one call to db.Order and some string validation).

BruceMacD (approved):
This looks good to me once other people follow up on their feedback. Great work.

internal/server/models/pagination.go
```diff
@@ -0,0 +1,26 @@
...
+	"math"
+)
+
+// Internal Pagination Data
+type Pagination struct {
+	Page  int
+	Limit int
+	Sort  string
+
+	TotalCount int
```
jmorganca, line 13: Do we need these fields yet (or the functions below)? It doesn't seem they are called + tested anywhere, and I'd be weary of needing to have "calculated" state. We already have `Count` in our `ListResponse` type.

internal/server/models/pagination.go
```diff
@@ -0,0 +1,26 @@
...
+	TotalPages int
+}
+
+func (p *Pagination) SetCount(count int64) {
```
jmorganca, line 17: Similar to above comment – having to "calculate" state can be a tough developer experience using this struct

internal/server/models/pagination.go
j-sneh: This is a good point and calculating the total count/pages is less than ideal (querying a whole table twice) and can contribute to some wonky design. We do need some sort of method of communicating to the user how many requests are necessary to get the full list

jmorganca (approved):

PR #2242, opened 2022-06-08
maintain: add user-agent to connector requests

mxyng:
## Summary

- automatically add User-Agent header to all api.Client requests
- refactor api client requests to use a single request func since there's duplication between existing `get`, `list`, and `request`
- update user agent string to include type of client
- update user agent to include api version

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -9,17 +9,20 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"runtime"
 
 	"github.com/ssoroka/slice"
 
 	"github.com/infrahq/infra/uid"
 )
 
-var clientVersion = "0.13.0"
+var apiVersion = "0.13.0"
 
 var ErrTimeout = errors.New("client timed out waiting for response from server")
 
 type Client struct {
+	Name      string
+	Version   string
 	URL       string
 	AccessKey string
 	HTTP      http.Client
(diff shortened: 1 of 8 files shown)


api/client.go
```diff
@@ -176,12 +131,20 @@ func request[Req, Res any](client Client, method string, path string, req *Req)
 	return &res, nil
 }
 
+func get[Res any](client Client, path string) (*Res, error) {
```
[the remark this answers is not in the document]
mxyng: That's a good call out. We talked about merging the two func previously but I didn't get around to doing it here. Should be simple though. I'll add it to the pull request

internal/cmd/cmd.go
```diff
@@ -90,11 +89,9 @@ func apiClient(host string, accessKey string, skipTLSVerify bool) (*api.Client,
 
 	u.Scheme = "https"
 
-	headers := http.Header{}
-	ua := fmt.Sprintf("Infra CLI/%v (%v/%v)", internal.Version, runtime.GOOS, runtime.GOARCH)
-	headers.Add("User-Agent", ua)
-
 	return &api.Client{
+		Name:      "cmd",
```
dnephin, line 93: `CLI` or `cli` seems like a more obvious name than `cmd` for our CLI user agent. This name doesn't need to match the package, right?

dnephin (commented):
Lots of great improvements here

internal/cmd/cmd.go
mxyng: You're probably right. I was going for consistency between the client name and their package, e.g. `internal/connector` is `connector` and `internal/cmd` is `cmd`

internal/cmd/list.go
```diff
@@ -102,12 +107,7 @@ func list(cli *CLI) error {
 	return writeKubeconfig(user, destinations.Items, grants.Items)
 }
 
-func getUserDestinationGrants() (*api.User, *api.ListResponse[api.Destination], *api.ListResponse[api.Grant], error) {
-	client, err := defaultAPIClient()
```
mxyng, line 106: Moving out the client so the cli and agent can report different user agents

PR #2245, opened 2022-06-09
fix: do not warn about the connector identity

### LINKED ISSUE #1878: Admin signup should be an email

BruceMacD:
## Summary
With the users requiring email change the connector loaded in config was logging a warning. For now this is the only identity that we won't want to log this warning for.

Also remove an admin identity name that is no longer used.

## Checklist

## Related Issues

Resolves #1878

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -753,9 +753,11 @@ func (s Server) loadUser(db *gorm.DB, input User) (*models.Identity, error) {
 			return nil, err
 		}
 
-		_, err := mail.ParseAddress(name)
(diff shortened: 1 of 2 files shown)


ssoroka (approved):
what happens when we have connectors with other names? :D

mxyng (approved):

mxyng:
This refers to the connector user. We don't want to output the log if that user is being created since it's not controlled by the user

PR #2246, opened 2022-06-09
fix: specify email fields in the API doc

BruceMacD:
## Summary
When users name became required as email I missed updating the generated API doc from a string field. Update the OpenAPI generation to specify that the name is of format "email".

Screen Shot 2022-06-09 at 3 10 25 PM

## Checklist


ssoroka (approved):

ssoroka:
regenerate the doc before merging `make docs/api/openapi3.json`

PR #2257, opened 2022-06-09
fix: hides expired keys by default

### LINKED ISSUE #2087: `infra keys list` shows expired keys
## Describe the bug

`infra keys list` still shows keys even after they have expired. This list can get really long if there are multiple login requests and so cleanup is necessary.

## Expected behavior

Either completely hiding the expired keys or only showing them with a verbose/all flag. 

## Screenshots

ple com yesterday

## Environment

```
$ infra version

Client: 0.13.0
Server: 0.13.0
```

j-sneh:
## Summary

Hides the listing of expired keys unless explicitly enabled.

## Checklist

## Related Issues

Resolves #2087


api/types.go
```diff
@@ -69,6 +69,10 @@ func (t Time) Equal(other Time) bool {
 	return time.Time(t).Equal(time.Time(other))
 }
 
+func (t Time) Expired(expiry Time) bool {
+	return time.Time(t).Before(time.Time(expiry))
```
BruceMacD, line 73: Does timezone have any effect on this (ie: should this be converting to UTC)?

api/types.go
```diff
@@ -71,6 +71,10 @@ func (t Time) Equal(other Time) bool {
 	return time.Time(t).Equal(time.Time(other))
 }
 
+func (t Time) Expired(expiry Time) bool {
+	return time.Time(t).UTC().Before(time.Time(expiry).UTC())
+}
```
dnephin, line 76: This `api.Time` type could be used for all kinds of things, not necessarily only for expiry times. 

Since this method is only used in tests, I think it would be better to keep this logic in the test rather than expose it as a method.

internal/cmd/keys.go
```diff
@@ -219,6 +221,6 @@ func newKeysListCmd(cli *CLI) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&options.UserName, "user", "", "The name of a user to list access keys for")
-
+	cmd.Flags().BoolVar(&options.ShowExpired, "show-expired", false, "Show expired access keys")
```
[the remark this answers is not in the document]
mxyng: I would call this `--all` instead since it'll list both expired and unexpired grants

internal/server/data/selectors.go
```diff
@@ -119,8 +121,13 @@ func ByUserID(userID uid.ID) SelectorFunc {
 	}
 }
 
-func ByPagination(pg models.Pagination) SelectorFunc {
+func ByNotExpired() SelectorFunc {
+	return func(db *gorm.DB) *gorm.DB {
+		return db.Where("expires_at is null or expires_at > ?", time.Now().UTC()).Or("expires_at is ?", time.Time{})
```
dnephin, line 126: Isn't the `OR` in here twice? Once in the first statement and again because of the `.Or(...)` ?

dnephin (commented):
Looks good, just a couple questions/suggestions

internal/cmd/keys.go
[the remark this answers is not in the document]
dnephin: The risk with `--all` is that in the future if we have other filters it's less obvious what `--all` should do. Should it change to show all the things hidden by default?

Something like `--show-expired` is nice because the behaviour should not need to change even when we default hide other keys.

internal/server/data/selectors.go
j-sneh: these are three different checks: expiration, null, and UTC 0. I can change all of them to `.Or` for consistency.

internal/cmd/keys.go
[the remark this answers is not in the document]
j-sneh: I think I'm going to go with show-expired because it's the most detailed/accurate. Although I'm also a fan of one word flags, this keeps it consistent with the API and we already have some multi-word flags.

BruceMacD (approved):
Looks good 👍

PR #2264, opened 2022-06-10
fix: improve grants list performance

### LINKED ISSUE #2231: CLI: infra grants list is too slow
with ~300 grants infra grants list command can take > 30s to respond.

Sounds like the client is doing too many requests. From the code it looks like it's doing O(2n) queries, which is way too many.

mxyng:
## Summary

Grants list queries the server for each user or group which is exceedingly slow when more than a few users/groups exist. It's also unnecessary in a lot of use cases because users and groups will be duplicated for multiple grants. Instead, retrieve a list of users and groups once and lookup the user when necessary

Old code:

```
$ /usr/bin/time -h infra grants list | grep USER
  USER (7702)        ACCESS         DESTINATION
        20.20s real             1.21s user              0.82s sys
```

New code:

```
$ /usr/bin/time -h infra grants list | grep USER
  USER (7702)        ACCESS         DESTINATION
        0.28s real              0.14s user              0.05s sys
```

Tested against local Infra server with a permutation of ~100 users, 7 roles, and 10 destinations

## Checklist

## Related Issues

Resolves #2231

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/ssoroka/slice"
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal/logging"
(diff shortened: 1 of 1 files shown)


internal/cmd/grants.go
```diff
@@ -132,6 +83,86 @@ func newGrantsListCmd(cli *CLI) *cobra.Command {
 	return cmd
 }
 
+func userGrants(cli *CLI, client *api.Client, grants *api.ListResponse[api.Grant]) (int, error) {
```
mxyng, line 86: would be nice to use generics here since the code is _very_ similar between this and `groupGrants` but the table headers are different so not sure how to genericize it

BruceMacD (approved):
I'm ok with the repetition between functions here, my solution would be to make the rows named types with a shared "append"-like function, but I am not sure if that's really any cleaner.

PR #2276, opened 2022-06-13
maintain: remove unused models and data for certs

dnephin:
## Summary

Follow up to #2174, I found a few more things that are not used and that can be a distraction for contributors trying to understand how infra uses certificates.


pki/certificates_ext_test.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 func TestCertificateSigningWorks(t *testing.T) {
+	t.Skip("persistence not implemented")
```
mxyng, line 25: Any reason why these aren't removed instead?

dnephin: Not really, I've kept the `pki` package for now since it's at least isolated from everything else, but it is also not used. So we could remove the `pki` package as well.

internal/server/data/migrations.go
```diff
@@ -506,8 +506,6 @@ func initializeSchema(db *gorm.DB) error {
 		&models.AccessKey{},
 		&models.Settings{},
 		&models.EncryptionKey{},
-		&models.TrustedCertificate{},
```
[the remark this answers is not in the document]
dnephin: If we expect to create these at some point, I think we can add a migration to remove these tables now.

BruceMacD (approved):

pki/native.go
```diff
@@ -386,82 +381,14 @@ func createCertSignedBy(signer, signee KeyPair, lifetime time.Duration) (*x509.C
...
-	return &models.RootCertificate{
-		KeyAlgorithm:     k.KeyAlgorithm,
-		SigningAlgorithm: k.SigningAlgorithm,
-		PublicKey:        models.Base64(k.PublicKey),
-		PrivateKey:       models.EncryptedAtRest(k.PrivateKey),
-		SignedCert:       models.EncryptedAtRest(b.Bytes),
-		ExpiresAt:        k.SignedCert.NotAfter,
-	}
+	// TODO:
+	return fmt.Errorf("persistence not impemeneted")
```
jmorganca, line 385: nit: typo `impemeneted` -> `implemented`

dnephin:
I fixed those typos in this PR. I worked on some migrations for this as well, but I did some cleanup of the docs at the same time, so I'll submit the migrations as a follow up PR.

PR #2279, opened 2022-06-13
fix: set logLevel from helm chart values

### LINKED ISSUE #2079: Missing way to change log level via `helm` values file
Currently setting the log level via helm values is hard – users need to specify environment variable named `INFRA_LOG_LEVEL`

It would be easier to have a value such as `--set server.config.logLevel=debug` or similar

mxyng:
## Summary

Provide a way for users to set log level using helm values. This change is necessary because `logLevel` is not read from config but retrieved directly from env var or command line parameter

## Checklist

## Related Issues

Resolves #2079

### DIFF
--- a/helm/charts/infra/templates/connector/deployment.yaml
+++ b/helm/charts/infra/templates/connector/deployment.yaml
@@ -31,7 +31,15 @@ spec:
 {{- toYaml .Values.connector.securityContext | nindent 12 }}
           image: "{{ include "connector.image.repository" . }}:{{ include "connector.image.tag" . }}"
           imagePullPolicy: {{ include "connector.image.pullPolicy" . }}
-          args: ["connector", "-f", "/etc/infrahq/infra.yaml"]
+          args:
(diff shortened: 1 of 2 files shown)


helm/charts/infra/templates/connector/deployment.yaml
```diff
@@ -31,7 +31,14 @@ spec:
 {{- toYaml .Values.connector.securityContext | nindent 12 }}
           image: "{{ include "connector.image.repository" . }}:{{ include "connector.image.tag" . }}"
           imagePullPolicy: {{ include "connector.image.pullPolicy" . }}
-          args: ["connector", "-f", "/etc/infrahq/infra.yaml"]
+          args:
```
BruceMacD, line 34: This seems kind of confusing to someone opening the helm chart without the context that we want to be able to set the log level from an external config file. Maybe a comment would help that?

jmorganca (approved):

PR #2283, opened 2022-06-14
maintain: document default connector key lifetime

BruceMacD:
## Summary
It's not clear in the quickstart that the access key you create for the connector will expire in 30 days. Note it at the time of creation.

## Checklist

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -81,7 +81,7 @@ infra login <INFRA_SERVER_HOSTNAME> --skip-tls-verify
 
 ### 3. Connect your first Kubernetes cluster
 
-Generate an access key:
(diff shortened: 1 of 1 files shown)


docs/getting-started/quickstart.md
```diff
@@ -81,7 +81,7 @@ infra login <INFRA_SERVER_HOSTNAME> --skip-tls-verify
 
 ### 3. Connect your first Kubernetes cluster
 
-Generate an access key:
+Generate an access key with a 30 day time-to-live:
```
mxyng, line 84: Alternatively, this can explicitly create an access key with a longer life time. The CLI would need to be updated to access TTL as an option

mxyng (approved):

Issue #2205, opened 2022-06-06
infra grants can


ssoroka:
## Is your feature request related to a problem? Please describe.

Add an `infra grants can` command that lets you query if a user has access to do something.

## Describe the solution you'd like

Here's an example with example output.

```sh
# infra grants can <identity> <privilege> <resource> 
$ infra grants can joe@example.com view production.web
Yes.
joe@example.com has the following grants that grant this access:

  USER                            ACCESS         DESTINATION                     
  joe@example.com                 admin          production                           
  joe@example.com                 cluster-admin  production.web      
  joe@example.com                 view           production.web      

  GROUP     ACCESS  DESTINATION                 
  Everyone  view    production.web
```

Rather than making the CLI do all this work, there should be a dedicated api endpoint, eg `/api/grants/can` that answers this.

PR #2287, opened 2022-06-14
improve: resetting passwords for other users should create a temp password

### LINKED ISSUE #2145: Add ability to reset password for other users via a temporary password
When a user loses their password, another admin user should be able to generate another temporary password for them. Today, admins can only set other users' password to a fixed value.

Example experience (cli):

```
$ infra users edit jeff@infrahq.com --reset-password
```

ssoroka:
## Examples 

```
# in this example, joe@example.com is an existing IDP user, and the current user is logged in as an admin.

$ ./infra users edit joe@example.com --password
Error: existing credential: record not found

# we can't add a password for joe because he's not a local user. Let's add him as a local user

$ ./infra users add joe@example.com
Added user "joe@example.com"
Password: SJXiwhr1Kvcy

$ ./infra users edit joe@example.com --password
  Temporary password for user "joe@example.com" set to: Y*Wv?|Ugv@mX

# logging in as joe, we can set our own password.
$ ./infra login localhost
? Select a login method: Login with username and password
? Username: joe@example.com
? Password: ************
  Your password has expired. Please update your password (min. length 8).
? Password: *************
? Confirm Password: *************
  Updated password.
  Logged in as joe@example.com

# now joe can edit his own password, which will not be temporary.

$ ./infra users edit joe@example.com --password
  Enter a new password (min. length 8):
? Password: ********
? Confirm Password: ********
  Updated password

```

## Checklist

## Related Issues

Resolves #2145

### DIFF
--- a/api/user.go
+++ b/api/user.go
@@ -24,9 +24,9 @@ type ListUsersRequest struct {
 	PaginationRequest
 }
 
+// CreateUserRequest is only for creating users with the Infra provider
 type CreateUserRequest struct {
-	Name               string `json:"name" validate:"email,required"`
-	SetOneTimePassword bool   `json:"setOneTimePassword"`
+	Name string `json:"name" validate:"email,required"`
 }
 
 type CreateUserResponse struct {
--- a/docs/api/openapi3.json
+++ b/docs/api/openapi3.json
@@ -3468,9 +3468,6 @@
                   "name": {
                     "format": "email",
                     "type": "string"
-                  },
-                  "setOneTimePassword": {
-                    "type": "boolean"
                   }
                 },
                 "required": [
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/generate"
 	"github.com/infrahq/infra/internal/logging"
 )
 
@@ -59,7 +60,7 @@ $ infra users add johndoe@example.com`,
 				return err
 			}
 
-			createResp, err := createUser(client, args[0], true)
+			createResp, err := createUser(client, args[0])
 			if err != nil {
 				if api.ErrorStatusCode(err) == 403 {
 					logging.S.Debug(err)
@@ -252,7 +253,10 @@ func updateUser(cli *CLI, name string) error {
 		return err
 	}
 
-	req := &api.UpdateUserRequest{}
+	config, err := currentHostConfig()
+	if err != nil {
+		return err
+	}
 
 	isSelf, err := isUserSelf(name)
 	if err != nil {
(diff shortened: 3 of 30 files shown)


docs/api/openapi3.json
```diff
@@ -3176,8 +3176,11 @@
                     "format": "email",
                     "type": "string"
                   },
-                  "setOneTimePassword": {
-                    "type": "boolean"
```
ssoroka, line 3180: this setting was superfluous. we know whether to generate one based on provider, and there's no harm setting a temp password that is never used.

api/user.go
```diff
@@ -24,8 +24,8 @@ type ListUsersRequest struct {
 }
 
 type CreateUserRequest struct {
-	Name               string `json:"name" validate:"email,required"`
-	SetOneTimePassword bool   `json:"setOneTimePassword"`
+	Name       string `json:"name" validate:"email,required"`
+	ProviderID uid.ID `json:"providerID"` // optional. if missing, assume infra. This is needed for custom IDP implementations
```
jmorganca, line 28: Nice to see `SetOneTimePassword` go.!

I'm confused as to why `ProviderID` is exposed here? Which custom IDP implementations require ProviderID? AFAIK no IDPs call this endpoint directly.

jmorganca (commented):
This is looking good - left some comments as to why we need `ProviderID` in some requests

internal/cmd/users.go
```diff
@@ -225,51 +227,68 @@ func updateUser(cli *CLI, name string) error {
...
+	if !ok {
+		return Error{Message: "No permission to change password for user " + name}
+	}
 
-	logging.S.Debugf("call server: update user %s", req.ID)
-	if _, err := client.UpdateUser(req); err != nil {
+	user, err := getUserByName(client, name)
+	if err != nil {
+		if errors.Is(err, ErrUserNotFound) {
+			logging.S.Debugf("user not found: %s", err)
```
BruceMacD, line 272: This debug log seems repetitive of the actual error

api/user.go
ssoroka: all IDPs will have to know their own providerID from install/setup. It's not possible to look it up if you have multiple IDPs of the same type.
What Bruce said is correct; while we currently have no IDPs that call this, it's impossible to integrate with us without allowing an IDP to call this endpoint, so we should at least check that you intend to create a local Infra user vs something else.

internal/cmd/users.go
ssoroka: we lose the actual root error; thought it might be helpful

internal/cmd/users.go
```diff
@@ -225,51 +227,68 @@ func updateUser(cli *CLI, name string) error {
...
-	if isSelf {
-		cli.Output("  Updated password")
-	} else {
-		cli.Output("  Updated password for %q", name)
+	tmpPassword, err := generate.CryptoRandom(12, generate.CharsetPassword)
+	if err != nil {
+		return err
 	}
 
+	if _, err := client.UpdateUser(&api.UpdateUserRequest{
```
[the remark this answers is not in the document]
ssoroka: Depends: 
1. There is an Infra user for this email: The API call assumes you mean that Infra user (since you can't update non-infra users)
2. There is not an infra user for this email: The API call fails. You can't update a non-local non-infra user. One option is to make this work by assuming you want to create the local infra user, but it seems really weird to do that. I'd rather have them explicitly create the local infra user first, then try to edit it.

api/user.go
jmorganca: Let's remove this `ProviderID` in this PR until we need it? Calling this endpoint should be for creating a local user. Adding SCIM will be a whole other change

ssoroka: SCIM aside, this is still important if you want to hook us up to your company's IDP. It's an optional field, so api consumers can ignore it, but the api is useless unless we already support your choice of IDP. That's not great API design.

internal/cmd/users.go
```diff
@@ -353,12 +368,52 @@ func isUserSelf(name string) (bool, error) {
...
+
+	myGroups, err := client.ListGroups(api.ListGroupsRequest{UserID: config.UserID})
+	if err != nil {
+		return false, err
+	}
+
+	for _, group := range myGroups.Items {
+		grants, err := client.ListGrants(api.ListGrantsRequest{
+			Group:     group.ID,
+			Privilege: models.InfraAdminRole,
```
dnephin, line 406: This is the CLI referencing models. I believe I've seen a few people suggest in PRs that the CLI should not reference models. I think my suggestion above would remove the need for this.

internal/generate/generate.go
```diff
@@ -8,26 +8,33 @@ import (
...
-// CryptoRandom generates a cryptographically-safe random number
-func CryptoRandom(n int) (string, error) {
+// CryptoRandom generates a cryptographically-safe random number. defaults to alphanumeric charset.
+func CryptoRandom(n int, charset ...string) (string, error) {
 	if n <= 0 {
 		return "", nil
 	}
+	useCharset := CharsetAlphaNumeric
+	if len(charset) > 0 {
+		useCharset = charset[0]
```
dnephin, line 27: Using a var args as an option argument is really error prone. The caller has no way to know that all the other arguments are going to be ignored.

I think we should either create two functions (one that accepts the charset and one that does not), or simply require that all callers specify the charset, which would be my preference.

There are only 6 other callers (2 in production code, 4 in tests) that don't pass a charset, so it should be easy to update all of them.

Another option could be to actually use all the charsets by appending them together. That seems fine as well.

internal/generate/generate.go
```diff
@@ -36,16 +43,20 @@ func CryptoRandom(n int) (string, error) {
...
 // not a drain on the entropy pool.
-func MathRandom(n int) string {
+func MathRandom(n int, charset ...string) string {
 	if n <= 0 {
 		return ""
 	}
+	useCharset := CharsetAlphaNumeric
+	if len(charset) > 0 {
+		useCharset = charset[0]
+	}
```
dnephin, line 53: same comment about using varargs as optional args.

dnephin (commented):
The behaviour here makes sense to me. My concerns are primarily with the CLI being too aware of the server's authz model, and doing the password generation in the CLI (which means we'll need to duplicate it in the UI and any other clients).

I believe there are 3 things in this PR:
1. cleanup of the client config fields (which is long overdue!), sadly this will conflict a decent amount with #2177 and any other PRs that add tests for login
2. cleanup of the `CreateUserRequest` fields
3. the behaviour change to resolve https://github.com/infrahq/infra/issues/2145

I suspect it would be easier to review if these were at least 2 (or maybe 3) separate PRs. If it's too much work to split that up I think it's fine to keep it all in this PR, but it will likely increase the time it takes for reviews.

api/user.go
ssoroka: > How would a developer use infra login to login using that provider? Wouldn't infra login assume it was an oidc provider, attempt a login that way, and fail?

Right, good question. Today I suppose they'd still also have to be oidc compatible. If they're OIDC-compatible then JIT user creation will work fine. 

This isn't a user facing thing at the moment, it's really a question of getting the backend architecture right. Regardless of when you expose it, the endpoint and the API are specific to the Infra provider, and the code should reflect that, rather than assume.

internal/cmd/users.go
```diff
@@ -353,12 +368,52 @@ func isUserSelf(name string) (bool, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
 	return user, nil
 }
+
+// check if the user has permissions to reset passwords for another user.
+// This might be handy for customizing error messages
+func hasAccessToChangePasswordsForOtherUsers(client *api.Client, config *ClientHostConfig) (bool, error) {
```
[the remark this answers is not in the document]
ssoroka: Yeah, that's how we had it before. the problem with that is it creates a weird workflow where the client asks for the new password, then errors that the user didn't have access to change it. This addresses that poor user experience.

internal/cmd/users.go
```diff
@@ -225,67 +227,80 @@ func updateUser(cli *CLI, name string) error {
...
+			return Error{Message: fmt.Sprintf("No user named %q", name)}
+		}
 		return err
 	}
 
-	if isSelf {
-		cli.Output("  Updated password")
-	} else {
-		cli.Output("  Updated password for %q", name)
+	tmpPassword, err := generate.CryptoRandom(12, generate.CharsetPassword)
```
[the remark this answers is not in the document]
ssoroka: I like that idea. We don't have a mechanism to update the user with no password and expect the server to set (and return!) one. Seems a bit odd for the server to be returning passwords, but I guess it's okay. let's address that in a separate issue if we still want it; the previous behavior worked this way as well and it's not new to this PR.

internal/cmd/users.go
ssoroka: if we eventually separate them we'll have to break these dependencies and/or copy the definitions they point to. Do we want them in two places right now?

api/user.go
dnephin: Could we remove the ambiguity with a comment? `CreateUserRequest`, and the `CreateUser` endpoint could both say "for the infra provider".

internal/cmd/users.go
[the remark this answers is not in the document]
dnephin: Ah right, makes sense.  I guess doing https://github.com/infrahq/infra/issues/2205 would move more of the logic to the server? That would be much nicer than having to query for all grants.

internal/cmd/users.go
jmorganca: I think that would be best, or to declare special roles in a mutual package like `api` until we can bring it to the server side

jmorganca (approved):
Looks great - thanks for tackling this one.

Small comment RE removing the `models` reference from the `cmd` package for now

internal/cmd/users.go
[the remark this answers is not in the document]
ssoroka: yeah

PR #2288, opened 2022-06-14
cmd/login: add --tls-trusted-cert flag 

dnephin:
## Summary

This PR adds a new `--tls-trusted-cert` flag to `infra login`. The flag can be used to specify either the server TLS certificate, or a CA that was used to create the certificate.

This PR is branched from #2177, which adds most of the support that was needed. That PR will need to merge first.

## Related Issues

Resolves #2177

### DIFF
--- a/docs/reference/cli-reference.md
+++ b/docs/reference/cli-reference.md
@@ -39,12 +39,13 @@ $ infra login --key 1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy
 #### Options
 
 ```
-      --key string                   Login with an access key
-      --no-agent                     Skip starting the Infra agent in the background
-      --non-interactive              Disable all prompts for input
-      --provider string              Login with an identity provider
-      --skip-tls-verify              Skip verifying server TLS certificates
-      --tls-trusted-cert file path   TLS certificate or CA used by the server
+      --key string                       Login with an access key
+      --no-agent                         Skip starting the Infra agent in the background
+      --non-interactive                  Disable all prompts for input
+      --provider string                  Login with an identity provider
+      --skip-tls-verify                  Skip verifying server TLS certificates
+      --tls-trusted-cert filepath        TLS certificate or CA used by the server
+      --tls-trusted-fingerprint string   SHA256 fingerprint of the server TLS certificate
 ```
 
 #### Options inherited from parent commands
--- a/docs/reference/cli-reference.md
+++ b/docs/reference/cli-reference.md
@@ -39,11 +39,12 @@ $ infra login --key 1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy
 #### Options
 
 ```
-      --key string        Login with an access key
-      --no-agent          Skip starting the Infra agent in the background
-      --non-interactive   Disable all prompts for input
-      --provider string   Login with an identity provider
-      --skip-tls-verify   Skip verifying server TLS certificates
+      --key string                   Login with an access key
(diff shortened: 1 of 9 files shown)


docs/reference/cli-reference.md
```diff
@@ -57,11 +57,12 @@ $ infra login --key 1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy
...
-      --no-agent          Skip starting the Infra agent in the background
-      --non-interactive   Disable all prompts for input
-      --provider string   Login with an identity provider
-      --skip-tls-verify   Skip verifying server TLS certificates
+      --key string                   Login with an access key
+      --no-agent                     Skip starting the Infra agent in the background
+      --non-interactive              Disable all prompts for input
+      --provider string              Login with an identity provider
+      --skip-tls-verify              Skip verifying server TLS certificates
+      --tls-trusted-cert file path   TLS certificate or CA used by the server
```
BruceMacD, line 65: I'd like to be able to pass a [certificate fingerprint](https://ldapwiki.com/wiki/Certificate%20Fingerprint) so I don't have to always have a file available.

dnephin: That's definitely another option. I can add another flag for `--tls-trusted-fingerprint`.

internal/cmd/types/file.go
```diff
@@ -0,0 +1,38 @@
+package types
+
+import (
+	"io"
+	"os"
+)
+
+// StringOrFile is a pflag.Value type that can be used to read a value either
+// from the command line flag, or as path to a file.
+// If the value is not an existing filepath, it will be used as the literal
```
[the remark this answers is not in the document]
dnephin: The idea here is to avoid having separate flags for a literal cert or a file which contains the cert.

Calling code should only ever see the final value:
1. `--tls-trusted-cert=cert.pem`, `options.TrustedCertificate` will have the contents of `cert.pem`
2. `--tls-trusted-cert="-----BEGIN CERTIFICATE----...."`,  `options.TrustedCertificate` will have the PEM encoded cert
3. `--tls-trusted-cert=non-existant-file.pem`, error "Failed to read trusted certificates for server"

docs/reference/cli-reference.md
dnephin: Added that flag in the latest commit

internal/cmd/login.go
```diff
@@ -82,6 +83,8 @@ $ infra login --key 1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy`,
 	cmd.Flags().StringVar(&options.AccessKey, "key", "", "Login with an access key")
 	cmd.Flags().StringVar(&options.Provider, "provider", "", "Login with an identity provider")
 	cmd.Flags().BoolVar(&options.SkipTLSVerify, "skip-tls-verify", false, "Skip verifying server TLS certificates")
+	cmd.Flags().Var((*types.StringOrFile)(&options.TrustedCertificate), "tls-trusted-cert", "TLS certificate or CA used by the server")
+	cmd.Flags().Var((*types.StringOrFile)(&options.TrustedFingerprint), "tls-trusted-fingerprint", "SHA256 fingerprint of the server TLS certificate")
```
BruceMacD, line 87: Should the fingerprint just be a string only, rather than a StringOrFile?

BruceMacD (approved):
Looks good with one minor comment

internal/cmd/login.go
dnephin: Could be, I'm not sure if people ever store a fingerprint in a file or if they expect to set it from command line flag

PR #2289, opened 2022-06-14
fix: normalize uids

### LINKED ISSUE #2171: Errors parsings Snowflake IDs that are not in the expected format should be handled
## Describe the bug
If an ID is encountered that is not in the expected base58 encoding this causes an error to be thrown. It has been observed in some cases that this happens unexpectedly, maybe from an older Infra version. 
```
infra login
ERROR  invalid base58: not in canonical form
```

## Expected behavior
This error should generally not be reachable since we control the creation of IDs and should validate this when IDs are generated. If it is reached we should log the error and attempt to continue or recover.

## Environment

```
infra version

 Client: 0.13.2+dev
 Server: disconnected
```

```
$ kubectl version
```

ssoroka:
## Checklist

## Related Issues

Resolves #2171

### DIFF
--- a/uid/snowid_test.go
+++ b/uid/snowid_test.go
@@ -301,6 +301,8 @@ func FuzzID_Parse_RoundTrip_FromString(f *testing.F) {
 	testCases := []string{
 		"self",
 		"abcdefghi",
+		"123456789",
+		"1",
 		"gbtNrmnJkvA",
 		"dbtNrmnJkvA",
 		"btNrmnJkvA",
@@ -308,6 +310,7 @@ func FuzzID_Parse_RoundTrip_FromString(f *testing.F) {
 		"A1111111111",
 		"X1111111111",
 		"JR111111111",
+		"111111111111",
 		"JPwcyDCgEuq",
 	}
 	for _, tc := range testCases {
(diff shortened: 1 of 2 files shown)


uid/snowid_test.go
```diff
@@ -301,8 +301,6 @@ func FuzzID_Parse_RoundTrip_FromString(f *testing.F) {
 	testCases := []string{
 		"self",
 		"abcdefghi",
-		"123456789",
-		"1",
```
dnephin, line 305: These cases will be found pretty quickly again when running in fuzz mode. You can try it out with `-fuzz FuzzID_Parse_RoundTrip_FromString`.

I think we need to keep these, but handle the normalization in the test logic below with something like this at the end:

```go
normalized := strings.TrimLeft(original, "1")
assert.Equal(t, id.String(), normalized, "int64=%d", id)
```

uid/snowid_test.go
```diff
@@ -331,10 +329,31 @@ func FuzzID_Parse_RoundTrip_FromString(f *testing.F) {
 	})
 }
 
+func TestNormalizesLeadingZeroValue(t *testing.T) {
```
dnephin, line 332: I always try to be explicit about which function or method is being tested in the name, to make it easier to find the tests for a particular function

```suggestion
func TestParse_NormalizesLeadingZeroValue(t *testing.T) {
```

dnephin (commented):
Looks good, but I think we need to make one more small changes to the fuzz test to keep it working.

dnephin (approved):
lgtm

PR #2290, opened 2022-06-14
ci: add timing and caching to go steps

dnephin:
Downloading dependencies only takes 5 seconds for me locally (with an empty GOMODCACHE), but it appears to take ~2m30s
for every CI job that runs or builds the project.

This change should give us some idea of why it is so slow in CI.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -23,8 +23,9 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: '1.18'
+          cache: true
+      - run: go mod download -x
       - run: go install gotest.tools/gotestsum@v1.8.0
-      - run: go mod tidy
 
       - name: go test
         run: ~/go/bin/gotestsum -ftestname -- -race ./...
@@ -70,6 +71,7 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: '1.18'
+          cache: true
       - uses: golangci/golangci-lint-action@v3.2.0
 
   helm-lint:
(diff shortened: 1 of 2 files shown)


dnephin:
Interesting, it's not `go mod download` that is slow, it seems to be compiling all the packages that takes some time.

BruceMacD (approved):

dnephin:
Using this


jq 'sort_by(.CmdReal) | reverse | .[] | select(.CmdReal != null) | {Time: (.CmdReal / 1000 / 1000 / 1000), Package}' compile.json | less

dnephin:
Not surprising a lot of the build time is in cgo calling out to gcc.

.github/workflows/pull_request.yml
```diff
@@ -86,6 +89,14 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: '1.18'
+          cache: true
+      - run: go mod download -x
+      - run: go build -debug-actiongraph=compile.json .
+      - uses: actions/upload-artifact@v3
+        with:
+          name: build-log
+          path: compile.json
```
mxyng, line 98: do we still need this?

.github/workflows/pull_request.yml
```diff
@@ -23,8 +23,10 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: '1.18'
+          cache: true
+      - run: go mod download -x
+      - run: go list ./...
```
mxyng, line 28: still need separate steps?

mxyng (approved):

.github/workflows/pull_request.yml
dnephin: I think it might be interesting to leave it for a bit to see how it changes with cache hits. I doesn't seem to slow things down much. We can remove it anytime if we are no longer interested in the timing.

.github/workflows/pull_request.yml
dnephin: I can drop this `go list ./..` stop, I don't think that adds anything. The separate `go mod download -x` step is nice because makes it easy to see how much time is spend downloading dependencies separate from the actual checks below. If things start to slow down we'll be able to see why without changing CI config.

PR #2291, opened 2022-06-15
Add and remove groups

pdevine:
## Summary

This change allows users to add and remove simple groups from the CLI, and also includes a new delete group endpoint.

## Checklist

### DIFF
--- a/docs/api/openapi3.json
+++ b/docs/api/openapi3.json
@@ -2073,6 +2073,90 @@
       }
     },
     "/api/groups/{id}": {
+      "delete": {
+        "description": "DeleteGroup",
+        "operationId": "DeleteGroup",
+        "parameters": [
+          {
+            "example": "4yJ3n3D8E2",
+            "in": "path",
+            "name": "id",
+            "required": true,
+            "schema": {
+              "example": "4yJ3n3D8E2",
+              "format": "uid",
+              "pattern": "[\\da-zA-HJ-NP-Z]{1,11}",
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "400": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            },
+            "description": "Bad Request"
+          },
+          "401": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            },
+            "description": "Unauthorized: Requestor is not authenticated"
+          },
+          "403": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            },
+            "description": "Forbidden: Requestor does not have the right permissions"
+          },
+          "404": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            },
+            "description": "Not Found"
+          },
+          "409": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            },
+            "description": "Duplicate Record"
+          },
+          "default": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/EmptyResponse"
+                }
+              }
+            },
+            "description": "Success"
+          }
+        },
+        "summary": "DeleteGroup",
+        "tags": [
+          "Groups"
+        ]
+      },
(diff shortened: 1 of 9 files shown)


docs/api/openapi3.json
```diff
@@ -2073,6 +2073,90 @@
...
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Error"
+                }
+              }
+            },
+            "description": "Not Found"
+          },
+          "409": {
```
BruceMacD, line 2134: This seems wrong, but it might be something in the generation code. I don't think a delete request should be able to return "duplicate record".

internal/cmd/groups.go
```diff
@@ -107,3 +109,64 @@ func newGroupsListCmd(cli *CLI) *cobra.Command {
...
 	}
 }
+
+func newGroupsAddCmd(cli *CLI) *cobra.Command {
+	return &cobra.Command{
+		Use:   "add GROUP",
+		Short: "Create a group",
+		Args:  ExactArgs(1),
+		Example: `# Create a group
+$ infra groups add 'Engineering'`,
```
BruceMacD, line 119: The quotes here around 'Engineering' are a bit confusing.

internal/server/data/identity.go
```diff
@@ -62,7 +62,10 @@ func AssignIdentityToGroups(db *gorm.DB, user *models.Identity, provider *models
 			}
 		}
 		if !found {
-			group := &models.Group{Name: name}
+			group := &models.Group{
+				Name:              name,
+				CreatedByProvider: true,
```
BruceMacD, line 67: Should we track the provider ID that actually created the group so we can have a full way to trace back to the origin?

internal/server/models/group.go
```diff
@@ -8,8 +8,9 @@ import (
 type Group struct {
 	Model
 
-	Name      string `gorm:"uniqueIndex:idx_groups_name,where:deleted_at is NULL"`
-	CreatedBy uid.ID
+	Name              string `gorm:"uniqueIndex:idx_groups_name,where:deleted_at is NULL"`
+	CreatedBy         uid.ID
```
BruceMacD, line 12: What does `CreatedBy` get set to when a provider creates the group?

pdevine: This is a good catch. I wasn't filling out `CreatedBy` at all, and it also doesn't get filled out when it gets created by the provider either. I think it probably makes the most sense to have two fields though; one specifically w/ the uid of the provider when it's automatically created, and the other one with the uid of the person who manually creates it.

pdevine: We could additionally fill in the `CreatedBy` field for the user who triggered a group to be automatically created by the provider, but I'm not certain how useful that is.

internal/server/data/identity.go
pdevine: Yes, I think that makes sense.

internal/cmd/groups.go
```diff
@@ -107,3 +109,64 @@ func newGroupsListCmd(cli *CLI) *cobra.Command {
...
+				return err
+			}
+			cli.Output("Added group %q", args[0])
+
+			return nil
+		},
+	}
+}
+
+func newGroupsRemoveCmd(cli *CLI) *cobra.Command {
```
mxyng, line 137: `rm` should support `--force` to ignore removing non-existent groups

docs/api/openapi3.json
pdevine: This is an issue in the autogen code, and it's happening against each of the delete endpoints (`DELETE /api/users/:id`, `DELETE /api/access-keys/:id`, `DELETE /api/grants/:id`, etc.

I'll file a separate issue.

internal/server/models/group.go
BruceMacD: I like your suggestion of two separate fields, one to track the up-stream provider the group came from and one for the account that created the group. Having the group "CreatedBy" whichever user first logs in with a given group seems confusing.

internal/cmd/groups.go
```diff
@@ -107,3 +109,70 @@ func newGroupsListCmd(cli *CLI) *cobra.Command {
...
 	}
 }
+
+func newGroupsAddCmd(cli *CLI) *cobra.Command {
+	return &cobra.Command{
+		Use:   "add GROUP",
+		Short: "Create a group",
+		Args:  ExactArgs(1),
+		Example: `# Create a group
+$ infra groups add Engineering`,
```
BruceMacD, line 119: Will group names be case sensitive? The capital 'E' in 'Engineering' implies that possibility to me.

BruceMacD (approved):
This seems like a solid iteration

PR #2294, opened 2022-06-15
maintain: add kind to providers

### LINKED ISSUE #2284: Track Identity Provider Kind
As we support more identity providers we have found that getting additional details about users (such as their groups) is more provider specific. We need to support custom logic for different identity providers in some cases, such as group synchronization.

We will still default to spec compliant OIDC flows, but we will also have custom identity provider specific flows based on IDP kind. To support this a `kind` field will be added as an optional field on identity providers.

BruceMacD:
- add a kind field to providers
- add kind to provider list

## Summary
As we support more identity providers we have found that getting additional details about users (such as their groups) is more provider specific. We need to support custom logic for different identity providers in some cases, such as group synchronization. 

This is the first step to identity provider specific functionality. Adding an optional provider kind that can be specified on the CLI or API. Defaults to general OIDC.

## Checklist

## Related Issues

Resolves #2284


internal/server/handlers_test.go
```diff
@@ -317,113 +317,6 @@ func TestListKeys(t *testing.T) {
 	})
 }
 
-func TestListProviders(t *testing.T) {
```
BruceMacD, line 320: moved to `providers_test.go`

jmorganca:
Possible to update the UI code as well to add the `kind` field?

internal/cmd/providers.go
```diff
@@ -63,11 +63,12 @@ func newProvidersListCmd(cli *CLI) *cobra.Command {
 				type row struct {
 					Name string `header:"NAME"`
 					URL  string `header:"URL"`
+					Kind string `header:"KIND"`
```
mxyng, line 66: nit: this should appear before URL. order should be NAME, KIND, URL

internal/cmd/providers.go
```diff
@@ -151,6 +154,7 @@ $ infra providers add okta --url example.okta.com --client-id 0oa3sz06o6do0muoW5
 	cmd.Flags().StringVar(&opts.URL, "url", "", "Base URL of the domain of the OIDC identity provider (eg. acme.okta.com)")
 	cmd.Flags().StringVar(&opts.ClientID, "client-id", "", "OIDC client ID")
 	cmd.Flags().StringVar(&opts.ClientSecret, "client-secret", "", "OIDC client secret")
+	cmd.Flags().StringVar(&opts.Kind, "kind", "", "The identity provider kind. One of 'oidc, okta, azure, or google'")
```
mxyng, line 157: Should there be a default?

internal/server/config.go
```diff
@@ -533,6 +534,7 @@ func (s Server) loadConfig(config Config) error {
 		// inject internal infra provider
 		config.Providers = append(config.Providers, Provider{
 			Name: models.InternalInfraProviderName,
+			Kind: models.InfraKind.String(),
```
mxyng, line 537: this actually makes any query for or excluding the infra provider much easier since the name no longer matters. instead it can filter on the kind. also should mean removing the static provider name; don't need both name and kind. might be worth a follow up

internal/cmd/providers.go
```diff
@@ -138,6 +140,7 @@ $ infra providers add okta --url example.okta.com --client-id 0oa3sz06o6do0muoW5
 				URL:          opts.URL,
 				ClientID:     opts.ClientID,
 				ClientSecret: opts.ClientSecret,
+				Kind:         opts.Kind,
```
mxyng, line 143: is validation being delegated to the server?

internal/server/config.go
BruceMacD: Created #2304. We will need names for other identity providers though to cover the case of 2 identity providers of the same kind (ex: two Oktas).

internal/cmd/providers.go
BruceMacD: It defaults to 'oidc' on the server-side, but I have set the default now here too for understand-ability.

BruceMacD:
> Possible to update the UI code as well to add the `kind` field?

UI has been updated. I didn't think of the UI dependencies here, my bad!

internal/cmd/providers.go
```diff
@@ -151,6 +154,7 @@ $ infra providers add okta --url example.okta.com --client-id 0oa3sz06o6do0muoW5
 	cmd.Flags().StringVar(&opts.URL, "url", "", "Base URL of the domain of the OIDC identity provider (eg. acme.okta.com)")
 	cmd.Flags().StringVar(&opts.ClientID, "client-id", "", "OIDC client ID")
 	cmd.Flags().StringVar(&opts.ClientSecret, "client-secret", "", "OIDC client secret")
+	cmd.Flags().StringVar(&opts.Kind, "kind", "oidc", "The identity provider kind. One of 'oidc, okta, azure, or google'")
```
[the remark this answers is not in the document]
BruceMacD: I had some "auto-detect" functionality in this change at one point but it seemed overkill. Now that I think about it that auto-detect could be useful for migration. I'll open another issue.

PR #2297, opened 2022-06-15
Replace go-prettytime

dnephin:
go-prettytime uses the wrong suffix when the value is between 1.5 and 2. Replace the library with a modified version of go-units formatting. The go-units implementation is much easier to follow and does not have that bug.

This PR also removes the `api.Time.Relative` method, and moves the implementation to the `cmd` package where it is used. Formatting values is the responsibility of the CLI.

### DIFF
--- a/internal/cmd/testdata/TestKeysListCmd/list_all
+++ b/internal/cmd/testdata/TestKeysListCmd/list_all
@@ -1,4 +1,4 @@
   NAME        ISSUED FOR  CREATED       EXPIRES           EXTENSION DEADLINE  
   front-door  admin       24 hours ago  never             never               
-  side-door   admin       24 hours ago  6 hours from now  tomorrow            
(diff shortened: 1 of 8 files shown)


internal/cmd/testdata/TestKeysListCmd/list_all
```diff
@@ -1,4 +1,4 @@
   NAME        ISSUED FOR  CREATED       EXPIRES           EXTENSION DEADLINE  
   front-door  admin       24 hours ago  never             never               
-  side-door   admin       24 hours ago  6 hours from now  tomorrow            
+  side-door   admin       24 hours ago  6 hours from now  26 hours from now
```
dnephin, line 3: I guess this was another bug with `go-prettytime`. 26 hours from now should not be "tomorrow" since it's more than 24 hours.

mxyng (approved):

jmorganca (approved):
Way better, thanks!

PR #2298, opened 2022-06-16
ci: use unique names for all workflows

dnephin:
## Summary

I've been using the "Actions tab" (https://github.com/infrahq/infra/actions) view to find workflows runs. You'll notice that 3 workflows have the same name ("pull request") which makes it hard to find the right one.

This PR removes `human_pull_request.yml` and moves the one job from that workflow into `pull_request.yml`. We will still skip that job on dependabot PRs.

The third one is renamed to "release pull request", which should make it easier to find the workflows from that view.


dnephin:
Hmm, the workflow did not run, and there is no visible error about why.

I'm not sure if removing one of the files with the duplicate name confused it, or if there is a silent error somewhere else.

dnephin:
Ok, there was an error, I missed a `$`

.github/workflows/human_pull_request.yml
```diff
@@ -1,22 +0,0 @@
-name: pull request
-
-on:
-  pull_request:
```
mxyng, line 4: we can set this to `on: push: branches-ignore: [dependabot/*]` so it doesn't show up at all for dependabot prs

.github/workflows/release_pull_request.yml
```diff
@@ -1,4 +1,4 @@
-name: pull request
+name: release pull request
```
mxyng, line 1: can change this to `release candidate`

.github/workflows/release_pull_request.yml
```diff
@@ -1,4 +1,4 @@
...
 
 on:
   pull_request:
```
mxyng, line 4: similar to the other comment, this can be updated to `on: push: branches: [release-please-*]` to not create any jobs. not critical since I'm changing it in #2300 anyways

.github/workflows/human_pull_request.yml
mxyng: nvm this doesn't work in this case because env vars aren't being set correctly (`GITHUB_BASE_REF`) for the action to work

mxyng (approved):

.github/workflows/human_pull_request.yml
dnephin: Ya, I think we had tried that a while ago. I looked for some way to filter on the PR branch, but the docs say there is none. Has to be an `if`.

mxyng: Unfortunately GitHub Actions is lacking on this front. GitLab allows you to reference both the HEAD ref and the BASE ref for merge requests

PR #2299, opened 2022-06-16
maintain: add website

jmorganca:


docs/api/openapi3.json
```diff
@@ -2215,90 +2215,6 @@
       }
     },
     "/api/groups/{id}": {
-      "delete": {
```
mxyng, line 2218: not sure why this is being removed

FSHA (approved):

docs/api/openapi3.json
jmorganca: Great catch! Thanks!

technovangelist (approved):
this looks like an amazing next step on our road to great docs

PR #2300, opened 2022-06-16
improve: build and release candidate artifacts

### LINKED ISSUE #997: Build artifacts in release candidate branch
Now that we have a continuous release branch, we can create a beta release channel with the incoming changes. This allows internal users to always use and test to-be-released changes while also giving some external users a sneak peek of new features and bug fixes.

- beta channel version should be `<last tag>-<# commits since last tag>`, e.g. if most recent release is 0.5.5, a beta version could be 0.5.5-25 where 25 is the number of commits between v0.5.5 and HEAD
- version should need to comply with package versioning requirements, e.g. helm requires chart version to be semvers
- release branch should build and publish Docker images, helm charts, and other artifacts to their respective repositories

mxyng:
## Summary

Update CI to build and archive release candidate artifacts. Refactor existing build and release jobs into a reusable action which is invoked both on main and in the release branch.

Minor change of internal/version.go to not override Prerelease and Metadata components if they exist in the version string.

Use goreleaser action to build artifacts.

Use docker/metadata-action and docker/build-push-action to build images. The build-push-action comes with caching which should improve build time

## TODO

- [x] update homebrew-tap https://github.com/infrahq/homebrew-tap/pull/1
- [x] update scoop https://github.com/infrahq/scoop/pull/1
- [x] update deb/rpm

The linked update scripts can be implemented as github actions or reusable workflows but that's out of scope of this change

## Checklist

## Related Issues

Resolves #997


.github/workflows/pull_request.yml
```diff
@@ -4,65 +4,6 @@ on:
   pull_request:
 
 jobs:
-  test:
```
dnephin, line 7: It looks like we lost these 2 test jobs

.github/workflows/upload.yml
```diff
@@ -0,0 +1,77 @@
...
+        required: true
+      TELEMETRY_WRITE_KEY:
+        required: true
+
+jobs:
+  upload-binaries:
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.ENVIRONMENT }}
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v1
```
mxyng, line 37: Ideally this uses YAML anchors so it doesn't need to be repeated in `upload-helm` but that's not support currently

.github/workflows/release_candidate.yml
```diff
@@ -0,0 +1,50 @@
...
+    runs-on: ubuntu-latest
+    needs: [candidate-upload]
+    environment: development
+    steps:
+      - run: brew install infrahq/tap/infra
+      - run: |
+          infra login ${{ secrets.INFRA_SERVER }} --key ${{ secrets.INFRA_ACCESS_KEY }}
+          infra use ${{ secrets.INFRA_DESTINATION }}
+      - uses: azure/setup-kubectl@v2.0
+      - run: |
```
mxyng, line 48: This doesn't work at the moment due to #2310

internal/version.go
```diff
@@ -21,17 +19,15 @@ var (
 // This is because release-please keeps this at the released version, and not the upcoming next version.
 // While the next version may not match the patch release, it causes the right behavior for semver version comparisons.
 func FullVersion() string {
-	v, err := semver.NewVersion(Version)
-	if err != nil {
-		panic(fmt.Sprintf("invalid version %v: %v", Version, err))
-	}
-
-	if Metadata == "dev" {
+	v := semver.MustParse(Version)
```
jmorganca, line 22: Not necessary in this PR but we should write some tests for this function

.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - run: |
+          [ "${{ inputs.NIGHTLY }}" = 'true' ] && NIGHTLY=${{ inputs.NIGHTLY }}
+          [ "${{ inputs.LATEST }}" = 'true' ] && LATEST=${{ inputs.LATEST }}
```
jmorganca, line 74: should this just be for `inputs.ENVIRONMENT == production` vs a separate latest input?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+            curl -fs -F package=@$PACKAGE https://${{ secrets.GEMFURY_TOKEN }}@push.fury.io/infrahq/
+          done
+        if: ${{ inputs.ENVIRONMENT == 'production' }}
+      - run: aws s3 sync . s3://${{ secrets.RELEASES_BUCKET }}/infra/v${{ inputs.RELEASE_NAME }} --exclude '*' --include '*.zip' --include '*.deb' --include '*.rpm' --include 'infra-checksums.txt' --acl public-read
+      - uses: actions/download-artifact@v3
+        with:
+          name: helm
+      - run: aws s3 sync . s3://${{ secrets.HELM_BUCKET }} --exclude '*' --include 'index.yaml' --include '*.tgz' --acl public-read
+      - run: |
+          for REPO in homebrew-tap scoop; do
```
jmorganca, line 58: Deos this provide benefits to sticking with `goreleaser` for publishing `homebrew-tap` and `scoop` updates (not sure which options it has for publishing pre-built binaries)? Thinking from a maintainability standpoint on bash scripts + the multi-repo action events

.github/workflows/publish.yml
mxyng: good call. initial revisions didn't include ENVIRONMENT inputs which is why these are here but I don't think that's necessary anymore

.github/workflows/publish.yml
mxyng: This allows more control over how artifacts are published. From what I can find, `goreleaser` doesn't support publishing without building. Coupling the build with publish means it's more difficult to migrate build tools or publishers and constrains  artifacts to whatever goreleaser supports. The release process already implements a custom publisher to archive deb and rpm to Gemfury which is essentially a bash script implemented in yaml. This only takes it a small step further.

The multi-repo action was unnecessary and has been replaced with a checkout of the target repo to invoke the update script. It's a good idea to keep the update script in the target repo since it's tightly coupled with the recipe/manifest.

.github/workflows/release_candidate.yml
BruceMacD: Should we leave this part out for now until #2310 is fixed?

BruceMacD (approved):

.github/workflows/release_candidate.yml
mxyng: For sure. This change exists as a separate commit so it's trivial to pull out

.github/workflows/publish.yml
```diff
@@ -0,0 +1,117 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - uses: actions/download-artifact@v3
+        with:
+          name: binaries
+      - run: |
+          for PACKAGE in *.deb *.rpm; do
+            curl -fs -F package=@$PACKAGE https://${{ secrets.GEMFURY_TOKEN }}@push.fury.io/infrahq/
+          done
+        if: ${{ inputs.ENVIRONMENT == 'production' }}
```
mxyng, line 49: These comparisons are case insensitive

.github/workflows/publish.yml
```diff
@@ -0,0 +1,117 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - id: metadata
+        uses: docker/metadata-action@v3
+        with:
+          images: infrahq/infra
+          flavor: latest=false
+          tags: |
+            type=raw,value=${{ inputs.RELEASE_NAME }}
+            type=raw,value=latest,enable={{is_default_branch}}
```
mxyng, line 103: This should, in theory, tag `latest` if the action is trigger on the main branch

PR #2306, opened 2022-06-16
Fix a couple problems with login tests

dnephin:
**Stop http servers on shutdown**

This shouldn't impact production at all, but when running a server for tests we were not
releasing the ports until the process exited. When running with `-count=`100 this could
consume all ports.

With this change the server is closed before the test ends, which closes the listener
and releases the ports.

**Reduce the chance of a timeout**

When the login test times out the t.Cleanup function could attempt to
close the pty at the same time as the login command running in a
separate goroutine is attempting to read from it.

We can't fix the data race because the survey library does not provide
any way to cancel the prompt. So instead we have to try and prevent
the timeout.

This PR increases the timeout used with expect when the race detector is
enabled.

### DIFF
--- a/internal/cmd/login_test.go
+++ b/internal/cmd/login_test.go
@@ -19,11 +19,12 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/race"
 	"github.com/infrahq/infra/internal/server"
 	"github.com/infrahq/infra/uid"
 )
 
-func TestLoginCmdSignup(t *testing.T) {
+func TestLoginCmd_SetupAdminOnFirstLogin(t *testing.T) {
 	dir := setupEnv(t)
 
 	opts := defaultServerOptions(dir)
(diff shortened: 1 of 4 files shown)


internal/cmd/login_test.go
```diff
@@ -61,8 +62,33 @@ func TestLoginCmdSignup(t *testing.T) {
 		exp.ExpectString(t, "Confirm")
 		exp.Send(t, "password\n")
 		exp.ExpectString(t, "Logged in as")
+	})
+
+	runStep(t, "login updated infra config", func(t *testing.T) {
```
dnephin, line 67: These two steps were moved to another test case in #2280, so I'm restoring them here.

internal/cmd/login_test.go
```diff
@@ -93,21 +119,11 @@ func TestLoginCmd(t *testing.T) {
...
-			return Run(ctx, "login", srv.Addrs.HTTPS.String(), "--skip-tls-verify", "--no-agent", "--key", adminAccessKey)
-		})
-
-		exp := expector{console: console}
-		exp.ExpectString(t, "Logged in as")
-
-		assert.NilError(t, g.Wait())
+		// TODO: remove --skip-tls-verify
+		err := Run(ctx, "login", srv.Addrs.HTTPS.String(), "--skip-tls-verify", "--no-agent", "--key", adminAccessKey)
+		assert.NilError(t, err)
```
dnephin, line 124: Since we are only testing that the agent is not started we don't need the console and goroutine in this case. We can run the full command and check the error.

internal/cmd/login_test.go
```diff
@@ -148,7 +164,7 @@ var cmpClientHostConfig = cmp.Options{
 		cmpPolymorphicIDNotZero),
 	cmp.FilterPath(
 		opt.PathField(ClientHostConfig{}, "Expires"),
-		cmpApiTimeWithThreshold(5*time.Second)),
+		cmpApiTimeWithThreshold(20*time.Second)),
```
dnephin, line 167: Since the expect can take up to 30s I was seeing a few flakes when the config was written more than 5 seconds later. I increased this value to remove the flakes.

jmorganca (approved):

internal/race/norace.go
```diff
@@ -0,0 +1,6 @@
+//go:build !race
```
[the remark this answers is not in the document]
dnephin: Ya, it's used in `login_test.go` above, `|| race.Enabled`

PR #2311, opened 2022-06-16
fix: delete access keys and users when providers are removed

### LINKED ISSUE #2260: Removing provider should revoke all access keys issued using that provider
## Describe the bug
When an identity provider is removed from Infra the access keys issued using that IDP to login should be revoked.

## Expected behavior
When an identity provider is removed from Infra the access keys issued using that IDP should be deleted also.

## Additional context
```
$ infra login localhost
  The authenticity of the host can't be established.
? Are you sure you want to continue? Yes
? Select a login method: microsoft (login.microsoftonline.com)
  Logging in with microsoft...
  Logged in as bruce@example.com

 $ infra providers remove microsoft
Disconnected provider "microsoft" (login.microsoftonline.com)

# session should now be revoked, but its not
$ infra users list                              
  NAME                       LAST SEEN  
  admin@example.com  never      
  bruce@example.com  just now   
  connector                  never 
```

ssoroka:
## Summary

- remove access keys when a provider is removed
- remove provider_users when a provider is removed
- remove users if they have no more provider_users

## Checklist

## Related Issues

Resolves #2260

### DIFF
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/mail"
+	"strings"
 
 	survey "github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/cobra"
@@ -116,6 +117,7 @@ func newUsersListCmd(cli *CLI) *cobra.Command {
 			type row struct {
 				Name       string `header:"Name"`
 				LastSeenAt string `header:"Last Seen"`
+				Providers  string `header:"Provided By"`
 			}
 
 			var rows []row
@@ -138,6 +140,7 @@ func newUsersListCmd(cli *CLI) *cobra.Command {
 					rows = append(rows, row{
 						Name:       user.Name,
 						LastSeenAt: HumanTime(user.LastSeenAt.Time(), "never"),
+						Providers:  strings.Join(user.ProviderNames, ", "),
 					})
 				}
 
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -575,7 +575,8 @@ func (s Server) loadConfig(config Config) error {
 }
 
 func (s Server) loadProviders(db *gorm.DB, providers []Provider) error {
-	keep := make([]uid.ID, 0, len(providers))
+	infraProvider := data.InfraProvider(db)
+	keep := []uid.ID{infraProvider.ID}
 
 	for _, p := range providers {
 		provider, err := s.loadProvider(db, p)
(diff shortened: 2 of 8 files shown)


internal/cmd/users.go
```diff
@@ -138,6 +140,7 @@ func newUsersListCmd(cli *CLI) *cobra.Command {
 					rows = append(rows, row{
 						Name:       user.Name,
 						LastSeenAt: HumanTime(user.LastSeenAt.Time(), "never"),
+						Providers:  strings.Join(user.ProviderNames, ", "),
```
jmorganca, line 143: Does this include the `Infra` provider? We should filter that out

jmorganca (approved):
Great fix! Small comment on hiding the `Infra` provider

jmorganca:
Rebase `main` to hide the Vercel check error (sorry about that)

internal/server/config.go
```diff
@@ -573,7 +573,8 @@ func (s Server) loadConfig(config Config) error {
 }
 
 func (s Server) loadProviders(db *gorm.DB, providers []Provider) error {
-	keep := make([]uid.ID, 0, len(providers))
+	infraProvider := data.InfraProvider(db)
```
BruceMacD, line 576: I think this should have been added before this function was called

jmorganca (approved):

internal/cmd/users.go
ssoroka: I think that kills a lot of the usefulness of this feature. It's super useful to see where the users are coming from, and you wouldn't be able to tell that the user was in both an IDP and Infra without it.

internal/server/data/access_key.go
```diff
@@ -94,10 +95,9 @@ func DeleteAccessKeys(db *gorm.DB, selectors ...SelectorFunc) error {
 		return err
 	}
 
-	ids := make([]uid.ID, 0)
-	for _, k := range toDelete {
-		ids = append(ids, k.ID)
-	}
+	ids := slice.Map[models.AccessKey, uid.ID](toDelete, func(k models.AccessKey) uid.ID {
+		return k.ID
+	})
```
dnephin, line 100: I wouldn't block merging on this, but I definitely feel the old for loop was easier to read than the generic `slice.Map`.  Do we really need this?

internal/server/data/data.go
```diff
@@ -238,6 +238,13 @@ func InfraProvider(db *gorm.DB) *models.Provider {
 	if infraProviderCache == nil {
 		infra, err := get[models.Provider](db, ByName(models.InternalInfraProviderName))
 		if err != nil {
+			if errors.Is(err, internal.ErrNotFound) {
+				p := &models.Provider{Name: models.InternalInfraProviderName}
+				if err := add(db, p); err != nil {
+					logging.S.Panic(err)
+				}
```
dnephin, line 245: Shouldn't this be created at server startup?

PR #2324, opened 2022-06-17
maintain(website): add should-build script for previews

jmorganca:
Stops our release branch from building previews (for now – we may want to have this later for better versioning)

### DIFF
--- a/website/scripts/should-build.sh
+++ b/website/scripts/should-build.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+if [ "$VERCEL_GIT_COMMIT_REF" = "release-please--branches--main" ] ; then
+  # Don't build
+  exit 0
+fi
+
+if [ "$VERCEL_ENV" = "production" ] ; then
+  # Proceed with the build
+  exit 1
+else
+  # only build if website or docs files have changes
+  git diff --quiet HEAD^ HEAD ../docs .
+fi


website/scripts/should-build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
mxyng, line 1: I like to avoid bash specifically whenever possible. Since this doesn't need any bash-only functions, it can use /bin/sh instead, for portability.

You'll need to change the tests to `[ "$VERCEL_GIT_COMMIT_REF" = "..." ]`

website/scripts/should-build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+if [[ "$VERCEL_GIT_COMMIT_REF" == "release-please--branches--main" ]] ; then
+  # Don't build
+  exit 0;
```
mxyng, line 5: This and the `exit 1` below are counterintuitive. I would've expected 0 to build and 1 to not build

jmorganca: Totally fair. I think Vercel modelled this around `git diff`'s return value.

mxyng: Ah ok. That's what I was afraid of. nvm then

website/scripts/should-build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+if [ "$VERCEL_GIT_COMMIT_REF" = "release-please--branches--main" ] ; then
+  # Don't build
+  exit 0;
```
mxyng, line 5: v minor: you don't need the ;

mxyng (approved):

website/scripts/should-build.sh
jmorganca: Ah! I was hoping so. Thanks!

PR #2327, opened 2022-06-17
api: provide more detail in authz errors

### LINKED ISSUE #2312: api: return detailed errors about authz failures
Currently our `sendAPIError` function always returns very unhelpful "forbidden" errors for authorization failures. To work around that problem PR #2213 adds some really nice error messages for "403 Forbidden" responses, but this has a couple disadvantages:

1. none of our other clients (the UI, or third party clients) will benefit from these messages
2. it embeds details about our server authorization system in the client

Hiding the details about an authentication failure is a good idea (we don't want to leak anything about the credentials or existence of accounts), but once authentication is complete we know the user is who they claim to be.

Authorization happens after authentication, so we should be able to provide at least these details about any authz failures without leaking any sensitive information:

* the name of the operation (ex: "create user", "update access key", "list groups")
* the permission that was missing, which right now is either "admin role", or "view role" (we can hide details about the connector role). In the future if we change our authorization system, we only need to change these errors, not the CLI

This change would allow the CLI use to use the API errors more directly, without having to provide a lot of additional context.

dnephin:
## Summary

By providing these details about an authorization failure, the CLI and UI can use the `Message` field directly. The error should include:
* the name of the operation
* the name of the resource
* the role required to perform the operation

This PR changes a bunch of errors from 403 forbidden to 400 bad request. I believe these should be 400 errors because no one is allowed to perform the operation. A 403 error indicates that someone would be allowed to.

TODO:
* [ ] figure out what error to use for `UpdateIdentityInfoFromProvider`
* [ ] convert all the other `access` functions

## Related Issues

Resolves #2312


internal/access/access_key.go
```diff
@@ -20,9 +20,10 @@ func currentAccessKey(c *gin.Context) *models.AccessKey {
 }
 
 func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired bool, pg models.Pagination) ([]models.AccessKey, error) {
-	db, err := RequireInfraRole(c, models.InfraAdminRole, models.InfraViewRole)
+	roles := []string{models.InfraAdminRole, models.InfraViewRole}
+	db, err := RequireInfraRole(c, roles...)
 	if err != nil {
-		return nil, err
+		return nil, HandleAuthErr(err, "access keys", "list", roles...)
```
dnephin, line 26: I think ideally instead of a bunch of separate methods in `access` we would have a single method that accepted a struct, and returned the `AuthorizationError`. And our data access would be either directly in the API handler, or some other function. Something like this:

```go

type Request struct {
    Resource string
    Operation string
    RequiredRoles []string
}

func Authorize(req Request) error {
    ...
    return AuthorizationError{...}
}

```

That change would be much too large to make right now, so we're forced to keep the `RequireInfraRole` and `HandleAuthErr` calls in-sync. That shouldn't be too bad because they should always be just a couple lines apart in the same function.

internal/server/handlers.go
```diff
@@ -329,7 +329,8 @@ func (a *API) CreateToken(c *gin.Context, r *api.EmptyRequest) (*api.CreateToken
 	if access.AuthenticatedIdentity(c) != nil {
 		err := a.UpdateIdentityInfoFromProvider(c)
 		if err != nil {
-			return nil, fmt.Errorf("%w: update ident info from provider: %s", internal.ErrForbidden, err)
+			// TODO: why would this fail? seems like this should be a 5xx error
+			return nil, fmt.Errorf("update ident info from provider: %w", err)
```
dnephin, line 333: From my reading any error here would likely be a runtime error like the database or provider being unavailable because of a network problem or outage. That seems like it should be a 5xx error. Are there other errors that I've missed?

internal/server/handlers.go
```diff
@@ -465,7 +466,7 @@ func (a *API) DeleteGrant(c *gin.Context, r *api.Resource) (*api.EmptyResponse,
 		}
 
 		if len(infraAdminGrants) == 1 {
-			return nil, fmt.Errorf("%w: cannot remove the last infra admin", internal.ErrForbidden)
+			return nil, fmt.Errorf("%w: cannot remove the last infra admin", internal.ErrBadRequest)
```
dnephin, line 469: This error could be a `409 Conflict`, but I think we want to save that for "duplicate entry", so bad request is probably acceptable. Otherwise we'll need to provide more context about the type of conflict error.

internal/access/access_key.go
jmorganca: Love this idea to consolidate the layers here. It would remove quite a bit of bulk.

jmorganca (approved):

jmorganca:
Also: please ignore the outdated Vercel deployment error

BruceMacD (approved):

internal/server/handlers.go
```diff
@@ -329,6 +329,7 @@ func (a *API) CreateToken(c *gin.Context, r *api.EmptyRequest) (*api.CreateToken
 	if access.AuthenticatedIdentity(c) != nil {
 		err := a.UpdateIdentityInfoFromProvider(c)
 		if err != nil {
+			// TODO: why would this fail? seems like this should be a 5xx error
```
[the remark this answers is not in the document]
dnephin: I added the missing timeout in #2356.

I realized this is only the `CreateToken` endpoint, so the CLI never calls this directly. I'm not sure how the kubectl CLI handles the different errors. It likely just echo's back the response body. There will be more work to separate out the "user was deleted" error from other errors, so maybe we can do that in a follow up?

I don't think we can simply say "all other errors are unauthorized", because if the error is the DB not being available, it should still be a 500.

PR #2328, opened 2022-06-17
improve: add default sort on server side

### LINKED ISSUE #1341: API sorting

j-sneh:
## Summary

Changes the sorting for every list response to be: `name ASC` if the resource has a name and `id ASC` otherwise. Also removes some unnecessary client-side sorts

## Checklist

## Related Issues

Resolves #1341

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"reflect"
 	"strings"
 	"unicode"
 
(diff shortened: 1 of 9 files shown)


internal/server/grants_test.go
```diff
@@ -226,6 +227,10 @@ func TestAPI_ListGrants(t *testing.T) {
 						Resource:  "res1",
 					},
 				}
+				// check sort
+				assert.Assert(t, sort.SliceIsSorted(expected, func(i, j int) bool {
```
BruceMacD, line 231: What sets the default sort for grants?

j-sneh: good catch! This is outdated

internal/server/data/data.go
```diff
@@ -111,6 +121,7 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 }
 
 func list[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) ([]T, error) {
+	db = db.Order(getDefaultSortFromType((*T)(nil)))
```
mxyng, line 124: Why not set default sort to ID instead?

ssoroka: some people have asked for name to be the default, which is fine as long as you account for tables that don't have a name field.

ssoroka (approved):
looks great

PR #2348, opened 2022-06-20
maintain(docs): replace key name argument w/ flag

j-sneh:
Replaced outdated references to required key name when using `infra keys add`


jmorganca:
Great catch!

jmorganca (approved):

PR #2352, opened 2022-06-20
fix(website): nav button background on hover

jmorganca:


BruceMacD (approved):
This new transition is pretty subtle, but looks good.

jmorganca:
Thanks! This restores to what it was previously before this bug was introduced

PR #2355, opened 2022-06-20
feat: azure groups

BruceMacD:
- move provider logic to its own package
- move user info check into a general user sync function
- move token check to user sync

## Summary
This change enables calling the Microsoft graph API to synchronize groups in the case of using Azure as an identity provider. It moves the provider logic into its own `provider` package and out of the `authn` package.

This change is notably lacking in tests because the provider code-paths rely so heavily on external APIs. I've opened #2354 for getting better test coverage here, and I'll work on that next. Testing these is probably going to require standing up a local server that responds to the requests as expected.

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -1,7 +1,6 @@
 package access
 
 import (
-	"context"
 	"errors"
 	"fmt"
 
@@ -9,9 +8,9 @@ import (
 
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/logging"
-	"github.com/infrahq/infra/internal/server/authn"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/internal/server/providers"
 	"github.com/infrahq/infra/uid"
 )
 
@@ -146,8 +145,7 @@ func GetContextProviderIdentity(c *gin.Context) (*models.Provider, string, error
 }
 
 // UpdateIdentityInfoFromProvider calls the identity provider used to authenticate this user session to update their current information
-func UpdateIdentityInfoFromProvider(c *gin.Context, oidc authn.OIDC) error {
-	ctx := c.Request.Context()
+func UpdateIdentityInfoFromProvider(c *gin.Context, oidc providers.OIDC) error {
 	// added by the authentication middleware
 	identity := AuthenticatedIdentity(c)
 	if identity == nil {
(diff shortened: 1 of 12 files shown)


internal/access/identity.go
```diff
@@ -157,50 +156,24 @@ func UpdateIdentityInfoFromProvider(c *gin.Context, oidc authn.OIDC) error {
...
-		return err
-	}
-
-	provider, err := data.GetProvider(db, data.ByID(providerUser.ProviderID))
+	provider, err := data.GetProvider(db, data.ByID(accessKey.ProviderID))
 	if err != nil {
 		return fmt.Errorf("user info provider: %w", err)
 	}
 
-	if provider.Name == models.InternalInfraProviderName {
```
BruceMacD, line 170: This removed logic is handled in `SyncProviderUser` now.

internal/server/providers/azure.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return fmt.Errorf("could not get user info from provider: %w", err)
+	}
+
+	newGroups, err := checkMemberOfGraphGroups(string(providerUser.AccessToken))
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return fmt.Errorf("%w: %s", internal.ErrBadGateway, err.Error())
+		} else if errors.Is(err, ErrUnauthorized) {
+			newGroups = &[]string{} // set the groups empty to clear them
```
BruceMacD, line 79: If the Azure client isn't configured to allow getting groups from the graph API clear the user groups, log a warning, and continue.

internal/server/providers/oidc.go
```diff
@@ -0,0 +1,301 @@
...
+	// guard against this by validating the info in the response is what we expect
+	if err := claims.validate(); err != nil {
+		return nil, err
+	}
+
+	return claims, nil
+}
+
+// validate checks if the user info response claims have the information we expect
+func (ic *InfoClaims) validate() error {
```
jmorganca, line 271: Should we use our existing way of validating structs?

BruceMacD: Made the change to use our standard validator pattern.

internal/server/providers/azure.go
```diff
@@ -0,0 +1,145 @@
...
+
+	if err := data.AssignIdentityToGroups(db, user, provider, *newGroups); err != nil {
+		return fmt.Errorf("assign identity to groups: %w", err)
+	}
+
+	return nil
+}
+
+// checkMemberOfGraphGroups calls the Microsoft Graph API to find out what groups a user belongs to
+func checkMemberOfGraphGroups(accessToken string) (*[]string, error) {
```
jmorganca, line 98: Consider returning just a slice – i.e `([]string, error)`

mxyng (approved):

internal/server/providers/oidc.go
```diff
@@ -0,0 +1,291 @@
...
+const oidcProviderRequestTimeout = time.Second * 10
+
+// InfoClaims captures the claims fields from a user-info response that we care about
+type InfoClaims struct {
+	Email  string   `json:"email" validate:"required_without=Name"` // returned by default for Okta user info
+	Groups []string `json:"groups"`
+	Name   string   `json:"name" validate:"required_without=Email"` // returned by default for Azure user info
+}
+
+type OIDC interface {
```
jmorganca, line 31: Not important for this PR, but we may want to rename this type since `oidc` is also a kind of provider

BruceMacD: Agreed, Im going to rename it `OIDCClient` or something similar. I'll cover that in my upcoming test coverage change.

Issue #1793, opened 2022-04-29
testing: use VCR record/playback to test integration with identity providers


dnephin:
Using [VCR testing](https://en.wikipedia.org/wiki/Software_testing#VCR_testing)  likely with https://github.com/dnaeon/go-vcr we can test our integration with different oidc identity providers.

We could setup some test accounts with the different OIDC providers, and record their responses to different requests. Then in our normal CI and local test runs we could use the recorded responses. That should allow our tests to run quickly without having to build a fake implementation that is unlikely to be close enough to the real implementation. It also gives us an easy way to regenerate the canned responses occasionally when we need to (either because we make changes or the provider does).

It looks like both the [oidc](https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#ClientContext) and [oauth2](https://pkg.go.dev/golang.org/x/oauth2#pkg-variables) libraries allow us to inject an http.Client, which is all we need to make this work.

PR #2356, opened 2022-06-20
Use request context for oidc calls

dnephin:
By passing the context along from the API request

1. The `odic` and `oauth2` libraries can make use of an `http.Client` stored in the context values. We can set the `http.Client` in some API middelware. This allows us to log all the request, and possibly do something like #1793
2. if the client disconnects we can also cancel the request to the provider, so that the request can exit instead of continue to block waiting on the provider.

Using a logging transport I noticed we were requesting the `.well-known/openid-configuration` 3 times from the oidc provider as part of a single API login request. The last commit removes one of these calls by refactoring `tokenSource` a bit. I think to remove the second call we would need to create the `oidc.Provider` in `NewOIDC`, instead of doing it in each of the methods.


BruceMacD (approved):
Looks good thanks, I'll make sure that these changes aren't clobbered by #2355

PR #2358, opened 2022-06-21
improve: added links to pagination response

j-sneh:
## Summary

Pagination responses now have three (or less) links returned: Self, Next, and Prev. These are the proper links for the corresponding API to get the current, next, or previous page.

Previous gets hidden if there is no previous link, Next will be hidden in a future update.

## Checklist

### DIFF
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -9,5 +9,8 @@ type PaginationResponse struct {
 	Page  int `json:"page,omitempty"`
 	Limit int `json:"limit,omitempty"`
 
+	Next    string `json:"next,omitempty"`
+	Current string `json:"current,omitempty"`
+	Prev    string `json:"previous,omitempty"`
 	//TODO: add some indication of number of records/pages or if a next page exists
 }
--- a/api/pagination.go
+++ b/api/pagination.go
@@ -9,8 +9,8 @@ type PaginationResponse struct {
 	Page  int `json:"page,omitempty"`
 	Limit int `json:"limit,omitempty"`
 
-	Next    string `json:"next,omitempty"`
-	Current string `json:"current,omitempty"`
-	Prev    string `json:"previous,omitempty"`
+	Next string `json:"next,omitempty"`
+	Self string `json:"self,omitempty"`
+	Prev string `json:"prev,omitempty"`
 	//TODO: add some indication of number of records/pages or if a next page exists
 }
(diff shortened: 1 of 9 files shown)


internal/server/routes.go
```diff
@@ -184,7 +184,7 @@ func add[Req, Res any](a *API, r *gin.RouterGroup, route route[Req, Res]) {
 			a.t.RouteEvent(c, route.path, Properties{"method": strings.ToLower(route.method)})
 		}
 
-		c.JSON(defaultResponseCodeForMethod(route.method), resp)
+		c.PureJSON(defaultResponseCodeForMethod(route.method), resp)
```
j-sneh, line 187: This prevents html special-character escaping.

api/pagination.go
```diff
@@ -9,5 +9,8 @@ type PaginationResponse struct {
 	Page  int `json:"page,omitempty"`
 	Limit int `json:"limit,omitempty"`
 
+	Next string `json:"next,omitempty"`
+	Self string `json:"self,omitempty"`
```
ssoroka, line 13: doesn't seem useful?

internal/server/routes.go
j-sneh: PureJSON actually has another odd side effect of adding a newline character to the created JSON

internal/server/grants_test.go
```diff
@@ -253,7 +254,15 @@ func TestAPI_ListGrants(t *testing.T) {
 					},
 				}
 				assert.DeepEqual(t, grants.Items, expected, cmpAPIGrantShallow)
-				assert.Assert(t, grants.PaginationInfo == api.PaginationResponse{Limit: 2, Page: 2})
+				assert.Assert(t, grants.PaginationInfo.Limit == 2 && grants.PaginationInfo.Page == 2)
+
+				self, _ := url.Parse("http:///api/grants?page=2&limit=2")
+				next, _ := url.Parse("http:///api/grants?page=3&limit=2")
+				prev, _ := url.Parse("http:///api/grants?page=1&limit=2")
```
ssoroka, line 261: note the links probably shouldn't include the domain

internal/server/routes.go
ssoroka: really don't think we can do this

ssoroka (changes requested):
i think we are opening up security problems with this approach.

j-sneh:
Links are convenient and nice to have, but due to the tradeoff between security and functionality, I'm tabling this for now.

PR #2367, opened 2022-06-22
improve: add indications for end page

j-sneh:
## Summary

Users need some sort of indication that there are no more pages and further requests will be empty. To do this, the data layer has to count the number of records before pagination is applied. This is done in `data.list`. As a consequence, the Pagination struct has to be passed down more layers.

Currently, `&models.Pagination{}` (an empty Pagination) is used to return a list response of everything and is a placeholder for what will either be `nil` or a better change (different function call).
## Checklist

## Related Issues

Resolves #


BruceMacD (approved):
Looks good, just make sure to not have the capital letter at the start the commit message when you squash this.

Ie, the linter doesn't like this commit message:
```
improve: TotalCount + TotalPages fields]
```

api/pagination.go
```diff
@@ -9,5 +9,7 @@ type PaginationResponse struct {
 	Page  int `json:"page,omitempty"`
 	Limit int `json:"limit,omitempty"`
 
-	//TODO: add some indication of number of records/pages or if a next page exists
+	TotalCount int  `json:"total_count,omitempty"`
```
jmorganca, line 12: We have a `count` field that is returned by the API today – should that be removed in this change?

Suggestion on cleaner naming:

`count` for this field and `pages` below

api/pagination.go
```diff
@@ -9,5 +9,7 @@ type PaginationResponse struct {
...
+	TotalPages int  `json:"total_pages,omitempty"`
+	HasNext    bool `json:"has_next,omitempty"`
```
jmorganca, line 14: would suggest `hasNext` based on our [API guidelines](https://github.com/infrahq/infra/blob/main/docs/api/guidelines.md#field-naming)

j-sneh: Should we also rename `pagination_info`? It is technically breaking, but very likely unused since pagination is currently optional and buried in the API.

jmorganca: Great catch. Since we haven't published paginationInfo as part of our API docs or referenced we are okay to edit this to `paginationInfo` for consistency

ssoroka (approved):
approved, but see comment. could be a potential problem around paginating list for deletes.

internal/server/data/access_key.go
```diff
@@ -90,7 +90,7 @@ func DeleteAccessKey(db *gorm.DB, id uid.ID) error {
 }
 
 func DeleteAccessKeys(db *gorm.DB, selectors ...SelectorFunc) error {
-	toDelete, err := list[models.AccessKey](db, selectors...)
+	toDelete, err := list[models.AccessKey](db, &models.Pagination{}, selectors...)
```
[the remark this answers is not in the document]
j-sneh: Currently, an empty pagination struct returns all records. This is sort of an intermediary step while I/we transition to sending lists in multiple pages (or all at once internally).

PR #2370, opened 2022-06-23
feat(ui): Improved experience around granting and listing access

### LINKED ISSUE #1840: ui-refactor: switch `select` `option` dropdown to `button` and `a` list
Sharing a connected cluster drop-down is white text on a white background in FireFox
Untitled

Firefox has its own default browser dropdown css, and it is very hard to customize it. So in order to fix this ui bug, we will need to switch `select` & `option` dropdown to a `button` & `a` tag list

### LINKED ISSUE #2060: ui-refactor: `grant` components

### LINKED ISSUE #2063: UI should handle inherited grants from groups
## Describe the bug

UI is missing grants inherited from groups when checking if user has permissions.

User X is part of group A and group A is granted admin access to Infra. X has admin access through the CLI but if they login through the UI, options such as Providers and Users will be greyed out even though they are an admin.

When inspecting a destination, the grants view only shows users who are granted direct access to the destination. Group grants are missing as are the users that are a part of that group.

Similarly when inspecting a user, the grants view only shows the destinations they have been granted direct access to while destinations granted through groups are missing

## Expected behaviour

Group grants are accurately depicted. In the first scenario, users with a group Infra admin grant should have admin privileges in Infra. In the second scenario, either the group or members of the group should be visible in the destination grant view. Lastly, the destination should be visible in the user grants view regardless if the grant is direct or inherited

### LINKED ISSUE #2065: UI should dynamically list roles for each cluster

### LINKED ISSUE #2077: ui: removing one's own access in the UI should also remove instructions on how to access the cluster
Removing one's own access in the UI should also remove instructions on how to access the cluster

Steps to reproduce:
1. Remove access for yourself
2. Cluster access pane at the top should disappear or change to "no access"

Screen Shot 2022-05-26 at 1 44 12 PM

### LINKED ISSUE #2166: Safari form inputs using default system unstyled i.e. dropdown on user role

### LINKED ISSUE #2221: UI dropdowns should be typeahead
When adding access for a user - the infrastructure input should be typeahead

Image

### LINKED ISSUE #2222: UI username should be typeahead
When typing a users name - there needs to be typeahead

Image

### LINKED ISSUE #2236: ui: roles are static/enumerated in the ui and unrecognized roles default to "view"
## Screenshots

Screen Shot 2022-06-08 at 1 32 03 PM

Here, the same email has different "view" roles.

## Additional context

Could be the namespace permissions?

jmorganca:
This change introduces several improvements for the UI around granting and listing access:

* Fixes #2060 by refactoring grants into a hook for fetching data - leaving rendering up the components / product areas
* Fixes #2065 by looking up valid roles in the API and showing them in a list when granting access (see below)
Screen Shot 2022-06-23 at 9 53 55 AM

* Fixes #2221 Adds typeahead lookup when granting access to make sure users are selecting valid users or groups 
Screen Shot 2022-06-23 at 9 53 44 AM

* Fixes #2063 The UI now accounts for group-based grants (including detecting if a user is an admin or not)
* Fixes #2166 and fixes #1840 By switching to a non-native dropdown 
Screen Shot 2022-06-23 at 9 53 59 AM

* Fixes #2077 
* Fixes #2222 
* Fixes #2236


ui/components/grant-form.js
```diff
@@ -0,0 +1,133 @@
+import { useEffect, useState, useRef } from 'react'
+import useSWR, { useSWRConfig } from 'swr'
+import { Combobox } from '@headlessui/react'
+import { CheckIcon } from '@heroicons/react/solid'
+import { PlusIcon } from '@heroicons/react/outline'
+
+import RoleDropdown from './role-dropdown'
+
+export default function ({ resource, roles }) {
```
BruceMacD, line 9: Nit: this component should named rather than an unnamed function.

Ex: `export default function GrantForm`

ui/components/role-dropdown.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (b === 'cluster-admin') {
+      return 1
+    }
+
+    return 0
+  })
+}
+
+export default function ({ resource, role, roles, onChange, remove, direction = 'right' }) {
```
BruceMacD, line 34: nit: another anonymous function which is exported

ui/lib/admin.js
```diff
@@ -1,11 +1,15 @@
 import useSWR from 'swr'
 
+import { useGrants } from './grants'
+
 export function useAdmin () {
```
BruceMacD, line 5: Does this get called every time a page is loaded? Wondering if it should just be cached somewhere.

BruceMacD (commented):
Look good, couple nits. The comments you put in `ui/lib/grants.js` are really nice, we are definitely feeling the need for a `/api/grants/self` endpoint.

BruceMacD (approved):

ui/lib/admin.js
jmorganca: It gets cached by [`swr`](https://swr.vercel.app/)

jmorganca: This is a great catch! It was still revalidating on every page. Fixed

PR #2371, opened 2022-06-23
fix(ui): fix a few edge cases around missing namespaces and no admin access

jmorganca:
This fixes an issue where UI rendering feel erroneous:

1. When a cluster doesn't publish namespaces (e.g. an old connector)
2. When a non-admin user browses namespaces, the sidebar is currently blank, this adds back a (mostly empty) metadata section


BruceMacD (approved):
looks good to me

Issue #296, opened 2021-09-21
security: remove all non-test uses of InsecureSkipVerify


ssoroka:
InsecureSkipVerify is pretty dangerous. It allows someone to [MITM](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) the connection, pretend to be a host or client, and undermine all security.

It should never be a default, and we should probably just not allow it; solving whatever problems that creates.

Places we do this:
- [x] Connector <> Server connection #2485
- [x] CLI <> Server login #1541
- [ ] update the quickstart docs to remove reference to skip-tls-verify #2638

dnephin:
#1541 addresses CLI -> server. That trust can be established by either a prompt, or by allowing the user to specify the cert with a command line flag.

That leaves connector -> server. A command line flag on the connector would let us pass the trusted CA to the connector, instead of using `--skip-tls-verify`.

PR #2372, opened 2022-06-23
improve: add fingerprint lookup to quickstart

dnephin:
Related to #296

Give users a `kubectl` command in the quickstart to look up the fingerprint of the server's TLS certificate.

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -32,6 +32,13 @@ infra login localhost
 ```
 
 {% callout type="info" %}
+You may be prompted to verify the fingerprint of the server's TLS certificate. The
+fingerprint can be found in the server logs:
+
+```
+kubectl logs -l 'app.kubernetes.io/name=infra-server' | grep fingerprint
+```
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/quickstart.md
```diff
@@ -31,6 +31,13 @@ Next, log into your instance of Infra to setup your admin account:
 infra login localhost
 ```
 
+You may be prompted to verify the fingerprint of the server's TLS certificate. The
```
jmorganca, line 34: Move this to the callout below?

dnephin: Ya, that could work. Do you think we should have two different callouts together in the same block, or have two separate callout blocks?

jmorganca: One block may be better to avoid "callout exhaustion"

dnephin: Cool, let's try that. It might be less obvious that there are two different callouts in that block, but easy to change later if we need to.

jmorganca (approved):

PR #2382, opened 2022-06-23
fix(website): break words if necessary to avoid layout issues

jmorganca:
Fixes loooong words in markdown causing layout issues because they don't break


mchiang0610 (approved):
LGTM

PR #2385, opened 2022-06-23
fix(docs): revert quickstart login command

jmorganca:


mchiang0610 (approved):
LGTM

dnephin:
Let's do #2387 instead. AS @jmorganca mentioned, this doesn't fix the problem entirely because without this flag they don't get a prompt to skip verification.

dnephin:
Going to close, since we have the other solution

PR #2392, opened 2022-06-24
fix: db migration from 0.13.3

mxyng:
## Summary

`==` is a valid sqlite operator but not a valid postgresql operator. change it to `=` which is more portable

```
postgres      | 2022-06-24 01:29:55.564 UTC [34] ERROR:  operator does not exist: text == unknown at character 62
postgres      | 2022-06-24 01:29:55.564 UTC [34] HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
postgres      | 2022-06-24 01:29:55.564 UTC [34] STATEMENT:  UPDATE "providers" SET "kind"=$1 WHERE kind IS NULL AND name == $2
postgres      | 2022-06-24 01:29:55.564 UTC [34] ERROR:  current transaction is aborted, commands ignored until end of transaction block
postgres      | 2022-06-24 01:29:55.564 UTC [34] STATEMENT:  UPDATE "providers" SET "kind"=$1 WHERE kind IS NULL
```

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -534,10 +534,8 @@ func addKindToProviders() *gormigrate.Migration {
 			}
 
 			db := tx.Begin()
-			db.Table("providers").Where("kind IS NULL AND name == ?", "infra").
-				Update("kind", models.InfraKind)
-			db.Table("providers").Where("kind IS NULL").
-				Update("kind", models.OktaKind)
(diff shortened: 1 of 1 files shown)


mxyng:
That's likely a visual bug with the mobile app. There's a single change in this PR and nothing related to Docker

PR #2394, opened 2022-06-24
OIDC Provider Tests

### LINKED ISSUE #2354: OIDC testing

BruceMacD:
## Summary
This change adds some comprehensive testing to OIDC provider authentication and information syncing. The testing works by standing up a test server and using that to simulate the responses from various oidc providers.

## Checklist

## Related Issues

Resolves #2354


internal/server/providers/_testdata/test-server-sec.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
BruceMacD, line 1: This looks scary, but its just a dummy private key I made specifically for the test server. It must sign test ID tokens and use HTTPS.

BruceMacD, line 1: This looks alarming but its only used by the test server for signing test ID tokens.

internal/server/providers/oidc_test.go
```diff
@@ -2,16 +2,850 @@ package providers
...
+		}
+	})
+
+	// azure endpoints
+	newMux.HandleFunc("/v1.0/me/memberOf/fail", func(w http.ResponseWriter, req *http.Request) {
+		w.WriteHeader(403)
+	})
+	newMux.HandleFunc("/v1.0/me/memberOf", func(w http.ResponseWriter, req *http.Request) {
+		w.Header().Add("Content-Type", "application/json")
+		_, err := io.WriteString(w, azureGroupResponse)
```
dnephin, line 140: We should probably only add provider specific endpoints when testing that one provider. Otherwise we risk the tests of other providers not being exactly right.

Adding these for only the azure tests should be easy by accepting a func to add additional handlers:

```go
func (t *testing.T, mux *http.ServeMux) {
    ...
}
```

I made those changes in a1dde1bfa8222010e9a6a63052afe7de9cdbf388 if you want to check it out, or cherry-pick it to apply the changes.

internal/server/providers/oidc_test.go
```diff
@@ -2,16 +2,850 @@ package providers
...
+			w.WriteHeader(500)
+		}
+	})
+	newMux.HandleFunc("/token", func(w http.ResponseWriter, req *http.Request) {
+		w.Header().Add("Content-Type", "application/json")
+		w.WriteHeader(ts.tokenResponse.code)
+		_, err := io.WriteString(w, ts.tokenResponse.body)
+		if err != nil {
+			w.WriteHeader(500)
+		}
```
dnephin, line 124: I don't think this works to report an error. If there was already a status code written (2 lines above) the second call to `WriteHeader` will fail and log an error (https://github.com/golang/go/blob/go1.18.3/src/net/http/server.go#L1140-L1143), but that log won't fail the test for us.

In the other cases where we don't write a header, the call to `io.WriteString`, will write a 200 OK header (https://github.com/golang/go/blob/go1.18.3/src/net/http/server.go#L1600-L1601), so before we get an error from writing the body there will always be a header.

I think instead of trying to write a 500 response, we can fail the test:

```go
assert.Check(t, err, "failed to write response")
```

https://pkg.go.dev/gotest.tools/v3/assert#Check uses `t.Error` so it's safe to call from other goroutines (where as `Assert`, and `t.Fatal` are not, so we shouldn't use those here).

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+				assert.Assert(t, groups["Everyone"])
+				assert.Assert(t, groups["Developers"])
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			user := test.setupFunc(t)
+			server.userInfoResponse = test.infoResponse
```
dnephin, line 323: (Not a blocker) Since this modifies the test fixture used by all test cases it might make sense to move `setupOIDCTest` and anything that depends on the server into this run function. It seems like all of that runs pretty fast, so we don't pay much of a penalty for repeating it, and that might prevent surprising cases where the server state is saved between tests.

That would make the `assert.Check` suggestion for errors better, because then the `testing.T` passed to the `server.run` would be for only the test case, not the full test function.

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+		ClientSecret: "whatever",
+		RedirectURL:  "http://localhost:8301",
+	}
+	azure := &azure{OIDC: oidc}
+
+	now := time.Now().UTC()
+
+	claims := jwt.Claims{
+		Audience:  jwt.Audience([]string{server.knownClientID}),
+		NotBefore: jwt.NewNumericDate(now.Add(time.Minute * -5)), // adjust for clock drift
```
dnephin, line 125: Very minor, I generally see durations written as `-5 * time.Minute`, so that it reads from left to right. I noticed a lot of these here are reversed.

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+				"given_name": "Steve",
+				"picture": "https://graph.microsoft.com/v1.0/me/photo/$value"
+			}`,
+			verifyFunc: func(t *testing.T, err error, user *models.Identity) {
+				assert.NilError(t, err)
+
+				pu, err := data.GetProviderUser(db, provider.ID, user.ID)
+				assert.Assert(t, err)
+
+				assert.Assert(t, pu.AccessToken != "aaa")
```
dnephin, line 185: Are there other `ProviderUser` fields that are modified that we should check here? Should `ExpiredAt` be updated as well? `Groups` ? `LastUpdated` ?

Check out `cmpClientHostConfig` for an example of how we can customize a comparison. You can set time thresholds so that checking a `time.Time` doesn't fail if it's only a few seconds different.  That gets passed to `assert.DeepEqual`, so that we can check the whole structure is updated as expected.

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+				pu := &models.ProviderUser{
+					ProviderID: provider.ID,
+					IdentityID: user.ID,
+
+					Email:        user.Name,
+					RedirectURL:  "http://example.com",
+					AccessToken:  models.EncryptedAtRest("aaa"),
+					RefreshToken: models.EncryptedAtRest("bbb"),
+					ExpiresAt:    time.Now().Add(time.Minute * -5),
+					LastUpdate:   time.Now(),
```
dnephin, line 164: Maybe use a `LastUpdated` time in the last so that we can check that it was updated?

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+			infoResponse: `{
+				"sub": "o_aaabbbccc",
+				"sub": "o_aaabbbccc",
+				"name": "Nancy Wheeler",
+				"family_name": "Wheeler",
+				"given_name": "Nancy",
+				"picture": "https://graph.microsoft.com/v1.0/me/photo/$value"
+			}`,
+			verifyFunc: func(t *testing.T, err error, user *models.Identity) {
+				assert.NilError(t, err)
```
dnephin, line 261: Anything else we can check here? We expect the `ProviderUser` record to not have changed?

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+				pu := &models.ProviderUser{
+					ProviderID: provider.ID,
+					IdentityID: user.ID,
+
+					Email:        user.Name,
+					RedirectURL:  "http://example.com",
+					AccessToken:  models.EncryptedAtRest("aaa"), // this is used to fail the groups call, in reality this token should be valid
+					RefreshToken: models.EncryptedAtRest("bbb"),
+					ExpiresAt:    time.Now().Add(time.Minute * -5),
+					LastUpdate:   time.Now(),
```
dnephin, line 244: Should this be expired? And should we use `LastUpdated` in the past so that we can test that it doesn't change?

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+				pu := &models.ProviderUser{
+					ProviderID: provider.ID,
+					IdentityID: user.ID,
+
+					Email:        user.Name,
+					RedirectURL:  "http://example.com",
+					AccessToken:  models.EncryptedAtRest("aaa"), // this is used to fail the groups call, in reality this token should be valid
+					RefreshToken: models.EncryptedAtRest("bbb"),
+					ExpiresAt:    time.Now().Add(time.Minute * -5),
```
dnephin, line 284: Should this be expired?

internal/server/providers/oidc_test.go
```diff
@@ -2,16 +2,850 @@ package providers
...
+
+// testOIDCServer mocks the expected responses from an OIDC provider
+type testOIDCServer struct {
+	userInfoResponse string
+	tokenResponse    tokenResponse
+	signingKey       *rsa.PrivateKey
+
+	knownClientID     string
+	knownClientSecret string
+	knownAuthCode     string // in this test case the auth code can be re-used
```
dnephin, line 42: I believe `knownAuthCode` is unused. Should it be, or can we remove it?

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,328 @@
...
+			"securityEnabled": true,
+			"securityIdentifier": "qqq",
+			"theme": null,
+			"visibility": null,
+			"onPremisesProvisioningErrors": []
+		}
+	]
+}`
+
+func TestSyncAzureProviderUser(t *testing.T) {
```
dnephin, line 98:
```suggestion
func TestAzure_SyncProviderUser(t *testing.T) {
```

This is `Test[StructName]_[MethodName]` convention for naming tests.

internal/server/providers/oidc_test.go
```diff
@@ -2,16 +2,850 @@ package providers
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			server.tokenResponse = test.tokenResponse
+			accessToken, exp, err := provider.RefreshAccessToken(ctx, test.providerUser)
+			test.verifyFunc(t, accessToken, exp, err)
+		})
+	}
+}
+
+func TestSyncProviderUser(t *testing.T) {
```
dnephin, line 609:
```suggestion
func TestOIDC_SyncProviderUser(t *testing.T) {
```

This case is specific to the OIDC implementation, is that right, where as the other one is for azure

dnephin (commented):
Looks great! A few questions, and some small suggestions for making sure the fake behaves well, but generally this looks like great test coverage.

internal/server/providers/azure_test.go
BruceMacD: Good call, this actually caught a bug where this wasn't always set on update

internal/server/providers/azure_test.go
BruceMacD: Fixed now from your example, doing the deep equal compares

dnephin (approved):
Looks great! Still wondering about some additional assertions in error cases, but nothing blocking a merge.

internal/server/providers/azure_test.go
```diff
@@ -0,0 +1,395 @@
...
+				assert.NilError(t, err)
+
+				return user
+			},
+			infoResponse: `{
+				"sub": "o_aaabbbccc",
+				"picture": "https://graph.microsoft.com/v1.0/me/photo/$value"
+			}`,
+			verifyFunc: func(t *testing.T, err error, user *models.Identity) {
+				assert.ErrorContains(t, err, "could not get user info from provider")
```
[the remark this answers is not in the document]
BruceMacD: Any updates to the provider user are handled at a higher level than this test, so this one is ok

PR #2397, opened 2022-06-24
fix: connector service query

### LINKED ISSUE #2365: Connector Endpoint query results are undefined if multiple services exist
## Describe the bug

If multiple connectors exist in the same cluster and namespace, the services query returns multiple matches when queried with label matching `app.infrahq.com/component=connector` but only the first item is returned. When the connector reports to the server its connection details, it can potentially use the connection information for another connector which renders the information useless.

https://github.com/infrahq/infra/blob/main/internal/kubernetes/kubernetes.go#L521-L545=

## Environment

```
$ infra version
```

```
$ kubectl version
```

mxyng:
## Summary

When more than one deployment of infra exists, the service query will possibly return the incorrect service endpoint since the selector it uses is common to all deployments. Instead mount and add the instance label to the selector as well. The instance selector, when deployed with Helm, is the name of the deployment and will be unique between deployments. This will allow the connector to find the right service.

## Checklist

## Related Issues

Resolves #2365

### DIFF
--- a/helm/charts/infra/templates/connector/deployment.yaml
+++ b/helm/charts/infra/templates/connector/deployment.yaml
@@ -45,6 +45,9 @@ spec:
           envFrom:
 {{- include "connector.envFrom" . | nindent 12 }}
           volumeMounts:
+            - name: podinfo
+              mountPath: /etc/podinfo
+              readOnly: true
             - name: conf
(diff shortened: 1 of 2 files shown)


BruceMacD (approved):
Looks good 👍

PR #2398, opened 2022-06-24
server: accept TLS certificate and key

dnephin:
## Summary

Allow a user to specify the certificate and key to use for TLS.

Also fixes a couple bugs:
* the CA we were generating was not a valid CA certificate
* there was a bug in a `cliopts` hook for using a `Set(string) error` function to parse a config field from a string

## Related Issues

This is part 1 of #2176


internal/certs/certs.go
```diff
@@ -162,12 +168,17 @@ func newCA() (*x509.Certificate, crypto.PrivateKey, error) {
 		NotAfter:              time.Now().AddDate(0, 0, 365).UTC(),
 		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment | x509.KeyUsageCertSign,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		IsCA:                  true,
 		BasicConstraintsValid: true,
 	}
 
-	caPrivKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	caBytes, err := x509.CreateCertificate(rand.Reader, ca, ca, &caPrivKey.PublicKey, caPrivKey)
```
dnephin, line 175: Previously we were only creating the `x509.Certificate` manually, which does not give us any way to serialize it (as far as I can tell).

By calling `CreateCertificate` it creates the asn.1 DER bytes we need to serialize the certificate. When doing this I ran into an error which I believe is because we were not setting `IsCA` above.

I left a TODO below about this, but I could not find any other way to allow us to serialize this CA.

internal/server/server.go
```diff
@@ -73,6 +75,16 @@ type UIOptions struct {
 	FS fs.FS `config:"-"`
 }
 
+type TLSOptions struct {
+	// CA is a PEM encoded public certificate for the CA that signed the
```
BruceMacD, line 79: nit: this comment implies that the CA should be a public authority which should not be required.
```suggestion
	// CA is a PEM encoded certificate for the CA that signed the
```

BruceMacD (approved):

internal/server/server.go
```diff
@@ -73,6 +75,16 @@ type UIOptions struct {
...
+	// certificate, or that will be used to generate a certificate if one was
+	// not provided.
+	CA           types.StringOrFile
+	CAPrivateKey string
```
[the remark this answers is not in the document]
dnephin: Good question, it gets used in #2401 which is the follow up to this

PR #2399, opened 2022-06-24
maintain: revert build changes temporarily

mxyng:
## Summary

Reverting to make way for a patch release


BruceMacD (approved):
Good once tests pass

mxyng:
Failures are with the version in openapi3.json. Going to ignore them since fixing it will require changing the reverts

PR #2400, opened 2022-06-24
improve: log authn errors

dnephin:
## Summary

First I audited all the places we use `internal.ErrUnauthorized` to make sure all of them provide additional context. I fixed one error that did not have additional context, and one place where we hid the error (and only logged it with `Debug`).

Then I changed `sendAPIError` to make sure that we log the full error at `Warn` level, along with relevant request details.

Testing this with an automated test is difficult until we do #1609. I tested manually that these messages are at the warn level and contain the relevant context.

### DIFF
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -39,6 +39,8 @@ func sendAPIError(c *gin.Context, err error) {
 		resp.Code = http.StatusUnauthorized
 		// hide the error text, it may contain sensitive information
 		resp.Message = "unauthorized"
+		// log the error at info because it is not in the response
+		log = logging.L.WithOptions(zap.AddCallerSkip(1)).Info
 
 	case errors.As(err, &authzError):
 		resp.Code = http.StatusForbidden
@@ -81,7 +83,12 @@ func sendAPIError(c *gin.Context, err error) {
 		log = logging.L.WithOptions(zap.AddCallerSkip(1)).Error
 	}
 
-	log("api request error", zap.Error(err), zap.Int32("statusCode", resp.Code))
+	log("api request error",
+		zap.String("method", c.Request.Method),
+		zap.String("path", c.Request.URL.Path),
+		zap.Int32("statusCode", resp.Code),
+		zap.String("remoteAddr", c.Request.RemoteAddr),
+		zap.Error(err))
 
 	c.JSON(int(resp.Code), resp)
 	c.Abort()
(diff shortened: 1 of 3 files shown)


internal/server/handlers.go
```diff
@@ -565,9 +565,8 @@ func (a *API) Login(c *gin.Context, r *api.LoginRequest) (*api.LoginResponse, er
 			// this means an external request failed, probably to an IDP
 			return nil, err
 		}
-		logging.S.Debug(err)
```
dnephin, line 568: Removed this so that we log all `ErrUnauthorized` in a consistent way. At `Warn` level, from `sendAPIError` with relevant request details.

docs/api/openapi3.json
```diff
@@ -785,7 +785,7 @@
       "url": "https://www.elastic.co/licensing/elastic-license"
     },
     "title": "Infra API",
-    "version": "0.13.4"
+    "version": "0.13.5+dev"
```
dnephin, line 788: We still have this problem, where the main branch has the wrong version after a release, so any PRs have to fix this post release.

internal/server/errors.go
```diff
@@ -39,6 +39,8 @@ func sendAPIError(c *gin.Context, err error) {
 		resp.Code = http.StatusUnauthorized
 		// hide the error text, it may contain sensitive information
 		resp.Message = "unauthorized"
+		// log the error at warn, since it is not in the response
+		log = logging.L.WithOptions(zap.AddCallerSkip(1)).Warn
```
BruceMacD, line 43: Seems more like an info log than a warn log?

BruceMacD (approved):

internal/server/errors.go
dnephin: Ya, info would probably be fine too. My thinking with warning is that someone who wants to hide the request logs that are at INFO level might set their logging to WARN, and then lose the ability to debug any authn errors.  But maybe that's an unlikely scenario. Let's see if anyone else has thoughts on warn vs info.

BruceMacD: I would personally prefer info for the reason that any outside party can trigger this log by attempting to login with invalid an username/password, which could flood to logs.

ssoroka (approved):

internal/server/errors.go
dnephin: I changed it to INFO. It might be interesting to see if we can combine this logging of the error with the request log. I think they are effectively logging the same event, so it would be cool if they were fields on the same log line. I'll probably look into that as a follow up.

Issue #2362, opened 2022-06-22
server: improve TLS certificate generation


dnephin:
Today we have a few problems with the TLS certificate generation in the server:
* #2067
* we throw away the CA certificate and key, so each certificate is generated with a different CA, and we lose both the CA and certificate on restart

### Proposal

1. Generate the CA in helm, and save it to a kubernetes secret. Include that secret in the server deployment so the server can access it. Depends on #2176 so that we can tell the server about the path to the CA from the helm chart.
2. On first request, if we don't have a certificate, generate one using the CA. We should use `x509.Ed25519` instead of RSA to reduce the generation time. We also need to add some kind of synchronization to solve #2067.
3. In the CLI, use `tls.Config.VerifyPeerCertificate` (instead of the cert we get back from `x509.UnknownAuthorityError`) to prompt for trust of the CA (instead of prompting for trust of the server certificate). Since the server certificate may change, but the CA remains across restarts, we need to trust the CA instead of the leaf cert.

### Related issues

* #2176 - to allow the user to specify their own CA and certificates

PR #2401, opened 2022-06-24
tls: generate server TLS cert from CA

### LINKED ISSUE #2067: SelfSignedOrLetsEncryptCert can race with itself, creating multiple primary keys and certs
`SelfSignedOrLetsEncryptCert` is used as the `TLSConfig.GetCertificate` function, which is called by the `http.Server` when it receives a TLS request. It will create a primary key for the CA, and a primary key for the cert, which can take 2-3 seconds. During this time, if another request is receives the cache will still be empty, so the next request will also create another primary key and cert. Whichever generation ends last will end up being the cert and key that gets used.

I don't think this is a security problem right now, but it could be surprising if someone is making multiple requests immediately after a new install is started.

We should either generate the certificate before starting the HTTP server (my preference), or put a lock around the critical section to prevent it from running twice.

### LINKED ISSUE #2176: server: allow user to provide their own TLS cert and key

dnephin:
## Summary

The first commit moves some functions to `tls.go`, so best viewed by individual commit.

With the changes in this PR, a user has 3 options for configuring TLS:

1. set `tls.ACME=true`, to use ACME to create a publicly trusted certificate
2. set `tls.ca` and `tls.caPrivateKey`, which will be used to generate a TLS ceritifcate for any hostname the client requests.
3. set `tls.certificate` and `tls.privateKey` to use that certificate for TLS

We plan on setting a CA as part of the helm deploy, so the default for any user that uses our helm charts will be to specify nothing and use option 2.

ACME is no longer the default option. It has to be explicitly enabled, which is probably better because currently every single TLS request could make an ACME API request, even when it should never be used.

Resolves #2067
And addresses parts 1 and 2 of #2362

TODO:
* [x] fix some CLI tests that are failing due to missing CA option
* [x] add test case for generating certs from a CA
* [x] update helm to generate the CA as part of deploy

### DIFF
--- a/helm/charts/infra/templates/server/deployment.yaml
+++ b/helm/charts/infra/templates/server/deployment.yaml
@@ -48,6 +48,10 @@ spec:
             - name: conf
               mountPath: /etc/infrahq
               readOnly: true
+{{- if (not .Values.server.config.tls) }}
+            - name: tls-ca
+              mountPath: /var/run/secrets/infrahq.com/tls-ca
+{{- end }}
 {{- if .Values.server.persistence.enabled }}
             - name: data
               mountPath: /var/lib/infrahq/server
@@ -89,6 +93,11 @@ spec:
         - name: conf
           configMap:
             name: {{ include "server.fullname" . }}
+{{- if (not .Values.server.config.tls) }}
+        - name: tls-ca
+          secret:
+            secretName: {{ include "server.fullname" . }}-ca
+{{- end }}
 {{- if .Values.server.persistence.enabled }}
         - name: data
           persistentVolumeClaim:
--- a/helm/charts/infra/templates/server/secret.yaml
+++ b/helm/charts/infra/templates/server/secret.yaml
@@ -1,0 +1,29 @@
+
+{{- if include "server.enabled" . | eq "true" }}
+{{- if (not .Values.server.config.tls) }}
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "server.fullname" . }}-ca
+  labels:
+{{- include "server.labels" . | nindent 4 }}
+data:
+
+{{- $secret := lookup "v1" "Secret" .Release.Namespace (printf "%s-ca" (include "server.fullname" .)) -}}
+{{- if $secret.data }}
+  ca.crt: |
+{{- get $secret.data "ca.crt" | nindent 4 }}
+  ca.key: |
+{{- get $secret.data "ca.key" | nindent 4 }}
+
+{{- else }}
+{{- $ca := genCA "Infra Server" 3650 }}
+  ca.crt: |
+{{- $ca.Cert | b64enc | nindent 4 }}
+  ca.key: |
+{{- $ca.Key | b64enc | nindent 4 }}
+
+{{- end }}{{/* if secret.data */}}
+{{- end }}{{/* if not tls */}}
+{{- end }}{{/* if server.enabled */}}
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -531,6 +531,19 @@ server:
     # - name: dev@example.com
     #   password: file:/var/run/secrets/dev@example.com
 
+    # TLS configuration for the API server. Defaults to generating a self-signed CA and
+    # generating certificates from that CA.
+    tls: {}
+
+    # Configure a CA and private key using files
+    # ca: /path/to/ca.crt
+    # caPrivateKey: file:/path/to/ca.key
+
+    # Configure a TLS certificate and private key using files
+    # certificate: /path/to/server.crt
+    # privateKey: file:/path/to/server.key
+
+
 ## Default connector configurations
 connector:
 
--- a/internal/certs/certs.go
+++ b/internal/certs/certs.go
@@ -5,7 +5,6 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/sha256"
-	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
@@ -14,10 +13,6 @@ import (
 	"net"
 	"strings"
 	"time"
-
-	"golang.org/x/crypto/acme/autocert"
-
-	"github.com/infrahq/infra/internal/logging"
 )
 
 func GenerateCertificate(hosts []string, caCert *x509.Certificate, caKey crypto.PrivateKey) (certPEM []byte, keyPEM []byte, err error) {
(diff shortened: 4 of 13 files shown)


internal/certs/certs.go
```diff
@@ -63,64 +57,6 @@ func GenerateCertificate(hosts []string, caCert *x509.Certificate, caKey crypto.
 	return PEMEncodeCertificate(certBytes), keyBytes, nil
 }
 
-func SelfSignedOrLetsEncryptCert(manager *autocert.Manager) func(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
dnephin, line 66: I moved this into `internal/server/tls.go` and renamed it to `getCertificate`. That's the only place it is used.

internal/certs/certs.go
```diff
@@ -146,39 +77,3 @@ func PEMEncodeCertificate(raw []byte) []byte {
 func pemEncodePrivateKey(raw []byte) []byte {
 	return pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: raw})
 }
-
-func newCA() (*x509.Certificate, *rsa.PrivateKey, error) {
```
dnephin, line 150: No longer used, we'll rely on helm or the user to provide the CA.

internal/server/server.go
```diff
@@ -299,63 +299,6 @@ type routine struct {
 	stop func()
 }
 
-func tlsConfigFromOptions(
```
dnephin, line 302: Moved to `internal/server/tls.go`

internal/server/tls.go
```diff
@@ -0,0 +1,170 @@
...
+	if opts.CA == "" || opts.CAPrivateKey == "" {
+		return nil, fmt.Errorf("either a TLS certificate and key or a TLS CA and key is required")
+	}
+
+	cfg.GetCertificate = getCertificate(autocert.DirCache(tlsCacheDir), opts)
+	return cfg, nil
+}
+
+func getCertificate(cache autocert.Cache, opts TLSOptions) func(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	var lock sync.RWMutex
```
dnephin, line 89: This lock should fix #2067

internal/certs/certs.go
dnephin: Right, good question. I added another TODO to the list in the description. We need to generate the CA in helm as part of the deploy. That should make it invisible to the user. They won't have to do anything new, but helm will take care of creating the secret which contains the CA and CA private key.

internal/server/tls.go
```diff
@@ -50,82 +73,86 @@ func tlsConfigFromOptions(
...
-			}
+		caCert, err := x509.ParseCertificate(ca.Certificate[0])
+		if err != nil {
+			return nil, err
+		}
 
-			if err := manager.Cache.Put(ctx, serverName+".crt", certBytes); err != nil {
-				return nil, err
-			}
+		hosts := []string{"127.0.0.1", "::1", serverName}
```
dnephin, line 138: I change this to always include localhost in the list of valid IP SANs since it's common for a cert to be used on localhost as well.

internal/certs/certs.go
dnephin: I've pushed a commit which does this (generates the CA in helm and stores it in a secret)

internal/server/tls.go
```diff
@@ -26,21 +23,47 @@ func tlsConfigFromOptions(
 	tlsCacheDir string,
 	opts TLSOptions,
 ) (*tls.Config, error) {
-	// TODO: print CA fingerprint when the client can trust that fingerprint
+	// TODO: how can we test this?
```
dnephin, line 26: Not really sure how to test the ACME integration, but it wasn't tested before and it's still not documented as a supported feature, so probably not a requirement for this PR.

dnephin (commented):
This PR should be ready for review now

helm/charts/infra/templates/server/secret.yaml
```diff
@@ -0,0 +1,29 @@
...
+stringData:
+
+{{- if $secret.data }}
+  ca.crt: |
+{{- get $secret.data "ca.crt" | b64dec | nindent 4 }}
+  ca.key: |
+{{- get $secret.data "ca.key" | b64dec | nindent 4 }}
+
+{{- else }}
+{{- $ca := genCA "Infra Server" 3650 }}
```
dnephin, line 21: This is the name people will see in the Subject line of the certificate. Should we use "Infra" , or "Infra Server" ?

BruceMacD: "Infra Server" seems fine to me 👍

helm/charts/infra/templates/server/deployment.yaml
```diff
@@ -89,6 +93,11 @@ spec:
         - name: conf
           configMap:
             name: {{ include "server.fullname" . }}
+{{- if (not .Values.server.config.tls) }}
+        - name: tls-ca
```
BruceMacD, line 97: Is this the CA or a private key? Doesnt seem like it needs to be a secret if its just info about the CA

helm/charts/infra/templates/server/secret.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: {{ include "server.fullname" . }}-ca
+  labels:
+{{- include "server.labels" . | nindent 4 }}
+data:
+
+{{- $secret := lookup "v1" "Secret" .Release.Namespace (printf "%s-ca" (include "server.fullname" .)) -}}
+{{- if $secret.data }}
+  ca.crt: |
+{{- get $secret.data "ca.crt" | nindent 4 }}
+  ca.key: |
```
BruceMacD, line 17: Can I specify the CA without specifying the CA's key? Or is this CA cert used to generate other certificates?

helm/charts/infra/values.yaml
```diff
@@ -531,6 +531,10 @@ server:
     # - name: dev@example.com
     #   password: file:/var/run/secrets/dev@example.com
 
+    # TLS configuration for the API server. Defaults to generating a self-signed CA and
+    # generating certificates from that CA.
+    tls: {}
```
BruceMacD, line 536: What does this configuration look like?

internal/server/tls.go
```diff
@@ -25,21 +22,47 @@ func tlsConfigFromOptions(
...
 
-	if opts.Certificate != "" && opts.PrivateKey != "" {
-		roots, err := x509.SystemCertPool()
-		if err != nil {
-			logging.Warnf("failed to load TLS roots from system: %v", err)
-			roots = x509.NewCertPool()
+		manager := &autocert.Manager{
+			Prompt: autocert.AcceptTOS,
+			Cache:  autocert.DirCache(tlsCacheDir),
+			// TODO: according to the docs HostPolicy should be set to prevent
```
BruceMacD, line 34: We should open an issue for this, dont want it to get overlooked

helm/charts/infra/values.yaml
mxyng, line 536: similar to other keys, it'll be nice to add some example values

```
tls: {}
# ca: ...
# caPrivateKey: ...
```

internal/certs/certs.go
mxyng: The other implication is that the server can no longer be served outside of a helm install, e.g. docker, compose, or local build, unless the CA is manually created and configured. It'll be nice to keep this to support those use cases but not critical

helm/charts/infra/templates/server/deployment.yaml
dnephin: The kube secret has both the CA cert and private key. The CA cert is not secret, but the private key for the CA is. I think it's easier to manage them both together in the same kube secret, like we do with the connector.  If it were just a CA cert then definitely we wouldn't need a kube secret for it.

helm/charts/infra/templates/server/secret.yaml
dnephin: In the default setup provided by these helm charts it generates both, so both will always be specified. A user can specify only the CA if they create their own cert and key, but in that case the CA only gets used to verify client certs for mTLS.

ssoroka (approved):
looks reasonable. no concerns at this point

internal/certs/certs.go
dnephin: Ya, that's true. We could keep the ability to generate a CA if none is provided for development and demo purposes. I think in practice no real production use case should have the server generate a CA on first install. The operator really needs to save that CA and private key somewhere, and they aren't going to know it exists or where to find it if it was generated on first start.

I'll see about re-adding this for development.

internal/server/tls.go
dnephin: https://github.com/infrahq/infra/issues/2484, I'll add it to the comment as well.

internal/certs/certs.go
dnephin: One of the disadvantages to generating the CA by default is that if the user accidentally mis-configured either `ca` or `caPrivateKey`. For example, maybe they set only one of the two, or they set `primaryKey` instead of `caPrimaryKey`. With the changes in this PR they will get a nice error and know how to fix it.

If we always generate things for them by default then it won't fail, and they'll have a harder time debugging.  

I think it might be better to require these are set ahead of starting the server. For dev maybe we could add a `--dev` flag and do some setup in the CLI before starting the server.

mxyng: Either Helm or infra can exit error if only one of `ca` or `caPrivateKey` is supplied since both are required. It'll reduce the impact of the scenario you're suggesting

mxyng (approved):

PR #2404, opened 2022-06-25
maintain(ui): lint and format ui code

jmorganca:
Formats and lints the UI code


ui/.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": ["eslint:recommended", "next/core-web-vitals", "prettier"],
+  "rules": {
+    "@next/next/no-img-element": "off"
```
BruceMacD, line 4: We sure we want to disable this rather than using next.js image components? They bring a lot of potential benefits for speed.

BruceMacD (approved):
Looks good, just the one open question about Next images

hoyyeva (approved):

ui/.eslintrc.json
jmorganca: The `<Image />` component doesn't support `next export` (which is how we bundle the UI). That said, we should consider this if we do consider allowing the UI to run as an external process

PR #2414, opened 2022-06-27
feat: password requirements api

kimskimchi:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -96,6 +96,7 @@ func numberCheck(n int, password string) bool {
 	return match
 }
 
+// list is from OWASP, wikipedia
 func symbolCheck(n int, password string) bool {
 	match, _ := regexp.MatchString(fmt.Sprintf(`(.*[ !"#$%%&'()*+,-./\:;<=>?@^_{}|~%s%s]){%d,}`, regexp.QuoteMeta(`/\[]`), "`", n), password)
 	return match


internal/access/credential.go
```diff
@@ -73,3 +79,71 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
...
+func symbolCheck(n int, password string) bool {
+	match, _ := regexp.MatchString(fmt.Sprintf(`(.*[ !"#$%%&'()*+,-./\:;<=>?@^_{}|~%s%s]){%d,}`, regexp.QuoteMeta(`/\[]`), "`", n), password)
+	return match
+	// /\[]`
+}
+
+func checkPasswordRequirements(password string) error {
+	var errs []string
+
+	kind := "custom"
```
kimskimchi, line 109: These are temp variables; they will be replaced with what is read from the settings config

internal/access/credential.go
```diff
@@ -73,3 +79,71 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
 
 	return nil
 }
+
+func lowercaseCheck(n int, password string) bool {
+	match, _ := regexp.MatchString(fmt.Sprintf(`(.*\p{Ll}){%d,}`, n), password)
```
mxyng, line 84: These regex rules are hard to read and maintain. It'll be easier to understand if it doesn't use raw regex for everything. Instead, it can use something like [regexp.FindAllStringIndex](https://pkg.go.dev/regexp#Regexp.FindAllStringIndex) on an expression like `[a-z]`. You can compare the output as `len(result) >= n`

internal/access/credential.go
```diff
@@ -73,3 +79,71 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
...
+}
+
+func numberCheck(n int, password string) bool {
+	match, _ := regexp.MatchString(fmt.Sprintf(`(.*\p{N}){%d,}`, n), password)
+	return match
+}
+
+// list is from OWASP, wikipedia
+func symbolCheck(n int, password string) bool {
+	match, _ := regexp.MatchString(fmt.Sprintf(`(.*[ !"#$%%&'()*+,-./\:;<=>?@^_{}|~%s%s]){%d,}`, regexp.QuoteMeta(`/\[]`), "`", n), password)
```
mxyng, line 101: Might be better to negate letters and digits instead of listing out each symbol `[^a-zA-Z0-9]`

internal/access/credential.go
```diff
@@ -73,3 +79,71 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
...
+	return match
+	// /\[]`
+}
+
+func checkPasswordRequirements(password string) error {
```
mxyng, line 106: One requirement that isn't demonstrated here is the NIST password recommendations:

* [minimum 8 character length without additional complexity requirements](https://pages.nist.gov/800-63-3/sp800-63b.html#5111-memorized-secret-authenticators)
* [check passwords against a corpus of breached passwords and a list of common words/passwords.](https://pages.nist.gov/800-63-3/sp800-63b.html#5112-memorized-secret-verifiers)

The second recommendation is missing and I don't a good way of integrating with this implementation

internal/access/credential.go
dnephin: It would be great to avoid regex in general. I'm sure I could look up what all of this does, but it's generally not obvious form the regex string.

Could we use something like this instead?:

```go
var count int
for _, r := range password {
    if unicode.IsLower(r) {
        count++
    }
}
return count >= n
```

This pattern should work for uppercase and numbers as well (https://pkg.go.dev/unicode#IsNumber). https://pkg.go.dev/unicode#IsPunct or https://pkg.go.dev/unicode#IsSymbol may work for `symbolCheck`.

We can make it generic like this:

```go
func hasMinimumCount(min int, password string, check func(rune) bool) bool {
	var count int
	for _, r := range password {
		if check(r) {
			count++
		}
	}
	return count >= min
}
```

```go
if lowercase && !hasMinimumCount(minLowercase, password, unicode.IsLower) { ... }
```

dnephin (commented):
Great test coverage! Looks good, but I think we can avoid using regex

PR #2418, opened 2022-06-27
improve: add postgres migration test

### LINKED ISSUE #2395: data: run migration tests against postgres
Right now we only test our migrations against sqlite. We should test them against postgres as well.

Making this work is more involved than other tests in the `data` package because we have to generate a separate `.sql` file for the initial state. A dump from sqlite doesn't work in postgres.

This would prevent problems like #2392

mxyng:
## Summary

Add sqlite and postgres migration test for adding kind to providers

## Checklist

## Related Issues

Resolves #2395


internal/server/data/migrations_test.go
```diff
@@ -178,3 +178,21 @@ func tableExists(t *testing.T, db *gorm.DB, name string) bool {
 	assert.NilError(t, err)
 	return count > 0
 }
+
+func TestMigration_AddKindToProvider(t *testing.T) {
```
dnephin, line 182: I think `Test202206151027` was already covering this migration for sqlite, but I like this test name better.

Can we keep assertions from that test (the assertion is more strict and tests both cases), but update the test name to this one.

internal/server/data/migrationdata/202206151027-sqlite.sql
```diff
@@ -0,0 +1,46 @@
+PRAGMA foreign_keys=OFF;
```
dnephin, line 1: We should delete `internal/server/data/migrationdata/202206151027.sql` if we're not going to use it.

dnephin (commented):
Looks good! I think we just need to remove or integrate the existing test case

internal/server/data/migrationdata/202206151027-sqlite.sql
mxyng: Good call. I wasn't looking for an existing test case. Will update accordingly

ssoroka (approved):

PR #2419, opened 2022-06-27
improve: build and release dev packages from release branch

mxyng:
## Summary

Re-add build release changes merged in #2390 and reverted in #2399


BruceMacD (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,117 @@
+name: publish
+
+on:
```
jmorganca, line 3: To confirm: no PRs can trigger this workflow, right?

mxyng: The environment variables needed for publishing are also locked behind the production environment which is only accessible from main

Issue #2422, opened 2022-06-27
LIstGrants should support wildcard issue searches


ssoroka:
eg a search for a resource named `production.*` should return grants with resources named:

production
production.mynamespace
production.mynamespace.secrets.foo

see https://github.com/infrahq/infra/blob/58167809cda92d512bcb35eee9ade834a2c376d6/internal/connector/connector.go#L606-L620

PR #2423, opened 2022-06-27
feat: support requesting inherited grants

### LINKED ISSUE #2165: Inherited grants in the API
Currently to answer "can user X access resource Y?", API consumers must first query `/api/grants?user=<user_id>` and `/api/grants?group=<group_id>` for every user and group that user is in. This results in a complex query with quite a few calls.

Instead, it would be preferred if the API can show "inherited" grants. I.e. `/api/grants?user=<user_id>` takes into account:
1. This user's groups
2. This user's grants to _parent_ resources (e.g. a cluster -> namespace)

ssoroka:
## Summary

allows ListGrants to take an `includeInherited` flag that when set to `true` or `1` via query parameter includes grants for a user that they have due to group membership.

Note that I split one of the mentioned items from the original issue into a new issue as it didn't quite fit as well. see https://github.com/infrahq/infra/issues/2422 for that.

## Checklist

## Related Issues

Resolves #2165


internal/access/grant.go
```diff
@@ -30,7 +30,15 @@ func ListGrants(c *gin.Context, subject uid.PolymorphicID, resource string, priv
 	roles := []string{models.InfraAdminRole, models.InfraViewRole, models.InfraConnectorRole}
 	db, err := RequireInfraRole(c, roles...)
 	if err == nil {
-		selectors = append(selectors, data.ByOptionalSubject(subject))
+		if inherited && subject.IsIdentity() {
```
BruceMacD, line 33: I feel like a user should be able to request their own inherited roles too

internal/server/data/identity.go
```diff
@@ -92,6 +92,10 @@ func AssignIdentityToGroups(db *gorm.DB, user *models.Identity, provider *models
 	return nil
 }
 
+func AddUserToGroup(db *gorm.DB, user *models.Identity, group *models.Group) error {
```
BruceMacD, line 95: Could this functionality be in the group changes Patrick has been working on?

api/client.go
```diff
@@ -267,10 +267,11 @@ func (c Client) DeleteProvider(id uid.ID) error {
...
 	return get[ListResponse[Grant]](c, "/api/grants", Query{
-		"user":      {req.User.String()},
-		"group":     {req.Group.String()},
-		"resource":  {req.Resource},
-		"privilege": {req.Privilege},
+		"user":             {req.User.String()},
+		"group":            {req.Group.String()},
+		"resource":         {req.Resource},
+		"privilege":        {req.Privilege},
+		"includeInherited": {boolToStr(req.IncludeInherited)},
```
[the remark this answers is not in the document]
dnephin: https://pkg.go.dev/strconv#FormatBool is another option, does the same thing as the `fmt.Sprintf`

internal/server/handlers_test.go
```diff
@@ -1458,3 +1458,47 @@ func TestAPI_LoginResponse(t *testing.T) {
 	assert.Equal(t, loginResp.Name, "steve")
 	assert.Equal(t, loginResp.PasswordUpdateRequired, false)
 }
+
+func TestListGrants(t *testing.T) {
```
dnephin, line 1462: Is this different from the test cases in `grants_test.go` ?

internal/server/grants_test.go
```diff
@@ -339,6 +345,125 @@ func TestAPI_ListGrants(t *testing.T) {
 	}
 }
 
+func TestAPI_ListInheritedGrants(t *testing.T) {
```
dnephin, line 348:
```suggestion
func TestAPI_ListGrants_InheritedGrants(t *testing.T) {
```

internal/access/grant.go
ssoroka: Not sure what you're talking about here. you're talking about grants that grant access to roles, then those roles have their own grants? We don't have any of those today, but I agree with you.

BruceMacD: I mean the case like this:
```
if subject.IsGroup() && userBelongsToGroup(user, subject) {
  // add selector for this group
}
```

I don't think users could list grants for groups they belong to with this auth check?

api/grant.go
```diff
@@ -18,10 +18,11 @@ type Grant struct {
...
 type ListGrantsRequest struct {
-	User      uid.ID `form:"user" validate:"excluded_with=Group"`
-	Group     uid.ID `form:"group" validate:"excluded_with=User"`
-	Resource  string `form:"resource" example:"production"`
-	Privilege string `form:"privilege" example:"view"`
+	User             uid.ID `form:"user" validate:"excluded_with=Group"`
+	Group            uid.ID `form:"group" validate:"excluded_with=User"`
+	Resource         string `form:"resource" example:"production"`
+	Privilege        string `form:"privilege" example:"view"`
+	IncludeInherited bool   `form:"includeInherited" note:"if true, this field includes grants that the user inherits through groups"`
```
mxyng, line 25: Can this be `show_inherited`? We have `show_expired` for access keys and I'd like to have some consistency

internal/server/data/grant.go
```diff
@@ -56,6 +57,21 @@ func ByOptionalPrivilege(s string) SelectorFunc {
 	}
 }
 
+func GrantsInheritedByUser(userID uid.ID) SelectorFunc {
+	return func(db *gorm.DB) *gorm.DB {
+		var groupIDs []uid.ID
+		err := db.Session(&gorm.Session{Logger: db.Logger}).Raw("select distinct group_id from identities_groups where identity_id = ?", userID).Pluck("group_id", &groupIDs).Error
```
mxyng, line 63: Is this new session necessary?

internal/server/handlers_test.go
dnephin: This is testing the `ListGrants` handler, right? Can we move this into `grants_test.go` next to the other tests for this function?

It would also be good to include the particular case in the name of the test. From reading the test, it seems like it's testing inherited grants, so I don't think I understand why this wouldn't be another sub test in `TestAPI_ListGrants_InheritedGrants`

internal/access/grant.go
dnephin, line 33: I think what Bruce was trying to get at here is that this only looks at `inherited` if the request is being authorized by a grant.

If the request is being authorized because the request is for the user's own grants, then the `inherited` flag is being ignored.

I think this logic needs to happen above, so that the selector is added before going into this `err == nil` block.

dnephin: I don't think `inherited` means anything if the query is for a group. A group can't inherit grants, only a user inherits grants from a group.

But I think there is a case missing here. We're only using this flag when the request is authorized by a grant. The block below (where the request is authorized by authn identity matching the query) is ignoring `inherited`.

ssoroka: I added a test for this and it works as is

dnephin: If you mean `TestListGrants`, I think that is misleading. It's being authorized by the group grant, not by identity.

I'm pretty sure the scenario I described does not work.

internal/server/grants_test.go
```diff
@@ -177,7 +178,10 @@ func TestAPI_ListGrants(t *testing.T) {
 						Resource:  "res1",
 					},
 				}
-				assert.DeepEqual(t, grants.Items, expected, cmpAPIGrantShallow)
+				items := slice.Select(grants.Items, func(g api.Grant) bool {
+					return g.Resource != "infra"
+				})
```
dnephin, line 184: Why not add to the expected list instead of filtering it out? Seems like this is likely to hide bugs.

api/grant.go
ssoroka: yes

internal/access/grant.go
ssoroka: got it, thanks. added a test case for this.

internal/access/grant.go
```diff
@@ -29,30 +29,43 @@ func ListGrants(c *gin.Context, subject uid.PolymorphicID, resource string, priv
...
 		switch {
 		case identity == nil:
 			return nil, err
 		case subject.IsIdentity() && identity.ID == subjectID:
-			selectors = append(selectors, data.BySubject(subject))
+			if inherited {
+				selectors = append(selectors, data.GrantsInheritedBySubject(subject))
+			} else {
+				selectors = append(selectors, data.BySubject(subject))
+			}
```
dnephin, line 46: Are all 3 of these `if/else` blocks the same? I guess the last one is different in that it uses `ByOptionalSubject`.

What happens if `inherited` is used without a subject? I guess we should add no selector in that case? I think the `inherited && subject.IsIdentity()` you had originally was correct, the problem was only that it was not being applied in both authz scenarios. So maybe:

```go
if subject != "" {
    if inherited && subject.IsIdentity() {
        selectors = append(selectors, data.GrantsInheritedBySubject(subject))
    } else {
        selectors = append(selectors, data.BySubject(subject))
    }
}
```

I think if we did that once above, then we wouldn't need the `selectors = append(selectors, data.BySubject(subject))` we used to have here. We could perform the query with the already constructed list of selectors.

internal/server/apimigrator.go
```diff
@@ -130,6 +130,12 @@ func rebuildRequest(c *gin.Context, newReqObj interface{}) {
 				query.Add(fieldName, fmt.Sprintf("%d", f.Int()))
 			case reflect.Uint, reflect.Uint64:
 				query.Add(fieldName, fmt.Sprintf("%d", f.Int()))
+			case reflect.Bool:
+				if f.Bool() {
+					query.Add(fieldName, "1")
+				} else {
+					query.Add(fieldName, "0")
```
dnephin, line 137: Hopefully it's not a problem that we use `1`/`0` here and  `true`/`false` in the `api.Client`. I guess `strconv.FormatBool` could be used here too. Not a blocker.

internal/server/grants_test.go
```diff
@@ -344,6 +345,178 @@ func TestAPI_ListGrants(t *testing.T) {
...
+
+	loginAs := func(userID uid.ID) {
+		token := &models.AccessKey{
+			IssuedFor:  userID,
+			ProviderID: data.InfraProvider(srv.db).ID,
+			ExpiresAt:  time.Now().Add(10 * time.Minute),
+		}
+
+		var err error
+		accessKey, err = data.CreateAccessKey(srv.db, token)
```
dnephin, line 401: This modification of `accessKey` is really surprising, and is going to make test failures hard to debug. The next test case added to this test function will use this `accessKey` instead of the default one.

There should be an easy way to avoid this by changing this to modify the `http.Request`:


```go
func loginAs(db *gorm.DB, userID uid.ID) func(t *testing.T, req *http.Request) {
    ... // the current logic in this function
    return func(t *testing.T, req *http.Request) {
        req.Header.Set("Authorization", "Bearer "+accessKey)
    }
}
```

This would allow `loginAs` to be used directly as the setup function.  But there are lots of other options that allow you to create an access key directly inline and pass it to the request, without  modifying variables that are used in other test cases.

dnephin (commented):
Looks good!

Just two concerns:

Does a query with `inherited=true` and no subject panic? Seems like it might. Maybe it should be a 400 error, or maybe it should ignore `inherited`.

The test should be able to create an access key in the setup function without modifying the key used by other tests.

internal/access/grant.go
ssoroka: that's not right because in some cases (requesting own grants) the subject is required. Merging this endpoint with the /users/x/grants endpoints added some of this complexity.

internal/server/grants_test.go
ssoroka: nice suggestion.

internal/access/grant.go
dnephin: The subject is required for authorization, but that check already exists.

internal/access/grant.go
```diff
@@ -29,30 +30,47 @@ func ListGrants(c *gin.Context, subject uid.PolymorphicID, resource string, priv
...
-
 	if errors.Is(err, ErrNotAuthorized) {
 		// Allow an authenticated identity to view their own grants
-		db := getDB(c)
-		subjectID, _ := subject.ID()
+		db = getDB(c)
+		subjectID, err2 := subject.ID()
+		if err2 != nil {
+			// user is only allowed to select their own grants, so if the subject is missing or invalid, this is an access error
+			return nil, fmt.Errorf("%w: %s", AuthorizationError{Resource: "grants", Operation: "list", RequiredRoles: []string{"view"}}, err)
```
dnephin, line 40: I think this can return `err` directly like it did previously (at the end of the function). That should be the correct authz error:

```suggestion
			// user is only allowed to select their own grants, so if the subject is missing or invalid, this is an access error
			return nil, err
```

internal/server/data/grant.go
```diff
@@ -56,6 +57,34 @@ func ByOptionalPrivilege(s string) SelectorFunc {
...
+		case subjectID.IsIdentity():
+			userID, err := subjectID.ID()
+			if err != nil {
+				logging.Errorf("invalid subject id %q", subjectID)
+				return db.Where("1 = 0")
+			}
+			var groupIDs []uid.ID
+			err = db.Session(&gorm.Session{Logger: db.Logger}).Raw("select distinct group_id from identities_groups where identity_id = ?", userID).Pluck("group_id", &groupIDs).Error
+			if err != nil {
+				logging.Errorf("GrantsInheritedByUser: %s", err)
```
dnephin, line 72: Using `db.Session` here looks correct, it's a separate query so needs to be a separate session.

Doing a separate independent query in a selector is a bit strange, but unless we change this to a join I don't see any way around it until we replace selectors. So I think this works for now.

Maybe we should `return db.Where("1 = 0")` here as well , like above?

internal/server/routes.go
```diff
@@ -47,7 +47,7 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	// This group of middleware will apply to everything, including the UI
 	router.Use(
 		logging.Middleware(),
-		TimeoutMiddleware(1*time.Minute),
```
dnephin, line 50: Did you mean to change this to 2 minutes, or was this for debugging? Might be good to make this in a separate PR if you did mean to change it.

dnephin (approved):
LGTM

internal/server/data/grant.go
```diff
@@ -56,6 +57,34 @@ func ByOptionalPrivilege(s string) SelectorFunc {
...
+	return func(db *gorm.DB) *gorm.DB {
+		switch {
+		case subjectID.IsIdentity():
+			userID, err := subjectID.ID()
+			if err != nil {
+				logging.Errorf("invalid subject id %q", subjectID)
+				return db.Where("1 = 0")
+			}
+			var groupIDs []uid.ID
+			err = db.Session(&gorm.Session{Logger: db.Logger}).Raw("select distinct group_id from identities_groups where identity_id = ?", userID).Pluck("group_id", &groupIDs).Error
```
dnephin, line 70: Did this not work without setting `Logger`? It seems like `Session` should copy `db.Config`, which should have the logger already set.

internal/server/data/grant.go
dnephin: Looking at this again, I think we probably want `gorm.Session{NewDB: true}`. That should erase any previous statements added to the session by other selectors.

internal/server/routes.go
ssoroka: I found it annoying to debug to step through the code in less than a minute and figured 2 minutes was still a reasonable timeout for everyone. can remove it

PR #2424, opened 2022-06-27
feat: require Infra-Version header (#2041)

ssoroka:
## Summary

Brings back the required infra version header. Has already been reviewed and approved.

### DIFF
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -433,7 +433,9 @@ func buildRequest(r reflect.Type, op *openapi3.Operation) {
 
 			buildRequest(f.Type, tmpOp)
 			for _, param := range tmpOp.Parameters {
-				op.AddParameter(param.Value)
+				if param.Value.Name != "Infra-Version" {
+					op.AddParameter(param.Value)
+				}
 			}
 			continue
 		}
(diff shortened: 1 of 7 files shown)


BruceMacD (approved):
Let's get this one in

internal/server/openapi.go
```diff
@@ -430,6 +431,20 @@ func buildResponse(schemas openapi3.Schemas, rst reflect.Type) openapi3.Response
 func buildRequest(r reflect.Type, op *openapi3.Operation) {
 	op.Parameters = openapi3.NewParameters()
 
+	op.AddParameter(&openapi3.Parameter{
+		Name:     "Infra-Version",
+		In:       "header",
+		Required: true,
+		Schema: &openapi3.SchemaRef{
+			Value: &openapi3.Schema{
+				Example:     internal.FullVersion(),
```
dnephin, line 440: This is going to re-introduce the problem that was fixed in #2428. Every time we do a release, and every time this ends up on the `release-please` branch, this version will change, which will cause the `TestWriteOpenAPIDocToFile` test to fail.

We need some way to pass in the version so that we don't rely on the static `internal.FullVersion`.

If we can't pass in the version, we'll have to patch it in `TestWriteOpenAPIDocToFile` to set it to a hardcoded value like `0.0.0`.

dnephin (changes requested):
Sorry, blocking since I think this will cause the build to fail once merged

PR #2430, opened 2022-06-28
cleanup checking IDs while updating group membership

pdevine:
## Summary

The way we were looking up user IDs when adding or removing them from groups was fairly costly. This change uses the `ListIdentities()` call with the `ByIDs()` selector to verify that an ID exists. It also adds a couple of unit tests, and cleans up the `INSERT OR IGNORE` statement.

## Checklist

### DIFF
--- a/internal/access/group.go
+++ b/internal/access/group.go
@@ -89,15 +89,6 @@ func DeleteGroup(c *gin.Context, id uid.ID) error {
 }
 
 func checkIdentitiesInList(db *gorm.DB, ids []uid.ID) ([]uid.ID, error) {
-	contains := func(ids []models.Identity, id uid.ID) bool {
-		for _, i := range ids {
-			if i.ID == id {
-				return true
-			}
-		}
-		return false
-	}
-
 	identities, err := data.ListIdentities(db, data.ByIDs(ids))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


internal/access/group.go
```diff
@@ -84,6 +88,36 @@ func DeleteGroup(c *gin.Context, id uid.ID) error {
 	return data.DeleteGroups(db, selectors...)
 }
 
+func checkIdentitiesInList(db *gorm.DB, ids []uid.ID) ([]uid.ID, error) {
+	contains := func(ids []models.Identity, id uid.ID) bool {
```
BruceMacD, line 92: This seems like it could be slow for a big look-up, maybe we should sort or map the ids slice first.

internal/server/data/group.go
```diff
@@ -62,7 +62,9 @@ func DeleteGroups(db *gorm.DB, selectors ...SelectorFunc) error {
 
 func AddUsersToGroup(db *gorm.DB, groupID uid.ID, idsToAdd []uid.ID) error {
 	for _, id := range idsToAdd {
-		err := db.Exec("INSERT INTO identities_groups (group_id, identity_id) select ?, ? WHERE NOT EXISTS (SELECT 1 FROM identities_groups WHERE group_id = ? AND identity_id = ?)", groupID, id, groupID, id).Error
+		// This is effectively an "INSERT OR IGNORE" or "INSERT ... ON CONFLICT ... DO NOTHING" statement which
+		// works across both sqlite and postgres
```
BruceMacD, line 66:
```suggestion
```

This should hopefully always be the case at the moment, so we don't need to state it

pdevine: My thinking on this is it's a non-standard-ish SQL call that could be replaced if we standardized on one database type.  If we only supported postgres, we should instead of "INSERT ... ON CONFLICT", but that won't work w/ sqlite.

internal/access/group.go
pdevine: I'm not super happy with this being n^2, but even if you sorted it w/ n log(n) you still run into n^2 performance if the UID gets sorted to the end of the list. My thinking for not using a map originally was just that the slice would never be that big, but I guess I could see if there were thousands of IDs in a list and one wasn't found that it could be slow. I'm not sure how likely that is, but it's easy enough to change to a map.

BruceMacD (approved):

PR #2432, opened 2022-06-28
feat: 

mxyng:
## Checklist

## Related Issues

Resolves #


internal/logging/logger.go
```diff
@@ -2,205 +2,104 @@
...
+	zerolog.CallerMarshalFunc = func(file string, line int) string {
+		short := file
+		for i := len(file) - 1; i > 0; i-- {
+			if file[i] == '/' {
+				short = file[i+1:]
+				break
+			}
+		}
+		file = short
+		return file + ":" + strconv.Itoa(line)
```
dnephin, line 34: I see this is from their examples, but we can probably use https://pkg.go.dev/path/filepath#Base for this?

```suggestion
		return filepath.Base(file) + ":" + strconv.Itoa(line)
```

internal/logging/logger.go
```diff
@@ -2,205 +2,104 @@
...
-		}
+func (l *logger) Warnf(format string, v ...interface{}) {
+	l.Warn().Msgf(format, v...)
+}
 
-		if b[0] != '{' {
-			// not json; free to truncate.
-			return w.dest.Write(b[:idx])
-		}
+func (l *logger) Errorf(format string, v ...interface{}) {
```
dnephin, line 64: Should we document these compatibility helpers as deprecated? I suspect we probably want to encourage everyone to use the `zerolog.Logger` methods directly so that we can add context fields. And  we probably want to avoid having yet another way of logging (since zerolog seems to have at least a couple ways already)

dnephin (commented):
Nice! I think zerolog fixes a lot of issues I had with zap logger.  Not having both a global `S` and `L` is really nice (while still being able to use structured logging). And the console logging format looks much nicer.

There are a couple minor things I think we could fix for the ConsoleWriter. I suspect we probably want a `FormatLevel` setting to print `INFO` instead of `INF`, but I think that's easy to add later.

I'll have a more detailed look once the build is passing.

internal/logging/logger.go
mxyng: Yeah I took it verbatim without much thought. This is a much better implementation

internal/logging/logger.go
mxyng: Possibly. I like this simplified interface since the `zerolog` API requires a `Msg` to finalize the log, it's easy to forget and end up with nothing being logged. There's definitely scenarios where we'd want to use the `zerolog` API directly, particularly for structured logging or contexts but that might be overkill for most cases

mxyng:
The `ConsoleWriter` definitely needs some tuning. I didn't spend any time with the console outputs, instead focusing on the server log outputs. That'll will be the next step if we decide to go in this direction

internal/server/errors.go
```diff
@@ -78,20 +72,11 @@ func sendAPIError(c *gin.Context, err error) {
 	case errors.Is(err, context.DeadlineExceeded):
 		resp.Code = http.StatusGatewayTimeout // not ideal, but StatusRequestTimeout isn't intended for this.
 		resp.Message = "request timed out"
-
-	default:
-		log = logging.L.WithOptions(zap.AddCallerSkip(1)).Error
 	}
 
-	log("api request error",
```
mxyng, line 86: This log is extremely similar to the log line produced by the logging middleware with the addition of the error. Instead of producing the same line twice, attach the error to the context and consume it in the middleware

internal/logging/http.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/rs/zerolog"
+)
+
+var strInvalidHeaderFieldValue = []byte("invalid header field value")
+
+type FilteredHTTPLogger struct {
+	zerolog.Logger
+}
+
+func (l FilteredHTTPLogger) Write(b []byte) (int, error) {
```
BruceMacD, line 17: What does this filter? I feel like we could use some tests around this, it seems high risk for logging things that should not be logged.

internal/logging/database.go
```diff
@@ -0,0 +1,45 @@
...
+type DatabaseLogger struct {
+	zerolog.Logger
+}
+
+func (l DatabaseLogger) LogMode(level gormlogger.LogLevel) gormlogger.Interface {
+	return &DatabaseLogger{l.Logger}
+}
+
+func (l DatabaseLogger) Info(ctx context.Context, format string, v ...interface{}) {
+	l.Logger.Info().Msgf(format, v...)
```
BruceMacD, line 20: How much does this log at the info level?

internal/server/errors.go
```diff
@@ -78,20 +72,11 @@ func sendAPIError(c *gin.Context, err error) {
...
-		zap.String("method", c.Request.Method),
-		zap.String("path", c.Request.URL.Path),
-		zap.Int32("statusCode", resp.Code),
-		zap.String("remoteAddr", c.Request.RemoteAddr),
-		zap.Error(err))
-
 	c.JSON(int(resp.Code), resp)
-	c.Abort()
+	//nolint:errcheck // AbortWithError returns the error it pushed to the context or panics of err is nil
```
BruceMacD, line 78:
```suggestion
	//nolint:errcheck // AbortWithError returns the error it pushed to the context or panics if err is nil
```

internal/server/errors.go
```diff
@@ -78,20 +72,11 @@ func sendAPIError(c *gin.Context, err error) {
...
+	c.AbortWithError(int(resp.Code), err)
```
BruceMacD, line 79: This means that every invalid login attempt gets logged as an error. I think it should be at the `info` level to avoid letting unauthorized outside actors flood the logs (it can be a separate change).

internal/logging/http.go
mxyng: It filters out unknown header values from logs for https://github.com/infrahq/infra/issues/655. The code was moved from `logging/logger.go` and the test lives in `logging/logger_test.go`. I guess I can move the test to `logging/http.go`

internal/logging/database.go
mxyng: This shouldn't actually log. The only statements I see are trace logs when trace level is enabled

internal/server/errors.go
mxyng: That was already the case before so this isn't introducing anything new

internal/server/errors.go
dnephin: I like this idea, I was thinking of a similar thing here: https://github.com/infrahq/infra/pull/2400#discussion_r908779508

But as Bruce said below there are other implications for this change.  

I think we should hold off on that for now. Switching logging libraries is already a huge change, we should probably try to limit any other changes in this PR.

internal/server/errors.go
dnephin: This did change the behaviour. Previously we used `Debug` level for most errors, `Info` for authn errors, and `Error` for 5xx errors.  With this change I think we'd have everything at info level in the request logs.

internal/logging/logger.go
```diff
@@ -103,6 +105,7 @@ func Middleware() gin.HandlerFunc {
 
 		log.WithLevel(levelName).
 			Errs("errors", errs).
+			Dur("took", time.Since(begin)).
```
dnephin, line 108: Maybe `elapsed` or `elapsed_time` ?

internal/logging/logger.go
```diff
@@ -91,7 +91,18 @@ func Middleware() gin.HandlerFunc {
 
 		c.Next()
 
-		log.Info().
+		levelName := zerolog.InfoLevel
+		if c.IsAborted() {
+			levelName = zerolog.ErrorLevel
+		}
```
dnephin, line 99: Because of how we implemented the `UIProxy` every request through the proxy is aborted. So this might not behave correctly in all cases. Maybe be better to check the length of `c.Errors`.

mxyng: I don't see this. UI requests are logged with info regardless if they're served from the static fs or proxied to an external server

internal/server/errors.go
mxyng: You're right. I forgot logs levels are dependent on the error type

internal/logging/logger.go
mxyng: Good catch. Checking `c.Errors` is probably the right solution

internal/server/errors.go
mxyng: Removing for now

internal/server/errors.go
mxyng: Splitting it out for now. The current implementation isn't ideal so we should figure something out

BruceMacD (approved):
Looks good to me now 👍

PR #2434, opened 2022-06-28
feat: providers edit

### LINKED ISSUE #2381: Infra providers should be editable
## Describe the bug
In some cases client secrets for a provider will expire. You should be able to run `infra providers edit NAME` to change the secret of a provider without deleting them.

## Expected behavior
The UI and CLI allow editing identity providers.

## Additional context
Azure secrets expire and they should be able to change it without deleting the provider.

kimskimchi:
## Summary

Implements `infra providers edit` command

## Checklis

## Related Issues

Resolves #2381


BruceMacD (approved):
I'd like to see an non-interactive option, but otherwise looks good.

internal/cmd/providers.go
```diff
@@ -28,11 +30,34 @@ func newProvidersCmd(cli *CLI) *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !secret {
+				return errors.New("Please specify a field to update. For options, run 'infra providers edit --help'")
+			}
+
+			return updateProvider(cli, args[0])
+		},
+	}
+
+	cmd.Flags().BoolVar(&secret, "secret", false, "Set a new secret")
```
jmorganca, line 57: We're missing some values here:

```
--client-secret
--client-id
--url
--kind
```

jmorganca (changes requested):

internal/cmd/providers.go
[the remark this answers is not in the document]
kimskimchi: Not sure, it was shut down for user passwords. Passing values as flags were strongly discouraged. Perhaps we can later support passing files in if automation is requested.

kimskimchi: The ticket was only for secret, do we want to support all? I was under the impression that users should not change other fields and should rather create a new provider instead.

jmorganca: That makes sense, since `--client-secret` is often rotated. Let's stick to only `--secret`, but we should rename it to `--client-secret`

internal/cmd/providers.go
```diff
@@ -28,11 +29,33 @@ func newProvidersCmd(cli *CLI) *cobra.Command {
...
 	return cmd
 }
 
+func newProvidersEditCmd(cli *CLI) *cobra.Command {
+	var secret bool
+
+	cmd := &cobra.Command{
+		Use:   "edit PROVIDER",
+		Short: "Update a provider",
+		Example: `# Set a new secret for a provider
```
jmorganca, line 44: nit: replace secret -> client secret

internal/cmd/providers.go
```diff
@@ -28,11 +29,33 @@ func newProvidersCmd(cli *CLI) *cobra.Command {
...
+$ infra providers edit okta --secret`,
```
jmorganca, line 45: Update `--secret` -> `--client-secret`

internal/cmd/providers.go
```diff
@@ -164,6 +187,61 @@ $ infra providers add okta --url example.okta.com --client-id 0oa3sz06o6do0muoW5
 	return cmd
 }
 
+func updateProvider(cli *CLI, name string) error {
+	secret, err := promptSetSecret(cli)
```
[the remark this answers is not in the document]
kimskimchi: We didn't allow user passwords to be passed in as flag values, are we changing the stance on this one? 
Back then it was discouraged, and rather decided that it should be passed in as files or prompts

mxyng:
+1 on both @BruceMacD and @jmorganca's points about editing other fields in non-interactive ways. I also think this pattern isn't sustainable and we should look for an alternative in the long run. Requiring each editable field as a parameter is an annoyance with a handful of fields but will be impossible as the number and complexity of fields increases

PR #2435, opened 2022-06-28
connector: refactor authentication middleware

dnephin:
## Summary

This PR refactors the connector middleware to combine `jwtMiddleware` and `proxyMiddleware` into a single middleware. This reduces the code a bit, because we handle response codes in a single place, and it improves the logging by including more detail.

This refactor also allows us to better test the authentication. Instead of only checking the response code, we can now ensure we are correctly hitting the error case we expect. We should also add tests for the `proxyMiddleware`, but those did not previously exist, so I haven't added them yet.

As part of this change I also:
* removed `BearerTransport`, this type was not doing anything because we never set the `Token`
* refactored the tests to use a table test, since most of the test logic is the same
* slightly increased test coverage by also testing part of `getJWK`
* renamed `jwkCache` to `authenticator`, which is the responsibility of this type in relation to the rest of the connector

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -167,7 +167,7 @@ func proxyMiddleware(
 	return func(c *gin.Context) {
 		claim, err := authn.Authenticate(c.Request)
 		if err != nil {
-			logging.L.Err(err).Msgf("failed to authenticate request")
+			logging.L.Info().Err(err).Msgf("failed to authenticate request")
 			c.AbortWithStatus(http.StatusUnauthorized)
 			return
 		}
--- a/internal/connector/connector_test.go
+++ b/internal/connector/connector_test.go
@@ -94,6 +94,15 @@ func TestAuthenticator_Authenticate(t *testing.T) {
 			expectedErr: "token is expired",
 		},
 		{
+			name: "no username in JWT",
+			setup: func(t *testing.T, req *http.Request) {
+				j := generateJWT(t, priv, "", time.Now().Add(time.Hour))
+				req.Header.Set("Authorization", "Bearer "+j)
+			},
+			fakeClient:  fakeClient{key: *pub},
+			expectedErr: "no username in JWT claim",
+		},
+		{
 			name: "valid JWT",
 			setup: func(t *testing.T, req *http.Request) {
 				j := generateJWT(t, priv, "test@example.com", time.Now().Add(time.Hour))


internal/claims/claims.go
```diff
@@ -1,7 +1,7 @@
 package claims
 
 type Custom struct {
-	Name   string   `json:"name" validate:"required"`
```
dnephin, line 4: We already had code that checked if `Name == ""`. "required" is a really easy thing to validate with an `if` statement. I don't think we need to involve any validation libraries here. 

So I removed the `validate` tag and left the existing `if` check. I did however move it into a different function.

internal/connector/connector.go
```diff
@@ -46,16 +45,20 @@ type Options struct {
...
 }
 
-type jwkCache struct {
+type authenticator struct {
 	mu          sync.Mutex
 	key         *jose.JSONWebKey
 	lastChecked time.Time
 
-	client  *http.Client
+	client  httpClient
```
dnephin, line 53: I made this an interface so that we can test it with a fake client implementation. In the future we could test it with an in-process server, but this was already an improvement in test coverage, so I left it for now.

internal/connector/connector.go
```diff
@@ -98,123 +101,79 @@ func (j *jwkCache) getJWK() (*jose.JSONWebKey, error) {
...
-
-type BearerTransport struct {
-	Token     string
-	Transport http.RoundTripper
-}
-
-func (b *BearerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	if b.Token != "" {
-		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", b.Token))
+func newAuthenticator(url string, options Options) *authenticator {
```
dnephin, line 104: I created a constructor for this to remove some of the logic from the large `Run` function.

internal/connector/connector.go
```diff
@@ -98,123 +101,79 @@ func (j *jwkCache) getJWK() (*jose.JSONWebKey, error) {
...
-			return
-		}
-
-		var claims struct {
-			jwt.Claims
-			claims.Custom
-		}
-
-		out := make(map[string]interface{})
-		if err := tok.Claims(key, &claims, &out); err != nil {
```
dnephin, line 152: I'm guessing `out` was only used for debugging, I did not see it referenced anywhere else. I removed it for now.

internal/connector/connector_test.go
```diff
@@ -122,70 +155,42 @@ func generateJWT(priv *jose.JSONWebKey, email string, expiry time.Time) (string,
...
+		Body:       ioutil.NopCloser(&buf),
+	}
+	return resp, nil
 }
 
-func TestCertificate(t *testing.T) {
+func TestCertCache_Certificate(t *testing.T) {
+	if testing.Short() {
+		t.Skip("too slow for short run")
+	}
```
dnephin, line 193: Runtime is in the seconds because it generates private keys.

internal/connector/connector.go
```diff
@@ -98,123 +101,79 @@ func (j *jwkCache) getJWK() (*jose.JSONWebKey, error) {
...
-			return
-		}
-
-		if name != "" {
-			c.Request.Header.Set("Impersonate-User", name)
-		} else {
-			logging.S.Debug("unable to determine identity")
+		claim, err := authn.Authenticate(c.Request)
+		if err != nil {
+			logging.S.Infof("failed to authenticate request: %v", err)
```
dnephin, line 170: Previously we only logged these errors at debug level. I changed it to `Info` because there's no other way to see problems like failing to investigate these problems, and a user shouldn't need to enable debug logs to figure out why they can't connect.

jmorganca (approved):
LGTM! Thanks for cleaning this up

mxyng (approved):

PR #2438, opened 2022-06-29
fix: strip port from self-signed certs

### LINKED ISSUE #2437: Self-signed certificates created by infra server use wrong server name
The port isn't stripped from the incoming host address which causes it to fail the `net.ParseIP` check:

```
2022-06-28T20:43:29.732-0400	INFO	certs/certs.go:110	new server certificate	{"Server name": "127.0.0.1:443", "SHA256 fingerprint": "4B:89:21:87:69:03:C7:D2:54:61:85:8C:DF:04:E1:B0:2C:FA:70:74:50:D7:19:24:ED:85:9D:A9:E9:86:EE:8D"}
```

jmorganca:
Resolves #2437


mxyng (approved):

dnephin (approved):
Looks good!  A test for this would be great, but it is currently very difficult to test `SelfSignedOrLetsEncryptCert`. I'll see if I can add a test for it after the changes in #2401

jmorganca:
Thanks for the reviews!

PR #2441, opened 2022-06-29
feat: destination life cycle

### LINKED ISSUE #2151: Need to show destination status to know if a cluster is still alive

Example request:

```
GET /api/destinations/<id>

{
  ...
  ...
}
```

mxyng:
## Summary

Add and Update Destination `LastSeenAt` timestamp when a request is received with header `Infra-Destination`. This value will be used to inform other fields like destination status as well as take action if destinations are not seen for an extended period of time

## Checklist

## Related Issues

Resolves #2151


api/client.go
```diff
@@ -154,38 +158,8 @@ func patch[Req, Res any](client Client, path string, req *Req) (res *Res, err er
 }
 
 func delete(client Client, path string) error {
```
mxyng, line 160: Noticed this wasn't adding the common headers. Probably should. Could be its own PR

internal/server/middleware.go
```diff
@@ -56,6 +56,37 @@ func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
...
+						c.Next()
+						return
+					}
+
+					sendAPIError(c, err)
+					return
+				}
+
+				// only save if there's significant difference between LastSeenAt and Now
+				if time.Since(destination.LastSeenAt) > time.Second {
```
mxyng, line 76: This isn't strictly necessary but the connector tends to send bursts of requests. Seem unreasonable to update the destination for each request in a short window

internal/server/routes.go
```diff
@@ -57,6 +57,7 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	apiGroup := router.Group("/",
 		metrics.Middleware(promRegistry),
 		DatabaseMiddleware(a.server.db), // must be after TimeoutMiddleware to time out db queries.
+		DestinationMiddleware(),
```
dnephin, line 60: I think we'll probably need to put this after the `AuthenticationMiddleware` so that we can authorize the request to update the connector last seen.

internal/server/middleware.go
```diff
@@ -56,6 +56,37 @@ func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
 	}
 }
 
+func DestinationMiddleware() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		destinationID := c.GetHeader("Infra-Destination")
+		if destinationID != "" {
+			if db, ok := c.MustGet("db").(*gorm.DB); ok {
+				destination, err := data.GetDestination(db, data.ByOptionalUniqueID(destinationID))
```
dnephin, line 64: I think we probably want to add a call to `access.RequireInfraRole` to make sure this has the connector role before updating the `LastSeenAt`.

dnephin (commented):
Looks great! Just one suggestion about authorization

internal/server/routes.go
mxyng: Good call. For this draft, I elected to make a minimal set of changes. The auth check will definitely be in the final version

internal/server/data/migrationdata/202207041724-postgres.sql
```diff
@@ -0,0 +1,225 @@
...
+);
+
+
+ALTER TABLE testing.migrations OWNER TO postgres;
+
+--
+-- Name: providers; Type: TABLE; Schema: testing; Owner: postgres
+--
+
+CREATE TABLE testing.providers (
```
mxyng, line 79: The db migration checks if the provider table exists and creates all schemas if it doesn't. This is problematic for tests since it'll create the destinations table at the current version. To work around this, the SQL needs to include the providers table. Alternatively, the migrator should check the migrations table instead.

api/client.go
```diff
@@ -105,6 +105,10 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 	req.Header.Set("Infra-Version", apiVersion)
 	req.Header.Set("User-Agent", fmt.Sprintf("Infra/%v (%s %v; %v/%v)", apiVersion, clientName, clientVersion, runtime.GOOS, runtime.GOARCH))
 
+	for k, v := range client.Headers {
```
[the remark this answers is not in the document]
ssoroka: could check first if there's a value..
```go
if _, exists := req.Header[k]; !exists {
  req.Header[k] = v
}
```

api/destination.go
```diff
@@ -14,6 +14,8 @@ type Destination struct {
 
 	Resources []string `json:"resources"`
 	Roles     []string `json:"roles"`
+
+	LastSeen Time `json:"lastSeen"`
```
ssoroka, line 18: can we do LastSeenAt ?

internal/server/data/migrations.go
```diff
@@ -474,6 +474,21 @@ func migrate(db *gorm.DB) error {
 		addKindToProviders(),
 		dropCertificateTables(),
 		addAuthURLAndScopeToProviders(),
+		// create Destination.LastSeenAt and set to UpdatedAt
```
ssoroka, line 477: should we continue the pattern of putting these in named functions?

ssoroka (approved):

internal/server/data/migrationdata/202207041724-postgres.sql
mxyng: The postgres dump for all tables is large and most of it is irrelevant so I cherry picked certain tables. Previously, it was sufficient to just dump the providers and migrations table. This time, the dump needs to contain the destinations, providers, and migrations table

mxyng: nvm the partial dump is causing problems else where. will update the postgres sql with a full dump

PR #2443, opened 2022-06-29
feat: expose provider auth url for ui sso

### LINKED ISSUE #2353: Send identity provider auth URL in providers list
## Describe the bug
To login with a given identity provider in the UI we need to know their auth URL.

## Expected behavior
The URL for auth endpoint needs to be returned from the server. The UI can't retrieve it itself due to CORS.

### LINKED ISSUE #2384: ui: login page shows Azure AD as generic single sign on
Screen Shot 2022-06-23 at 12 27 52 PM

Also, it looks a bit off

BruceMacD:
## Summary
To login with a given identity provider in the UI we need to know their auth URL. This change exposes the auth URL for an identity provider and the scopes it supports from the `/api/providers` endpoint. This means we do not need to look these values up on each login in the CLI also.

Best viewed as 2 separate commits:
Expose the auth URL and scopes: b779919f00c149da41d4f0e1a3228604feeff70b
Rename OIDC to OIDCClient in the code: b779919f00c149da41d4f0e1a3228604feeff70b

## UI Updates
image

## Checklist

## Related Issues

Resolves #2353
Resolves #2384


api/provider.go
```diff
@@ -5,13 +5,15 @@ import (
...
-	ClientID string `json:"clientID" validate:"required" example:"0oapn0qwiQPiMIyR35d6"`
-	Kind     string `json:"kind" example:"oidc"`
+	ID       uid.ID   `json:"id"`
+	Name     string   `json:"name" example:"okta"`
+	Created  Time     `json:"created"`
+	Updated  Time     `json:"updated"`
+	URL      string   `json:"url" validate:"required" example:"infrahq.okta.com"`
+	ClientID string   `json:"clientID" validate:"required" example:"0oapn0qwiQPiMIyR35d6"`
+	Kind     string   `json:"kind" example:"oidc"`
+	AuthURL  string   `json:"authURL" example:"https://example.com/oauth2/v1/authorize"`
```
BruceMacD, line 15: This isn't sensitive info, it is retrieved from a public endpoint:
https://example.okta.com/.well-known/openid-configuration

internal/server/config.go
```diff
@@ -612,10 +616,38 @@ func (Server) loadProvider(db *gorm.DB, input Provider) (*models.Provider, error
...
 			ClientSecret: models.EncryptedAtRest(input.ClientSecret),
+			AuthURL:      input.AuthURL,
+			Scopes:       input.Scopes,
 			Kind:         kind,
 			CreatedBy:    models.CreatedBySystem,
 		}
 
+		if provider.Kind != models.InfraKind {
+			// only call the provider to resolve info if it is not known
+			if input.AuthURL == "" && len(input.Scopes) == 0 {
```
BruceMacD, line 627: In config the auth URL and scopes can be specified to skip the external look-up.

internal/server/data/migrationdata/202204111503.sql
```diff
@@ -24,9 +24,9 @@ INSERT INTO grants VALUES(36544743275307009,'2022-04-11 20:15:45.765891717+00:00
...
-CREATE TABLE `providers` (`id` integer,`created_at` datetime,`updated_at` datetime,`deleted_at` datetime,`name` text,`url` text,`client_id` text,`client_secret` text,PRIMARY KEY (`id`));
-INSERT INTO providers VALUES(36544743090757632,'2022-04-11 20:15:45.721963717+00:00','2022-04-11 20:15:45.763277384+00:00',NULL,'infra','','','');
-INSERT INTO providers VALUES(36544743262724096,'2022-04-11 20:15:45.762991842+00:00','2022-04-11 20:28:03.681333795+00:00',NULL,'okta','dev.okta.com','','');
+CREATE TABLE `providers` (`id` integer,`created_at` datetime,`updated_at` datetime,`deleted_at` datetime,`name` text,`url` text,`client_id` text,`client_secret` text, `scopes` text, `auth_ul` text, PRIMARY KEY (`id`));
+INSERT INTO providers VALUES(36544743090757632,'2022-04-11 20:15:45.721963717+00:00','2022-04-11 20:15:45.763277384+00:00',NULL,'infra','','','','','');
+INSERT INTO providers VALUES(36544743262724096,'2022-04-11 20:15:45.762991842+00:00','2022-04-11 20:28:03.681333795+00:00',NULL,'okta','dev.okta.com','','','','');
```
BruceMacD, line 29: Added the `auth_url` and `scopes` columns to some old migration snapshots to avoid the external migration look-up.

internal/server/data/provideruser.go
```diff
@@ -49,3 +53,39 @@ func DeleteProviderUsers(db *gorm.DB, selectors ...SelectorFunc) error {
 func GetProviderUser(db *gorm.DB, providerID, userID uid.ID) (*models.ProviderUser, error) {
 	return get[models.ProviderUser](db, ByProviderID(providerID), ByIdentityID(userID))
 }
+
+func SyncProviderUser(ctx context.Context, db *gorm.DB, user *models.Identity, provider *models.Provider, oidcClient providers.OIDCClient) error {
```
BruceMacD, line 57: Moved the `SyncProviderUser` function to the data package to the data package to avoid a circular dependency during testing and migrations. In general it makes more sense for this part of the user sync to live here because it interacts with the database, and that should only happen in the data package.

internal/server/handlers.go
```diff
@@ -637,5 +628,35 @@ func (a *API) providerClient(c *gin.Context, provider *models.Provider, redirect
 		return nil, fmt.Errorf("client secret not found")
 	}
 
-	return providers.NewOIDC(*provider, clientSecret, redirectURL), nil
+	return providers.NewOIDCClient(*provider, clientSecret, redirectURL), nil
+}
+
+// setProviderInfoFromServer checks information provided by an OIDC server
+func (a *API) setProviderInforFromServer(c *gin.Context, provider *models.Provider) error {
```
BruceMacD, line 635: The API provider creation/update doesn't specify the auth URL and scopes. We look them up here. This means the API is backwards compatible.

We can assume that the `/.well-known/openid-configuration` endpoint is publicly accessible, or clients would not be able to login.

internal/server/data/migrationdata/202206281027-postgres.sql
```diff
@@ -0,0 +1,902 @@
...
+
+SET default_tablespace = '';
+
+SET default_table_access_method = heap;
+
+--
+-- Name: access_keys; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.access_keys (
```
mxyng, line 27: The postgres test drivers creates a schema `testing` and also sets the search_path to `testing`. I suspect this may not populate the database the way you're expecting

mxyng: As an aside, you can reduce the size of this file significantly by only exporting relevant tables
```
pg_dump --column-insert --table providers --table migrations
```

internal/server/data/migrations_test.go
```diff
@@ -161,3 +161,26 @@ func TestMigration_AddKindToProvider(t *testing.T) {
...
 	}
 }
+
+func TestMigration_AddAuthURLAndScopesToProvider(t *testing.T) {
+	for _, driver := range dbDrivers(t) {
+		t.Run(driver.Name(), func(t *testing.T) {
+			db, err := newRawDB(driver)
+			assert.NilError(t, err)
+
+			loadSQL(t, db, "202206151027-"+driver.Name())
```
mxyng, line 171: The file is incorrect. Should be 202206281027

```suggestion
			loadSQL(t, db, "202206281027-"+driver.Name())
```

mxyng (changes requested):

internal/server/data/migrationdata/202206281027-postgres.sql
BruceMacD: You're right, this was not running the way I expected. Fixing it actually exposed some bugs in the code. Thanks!

hoyyeva (approved):

mxyng (approved):

PR #2444, opened 2022-06-29
improve: include the server TLS CA in the chain

### LINKED ISSUE #2362: server: improve TLS certificate generation

dnephin:
## Summary

Branched from #2401

This ended up being easier than I thought. By adding the CA cert to the chain returned by the server the client side error already has the right cert.  Also log the CA fingerprint in the server on startup so that it is visible in logs.

Resolves #2362

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -592,13 +592,17 @@ func promptVerifyTLSCert(cli *CLI, cert *x509.Certificate) error {
 	formatTime := func(t time.Time) string {
 		return fmt.Sprintf("%v (%v)", HumanTime(t, "none"), t.Format(time.RFC1123))
 	}
+	title := "Certificate"
+	if cert.IsCA {
+		title = "Certificate Authority"
+	}
 
 	// TODO: improve this message
 	// TODO: use color/bold to highlight important parts
 	fmt.Fprintf(cli.Stderr, `
 The certificate presented by the server is not trusted by your operating system.
 
-Certificate
+%[6]v
 
 Subject: %[1]s
 Issuer: %[2]s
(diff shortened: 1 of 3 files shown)


BruceMacD (approved):
Nice and simple 👍

internal/cmd/login.go
```diff
@@ -622,13 +622,17 @@ func promptVerifyTLSCert(cli *CLI, cert *x509.Certificate) error {
 	formatTime := func(t time.Time) string {
 		return fmt.Sprintf("%v (%v)", HumanTime(t, "none"), t.Format(time.RFC1123))
 	}
+	title := "Certificate"
+	if cert.IsCA {
+		title = "Certificate Authority"
```
dnephin, line 627: I pushed this one more small change. To indicate in the prompt if the user is trusting a leaf cert or a CA.

PR #2445, opened 2022-06-30
fix: unrecognized lint nolint directive

mxyng:
## Summary

```
WARN [runner/nolint] Found unknown linters in //nolint directives: forcedtypeassert
```

Should be `forcetypeassert`


dnephin (approved):
Thanks!

PR #2447, opened 2022-06-30
fix: client delete request

mxyng:
## Summary

The API client delete request does not use the common `request` function which means its missing things like `Infra-Version`, `User-Agent`, and other common configurations.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -130,8 +130,10 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 	}
 
 	var resBody Res
-	if err := json.Unmarshal(body, &resBody); err != nil {
-		return nil, fmt.Errorf("parsing json response: %w. partial text: %q", err, partialText(body, 100))
+	if len(body) > 0 {
+		if err := json.Unmarshal(body, &resBody); err != nil {
+			return nil, fmt.Errorf("parsing json response: %w. partial text: %q", err, partialText(body, 100))
+		}
 	}
 
 	return &resBody, nil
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -130,8 +131,10 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 	}
 
 	var resBody Res
-	if err := json.Unmarshal(body, &resBody); err != nil {
-		return nil, fmt.Errorf("parsing json response: %w. partial text: %q", err, partialText(body, 100))
+	if reflect.TypeOf(resBody).Size() > 0 {
```
dnephin, line 134: This is a fix for the problem where the response body is empty, right? Seems like a good thing to fix here. There's another related problem where the response body might not be JSON if it came from some proxy in-between the client and the API (or the API failed in some unexpected way that did not return JSON), but no reason to fix that other issue as part of this change.

We should be able to avoid reflection here. I believe the error happens when `body` is empty (it doesn't matter so much what type `Res` is). Since `body is a `[]byte` I think this should work:

```suggestion
	if len(body) > 0 {
```

dnephin (commented):
Looks great, thanks for fixing this! Just some minor suggestions

api/client.go
dnephin: What should be behaviour be in that case? If the server did not error, wouldn't an empty success response be appropriate?

api/client_test.go
```diff
@@ -123,3 +123,39 @@ func TestGet(t *testing.T) {
 		assert.DeepEqual(t, req.Header, expectedHeaders)
 	})
 }
+
+func TestDelete(t *testing.T) {
+	ch := make(chan *http.Request, 1)
+	handler := func(rw http.ResponseWriter, r *http.Request) {
+		ch <- r
+		assert.Equal(t, r.Method, http.MethodDelete)
+		assert.Equal(t, r.URL.Path, "/good")
```
dnephin, line 132: Since this runs in a different goroutine it's going to panic instead of give a nice error message.

I think these would be better off below next to the `assert.DeepEqual`, so that the failure message is easier to read.

api/client.go
mxyng: In that case the client expects an non-empty response and will run into the same json marshalling error. It would be a bug in the server if it erroneously returns an empty response. I guess one could argue it's out of scope of the client to check

dnephin (approved):
LGTM!

PR #2454, opened 2022-06-30
maintain: organize the connector code

dnephin:
This PR moves some of the connector code around. Doing it separate from any other refactors or code changes so that it's easier to review. 

* Move everything related to authenticating requests into authn.go
* Move the rest of the proxy code into proxy.go
* Move updateRoles to the bottom of connector.go so that the Run function is at the top

There are a few things that can be extracted out of `Run` and moved into the proxy file, but I want to do that separately so it's easier to see the changes.

In the future maybe we'll move the reconcile logic into a separate file as well, and `connector.go` can be just the generic parts.

### DIFF
--- a/internal/connector/authn.go
+++ b/internal/connector/authn.go
@@ -1,0 +1,133 @@
+package connector
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"fmt"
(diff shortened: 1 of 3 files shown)


mxyng (approved):

jmorganca (approved):
Thanks for this!

PR #2457, opened 2022-06-30
fix(docs): remove bash syntax highlighting

### LINKED ISSUE #1629: Docs: highlight or dont in the kubernetes docs
## Describe the bug
on this page: https://infrahq.com/docs/getting-started/introduction
CleanShot 2022-04-14 at 11 53 58@2x
Only the word 'in' is color coded here. Seems strange. Either do the syntax highlighting throughout or not at all. Would prefer none

## Environment

```
$ infra version
```

```
$ kubectl version
```

technovangelist:
Signed-off-by: Matthew Williams <m@technovangelist.com>

## Summary

When you have 3 back ticks and then the word bash then some
code, it auto highlights this. While it works well for yaml, it
looks horrible for bash. We only had 2 or 3 instances of this, so
removed them. Resolves #1629.

## Checklist

its just docs, so none of these are really relevant

## Related Issues

Resolves #1629


dnephin (commented):
I wonder if something like `shell` or `shell-session` works instead of `bash`. `bash` generally means a script, where as these are examples of interactive shells.

I'm not sure what library we use for this syntax highlighting , so I'm not sure how to look up if `shell` is an option or not.

dnephin:
Ah, I think we use prismjs, so according to https://prismjs.com/#supported-languages, `shell-session` should work. Want to try that out to see how it looks ? I guess the Vercel preview should work so we can preview it before merge.

mxyng:
This is probably what we want: https://prismjs.com/plugins/command-line/. `shell-session` seems to look very similar to bash

technovangelist:
looking at each of those suggestions i don't see much value add. i think we just keep it simple .

dnephin (approved):
> looking at each of those suggestions i don't see much value add.

Did you try it out?

I would expect `shell-session` to highlight the comments and `$` making it a bit easier to read. https://www.consul.io/commands/acl/auth-method#basic-examples is an example of it on another site, but I couldn't find any examples that have comments.

Definitely no syntax highlighting is better than wrong syntax highlighting, so I think it's worth merging this, but for future docs it would be great to figure out if `shell-session` is the right thing to use.

The command-line plugin looks interesting too, but it looks like that's probably more work to setup (requires the plugin and different syntax from what I can tell).

PR #2458, opened 2022-06-30
helm: refactor to not create a new CA on every connector deploy

dnephin:
This PR changes the infra-connector-ca secret helm template to only `genCA` if there is not an existing secret. Also adds a few comments to the `{{ end }}` tags to make it easier to follow.

This `secret.yaml` is modeled after the one for infra server, added in #2401.

### DIFF
--- a/helm/charts/infra/templates/connector/secret.yaml
+++ b/helm/charts/infra/templates/connector/secret.yaml
@@ -1,3 +1,4 @@
+
 {{- if include "connector.enabled" . | eq "true" }}
 {{- $accessKey := default "" .Values.connector.config.accessKey -}}
 {{- if or (not $accessKey) (and (not (hasPrefix "file:" $accessKey)) (not (hasPrefix "env:" $accessKey))) }}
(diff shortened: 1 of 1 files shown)


helm/charts/infra/templates/connector/secret.yaml
```diff
@@ -9,29 +10,32 @@ metadata:
 {{- include "connector.labels" . | nindent 4 }}
 data:
   access-key: {{ include "connector.accessKey" . | b64enc | quote }}
-{{- end }}
+{{- end }}{{/* if not accessKey */}}
 ---
 {{- if and (not .Values.connector.config.caCert) (not .Values.connector.config.caKey) }}
-{{- $secret := lookup "v1" "Secret" .Release.Namespace (printf "%s-ca" (include "connector.fullname" .)) -}}
-{{- $ca := genCA "infra-connector" 3650 -}}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace (printf "%s-ca" (include "connector.fullname" .)) }}
```
dnephin, line 16: maybe it would make sense to move this lower too, so it's closer to the if

helm/charts/infra/templates/connector/secret.yaml
```diff
@@ -9,29 +10,32 @@ metadata:
...
+
 apiVersion: v1
 kind: Secret
 metadata:
   name: {{ include "connector.fullname" . }}-ca
   labels:
 {{- include "connector.labels" . | nindent 4 }}
-data:
-  ca.crt: |
+stringData:
```
mxyng, line 24: Not sure why this is being changed from `data`. It's only exchanging `b64enc` on `genCA` outputs for `b64dec` on existing secrets

mxyng (approved):
Minor comment but otherwise looks good

PR #2461, opened 2022-07-02
fix(ui): show correct error message when using temporary password

jmorganca:
Fixes issue where temporary password verification would fail and user wouldn't receive an error message


BruceMacD (approved):
looks good

PR #2481, opened 2022-07-04
fix: find server endpoint for localhost

### LINKED ISSUE #2452: Connector attempt to list destinations results in "unauthorized" response
## Describe the bug
Observed on mainline. 

Steps to reproduce:
1. Stand up a new infra server (mainline build as of `563d6d1`)
2. Login and create a new connector key: `infra keys add connector`
3. Deploy a connector using the dev image and this access key:
```
helm install infra-connector helm/charts/infra --set connector.config.name=docker-desktop --set connector.config.accessKey=${CONNECTOR_KEY} --set connector.config.server=localhost --set connector.config.skipTLSVerify=true --set global.image.pullPolicy=Never --set global.image.tag=dev
```
4. Connector fails to register on the `GET /api/destinations` call due to a 401 unauthorized response.

## Expected behavior
The connector can access the `/api/destinations` endpoint.

## Environment

```
$ infra version
Client: 0.13.6+dev
Server: 0.13.5
```

## Additional context
Connector logs failing:
```
{"level":"error","time":1656619807887,"caller":"logging/logger.go:58","message":"initializing destination: error listing destinations: 401 unauthorized"}
```

It appears to succeed in the server logs:
```
{"level":"info","method":"GET","path":"/api/destinations","host":"localhost","remoteAddr":"192.168.65.3:60914","userAgent":"Infra/0.13.0 (agent 0.13.5; darwin/arm64)","contentLength":0,"errors":[],"elapsed":6.244791,"statusCode":200,"size":43,"time":1656619814614,"caller":"logging/middleware.go:41"}
```

BruceMacD:
- make service labels variadic

## Summary
When looking up "localhost" as the server name in the connector (as seen in the docs) the request was failing to find the infra server in the local pod. This was due to the labels being applied to the look-up, this change makes the labels optional and applies them where needed so that the local server look-up succeeds.

No tests at the moment because using a mock Kubernetes clientset will require some refactoring in our Kubernetes code.

## Checklist

## Related Issues

Resolves #2452

### DIFF
--- a/internal/kubernetes/kubernetes.go
+++ b/internal/kubernetes/kubernetes.go
@@ -529,18 +529,13 @@ func CA() ([]byte, error) {
 }
 
 // Find the first suitable Service, filtering on infrahq.com/component
-func (k *Kubernetes) Service(component string) (*corev1.Service, error) {
+func (k *Kubernetes) Service(component string, labels ...string) (*corev1.Service, error) {
 	clientset, err := kubernetes.NewForConfig(k.Config)
 	if err != nil {
 		return nil, err
 	}
 
 	namespace, err := Namespace()
-	if err != nil {
-		return nil, err
-	}
-
-	labels, err := PodLabels()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


mxyng (approved):
Nice! Thanks for fixing this

internal/kubernetes/kubernetes.go
```diff
@@ -529,7 +529,7 @@ func CA() ([]byte, error) {
 }
 
 // Find the first suitable Service, filtering on infrahq.com/component
-func (k *Kubernetes) Service(component string) (*corev1.Service, error) {
+func (k *Kubernetes) Service(component string, labels ...string) (*corev1.Service, error) {
```
dnephin, line 532: It seems really surprising to me that this label arg will be filtered and only the label matching `app.kubernetes.io/instance` will be used.

Maybe that filtering should also be done by the caller?

At the very least we should document how labels are used in the godoc above here.

BruceMacD: Good point, switched the filtering to be done by the callers

PR #2483, opened 2022-07-04
Logging improvements and fixes

dnephin:
## Summary

There are a few small things here, so best viewed by individual commit.

* restore the console logger when server is run in an interactive terminal
* improve the output of the console logger (better level names, and remove timestamp and caller)
* expose `PatchLogger` for tests, instead of `NewLogger` and having each package duplicate the patch function


internal/logging/logger.go
```diff
@@ -36,6 +37,10 @@ func NewLogger(writer io.Writer) *logger {
 }
 
 func UseServerLogger() {
+	// If the server is run from an interactive terminal, use the default ConsoleWriter
+	if os.Stdin != nil && term.IsTerminal(int(os.Stdin.Fd())) {
```
dnephin: We had this before #2432, I think it must have been lost in that change. If a server or connector is run with an interactive terminal then we want the ConsoleLogger.

This is different from short lived CLI commands like `infra grants list`. If short lived CLI commands are run without an interactive terminal we still want a ConsoleLogger, but we want color to be disabled (fixed in a later commit).

internal/logging/logger.go
```diff
@@ -44,6 +50,19 @@ func UseServerLogger() {
 	L = newLogger(os.Stderr)
 }
 
+// UseFileLogger changes L to a logger that writes log output to a file that is
+// rotated.
+func UseFileLogger(filepath string) {
```
dnephin, line 60: Moved from `file.go` since that file contained only this one function (the type was unused), and so that all things that modify `L` are close together in this one file.

internal/logging/zerolog.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+
+	"github.com/rs/zerolog"
+)
+
+// consoleDefaultFormatLevel is copied from zerolog/console.go to modify the
+// names and colors used for levels.
+func consoleFormatLevel(i interface{}) string {
```
dnephin, line 11: These were not exported, so I couldn't see any other way of fixing this than to copy it and modify it.

mxyng (approved):

PR #2485, opened 2022-07-04
connector: verify the server TLS

dnephin:
## Summary

Branched from #2401

This PR adds support to the connector to trust the TLS CA or certificate of the server.

With this change, someone deploying a connector can set the trusted certificate instead of `skipTLSVerify`:

```
helm upgrade --install infra-connector ... --set connector.config.serverTrustedCertificate="$(cat server.ca)"
```

I think we can improve this by fetching the CA for them, if they've already trusted it in the CLI. We might also want to fetch it out of a previous config map to remove the need to set it every time. I think those could be follow up PRs.

### DIFF
--- a/helm/charts/infra/templates/connector/configmap.yaml
+++ b/helm/charts/infra/templates/connector/configmap.yaml
@@ -7,35 +7,34 @@ metadata:
 {{- include "connector.labels" . | nindent 4 }}
 data:
   infra.yaml: |
-{{- range $key, $val := omit .Values.connector.config "accessKey" }}
-{{- if kindIs "invalid" $val }}
-    # skipping invalid value: {{ $val }} ({{ kindOf $val }})
-{{- else if kindIs "map" $val }}
-    {{ $key }}:
-{{- $val | toYaml | nindent 6 }}
-{{- else if kindIs "slice" $val }}
-    {{ $key }}:
-{{- $val | toYaml | nindent 6 }}
-{{- else if kindIs "string" $val }}
-    {{ $key }}: {{ tpl $val $ }}
-{{- else }}
-    {{ $key }}: {{ $val }}
-{{- end }}
+
+{{- with .Values.connector.config.name }}
+    name: {{ . }}
 {{- end }}
 
+    server:
 {{- $accessKey := default "" .Values.connector.config.accessKey }}
 {{- if and $accessKey (or (hasPrefix "file:" $accessKey) (hasPrefix "env:" $accessKey)) }}
-    accessKey: {{ $accessKey }}
+      accessKey: {{ $accessKey }}
 {{- else }}
-    accessKey: file:/var/run/secrets/infrahq.com/access-key/access-key
+      accessKey: file:/var/run/secrets/infrahq.com/access-key/access-key
+{{- end }}
+
+{{- with .Values.connector.config.serverTrustedCertificate }}
+      trustedCertificate: |
+{{ indent 8 . }}
+{{- end }}
+
+{{- with .Values.connector.config.server }}
+      url: {{ . }}
 {{- end }}
 
 {{- if include "server.enabled" . | eq "true" }}
-    server: {{ .Release.Name }}-server.{{ .Release.Namespace }}
+      url: {{ .Release.Name }}-server.{{ .Release.Namespace }}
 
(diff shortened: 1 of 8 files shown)


internal/connector/connector.go
```diff
@@ -30,12 +31,17 @@ import (
...
+	Name   string
+	CACert string
+	CAKey  string
+}
+
+type ServerOptions struct {
+	URL                string
+	AccessKey          string
+	SkipTLSVerify      bool
+	TrustedCertificate types.StringOrFile
```
dnephin, line 44: I think adding more TLS options to the top level config is not great. There are two separate trust roots here, and it needs to be clear which settings apply to each.

Unfortunately creating this extra structure in the config causes a few problems. It changes the name of the flags. We can make that backwards compatible if we need to.

It also makes the helm configmap for the connector more difficult because `accessKey` can come from different places.

internal/connector/connector.go
```diff
@@ -84,9 +90,9 @@ func Run(ctx context.Context, options Options) error {
 		"plaintext": secrets.NewPlainSecretProviderFromConfig(secrets.GenericConfig{}),
 	}
 
-	u, err := urlx.Parse(options.Server)
+	u, err := urlx.Parse(options.Server.URL)
 	if err != nil {
-		logging.Errorf("server: %s", err)
+		return fmt.Errorf("invalid server url: %s", err)
```
dnephin, line 95: Seems like this should be an error (not simply logging), because it will panic on the next line if we don't return here.

helm/charts/infra/templates/connector/configmap.yaml
```diff
@@ -7,35 +7,30 @@ metadata:
...
-{{- $val | toYaml | nindent 6 }}
-{{- else if kindIs "slice" $val }}
-    {{ $key }}:
-{{- $val | toYaml | nindent 6 }}
-{{- else if kindIs "string" $val }}
-    {{ $key }}: {{ tpl $val $ }}
-{{- else }}
-    {{ $key }}: {{ $val }}
-{{- end }}
-{{- end }}
```
dnephin, line 24: This generic "import all config" doesn't work anymore now that `server` changed from a string to a struct.  We already have explicit handling of all the options, so we don't need this I think. Users can always inject specific config settings using env vars, which I think works better here because it doesn't require type checking the value.

dnephin (commented):
This should be ready for review now

helm/charts/infra/templates/connector/configmap.yaml
```diff
@@ -7,35 +7,30 @@ metadata:
 {{- include "connector.labels" . | nindent 4 }}
 data:
   infra.yaml: |
```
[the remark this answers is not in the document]
dnephin: Done

internal/connector/connector.go
BruceMacD: What does this the upgrade process look like for this new config? Could this cause unexpected behavior?

BruceMacD (commented):
Looks good pending my upgrade question, a bit nervous of this breaking current configs on upgrade from your description of that change.

internal/connector/connector.go
dnephin: Good question! If someone is running the connector directly they would have to change how they configure it. I believe right now we are only supporting running the connector via the helm chart. In the helm chart we are using the old values and translating it into the new structure, so anyone using the helm chart won't notice the change in structure.

I think we should change the structure in the helm chart eventually as well. I tried it initially but it got pretty involved, so I thought we could wait a bit to try and bundle it with other changes.

BruceMacD: Thanks, that doesnt sound that concerning

BruceMacD (approved):

internal/connector/connector.go
dnephin: @mxyng does that sound right to you as well?

I guess an old configmap will not be compatible with a new connector, and vice versa, but any `helm upgrade` should change both together.

mxyng: The `configmap` will be updated automatically but the service isn't configured for dynamic reloads. However, Helm upgrades will roll out new pods if the config changes. If the pod image version is updated at the same time, everything will work as expected since the config and service are upgraded in sync. If, for whatever reason, the new pod starts with the old service version, e.g. if the user pinned the connector image tag, the new pod will error

PR #2486, opened 2022-07-05
feat: better feedback for connector key

### LINKED ISSUE #2293: Keys created for the connector expire after 30 days by default
## Describe the bug
In our current [connector setup flow](https://infrahq.com/docs/connectors/kubernetes) we create an access key that is valid for 30 days. After these 30 days the key will expire and the connector will need a new access key. This is not apparent from our setup flow.

## Expected behavior
Either:
 - The process to setup automatically adding a new connector is documented simply.
 - Access keys can be refreshed within a time period or exchanged
 - Not ideal: The access key has a longer lifetime. If the connector key ever leaked this would increase the amount of time it could be used to add destinations to Infra.

## Additional context
- https://infrahq.com/docs/connectors/kubernetes
- https://infrahq.com/docs/getting-started/quickstart

BruceMacD:
- output lifetime and extension deadline from keys add command
- specify when an access key was valid but is now expired

## Summary
The connector key expired after 30 days by default. This was not clear in our workflow or docs and caused an unexpected failure for connectors after the 30 day period ends. This change alleviates the problem by specifying the access key lifetime on creation and giving info in the error response for expired keys that detail the actual problem.

The connector log will look something like this when the access key expires:
```
{"level":"error","time":1657026989645,"caller":"logging/logger.go:58","message":"initializing destination: error listing destinations: unauthorized: access key expired: extension deadline exceeded"}
```
## Checklist

## Related Issues

Resolves #2293

### DIFF
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -11,7 +11,7 @@ import (
 	"github.com/infrahq/infra/internal/logging"
 )
 
-const ThirtyDays = 30 * (24 * time.Hour)
+const thirtyDays = 30 * (24 * time.Hour)
 
 func newKeysCmd(cli *CLI) *cobra.Command {
 	cmd := &cobra.Command{
@@ -97,15 +97,26 @@ $ infra keys add connector
 				return err
 			}
 
+			var expMsg strings.Builder
+			expMsg.WriteString("This key will expire in ")
+			expMsg.WriteString(ExactDuration(options.TTL))
+			if !resp.Expires.Equal(resp.ExtensionDeadline) {
+				expMsg.WriteString(", and must be used every ")
+				expMsg.WriteString(ExactDuration(options.ExtensionDeadline))
+				expMsg.WriteString(" to remain valid")
+			}
 			cli.Output("Issued access key %q for %q", resp.Name, userName)
+			cli.Output(expMsg.String())
+			cli.Output("")
+
 			cli.Output("Key: %s", resp.AccessKey)
 			return nil
 		},
 	}
 
 	cmd.Flags().StringVar(&options.Name, "name", "", "The name of the access key")
-	cmd.Flags().DurationVar(&options.TTL, "ttl", ThirtyDays, "The total time that the access key will be valid for")
-	cmd.Flags().DurationVar(&options.ExtensionDeadline, "extension-deadline", ThirtyDays, "A specified deadline that the access key must be used within to remain valid")
+	cmd.Flags().DurationVar(&options.TTL, "ttl", thirtyDays, "The total time that the access key will be valid for")
+	cmd.Flags().DurationVar(&options.ExtensionDeadline, "extension-deadline", thirtyDays, "A specified deadline that the access key must be used within to remain valid")
 
 	return cmd
 }
--- a/internal/server/data/errors.go
+++ b/internal/server/data/errors.go
@@ -1,0 +1,8 @@
+package data
+
+import "fmt"
+
+var (
+	ErrAccessKeyExpired          = fmt.Errorf("access key expired")
+	ErrAccessKeyDeadlineExceeded = fmt.Errorf("%w: extension deadline exceeded", ErrAccessKeyExpired)
+)
--- a/internal/server/data/errors.go
+++ b/internal/server/data/errors.go
@@ -1,8 +1,0 @@
-package data
-
-import "fmt"
-
-var (
-	ErrAccessKeyExpired          = fmt.Errorf("access key expired")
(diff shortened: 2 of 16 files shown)


internal/cmd/keys.go
```diff
@@ -97,15 +97,26 @@ $ infra keys add connector
...
+
 			cli.Output("Key: %s", resp.AccessKey)
 			return nil
 		},
 	}
 
 	cmd.Flags().StringVar(&options.Name, "name", "", "The name of the access key")
-	cmd.Flags().DurationVar(&options.TTL, "ttl", ThirtyDays, "The total time that the access key will be valid for")
-	cmd.Flags().DurationVar(&options.ExtensionDeadline, "extension-deadline", ThirtyDays, "A specified deadline that the access key must be used within to remain valid")
+	cmd.Flags().DurationVar(&options.TTL, "ttl", thirtyDays, "The total time that the access key will be valid for")
```
BruceMacD, line 118: Ended up leaving the default lifetime for all access keys as 30 days here, rather than creating a special case for connector keys. In order to check if the flag was specified we would have to set and invalid default value and check for that, but this solution messes with the go doc.

internal/server/data/errors.go
```diff
@@ -0,0 +1,8 @@
+package data
+
+import "fmt"
+
+var (
+	ErrAccessKeyExpired          = fmt.Errorf("access key expired")
+	ErrAccessKeyDeadlineExceeded = fmt.Errorf("%w: extension deadline exceeded", ErrAccessKeyExpired)
```
dnephin, line 7: minor: I usually try to keep package-level variables like this in the same file where they are used. It makes it much easier to split up a package without a large diff, because files can be `git mv` instead of having to copy bits from multiple files.

I find grouping things by feature or domain instead of grouping by kind often works out well.

internal/server/errors.go
```diff
@@ -41,6 +41,11 @@ func sendAPIError(c *gin.Context, err error) {
 		// log the error at info because it is not in the response
 		log = logging.L.Info()
 
+	case errors.Is(err, data.ErrAccessKeyExpired):
+		resp.Code = http.StatusUnauthorized
+		// this means the key was once valid, so include some extra details
+		resp.Message = fmt.Sprintf("%s: %s", internal.ErrUnauthorized, err)
```
dnephin, line 47: Nice! It would be great to add a test case for this as well, in `errors_test.go`

internal/cmd/keys.go
```diff
@@ -97,15 +97,26 @@ $ infra keys add connector
...
+			var expMsg strings.Builder
+			expMsg.WriteString("This key will expire in ")
+			expMsg.WriteString(ExactDuration(options.TTL))
+			if !resp.Expires.Equal(resp.ExtensionDeadline) {
+				expMsg.WriteString(", and must be used every ")
+				expMsg.WriteString(ExactDuration(options.ExtensionDeadline))
+				expMsg.WriteString(" to remain valid")
+			}
 			cli.Output("Issued access key %q for %q", resp.Name, userName)
+			cli.Output(expMsg.String())
```
dnephin, line 109: It would be great to capture this in `TestKeysAddCmd` so that we can see the rendered message. We probably don't even need new test cases, just a new assertion for the output.

We can use the new "golden variable" support in `gotest.tools/v3/assert`. Since we don't have any example of that yet, I put together a commit which adds this test coverage: https://github.com/infrahq/infra/compare/brucemacd/connector-key-feedback...dnephin/demo-golden

dnephin (commented):
Looks great! Lots of good test coverage here. I left a couple small suggestions for additional test coverage

internal/cmd/keys.go
BruceMacD: Thanks for the example, that would have taken me a bit to figure out. I've cherry-picked this commit.

internal/server/errors_test.go
```diff
@@ -41,6 +41,18 @@ func TestSendAPIError(t *testing.T) {
...
+			result: api.Error{Code: http.StatusUnauthorized, Message: "unauthorized: " + data.ErrAccessKeyExpired.Error()},
+		},
+		{
+			err:    data.ErrAccessKeyDeadlineExceeded,
+			result: api.Error{Code: http.StatusUnauthorized, Message: "unauthorized: " + data.ErrAccessKeyDeadlineExceeded.Error()},
+		},
+		{
+			err:    fmt.Errorf("hide this: %w", internal.ErrUnauthorized),
+			result: api.Error{Code: http.StatusUnauthorized, Message: "unauthorized"},
+		},
```
dnephin, line 55: I think this test case may have been duplicated. Seems like we have it just above the two new ones as well.

dnephin (approved):
Looks great! One small thing, and also +1 to using `errors.Is` for the test, but nothing blocking a merge

PR #2487, opened 2022-07-05
improve: change json output

### LINKED ISSUE #972: cli: Support YAML for `--format` flag

kimskimchi:
## Checklist

## Related Issues

Resolves #972

### DIFF
--- a/internal/cmd/destinations.go
+++ b/internal/cmd/destinations.go
@@ -51,7 +51,7 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
 
 			switch format {
 			case "json":
-				jsonOutput, err := json.Marshal(destinations)
+				jsonOutput, err := json.Marshal(destinations.Items)
 				if err != nil {
 					return err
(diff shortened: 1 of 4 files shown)


BruceMacD:
I notice the original ticket (#972) has a top level key for JSON output (resources, rather than items). Was getting rid of the top-level key discussed out-of-band? Otherwise looks good.

mxyng:
It's probably fine to omit `items`. I think the goal is to filter out meta field like `count` and `pagination` which this accomplishes. The open question is will there be other top-level fields this should output?

mxyng (approved):

PR #2490, opened 2022-07-05
feat: add destination status and last seen to CLI and UI

### LINKED ISSUE #2151: Need to show destination status to know if a cluster is still alive

mxyng:
## Summary

- Add Destination `Status` API field
- Add columns for Status and Last Seen to CLI and UI list outputs

```
$ infra destinations list
  NAME            URL                 STATUS        LAST SEEN
  docker-desktop  localhost:123       Disconnected  never
  docker-desktop  localhost:123       Disconnected  never
  docker-desktop  10.98.125.93:10443  Connected     1 second ago
```
image

## Checklist

## Related Issues

Resolves #2151

### DIFF
--- a/internal/cmd/destinations.go
+++ b/internal/cmd/destinations.go
@@ -10,6 +10,11 @@ import (
 	"github.com/infrahq/infra/internal/logging"
 )
 
+const (
+	DestinationStatusConnected    = "Connected"
+	DestinationStatusDisconnected = "Disconnected"
+)
+
 func newDestinationsCmd(cli *CLI) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "destinations",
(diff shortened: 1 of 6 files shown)


ui/pages/destinations/index.js
```diff
@@ -290,13 +290,33 @@ const columns = [
...
-    width: '33%',
+    width: '25%',
     Cell: ({ value }) => (
       <span className='rounded bg-gray-800 px-2 py-0.5 text-gray-400'>
         {value.kind}
       </span>
     ),
   },
+  {
+    Header: 'Status',
```
mxyng, line 301: This doesn't necessarily need a column. An alternative may be to add it as an icon in the `Last Seen` column @FSHA

ui/pages/destinations/index.js
```diff
@@ -305,13 +305,27 @@ const columns = [
...
+    accessor: v => v,
+    width: '15%',
+    Cell: ({ row, value }) => (
+      <div className='items-center py-2'>
```
jmorganca, line 320: Missing `flex`

internal/server/models/destination.go
```diff
@@ -21,6 +21,12 @@ type Destination struct {
 }
 
 func (d *Destination) ToAPI() *api.Destination {
+	connected := false
+	// TODO: this should be configurable
```
jmorganca, line 25: Nit: create an issue for this TODO?

internal/cmd/destinations.go
```diff
@@ -58,15 +58,24 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
...
-					URL  string `header:"URL"`
+					Name     string `header:"NAME"`
+					URL      string `header:"URL"`
+					Status   string `header:"STATUS"`
+					LastSeen string `header:"LAST SEEN"`
 				}
 
 				var rows []row
 				for _, d := range destinations.Items {
+					status := "Disconnected"
```
jmorganca, line 69: Nit: make these constants / enums?

jmorganca (approved):
LGTM – two smaller comments. Nice work!

PR #2492, opened 2022-07-05
fix: create a separate session for each db query

dnephin:
## Background

After reading https://gorm.io/docs/method_chaining.html I think I finally understand how gorm queries are supposed to work.  The important bit is this:

>  After a Chain method [or] Finisher Method, GORM returns an initialized `*gorm.DB` instance, which is NOT safe to reuse anymore, or new generated SQL might be polluted by the previous conditions.

It sounds like a "gorm session" is how we are supposed to be building queries. A session is created with `db.Session`, or `db.WithContext` (which is a thin wrapper around `db.Session`).

Today we create a new session in `DatabaseMiddleware` with `WithContext` . The first chain method called on an empty session will create a new clone, so in most cases we did not run into problems because of that session.

Previous to  #1871 we had this extra `db2` variable, but that did not actually create a new session, so that wasn't addressing the problem either.  The tests I added in that PR used `WithContext` (like our middleware) to create a new session, which is why it all looks good in that test. 

## Summary

This PR adds a call to `db.Session` before making any modifications to `gorm.DB`, which should prevent one query from interfering with others.

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -105,6 +105,7 @@ func getDefaultSortFromType(t interface{}) string {
 }
 
 func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error) {
+	db = db.Session(&gorm.Session{})
 	for _, selector := range selectors {
 		db = selector(db)
 	}
@@ -122,6 +123,7 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 }
 
 func list[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) ([]T, error) {
+	db = db.Session(&gorm.Session{NewDB: true})
 	db = db.Order(getDefaultSortFromType((*T)(nil)))
 	for _, selector := range selectors {
 		db = selector(db)
@@ -141,6 +143,7 @@ func save[T models.Modelable](db *gorm.DB, model *T) error {
 		return err
 	}
 
+	db = db.Session(&gorm.Session{NewDB: true})
 	err := db.Save(model).Error
 	return handleError(err)
 }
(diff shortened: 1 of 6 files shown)


internal/server/data/data.go
```diff
@@ -105,6 +105,7 @@ func getDefaultSortFromType(t interface{}) string {
 }
 
 func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error) {
+	db = db.Session(&gorm.Session{})
```
dnephin, line 108: None of the fields on this `gorm.Session` struct has any godoc, so I have no idea which ones should be set. I'm guessing none, since it clones the config of the original `gorm.DB`.

mxyng (approved):

dnephin:
On further testing I'm no longer sure this helps. It seems like the first chained method (`Where`, etc) will do a clone, and calling `Session` another time does not clone again.  Going to think about this one some more.

dnephin:
I amended the commit to add `NewDB: true` to the session. This seems to do the right thing and zero out previous statements.

dnephin:
Ok, I pushed another commit which fixes the two test failures caused by `Session{NewDB: true}`.

internal/server/data/selectors.go
```diff
@@ -210,3 +210,7 @@ func ByOptionalIdentityGroupID(groupID uid.ID) SelectorFunc {
 			Where("identities_groups.group_id = ?", groupID)
 	}
 }
+
+func IncludeSoftDeleted(db *gorm.DB) *gorm.DB {
+	return db.Unscoped()
```
ssoroka, line 215: may cause problems if we use other things with scopes later.

ssoroka (changes requested):
This seems mostly unnecessary. The first call on a `db` variable is already creating a copy of the db transparently. It's only in the selectors that we had any issue, and this doesn't resolve the issues that selectors have making additional queries.

ssoroka (approved):
this is good to go.

dnephin:
Going to close this out for now. I think we understand the session behaviour well enough to catch problems, and replacing selectors funcs should also mitigate the problem.

PR #2493, opened 2022-07-05
improve: support INFRA_SERVER and INFRA_ACCESS_KEY env variables

### LINKED ISSUE #2286: Replace login via access key with specifying access key as an argument or environment variable
Infra's access keys are meant to be used as API keys. They should not be used in the login flow.

1. Remove logging in `infra login` via access key
2. Allow `INFRA_ACCESS_KEY` and `INFRA_SERVER` (or similar) as environment variables
3. Optionally allow `--access-key` as a global flag for the CLI – although we shouldn't encourage using access key in plain text

ssoroka:
## Summary

This allows you to log in with an env-variable access key and removes the option from the menu to discourage it.  

## Checklist

## Related Issues

Resolves #2286

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -34,21 +34,48 @@ func Run(ctx context.Context, args ...string) error {
 	return cmd.ExecuteContext(ctx)
 }
 
-func mustBeLoggedIn() error {
+func tryLogin(cli *CLI) error {
+	if accessKey, isSet := os.LookupEnv("INFRA_ACCESS_KEY"); isSet {
+		server, isSet := os.LookupEnv("INFRA_SERVER")
+		if !isSet {
+			return errors.New("INFRA_ACCESS_KEY environment variable is set but INFRA_SERVER is not. please set INFRA_SERVER to log in automatically.")
+		}
+		return login(cli, loginCmdOptions{
+			Server:         server,
+			AccessKey:      accessKey,
+			NonInteractive: true,
+		})
+	}
+
+	return nil
+}
+
+func mustBeLoggedIn(cli *CLI) error {
+	tryLoginOrError := func(err Error) error {
+		if _, isSet := os.LookupEnv("INFRA_ACCESS_KEY"); isSet {
+			if err2 := tryLogin(cli); err2 != nil {
+				cli.Output(err2.Error())
+				return err
+			}
+			return nil
+		}
+		return err
+	}
+
 	config, err := currentHostConfig()
 	if err != nil {
 		if errors.Is(err, ErrConfigNotFound) {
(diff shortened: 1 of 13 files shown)


internal/cmd/cmd.go
```diff
@@ -34,7 +34,12 @@ func Run(ctx context.Context, args ...string) error {
 	return cmd.ExecuteContext(ctx)
 }
 
-func mustBeLoggedIn() error {
+func mustBeLoggedIn(cli *CLI) error {
+	if _, isSet := os.LookupEnv("INFRA_ACCESS_KEY"); isSet {
```
mxyng, line 38: Why `isSet` for some but `ok` for other?

internal/cmd/cmd.go
```diff
@@ -78,7 +83,18 @@ func defaultAPIClient() (*api.Client, error) {
...
+	server := config.Host
+	accessKey := config.AccessKey
+
+	if envAccessKey, isSet := os.LookupEnv("INFRA_ACCESS_KEY"); isSet {
+		accessKey = envAccessKey
+	}
+
+	if envServer, ok := os.LookupEnv("INFRA_SERVER"); ok {
+		server = envServer
+	}
```
mxyng, line 95: To be clear, this functions as an override correct? So it can be any combination of server from config or env var and access key from config or env var?

internal/cmd/cmd.go
```diff
@@ -34,7 +34,12 @@ func Run(ctx context.Context, args ...string) error {
 	return cmd.ExecuteContext(ctx)
 }
 
-func mustBeLoggedIn() error {
+func mustBeLoggedIn(cli *CLI) error {
```
mxyng, line 37: What's `cli` used for?

internal/cmd/login.go
```diff
@@ -65,10 +64,15 @@ $ infra login infraexampleserver.com
 $ infra login --provider okta
 
 # Login with an access key
-$ infra login --key 1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy`,
+$ export INFRA_ACCESS_KEY=1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy
+$ infra login`,
 		Args:  MaxArgs(1),
 		Group: "Core commands:",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if server, ok := os.LookupEnv("INFRA_SERVER"); ok {
```
mxyng, line 72: This means positional args > env vars right?

mxyng (approved):

internal/cmd/login.go
ssoroka: yeah

PR #2496, opened 2022-07-06
feat: yaml output

### LINKED ISSUE #972: cli: Support YAML for `--format` flag

kimskimchi:
## Summary

Introduces `yaml` for existing `json` outputs. 
Comes with free tests. 👍 

## Checklist

## Related Issues

Resolves #972


internal/cmd/providers_test.go
```diff
@@ -130,4 +130,29 @@ func TestProvidersAddCmd(t *testing.T) {
...
+		t.Setenv("INFRA_PROVIDER_CLIENT_ID", "okta-client-id")
+		t.Setenv("INFRA_PROVIDER_CLIENT_SECRET", "okta-client-secret")
+
+		err := Run(ctx, "providers", "add", "okta")
+		assert.NilError(t, err)
+
+		err = Run(ctx, "providers", "list", "--format=yaml")
+		assert.NilError(t, err)
+
+		assert.Assert(t, strings.Contains(bufs.Stdout.String(), `- id: ""
```
mxyng, line 148: Should consider using `golden` when testing these outputs

internal/cmd/destinations.go
```diff
@@ -56,6 +57,12 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
 					return err
 				}
 				cli.Output(string(jsonOutput))
+			case "yaml":
```
BruceMacD, line 60: Any way we could have a re-usable function for the output formatting to reduce the repetition in the cli output code here?

internal/cmd/providers_test.go
BruceMacD: Daniel did an example of what adding a golden test looks like in this commit of one of my recent PRs, good for reference:
https://github.com/infrahq/infra/pull/2486/commits/f8a29f49744f59f1b5ee1cb9d9464d5ed32305ab

internal/cmd/destinations.go
mxyng: Mentioned in the previous change. There should be a `Formatter` interface which `JSON`/`YAML`/`TableFormat` implements. The result is passed to `cli.Output` for printing. This will allow more formats to be added with minimal impact on the existing code base

kimskimchi: For the actual function, not sure how this can be achieved, could I have an example? 

The couple of commands here are all `list`, so it expects similar outputs, but different functions will not use the same `something.Items` but rather be completely different.

internal/cmd/providers_test.go
kimskimchi: done

internal/cmd/destinations.go
kimskimchi: resolved & done

dnephin (approved):
Looks good. We should move the `Conencted provider` output to `Stderr`, but otherwise this looks great.  I think we can refactor to a formatter interface in a follow up.

internal/cmd/destinations.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
```
dnephin, line 8: Oops, I missed this on earlier review. This should be `gopkg.in/yaml.v2`, which is the more standard yaml library. 


```suggestion
	"gopkg.in/yaml.v2"
```

internal/cmd/testdata/TestUsersCmd/list_with_json
```diff
@@ -0,0 +1,2 @@
+Added user "apple@example.com"
```
dnephin, line 1: This line too, should go to `Stderr`

internal/cmd/destinations.go
dnephin, line 8: For anyone else curious about this. It turns out we have to use `sigs.k8s.io/yaml` so that the field names come from the `json` tags. If we used the more standard yaml library we'd have to add `yaml` names to all the fields, and the output would look strange for some fields (ex: would use `{}` for the zero value of `api.Time` instead of `null`).

Although the second part of that we could fix by changing to `MarshalText` instead of `MashalJson`.

internal/cmd/destinations.go
dnephin: I put together an example of one option here: https://github.com/infrahq/infra/compare/cli-9...dnephin/formatter-example

But let's look into that in a follow up PR.

internal/cmd/users.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	survey "github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
```
dnephin, line 12: I think we need to use the k8s library consistently so that all the commands `yaml` output looks the same. Sorry for the back-and-forth on this.

```suggestion
	"sigs.k8s.io/yaml"
```

This will change the expected output again.

PR #2501, opened 2022-07-06
api: integrate validation library

dnephin:
Branched from #2158

This PR adds validation rules to `CreateAccessKeyRequest`. Also updates the test for `CreateAccessKey` and adds a test case for failed validation.

It also integrates the `internal/validate` library with both the `routes.go` framework, and openapi doc generation.

* Handle the validation error in sendAPIError.
* Call Validate from the handler wrapper.
* Call DescribeSchema from openapi doc generation.
* Change the text of the required validation rule failure.

Related to #1763

### DIFF
--- a/api/access_key.go
+++ b/api/access_key.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"github.com/infrahq/infra/internal/validate"
 	"github.com/infrahq/infra/uid"
 )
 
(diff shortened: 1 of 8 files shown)


api/access_key.go
```diff
@@ -39,3 +49,20 @@ type CreateAccessKeyResponse struct {
...
+		Value:     value,
+		Name:      "name",
+		MinLength: 3,
+		MaxLength: 256,
+		CharacterRanges: []validate.CharRange{
+			validate.AlphabetLower,
+			validate.AlphabetUpper,
+			validate.Numbers,
+			validate.Dash, validate.Underscore, validate.Dot,
+		},
```
dnephin, line 66: This is an important piece to review. I'd like to apply this same validation to all name fields.

Does a length of 3-256 characters seem reasonable?
This will restrict the characters to `a-z`, `A-Z`, `0-9`, and `-`, `_`, `.`.

dnephin: We might need to allow the `@` character as well, for users, because I think we use the name field for email addresses.

PR #2509, opened 2022-07-06
improve: eof errors are more user friendly

### LINKED ISSUE #2307: Better error message from CLI on EOF response
## Describe the bug
When you try to run an `infra ...` CLI command and the server is down or restarting the result is an EOF error:
```
$ infra login localhost
Error: Get "https://localhost": EOF
```

## Expected behavior
We should give some better information in the CLI case:
```
$ infra login localhost --log-level debug
DEBUG  call server: test tls for "localhost"
DEBUG  Cannot validate TLS due to an unexpected error: Get "https://localhost": EOF
Error: Could not reach Infra server, please try again
```

## Environment

```
$ infra version

Client: 0.13.4+dev
Server: 0.13.3
```

## Additional context
This can happen on initial deploy or restarting the server.

j-sneh:
## Summary

Earlier, the CLI would show EOF when the server wasn't ready/unable to connect to and the user tried to login (or use a different command upon update).

It now shows:
`Could not reach infra server, please wait a moment and try again.`

## Checklist

## Related Issues

Resolves #2307

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -113,7 +113,8 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 
 	resp, err := client.HTTP.Do(req)
 	if err != nil {
-		if connError := HandleConnError(err); connError != nil {
+		connError := HandleConnError(err)
+		if connError != nil {
 			return nil, connError
 		}
 		return nil, fmt.Errorf("%s %q: %w", method, path, err)
@@ -326,9 +327,6 @@ func partialText(body []byte, limit int) string {
 	return string(body[:limit]) + "..."
 }
 
-// HandleConnError translates common connection errors into more informative human
-// readable errors. Returns `nil` if the error was not handled, so it is the callers responsibility to
-// return the original error if `HandleConnError` returns nil.
 func HandleConnError(err error) error {
 	urlErr := &url.Error{}
 	if errors.As(err, &urlErr) {
--- a/api/client.go
+++ b/api/client.go
@@ -113,8 +113,7 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 
 	resp, err := client.HTTP.Do(req)
 	if err != nil {
-		connError := HandleConnError(err)
-		if connError != nil {
+		if connError := HandleConnError(err); connError != nil {
 			return nil, connError
 		}
 		return nil, fmt.Errorf("%s %q: %w", method, path, err)
@@ -327,6 +326,9 @@ func partialText(body []byte, limit int) string {
 	return string(body[:limit]) + "..."
 }
 
+// HandleConnError translates common connection errors into more informative human
+// readable errors. Returns `nil` if the error was not handled, so it is the callers responsibility to
+// return the original error if `HandleConnError` returns nil.
 func HandleConnError(err error) error {
 	urlErr := &url.Error{}
 	if errors.As(err, &urlErr) {
--- a/api/client.go
+++ b/api/client.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/ssoroka/slice"
 
+	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/uid"
 )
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -16,6 +17,9 @@ func main() {
 	if err := cmd.Run(context.Background(), os.Args[1:]...); err != nil {
 		var userErr cmd.Error
 		switch {
+		case strings.HasSuffix(err.Error(), ": EOF"):
+			logging.Debugf("%v\n", err)
+			fmt.Fprintln(os.Stderr, "Could not reach infra server, please wait a moment and try again.")
```
dnephin, line 22: A couple thoughts here:

Hopefully we can use `errors.Is(err, io.EOF)` here to detect the error. Checking for the `: EOF` suffix should be relatively safe as well, but if `errors.Is` does not work for some reason it might mean we are wrapping errors incorrectly, which would be a good thing to fix.

I think handling the error at this point (in the `main` function) risks that the EOF could come from other places than the http request to the server. It could have come from reading a file or other network operation.

I think this check would be safer if we moved it to `attemptTLSRequest`. I think the bottom of that function is where we generally see it, and we already have some error handling there.

If we want to apply it to other requests as well, we handle the request errors in `api/client.go:request`. 

Both of these functions have similar checks for `errors.As(err, urlErr)` to check for a timeout. Maybe we should refactor that out into a function something like this, and call it from the two places?

```go
func handleConnError(err error) error {
	if err == nil {
		return nil
	}
	urlErr := &url.Error{}
	if errors.As(err, &urlErr) {
		if urlErr.Timeout() {
			return nil, fmt.Errorf("%w: %s", ErrTimeout, err)
		}
	}
	if errors.Is(err, io.EOF) {
		return fmt.Errorf("could not reach infra server, please wait a moment and try again")
	}
	return err
}
```

dnephin (commented):
Nice! This looks like it's on the right track.

api/client.go
```diff
@@ -107,11 +108,9 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 
 	resp, err := client.HTTP.Do(req)
 	if err != nil {
-		urlErr := &url.Error{}
-		if errors.As(err, &urlErr) {
-			if urlErr.Timeout() {
-				return nil, fmt.Errorf("%w: %s", ErrTimeout, err)
-			}
+		connError := HandleConnError(err)
+		if connError != nil {
```
dnephin, line 112: very minor style comment. Generally we try to keep these in one line when we can:

```suggestion
		if connError := HandleConnError(err); connError != nil {
```

api/client.go
```diff
@@ -321,3 +320,19 @@ func partialText(body []byte, limit int) string {
 
 	return string(body[:limit]) + "..."
 }
+
+func HandleConnError(err error) error {
```
dnephin, line 324: Would be great to have a godoc comment on this since it's exported:

```suggestion
// HandleConnError translates common connection errors into more informative human
// readable errors. Returns `nil` if the error was not handled, so it is the callers responsibility to
// return the original error if `HandleConnError` returns nil.
func HandleConnError(err error) error {
```

In other places our `handleErr` function returns the original error if it doesn't find any special cases, so we should be clear that this one returns `nil` if it doesn't handle anything.

internal/cmd/login.go
```diff
@@ -490,17 +489,20 @@ func attemptTLSRequest(options loginCmdOptions) error {
 			TLSClientConfig: &tls.Config{RootCAs: pool, MinVersion: tls.VersionTLS12},
 		},
 	}
+
 	res, err = httpClient.Do(req)
-	urlErr := &url.Error{}
-	switch {
-	case err == nil:
+
+	if err == nil {
```
dnephin, line 495:
```suggestion
	if err == nil {
```

Minor style note, I think generally guidelines say to keep the `if err` statement on the line immediately after the one that assigned the `err`, to show the close relationship.

internal/cmd/login.go
```diff
@@ -490,17 +489,20 @@ func attemptTLSRequest(options loginCmdOptions) error {
...
 		return nil
-	case errors.As(err, &urlErr):
-		if urlErr.Timeout() {
-			return fmt.Errorf("%w: %s", api.ErrTimeout, err)
-		}
 	}
+
+	connError := api.HandleConnError(err)
+
+	if connError != nil {
```
dnephin, line 502:
```suggestion
	if connError := api.HandleConnError(err); connError != nil {
```

dnephin (approved):
Looks great! It's too bad we can't have `HandleConnError` return the original error instead of `nil`, but I see that is not possible due to the wrapping we use in `request`.  Maybe we can find a way to address that later.

Nothing blocking a merge, but left some minor style suggestions

PR #2510, opened 2022-07-07
feat: groups

### LINKED ISSUE #2417: Add groups to UI

hoyyeva:
## Summary
- adding groups feature to ui
- components refactoring

## Related Issues
Resolves #2417 

## 📷 📹 

https://user-images.githubusercontent.com/63033505/177809670-4df7366c-5c57-40a1-9468-235895b6e11c.mov


ui/components/access-form.js
```diff
@@ -0,0 +1,41 @@
+import { sortByResource } from '../lib/grants'
+
+import RoleSelect from './role-select'
+
+export default function AccessForm({ grants, mutate }) {
```
jmorganca, line 5: Should this be GrantsList? or similar?

ui/components/email-badge.js
```diff
@@ -0,0 +1,18 @@
+import { XIcon } from '@heroicons/react/outline'
+
+export default function EmailBadge({ email, onRemove }) {
```
jmorganca, line 3: Nit: should we generalize this to `Badge` and instead of `email` – you can render the `children` prop below

ui/components/remove.js
```diff
@@ -0,0 +1,32 @@
+import DeleteModal from './delete-modal'
+
+export default function Remove({
```
jmorganca, line 3: Not sure we want to couple the button + modal here since they are quite different – would it be better to leave them separate? – or better have the sidebar layout include these?

ui/components/remove.js
```diff
@@ -0,0 +1,32 @@
...
+  buttonText = 'Remove',
+  deleteModalOpen,
+  setDeleteModalOpen,
+  onSubmit,
+  deleteModalTitle,
+  deleteModalMessage,
+  hide = false,
```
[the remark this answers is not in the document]
hoyyeva: oh yeah, you are right! I was struggling with showing and hiding the remove button for user on sidebar - and now i see it and it should totally in the parent component.

ui/components/remove.js
hoyyeva: this is the remove functionality on the sidebar. In my opinion i think it is ok because the modal only appear when the button clicks

ui/components/empty-data.js
```diff
@@ -0,0 +1,3 @@
+export default function EmptyData({ text }) {
```
jmorganca, line 1: This one would be better staying decoupled vs as a component due to padding/margin

ui/components/typeahead-dropdown.js
```diff
@@ -0,0 +1,44 @@
+import { Combobox } from '@headlessui/react'
+import { CheckIcon } from '@heroicons/react/solid'
+
+export default function TypeaheadDropdown({ filtered, selected }) {
+  return (
+    <>
+      {filtered.length > 0 && (
+        <Combobox.Options className='absolute -left-[13px] z-10 mt-1 max-h-60 w-56 overflow-auto rounded-md border border-gray-700 bg-gray-800 py-1 text-2xs ring-1 ring-black ring-opacity-5 focus:outline-none'>
```
jmorganca, line 8: Wouldn't recommend we sever the `ComboBox` elements what if we create a component that goes inside the `Combobox.Options` (almost like a wrapper for the contents)?

ui/components/typeahead-dropdown.js
```diff
@@ -0,0 +1,44 @@
...
+            <Combobox.Option
+              key={f.id}
+              value={f}
+              className={({ active }) =>
+                `relative cursor-default select-none py-2 px-3 hover:bg-gray-700 ${
+                  active ? 'bg-gray-700' : ''
+                }`
+              }
+            >
+              <div className='flex flex-row'>
```
jmorganca, line 19: Similar to above, I would extract this into an element e.g. (`ListRow`)

ui/pages/destinations/index.js
```diff
@@ -196,51 +207,19 @@ function Details({ destination, onDelete }) {
...
-          <div className='flex flex-row items-center'>
-            <div className='w-1/3 text-2xs text-gray-400'>Updated</div>
-            <div className='text-2xs'>
-              {destination?.updated
-                ? dayjs(destination.updated).fromNow()
-                : '-'}
-            </div>
-          </div>
-        </div>
+        <Metadata data={metadata} />
```
jmorganca, line 210: Nice! Love this abstraction.

ui/components/metadata.js
```diff
@@ -0,0 +1,12 @@
+export default function Metadata({ data }) {
+  return (
+    <div className='flex flex-col space-y-2 pt-3'>
```
jmorganca, line 3: I'd recommend using `grid` here

jmorganca: Check out https://github.com/infrahq/infra/blob/main/ui/pages/providers/index.js#L52 for an example!

ui/components/remove.js
```diff
@@ -0,0 +1,29 @@
+import DeleteModal from './delete-modal'
+
+export default function Remove({
```
jmorganca, line 3: How repetitive would it be to leave this de-coupled? Another option would be to bake this into a Sidebar `<Footer />` component or similar which may be easier to test

ui/pages/groups/add.js
```diff
@@ -0,0 +1,185 @@
...
+  const inputRef = useRef(null)
+
+  const selectedEmailsId = selectedEmails.map(i => i.id)
+
+  const filteredEmail = [...users.map(u => ({ ...u, user: true }))]
+    .filter(s => s?.name?.toLowerCase()?.includes(query.toLowerCase()))
+    .filter(s => s.name !== 'connector')
+    .filter(s => !selectedEmailsId?.includes(s.id))
+
+  const removeSelectedEmail = email => {
```
jmorganca, line 24: This one's personal preference but we do  `function removeSelectedEmail(email) {` elsewhere

ui/pages/groups/index.js
```diff
@@ -0,0 +1,308 @@
...
+          </div>
+          <div className='ml-3 flex min-w-0 flex-1 flex-col leading-tight'>
+            <div className='truncate'>{group.name}</div>
+          </div>
+        </div>
+      )
+    },
+  },
+  {
+    Header: 'Team Size',
```
jmorganca, line 40: `Users`

ui/components/empty-data.js
hoyyeva: hmm but this is the empty data message for the sidebar. So isn't it safe to assume that the padding and margin would be the same for all the pages' sidebar

ui/pages/groups/add.js
hoyyeva: ugh i kept making this mistake - I am too used to the arrow function

ui/components/remove.js
hoyyeva: maybe the name of this component is misleading? maybe I should rename this to remove button? I think it is a great idea to break the sidebar footer into a `<Footer/>` component but this component is just the remove button - the positioning of the button is in the parent components

ui/components/grants-list.js
```diff
@@ -0,0 +1,41 @@
+import { sortByResource } from '../lib/grants'
+
+import RoleSelect from './role-select'
+
+export default function GrantsList({ grants, mutate }) {
```
jmorganca, line 5: Would it be possible to move the grant fetching and mutating within this component? The props would then be:

```
({ destination })
```

ui/pages/users/index.js
```diff
@@ -80,113 +93,91 @@ function Details({ user, admin, onDelete }) {
...
-          <button
-            type='button'
-            onClick={() => setDeleteModalOpen(true)}
-            className='flex items-center rounded-md border border-violet-300 px-6 py-3 text-2xs text-violet-100'
-          >
-            Remove
-          </button>
+          <RemoveButton
+            modalOpen={deleteModalOpen}
+            setModalOpen={setDeleteModalOpen}
```
jmorganca, line 167: In this case the parent component doesn't really need to hold the state for if the modal is open or not: we can remove these props (unless they are used elsewhere) for the `RemoveButton` component

ui/pages/users/index.js
```diff
@@ -80,113 +93,91 @@ function Details({ user, admin, onDelete }) {
...
+            onSubmit={async () => {
```
jmorganca, line 168: `onRemove` would be a better prop name

ui/components/grants-list.js
hoyyeva: hmm we definitely can. However, we are also using the `grants` in the parent component to determine if the grant list is empty or not - which it has a specific margin ( which we don't want to include that in the component) and also for the user sidebar - we also have inherited list as well. We can pass the inherited list as a children prop but I am not very confident that is a good idea.

jmorganca: We can keep it as: passing `mutate` is definitely something we should try to improve though since it isn't obvious what that is – I would suggest changing this to `onChange` and `onRemove` and handling edits in the parent component since that's where the data is coming from

ui/components/grants-list.js
```diff
@@ -0,0 +1,39 @@
...
+      {grants?.sort(sortByResource)?.map(g => (
+        <div key={g.id} className='flex items-center justify-between text-2xs'>
+          <div>{g.resource}</div>
+          <RoleSelect
+            role={g.privilege}
+            resource={g.resource}
+            remove
+            direction='left'
+            onRemove={async () => {
+              await fetch(`/api/grants/${g.id}`, { method: 'DELETE' })
```
jmorganca, line 17: This should be managed by the caller since they are supplying the `grants`

ui/components/list-row.js
```diff
@@ -0,0 +1,22 @@
+import { CheckIcon } from '@heroicons/react/solid'
+
+export default function ListRow({ item, selected }) {
```
jmorganca, line 3: Last improvements for this: `ComboboxItem` might be a better name given it always shows up in a combo box

Would be better to decouple this a bit more by changing the props to `title`, `subtitle`, `selected=true|false` to render it as selected or not?

ui/pages/groups/add.js
```diff
@@ -0,0 +1,185 @@
...
+  const [errors, setErrors] = useState({})
+
+  const router = useRouter()
+
+  const handleGroupNameInputChange = value => {
+    setGroupName(value)
+    setError('')
+  }
+
+  const addUsersToGroup = async groupId => {
```
jmorganca, line 64: nit: `async function addUsersToGroup { ...`

jmorganca (approved):

Issue #2140, opened 2022-05-30
ui: link to sidebar details


jmorganca:
Users should be able to link to the sidebar view of an individual item:

* `/destinations/<id>`
* `/providers/<id>`
* `/users/id`

jmorganca:
Closing as this is being fixed elsewhere

PR #2512, opened 2022-07-07
improve: move UI to a standalone service

jmorganca:
PR for running the UI as a standalone service.

By deploying the UI, we get the benefits of Next.js server-side rendering on first load and dynamic routes (e.g. enables #2140).

This also simplifies the packaging story for the API server since we no longer have to embed it into the Go binary.

This PR also includes a docker-compose file for developing the "full stack" including the UI locally, which can be moved to another PR


helm/charts/infra/values.yaml
```diff
@@ -5,17 +5,16 @@ nameOverride: ""
 fullnameOverride: ""
 
 global:
-
   image:
```
mxyng, line 8: Since the object is now empty, it's a good idea to set it to `{}`. Though I suggest keeping at least the `tag` and `pullPolicy` since they're still good defaults

internal/server/server_test.go
```diff
@@ -238,196 +236,6 @@ func TestServer_Run_UIProxy(t *testing.T) {
 	})
 }
 
-func TestServer_GenerateRoutes_NoRoute(t *testing.T) {
```
dnephin, line 241: I don't think we want to lose this whole test. A couple of the test cases can be deleted, but at least 2 of them should still be relevant.

dnephin (commented):
In `registerUIRoutes`, we should set the `Logger` on the  `proxy` (https://pkg.go.dev/net/http/httputil#ReverseProxy).

I noticed a little while ago that the logs from that proxy do not get turned into JSON properly, and I think that's why.

helm/charts/infra/values.yaml
jmorganca: Thanks!

Does `repository` make sense to keep at all?

internal/server/server_test.go
jmorganca: Great catch. Will restore these

internal/server/server.go
```diff
@@ -202,16 +186,14 @@ func registerUIRoutes(router *gin.Engine, opts UIOptions) {
 			req.URL.Scheme = remote.Scheme
 			req.URL.Host = remote.Host
 		}
+		proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
```
jmorganca, line 189: @dnephin does this look right? Or should we pass in the `httpErrorLog` from `server.Run` ?

helm/charts/infra/templates/server/configmap.yaml
```diff
@@ -59,6 +59,16 @@ data:
 {{- $users = append $users (dict "name" "connector" "accessKey" $accessKey) }}
 {{- end }}
 
+{{- if .Values.server.config.ui.proxyURL }}
+    ui:
+      proxyURL: {{ .Values.connector.config.ui.proxyURL }}
+{{- end }}
+
+{{- if and (include "ui.enabled" . | eq "true") (not .Values.server.config.ui.proxyURL) }}
```
jmorganca, line 67: Does this look right? Idea is to set a default if not specified by the user (and if the UI is on)

internal/server/server.go
dnephin: This looks right to me

helm/charts/infra/values.yaml
```diff
@@ -5,17 +5,15 @@ nameOverride: ""
 fullnameOverride: ""
 
 global:
```
mxyng, line 7: The structure of this object shouldn't change since it may break existing deployments. The UI image.repository doesn't need to use any global values. See #2602 which adds a postgres component to the change that's configured entirely separately from the main infra apps

.github/workflows/publish.yml
```diff
@@ -95,6 +95,13 @@ jobs:
 
   publish-docker:
     runs-on: ubuntu-latest
+    strategy:
```
jmorganca, line 98: @mxyng anything here that might not work?

dnephin: This looks right to me

helm/charts/infra/values.yaml
jmorganca: Gotcha. Should we still keep the global object around if some services won't use it?

helm/charts/infra/templates/server/configmap.yaml
mxyng: Yes that looks right. A simple way to verify is with helm template:

```
$ helm template helm/charts/infra -s templates/server/configmap.yaml --set ui.enabled=false
```

Or

```
$ helm template helm/charts/infra -s templates/server/configmap.yaml --set server.config.ui.proxyURL=somevalue```

**Review by dnephin — APPROVED** [review]
I did not review the helm changes or dockerfile changes too closely. I ran the docker-compose.dev.yaml which seemed to work.

Assuming you ran the helm charts and it did what you expected, I think this looks good.

File: helm/charts/infra/values.yaml (hunk 5 above)
Comment (mxyng) [reply helm/charts/infra/values.yaml:7]: There is. The main image is still shared so changing it in all places without a global would be painful. It's completely fine to have some deployments not use the global image

File: helm/charts/infra/templates/server/configmap.yaml [hunk 7]
```diff
@@ -61,6 +61,16 @@ data:
 
 {{- $users | uniq | toYaml | nindent 6 }}
 
+{{- if .Values.server.config.ui.proxyURL }}
+    ui:
+      proxyURL: {{ .Values.connector.config.ui.proxyURL }}
+{{- end }}
+
+{{- if and (include "ui.enabled" . | eq "true") (not .Values.server.config.ui.proxyURL ) }}
```
mxyng, line 69: Can group this with the previous condition to simplify it

```
{{- if .Values.server.config.ui.proxyURL }}
    ...
{{- else if include "ui.enabled" . | eq 'true' }}
    ...
{{- end }}
```

helm/charts/infra/templates/ui/_helpers.tpl
```diff
@@ -0,0 +1,151 @@
...
+UI image pull secrets. Merges global and local values.
+*/}}
+{{- define "ui.imagePullSecrets" -}}
+{{ concat .Values.ui.imagePullSecrets .Values.global.imagePullSecrets | uniq | toYaml }}
+{{- end }}
+
+{{/*
+UI 'env' values. Merges global and local values.
+*/}}
+{{- define "ui.env" -}}
```
mxyng, line 133: We can probably omit `env` and `envFrom` templates since there's no benefit in merging the global `env` values

mxyng (approved):

helm/charts/infra/templates/server/configmap.yaml
jmorganca: Much better. Thanks!

PR #2528, opened 2022-07-07
api: allow types to define their own default values for openapi schema

dnephin:
## Summary

This commit moves schema descriptions for api types from being hard-coded in the openapi doc generation, to being methods on the types being described. This `DescribeSchema` method is the same interface we use in `internal/validate` to allow validation rules to describe themselves.

By using an interface it makes it easy for us to support new api types without modifying the openapi doc generation.

### DIFF
--- a/internal/server/apimigrator.go
+++ b/internal/server/apimigrator.go
@@ -102,7 +102,7 @@ func rebuildRequest(c *gin.Context, newReqObj interface{}) {
 	for i := 0; i < r.NumField(); i++ {
 		f := r.Field(i)
 		if fieldName, ok := t.Field(i).Tag.Lookup("form"); ok {
-			if structNameWithPkg(f.Type()) == "uid.ID" {
+			if f.Type() == reflect.TypeOf(uid.ID(0)) {
 				query.Add(fieldName, uid.ID(f.Int()).String())
 				continue
 			}
@@ -114,8 +114,8 @@ func rebuildRequest(c *gin.Context, newReqObj interface{}) {
 				query.Add(fieldName, f.String())
 			case reflect.Slice:
 				// only type that does this is []uid.ID
-				switch structNameWithPkg(f.Type()) {
-				case "uid.ID":
+				switch f.Type() {
+				case reflect.TypeOf(uid.ID(0)):
 					for j := 0; j < f.Len(); j++ {
 						query.Add(fieldName, uid.ID(f.Index(j).Int()).String())
 					}
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -165,7 +165,7 @@ func buildProperty(f reflect.StructField, t, parent reflect.Type, parentSchema *
 	}
 
 	s := &openapi3.Schema{}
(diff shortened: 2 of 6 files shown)


internal/server/apimigrator.go
```diff
@@ -102,7 +102,7 @@ func rebuildRequest(c *gin.Context, newReqObj interface{}) {
 	for i := 0; i < r.NumField(); i++ {
 		f := r.Field(i)
 		if fieldName, ok := t.Field(i).Tag.Lookup("form"); ok {
-			if structNameWithPkg(f.Type()) == "uid.ID" {
+			if f.Type() == reflect.TypeOf(uid.ID(0)) {
```
dnephin, line 105: I also removed `structNameWithPkg` and replaced it with these more strict type checks.  I think the type comparison is nicer because an IDE can find these references, but it's not going to find `uid.ID`.

This `apimigrator` was the last place that used it after I removed it from `openapi.go`.

internal/server/openapi.go
```diff
@@ -217,10 +217,6 @@ func WriteOpenAPIDocToFile(openAPIDoc openapi3.T, version string, filename strin
 }
 
 func setTagInfo(f reflect.StructField, t, parent reflect.Type, schema, parentSchema *openapi3.Schema) {
-	if ex := getDefaultExampleForType(t); len(ex) > 0 {
-		schema.Example = ex
-	}
```
dnephin, line 222: This call wasn't necessary because we already set these defaults in `setTypeInfo`, which is always called in exactly the same place as `setTagInfo`, so I was able to remove this.

internal/server/openapi.go
```diff
@@ -252,56 +249,33 @@ func setTagInfo(f reflect.StructField, t, parent reflect.Type, schema, parentSch
...
-		schema.Type = "string"
-		schema.Format = "poly-uid"
-		schema.Pattern = `\w:[\da-zA-HJ-NP-Z]{1,11}`
-		schema.Example = "i:4yJ3n3D8E3"
+	switch value.(type) {
+	case time.Time:
+		panic("field must use api.Time")
+	case time.Duration:
+		panic("field must use api.Duration")
+	}
```
dnephin, line 271: These are new. Now that we have our own API types for time and duration we should warn a contributor if they accidentally use the ones from stdlib (instead of supporting the stdlib types which would make our API inconsistent).

internal/server/testdata/openapi3.json
```diff
@@ -865,7 +865,6 @@
             }
           },
           {
-            "example": "4yJ3n3D8E2",
```
dnephin, line 868: We had these examples in two places. If you expand the fold below this section, you should see all of these have a `schema.example` with the same value.

Reading the docs here https://swagger.io/specification/#fixed-fields-10 (second table) it says:

> Furthermore, if referencing a schema that contains an example, the example value SHALL override the example provided by the schema.

So we only need to set this example if we are overriding the `schema.example`.

mxyng (approved):

PR #2536, opened 2022-07-08
Support nested structs in request validation

dnephin:
Branched from #2527

This PR adds validation rules to `CreateDestinationRequest` and `UpdateDestinationRequest`.
These are the first two structs that have nested validation, so most of the PR is adding support for nested rules.

The challenge with nested structs is that openapi doc generation uses reflection (don't see any way around that), but `Validate` does not. The `ValidationRules` on a struct can't return the rules for nested or embedded structs, since we'd see them described the wrong way in openapi doc generation.  

So `Validate` ends up needing to use reflection as well, to match `DescribeSchema`.

In the current state:
* openapi doc generation calls `DescribeSchema` on the nested fields
* `Validate` will call `ValidationRules` on the top level struct, and use reflection to traverse the fields and call it on any nested or embedded structs.

Embedded structs are a challenge, because a `ValidationRules` method on an embedded struct could satisfy the interface for the parent struct. We'd end up calling it twice. So I updated all the list requests to return `nil` for validation rules.

### DIFF
--- a/api/access_key.go
+++ b/api/access_key.go
@@ -24,6 +24,8 @@ type ListAccessKeysRequest struct {
 }
 
 func (r ListAccessKeysRequest) ValidationRules() []validate.ValidationRule {
+	// no-op ValidationRules implementation so that the rules from the
+	// embedded PaginationRequest struct are not applied twice.
 	return nil
 }
 
--- a/api/access_key.go
+++ b/api/access_key.go
@@ -24,7 +24,7 @@ type ListAccessKeysRequest struct {
 }
 
 func (r ListAccessKeysRequest) ValidationRules() []validate.ValidationRule {
-	return r.PaginationRequest.ValidationRules()
+	return nil
 }
 
 type CreateAccessKeyRequest struct {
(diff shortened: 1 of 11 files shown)


api/access_key.go
```diff
@@ -24,7 +24,7 @@ type ListAccessKeysRequest struct {
 }
 
 func (r ListAccessKeysRequest) ValidationRules() []validate.ValidationRule {
-	return r.PaginationRequest.ValidationRules()
+	return nil
```
BruceMacD, line 27: I'd like to see some comments here that explain why this is needed for future reference

internal/validate/validate.go
```diff
@@ -11,19 +11,51 @@ import (
...
+			}
 		}
 	}
 
-	if len(err) > 0 {
-		return err
+	if v.Kind() == reflect.Struct {
+		for i := 0; i < v.NumField(); i++ {
+			f := v.Field(i)
+			if v.Type().Field(i).Anonymous {
```
BruceMacD, line 42: What does this anonymous check do?

BruceMacD (approved):
Looks good, I just have a question about a new (to me) bit of code

internal/validate/validate.go
dnephin: `Anonymous` means the field is an embedded structs: https://pkg.go.dev/reflect#StructField

I can add a comment to say that

BruceMacD: Thanks for the explanation

PR #2537, opened 2022-07-08
feat: connector version and apiVersion

### LINKED ISSUE #2506: Add version to destinations
Destinations need to report their versions to give users feedback when upgrading either server or connectors. This value should be reported by the UI

mxyng:
## Summary

Report a connector's version and apiVersion when creating or updating a destination. These values are useful metadata for the user when investigating destination issues.

Update server Prometheus metrics to attach labels for resources. The previous collectors only allowed constant labels so it's been refactored to a custom collector which can attach variable labels to the metric. This is important for destinations and providers. Destinatiosn should report version and apiVersion as labels while providers should report kind as labels. Other label values will be added as needed

This change makes some changes to the Destinations API. `UniqueID` is now a required field in both `CreateDestination` and `UpdateDestination` requests. While this is technically a breaking change, existing connectors already populate this field so the impact will be minimal.

This change removes uses of `promauto`. `promauto` was useful when service metrics were attached to the default registry since it removes the need to register but since it's creating and using a new registry, the benefits of `promauto` is greatly reduced. This change also removes the `database_*` metrics in favour of the Prometheus DB stats collector which reports the same information and much more.

image

## Checklist

## Related Issues

Resolves #2506


internal/server/models/destination.go
```diff
@@ -13,6 +13,9 @@ type Destination struct {
 	UniqueID   string `gorm:"uniqueIndex:idx_destinations_unique_id,where:deleted_at is NULL"`
 	LastSeenAt time.Time
 
+	Version    string
+	APIVersion string
```
jmorganca, line 17: Should we include this field if there's a clear mapping of Version: APIVersion in our releases?

ui/pages/destinations/index.js
```diff
@@ -221,6 +221,14 @@ function Details({ destination, onDelete }) {
                 : '-'}
             </div>
           </div>
+          <div className='flex flex-row items-center'>
+            <div className='w-1/3 text-2xs text-gray-400'> Version</div>
```
jmorganca, line 225: This is awesome. Suggestion: `Connector Version` so users don't confuse it with Kubernetes versions (especially once we get to 1.x)!

internal/server/models/destination.go
mxyng: I was on the fence about including this. It seems distinct enough to warrant including but can definitely be omitted for now

ui/pages/destinations/index.js
mxyng: I was under the impression this panel is destination metadata. In that case, it should be fairly obvious this is the destination version as nothing else in this view references the actual Kubernetes cluster

BruceMacD: I think "connector-version" is easier to understand too. There's a lot of versions in the mix so being more specific can't hurt.

internal/server/metrics.go
```diff
@@ -2,126 +2,144 @@ package server
...
-
-	factory.NewGaugeFunc(prometheus.GaugeOpts{
+	registry := metrics.NewRegistry()
+
+	rawDB, err := db.DB()
+	if err == nil {
+		registry.MustRegister(collectors.NewDBStatsCollector(rawDB, db.Dialector.Name()))
+	}
+
+	registry.MustRegister(prometheus.NewGaugeFunc(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
mxyng: You're probably right. It's lead to the connector build_info metric being missed in the past

PR #2550, opened 2022-07-11
fix: login provider button text

### LINKED ISSUE #2524: ui: login should show the name of provider instead of Single Sign-On
Image

hoyyeva:
## Summary
Since we are supporting more than just `okta` now. The login provider buttons will only display `single sign-on` when the kind of provider is not included in the default list 

## Related Issues
Resolves #2524


hoyyeva:
as what @jmorganca mentioned -  having "Single Sign-On" text for non-Okta was on purpose since we don't have a good image for those buttons, and it didn't look right just showing their "icon"

we are looking for an alternative to better display the buttons - closing the pr for now

PR #2555, opened 2022-07-11
feat: sample api request logs based on fields

### LINKED ISSUE #2368: Server log volume is high

Server logs every API request which can be very noisy since the connector calls a set of APIs periodically. Since this is logged at the INFO level, the default configurations produces a lot of logs.

Logs should be useful but not log excessively. Maybe rate limiting? Alternatively reduce level of API logs to DEBUG or raise default level to WARN

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

mxyng:
## Summary

Built-in zerolog samplers are not aware of the data they're sampling so
enabling it for even something as specific as api request logs may drop
useful information. Instead, create a simpler something that can sample
based on specified fields.

For API requests, group by request method (GET, POST, etc.) and request
path. The path is the template path known to gin, not the real path seen
my the request. e.g. /api/users/:id instead of /api/users/amNmdZKwwU

## Checklist

## Related Issues

Resolves #2368

### DIFF
--- a/internal/logging/middleware.go
+++ b/internal/logging/middleware.go
@@ -1,16 +1,51 @@
 package logging
 
 import (
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/rs/zerolog"
 )
 
+type Sampler struct {
+	fn       func() zerolog.Sampler
+	samplers map[string]zerolog.Sampler
+	mu       sync.Mutex
+}
+
+func NewSampler(fn func() zerolog.Sampler) *Sampler {
+	return &Sampler{
+		fn:       fn,
+		samplers: make(map[string]zerolog.Sampler),
+	}
+}
+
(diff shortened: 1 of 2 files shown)


internal/logging/middleware.go
```diff
@@ -1,16 +1,51 @@
...
+	"sync"
 	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/rs/zerolog"
 )
 
+type Sampler struct {
+	fn       func() zerolog.Sampler
+	samplers map[string]zerolog.Sampler
```
mxyng, line 14: Size of this map _may_ be an issue. Need to keep a close eye on service metrics

BruceMacD: I see why you're nervous, but I think it will be ok for the time being where we have ~20 endpoints 👍

BruceMacD (approved):

PR #2560, opened 2022-07-12
feat: add destination status as metric label

### LINKED ISSUE #2151: Need to show destination status to know if a cluster is still alive

mxyng:
## Summary

Add a label to `infra_destinations` that describes the status of the destination. Label values are either `connected` or `disconnected`

## Checklist

## Related Issues

Resolves #2151


internal/server/metrics.go
```diff
@@ -109,20 +111,28 @@ func setupMetrics(db *gorm.DB) *prometheus.Registry {
...
-			Count   int
+			Version   string
+			Connected bool
+			Count     float64
 		}
 
-		if err := db.Raw("SELECT COALESCE(version, '') AS version, COUNT(*) as count FROM destinations WHERE deleted_at IS NULL GROUP BY COALESCE(version, '')").Scan(&results).Error; err != nil {
+		cmp := time.Now().Add(-5 * time.Minute)
+
+		if err := db.Raw("SELECT COALESCE(version, '') AS version, last_seen_at >= ? AS connected, COUNT(*) AS count FROM destinations WHERE deleted_at IS NULL GROUP BY COALESCE(version, ''), last_seen_at >= ?", cmp, cmp).Scan(&results).Error; err != nil {
```
BruceMacD, line 123: Do we need to group by `last_seen_at >= ?` where we already selected based on that?

mxyng: We do because otherwise the aggregation breaks, incorrectly grouping items with different `connected` statuses together.

Example with `GROUP BY last_seen_at >= ?`:
```
┌─────────┬───────────┬───────┐
│ version │ connected │ count │
├─────────┼───────────┼───────┤
│ 0.13.6  │ 0         │ 1     │
│ 0.13.6  │ 1         │ 1     │
└─────────┴───────────┴───────┘
```

Example _without_ `GROUP BY last_seen_at >= ?`:
```
┌─────────┬───────────┬───────┐
│ version │ connected │ count │
├─────────┼───────────┼───────┤
│ 0.13.6  │ 1         │ 2     │
└─────────┴───────────┴───────┘
```

BruceMacD (approved):

PR #2561, opened 2022-07-12
maintain: docs/api/openapi3.json no longer exists

mxyng:
## Summary

Minor cleanup: `docs/api/openapi3.json` doesn't exist anymore so these configurations are unnecessary

### DIFF
--- a/.github/release-please.json
+++ b/.github/release-please.json
@@ -17,12 +17,7 @@
         "Dockerfile",
         "helm/charts/infra/Chart.yaml",
         "internal/version.go",
-        "docs/install/upgrading.md",
-        {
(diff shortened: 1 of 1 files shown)


jmorganca:
It's now located at `internal/server/testdata/openapi3.json` although we should reconsider as this would be hard to find for developers using our API

.gitignore
```diff
@@ -5,4 +5,3 @@
 
 # infra binary
 dist
-docs/api/openapi3.json
```
dnephin, line 8: We still want this line in `.gitignore` because `make docs` can create this file (for release).

dnephin (approved):

PR #2565, opened 2022-07-13
fix: remove email validation lib

hoyyeva:
## Summary
remove email validation lib since we can check the validation by wrapping the input field into a `<form>`


jmorganca (approved):
Great!!

PR #2567, opened 2022-07-13
api: only sample successful GET requests

### LINKED ISSUE #2566: Migrate to a sync.Map for log sampling
## Describe the bug

`map` is simple and easy and for a low volume of logs, wrapping in a `sync.Mutex` works fine. As the volume increases, it should be replaced with a `sync.Map` which seems to be designed for this use case

dnephin:
## Summary

This PR makes a few improvements to our API log sampling, added in #2555.

* never sample requests if the status code is >= 400, errors should always be logged. Previously we were looking at `gin.Context.Errors` to determine this, but we are not populated it yet
* never sample non-GET requests. Any POST, PUT, or DELETE should be logged as they modify data and should not happen as frequently
* turn off sampling if the log level is set to `DEBUG` or `TRACE`. In those cases the user probably does not want sampling. This gives us an easy way to see all requests in development.
* update the tests to be a bit more strict about the expected log output, and include some test cases that use parameters in the URL to show the sampling works correctly with parameters.
* use a `sync.Map` instead of a `sync.Mutex` to avoid locking in the common case. Once each sampler is created we should be able to fetch it without any locking.

## Related Issues

Resolves #2566

### DIFF
--- a/internal/logging/logger.go
+++ b/internal/logging/logger.go
@@ -89,6 +89,9 @@ func SetLevel(levelName string) error {
 		return err
 	}
 
+	// logging middleware depends on this level. If we stop using the global level
+	// make sure to adjust the logging middleware to check the level of the
+	// logger instead of the global level.
 	zerolog.SetGlobalLevel(level)
 	return nil
 }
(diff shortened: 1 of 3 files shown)


mxyng (approved):
Excellent improvements to the logging middleware!

PR #2568, opened 2022-07-13
feat: create and mount db encryption key

mxyng:
## Summary

When creating a server deployment, if a DB encryption key isn't supplied, create one as a kubernetes secret and mount it into the server pod. This is nicer than the existing default which is to create an encryption key in the pod itself which is only persisted if the key is in the path of a PVC.

It removes a dependency on PVCs. Once this is merged, the only remaining dependency is the database.

Note: This is a _breaking change_ since it discards the previous encryption. Manual steps are required to ensure the continued viability of the data.

## TODO

- [x] add migration steps

Before attempting an upgrade from <=0.14.x, follow the steps below to migrate the encryption key if using Helm release name `infra` and namespaces `infrahq`:

```
$ ENCRYPTION_KEY=$(mktemp)
$ kubectl -n infrahq exec -i deployment/infra-server -- cat /var/lib/infrahq/server/sqlite3.db.key >$ENCRYPTION_KEY
$ kubectl -n infrahq create secret generic infra-server-encryption-key --from-file=key=$ENCRYPTION_KEY
$ kubectl -n infrahq annotate secret infra-server-encryption-key meta.helm.sh/release-name=infra meta.helm.sh/release-namespace=infrahq
$ kubectl -n infrahq label secret infra-server-encryption-key app.kubernetes.io/managed-by=Helm
# upgrade as usual and verify the upgrade. once the upgrade is complete, remove $ENCRYPTION_KEY
```

### DIFF
--- a/helm/charts/infra/templates/server/configmap.yaml
+++ b/helm/charts/infra/templates/server/configmap.yaml
@@ -35,10 +35,10 @@ data:
 {{- if not (hasKey .Values.server.config "dbFile") }}
     dbFile: /var/lib/infrahq/server/sqlite3.db
 {{- end }}
-
-{{- if not (hasKey .Values.server.config "dbEncryptionKey") }}
-    dbEncryptionKey: /var/lib/infrahq/server/sqlite3.db.key
 {{- end }}
+
+{{- if not .Values.server.config.dbEncryptionKey }}
+    dbEncryptionKey: /var/run/secrets/infrahq.com/encryption-key/key
 {{- end }}
 
 {{- if not .Values.server.config.tls }}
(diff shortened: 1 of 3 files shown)


BruceMacD (commented):
This is going to be a much better experience, locally when trying to build this branch I get an error though:
```
Error: UPGRADE FAILED: failed to create resource: Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal number into Go struct field ObjectMeta.spec.template.metadata.annotations of type string
```

Any ideas?
```
$ helm version                  
version.BuildInfo{Version:"v3.9.0", GitCommit:"7ceeda6c585217a19a1131663d8cd1f7d641b2a7", GitTreeState:"clean", GoVersion:"go1.18.2"}
```

mxyng:
Can you link the values file you're using? I haven't seen this locally with either the default values or my local values

mxyng:
I'm going to hold off on this until there's more breaking changes we can bundle together.

PR #2569, opened 2022-07-13
feat: pagination in cli

j-sneh:
## Summary

The `infra list ___` command now compiles its list of Items through multiple API requests (multiple pages) if there are more than 100 records.

The main logic is added in the `Paginatable` interface and `listAll` function.
## Checklist

### DIFF
--- a/internal/cmd/pagination.go
+++ b/internal/cmd/pagination.go
@@ -19,10 +19,6 @@ func listAll[Item any, Req api.Paginatable](listItems func(Req) (*api.ListRespon
 
 	res, err := listItems(req)
 	if err != nil {
-		if api.ErrorStatusCode(err) == 403 {
-			logging.Debugf("%s", err.Error())
-			return nil, ErrMissingPrivileges
-		}
 		return nil, err
 	}
 	users := make([]Item, 0, res.TotalCount)
@@ -37,10 +33,6 @@ func listAll[Item any, Req api.Paginatable](listItems func(Req) (*api.ListRespon
 		logging.Debugf("call server: page %d", page)
 		res, err = listItems(req)
 		if err != nil {
-			if api.ErrorStatusCode(err) == 403 {
-				logging.Debugf("%s", err.Error())
-				return nil, ErrMissingPrivileges
-			}
 			return nil, err
 		}
 		users = append(users, res.Items...)
--- a/internal/cmd/pagination.go
+++ b/internal/cmd/pagination.go
@@ -1,0 +1,31 @@
+package cmd
+
+import (
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/logging"
+)
+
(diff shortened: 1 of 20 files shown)


internal/cmd/pagination.go
```diff
@@ -0,0 +1,48 @@
...
+		if handleError == nil {
+			return nil, err
+		}
+		return nil, handleError(err)
+	}
+	users := make([]Item, 0, res.TotalCount)
+	users = append(users, res.Items...)
+
+	for page := 2; page <= res.TotalPages; page++ {
+		req, ok := req.SetPage(1).(Req)
```
BruceMacD, line 31:
```suggestion
		req, ok := req.SetPage(page).(Req)
```
I think this is a typo and it should be set to the iterating page?

internal/cmd/pagination.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/logging"
+)
+
+// listAll is a helper function that handles pagination and calls the given list request function.
+// listItems is the corresponding function in the API client that handles the Request "req".
+// handleError is a function that handles the error returned by the API client.
+func listAll[Item any, Req api.Paginatable](client *api.Client, req Req, listItems func(api.Client, Req) (*api.ListResponse[Item], error)) ([]Item, error) {
```
dnephin, line 11: We can accept just the api client method, instead of accepting client and the function separately.

```suggestion
func listAll[Item any, Req api.Paginatable](req Req, listItems func(Req) (*api.ListResponse[Item], error)) ([]Item, error) {
```

Then call `listAll` like this:

```go
listAll(api.ListDestinationsRequest{}, client.ListDestinations)
```

internal/cmd/pagination.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	res, err := listItems(*client, req)
+	if err != nil {
+		if api.ErrorStatusCode(err) == 403 {
+			logging.Debugf("%s", err.Error())
+			return nil, ErrMissingPrivileges
+		}
+		return nil, err
+	}
```
dnephin, line 27: Do we need to handle the error here? It seems like we return `err` in the default case anyway.

So I think we could avoid the need for `ErrMissingPrivileges` by always just returning `err`.

dnephin (commented):
Looking good! Do we have any tests that exercise this new function by retrieving multiple pages? If not we should add at least one!

I left a couple suggestions for improvements

internal/cmd/pagination.go
j-sneh: The ErrMissingPrivileges exists to replicate the more user-friendly error messages that were originally in place for the 403 codes.

dnephin: The old logic should still continue to work without the new custom error. Does it not?

PR #2570, opened 2022-07-13
api: add request validation rules for grants, groups, and login

dnephin:
## Summary

Branched from #2527

TODO: make sure login has sufficient test coverage for field validation

## Related Issues

### DIFF
--- a/api/login.go
+++ b/api/login.go
@@ -1,23 +1,54 @@
 package api
 
-import "github.com/infrahq/infra/uid"
+import (
+	"github.com/infrahq/infra/internal/validate"
+	"github.com/infrahq/infra/uid"
+)
 
 type LoginRequestOIDC struct {
-	ProviderID  uid.ID `json:"providerID" validate:"required"`
-	RedirectURL string `json:"redirectURL" validate:"required"`
-	Code        string `json:"code" validate:"required"`
+	ProviderID  uid.ID `json:"providerID"`
+	RedirectURL string `json:"redirectURL"`
+	Code        string `json:"code"`
+}
+
+func (r LoginRequestOIDC) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("providerID", r.ProviderID),
+		validate.Required("redirectURL", r.RedirectURL),
+		validate.Required("code", r.Code),
+	}
 }
 
 type LoginRequestPasswordCredentials struct {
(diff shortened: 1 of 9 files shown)


api/login.go
```diff
@@ -1,23 +1,54 @@
...
-	PasswordCredentials *LoginRequestPasswordCredentials `json:"passwordCredentials" validate:"excluded_with=OIDC,excluded_with=AccessKey"`
-	OIDC                *LoginRequestOIDC                `json:"oidc" validate:"excluded_with=KeyExchange,excluded_with=PasswordCredentials"`
+	AccessKey           string                           `json:"accessKey"`
+	PasswordCredentials *LoginRequestPasswordCredentials `json:"passwordCredentials"`
+	OIDC                *LoginRequestOIDC                `json:"oidc"`
+}
+
+func (r LoginRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.RequireOneOf(
```
BruceMacD, line 46: Should we have a validation rule `ExactlyOneOf`? Only one of these should really be specified.

dnephin: I changed `RequireOneOf` to be exactly that "one and only one".  I renamed the old rule to `RequireAnyOf`, which matches the names the OpenAPI spec uses for these.

dnephin (commented):
Ok, I've added some more test coverage, so I think this is ready for review.

internal/server/handlers_test.go
```diff
@@ -1189,60 +1188,3 @@ func TestAPI_ListAccessKey(t *testing.T) {
 		assert.Assert(t, accessKey.IssuedForName == "")
 	})
 }
-
-func TestAPI_LoginResponse(t *testing.T) {
```
dnephin, line 1193: Moved this to `login_test.go` and added more cases to it.

internal/server/groups_test.go
```diff
@@ -302,7 +302,7 @@ func TestAPI_CreateGroup(t *testing.T) {
 				assert.Equal(t, resp.Code, http.StatusCreated, resp.Body.String())
 			},
 			body: api.CreateGroupRequest{
-				Name: "Awesome group",
+				Name: "AwesomeGroup",
```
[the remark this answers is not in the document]
mxyng: `@` is likely going to be needed for google group. iirc groups in google are essentially mailing lists, i.e. email addresses

mxyng (approved):

PR #2571, opened 2022-07-14
feat: load database configuration options using infra secrets

mxyng:
## Summary

This allows the configuration values to come from `file:`, `env:` or any other configured secret provider instead of only handling raw configuration values.

fix: infra_destinations metrics in postgres

The previous query failed locally when running with postgres database. It complained about the `last_seen_at` column not being a part of `GROUP BY`. This is due to the two separate query args seen by postgres as distinct values. I tried using `sql.NamedArgs` but it produced the same result. Ended up using a subquery which seems to be cleaner. For some reason the postgres variant of the unit test wasn't able to find this

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -310,18 +310,28 @@ func (s *Server) getPostgresConnectionString() (string, error) {
 
 	if s.options.DBHost != "" {
 		// config has separate postgres parameters set, combine them into a connection DSN now
-		fmt.Fprintf(&pgConn, "host=%s ", s.options.DBHost)
+		host, err := secrets.GetSecret(s.options.DBHost, s.secrets)
+		if err != nil {
+			return "", fmt.Errorf("postgres host: %w", err)
+		}
+
+		fmt.Fprintf(&pgConn, "host=%s ", host)
 
 		if s.options.DBUsername != "" {
-			fmt.Fprintf(&pgConn, "user=%s ", s.options.DBUsername)
+			username, err := secrets.GetSecret(s.options.DBUsername, s.secrets)
+			if err != nil {
+				return "", fmt.Errorf("postgres username: %w", err)
+			}
+
+			fmt.Fprintf(&pgConn, "user=%s ", username)
 
 			if s.options.DBPassword != "" {
-				pass, err := secrets.GetSecret(s.options.DBPassword, s.secrets)
+				password, err := secrets.GetSecret(s.options.DBPassword, s.secrets)
 				if err != nil {
-					return "", fmt.Errorf("postgres secret: %w", err)
+					return "", fmt.Errorf("postgres password: %w", err)
 				}
 
-				fmt.Fprintf(&pgConn, "password=%s ", pass)
+				fmt.Fprintf(&pgConn, "password=%s ", password)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


internal/server/metrics.go
```diff
@@ -118,9 +118,9 @@ func setupMetrics(db *gorm.DB) *prometheus.Registry {
 			Count     float64
 		}
 
-		cmp := time.Now().Add(-5 * time.Minute)
+		timeout := time.Now().Add(-5 * time.Minute)
 
-		if err := db.Raw("SELECT COALESCE(version, '') AS version, last_seen_at >= ? AS connected, COUNT(*) AS count FROM destinations WHERE deleted_at IS NULL GROUP BY COALESCE(version, ''), last_seen_at >= ?", cmp, cmp).Scan(&results).Error; err != nil {
+		if err := db.Debug().Raw("SELECT *, COUNT(*) AS count FROM (SELECT COALESCE(version, '') AS version, last_seen_at >= ? AS connected FROM destinations WHERE deleted_at IS NULL) AS d GROUP BY version, connected", timeout).Scan(&results).Error; err != nil {
```
BruceMacD, line 123:
```suggestion
		if err := db.Raw("SELECT *, COUNT(*) AS count FROM (SELECT COALESCE(version, '') AS version, last_seen_at >= ? AS connected FROM destinations WHERE deleted_at IS NULL) AS d GROUP BY version, connected", timeout).Scan(&results).Error; err != nil {
```

Not sure if you meant to commit this `Debug()` part

BruceMacD, line 123:
```suggestion
		if err := db.Debug().Raw("SELECT *, COUNT(*) AS count FROM (SELECT COALESCE(version, '') AS version, last_seen_at >= ? AS connected FROM destinations WHERE deleted_at IS NULL) GROUP BY version, connected", timeout).Scan(&results).Error; err != nil {
```
Not sure we need to name this where we dont use it

BruceMacD (commented):
I'm not sure the db host, name, and username need to be secret. Is there a scenario where they contains sensitive info?

internal/server/server.go
```diff
@@ -310,18 +310,28 @@ func (s *Server) getPostgresConnectionString() (string, error) {
 
 	if s.options.DBHost != "" {
 		// config has separate postgres parameters set, combine them into a connection DSN now
-		fmt.Fprintf(&pgConn, "host=%s ", s.options.DBHost)
+		host, err := secrets.GetSecret(s.options.DBHost, s.secrets)
```
dnephin, line 313: As Bruce was saying, I don't quite understand why dbhost, username, and dbname would be secrets. We can already read these values from environment variables.

If we wanted them to also come from files we could use `types.StringOrFile`, but I don't know why we'd want them to come from files.

Even if you use a kube secret to store them, can't we set values from [a kube secret as environment variables on the pod](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables)?

internal/server/metrics.go
mxyng: Oops good catch. Removed

mxyng:
Split the metrics change to https://github.com/infrahq/infra/pull/2580

PR #2574, opened 2022-07-14
infra chart: typo in server labels helper function

dkulchinsky:
## Summary

Small typo fix in server's labels helper function, was referencing a non-existing key in the global values block

## Checklist

## Related Issues

Resolves #


dnephin (approved):
Thank you for the fix!

PR #2590, opened 2022-07-14
Add validation rules to provider requests

dnephin:
## Summary

Branched from #2570

This PR adds a new `Enum` rule to `internal/validate`, and adds validation rules for `CreateProvider`, and `UpdateProvider`. Also adds API handler tests for those operations.

Best viewed by individual commit. There are a couple small cleanups in separate commits.

## Related Issues

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -635,11 +635,10 @@ func (a *API) UpdateIdentityInfoFromProvider(c *gin.Context) error {
 	return access.UpdateIdentityInfoFromProvider(c, oidc)
 }
 
-func (a *API) providerClient(c *gin.Context, provider *models.Provider, redirectURL string) (providers.OIDCClient, error) {
-	if val, ok := c.Get("oidc"); ok {
+func (a *API) providerClient(ctx context.Context, provider *models.Provider, redirectURL string) (providers.OIDCClient, error) {
+	if c := providers.OIDCClientFromContext(ctx); c != nil {
 		// oidc is added to the context during unit tests
-		oidc, _ := val.(providers.OIDCClient)
-		return oidc, nil
+		return c, nil
 	}
 
 	clientSecret, err := secrets.GetSecret(string(provider.ClientSecret), a.server.secrets)
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -233,7 +233,7 @@ func (a *API) CreateProvider(c *gin.Context, r *api.CreateProviderRequest) (*api
 	}
 	provider.Kind = kind
 
-	if err := a.setProviderInforFromServer(c, provider); err != nil {
+	if err := a.setProviderInfoFromServer(c, provider); err != nil {
 		return nil, err
 	}
 
@@ -261,7 +261,7 @@ func (a *API) UpdateProvider(c *gin.Context, r *api.UpdateProviderRequest) (*api
 	}
 	provider.Kind = kind
 
-	if err := a.setProviderInforFromServer(c, provider); err != nil {
+	if err := a.setProviderInfoFromServer(c, provider); err != nil {
 		return nil, err
 	}
 
@@ -652,7 +652,7 @@ func (a *API) providerClient(c *gin.Context, provider *models.Provider, redirect
 }
 
 // setProviderInfoFromServer checks information provided by an OIDC server
-func (a *API) setProviderInforFromServer(c *gin.Context, provider *models.Provider) error {
+func (a *API) setProviderInfoFromServer(c *gin.Context, provider *models.Provider) error {
 	// create a provider client to validate the server and get its info
 	oidc, err := a.providerClient(c, provider, "http://localhost:8301")
 	if err != nil {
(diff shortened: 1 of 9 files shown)


internal/server/providers_test.go
```diff
@@ -120,3 +128,214 @@ func TestAPI_DeleteProvider(t *testing.T) {
...
+			},
+		},
+		{
+			name: "not authorized",
+			setup: func(t *testing.T, req *http.Request) {
+				accessKey, _ := createAccessKey(t, srv.db, "usera@example.com")
+				req.Header.Set("Authorization", "Bearer "+accessKey)
+			},
+			expected: func(t *testing.T, resp *httptest.ResponseRecorder) {
+				t.Skip("TODO: requires either a fake oidc client, or to authorize before validation")
```
dnephin, line 178: The two "not authorized" tests in this PR are skipped right now. 

It's not easy to test this because we validate the request before we authorize it. Part of the validation is checking the `client-id`, and `client-secret` work when sent to the provider at `url`.

We don't want to check-in valid credentials to a provider, so that leaves us two options:

1. change the API handler to authorize before validation. This might be a good thing to do in general, but likely requires replacing the access package (https://github.com/infrahq/internal/issues/20).
2. inject a fake OIDC client so that validation of those fields passes.

We have a lot of the pieces required for 2. There's are a couple `mockOIDCImplementation`, we'd need to export one so that we can use it in other packages. 
Currently the mock is read from `gin.Context`. I think we'll want to change it to read from `context.Context`, so that we can inject it into the request. There's no way to inject a value into the `gin.Context`, which I think is a good thing. We don't want tests relying on `gin.Context` so that we can easily remove it if we need to.

BruceMacD: We could create another mock client in this package, they're pretty simple. I also like switching things around to authorize before validating, but as you said that's a big change for a small problem.

internal/validate/string.go
```diff
@@ -106,3 +107,35 @@ func inRange(ranges []CharRange, c rune) bool {
...
+}
+
+type enum struct {
+	Name    string
+	Value   string
+	Allowed []string
+}
+
+func (e enum) Validate() *Failure {
+	if e.Value == "" {
```
BruceMacD, line 124: Maybe this should be called `OptionalEnum`, it isnt super obvious that this will return valid for an empty value. That is the behavior we want in the providers case, but maybe not in general.

internal/server/handlers.go
```diff
@@ -652,7 +652,7 @@ func (a *API) providerClient(c *gin.Context, provider *models.Provider, redirect
 }
 
 // setProviderInfoFromServer checks information provided by an OIDC server
-func (a *API) setProviderInforFromServer(c *gin.Context, provider *models.Provider) error {
+func (a *API) setProviderInfoFromServer(c *gin.Context, provider *models.Provider) error {
```
BruceMacD, line 655: Good catch, thanks

BruceMacD (commented):
Looking good so far, some minor feedback. I'll let this get in before #2592 and update that PR with the validation in this change.

internal/validate/string.go
dnephin: Good call out! This is different from our previous library. 

I've been making all of the validation rules optional, for example:
* a `StringRule` with a minLength > 0 is still valid if the string is the empty string
* an int of `0` is still valid, even with an `IntRule` with a min > 0
* an email address is still valid if it's the empty string

Since we have a `Required` validation rule, it seems like we already have a good way of making a field required, apply both the required rule and some other rule. 

So it seemed to me like all rules should default to optional, which means we probably don't need to name them that way.  

But I think it might be good to call that out in the docs. Maybe the godoc for `ValidationRule` should say that?

BruceMacD: That seems reasonable to me 👍

internal/server/providers_test.go
dnephin: I pushed another commit which uses the fake OIDC client, and these tests are no longer skipped.

internal/server/handlers.go
```diff
@@ -635,11 +635,10 @@ func (a *API) UpdateIdentityInfoFromProvider(c *gin.Context) error {
 	return access.UpdateIdentityInfoFromProvider(c, oidc)
 }
 
-func (a *API) providerClient(c *gin.Context, provider *models.Provider, redirectURL string) (providers.OIDCClient, error) {
-	if val, ok := c.Get("oidc"); ok {
+func (a *API) providerClient(ctx context.Context, provider *models.Provider, redirectURL string) (providers.OIDCClient, error) {
```
dnephin, line 638: `gin.Context` implements `context.Context`, so none of the callers needed to change here.

internal/server/providers/oidc.go
```diff
@@ -38,6 +38,21 @@ type OIDCClient interface {
 	GetUserInfo(ctx context.Context, providerUser *models.ProviderUser) (*UserInfoClaims, error)
 }
 
+type key struct{}
+
+var ctxKey = key{}
```
dnephin, line 43: This is the recommended way of creating a key for storing things in `context.Context`. By creating a value of an unexported struct, no other place can read or set this value. It forces everything to use the two functions below, which return or accept the typed value.

mxyng (approved):

PR #2593, opened 2022-07-14
Add validation rules to user requests

dnephin:
## Summary

Branched from #2590

Best viewed by individual commit, as I moved tests into `users_test.go` and refactored one.

This PR adds validation rules for user requests, and add tests for the validation.

## Related Issues


internal/server/users_test.go
```diff
@@ -422,7 +422,8 @@ func TestAPI_CreateUser(t *testing.T) {
 		body := jsonBody(t, tc.body)
 		req, err := http.NewRequest(http.MethodPost, "/api/users", body)
 		assert.NilError(t, err)
-		req.Header.Add("Authorization", "Bearer "+adminAccessKey(srv))
+		req.Header.Set("Authorization", "Bearer "+adminAccessKey(srv))
+		req.Header.Set("Infra-Version", "0.13.6")
```
BruceMacD, line 426: Should we automatically set this to the current version to reveal breaking changes? I think that is possible.

BruceMacD (approved):

PR #2604, opened 2022-07-15
api: remove deprecated routes and old migrations

dnephin:
## Summary

This PR removes old deprecates routes. The last one was deprecated in 0.13.1, but most of them are from 0.9 and 0.12.

It also removes old migrations. All the migrations appear to be from 0.12.2, which is now two months old. No one should be using a client that old, so we should be able to remove these from old versions from our pre-public announcement era.

In the future, when we have users on older version, we can keep migrations along for longer.


jmorganca (approved):
Thanks!!

PR #2605, opened 2022-07-15
fix: make infra version header required

### LINKED ISSUE #2588: Require Infra-Version header for all API endpoints
While writing more API handler tests I noticed that the `Infra-Version` header is only required if there is one or more migrations for that url. Endpoints like `UpdateProvider` and `CreateProvider` that don't have migrations don't end up requiring the header.

I guess we want the header to be required for all endpoints, otherwise adding the first migration for a url risks breaking a client that was only using endpoints that didn't previously require a version.

dnephin:
## Summary

Last release we attempted to make the `Infra-Version` header a required header. Unfortunately it did not work as we expected. The header was required for any path that had a migration registered to it. But most of our paths are new and don't have migrations (the latest migration was changing from `/v1` to `/api`).

This meant that most requests did not actually require a version header.

This PR fixes the problem by checking the version header in all cases, regardless if a migration exists or not.

Fixing this required adding the header to many tests.

## Related Issues

Resolves #2588


jmorganca (approved):
Thanks!!

PR #2608, opened 2022-07-15
Fix TestMetrics failing on the release branch

dnephin:
## Summary

This test depends on the product version, so the expected value changes
every time we create a new release branch, and after every release.

This commit fixes the problem by using the existing shim we added for
openapi doc generation. We can patch that shim in a test to set a
hardcoded version.

This way the expected value does not change when the product version
changes.

The test failure can be seen here: https://github.com/infrahq/infra/runs/7362326501?check_suite_focus=true

### DIFF
--- a/internal/server/metrics.go
+++ b/internal/server/metrics.go
@@ -12,7 +12,7 @@ import (
 )
 
 func setupMetrics(db *gorm.DB) *prometheus.Registry {
-	registry := metrics.NewRegistry(apiVersion())
+	registry := metrics.NewRegistry(productVersion())
 
 	if rawDB, err := db.DB(); err == nil {
 		registry.MustRegister(collectors.NewDBStatsCollector(rawDB, db.Dialector.Name()))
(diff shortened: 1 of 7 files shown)


jmorganca (approved):

internal/server/metrics.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func setupMetrics(db *gorm.DB) *prometheus.Registry {
-	registry := metrics.NewRegistry()
+	registry := metrics.NewRegistry(apiVersion())
```
mxyng, line 15: This should be `internal.FullVersion()` as well

mxyng (commented):
Oops thanks for fixing this. Left a comment on the naming but otherwise LGTM

internal/server/metrics.go
dnephin: It can't reference `internal.FullVersion` directly. We need the variable to patch it.

But I can change the variable from `apiVersion`, to `productVersion`. Would that work?

dnephin: I pushed another commit which renames the variable and the patch function to `productVersion`

mxyng (approved):

PR #2610, opened 2022-07-15
fix: kube config overwrite temp file

### LINKED ISSUE #2586: Windows CLI error updating Kube config
## Describe the bug
When running a command that updates the Kube config on Windows an error about concurrent file access may be output:
```
Error: could not overwrite kube config: rename ...\.kube\infra-kube-config-34243432 ...\.kube\config: The process cannot access the file because it is being used by another process.
```

## Expected behavior
Kube config is successfully updated.

## Environment
Windows 11

BruceMacD:
- close temp file before rename to avoid err on windows

## Summary
In the case of Window overwriting the Kube config using a temp file was not possible, as the temp file was still considered to be open. This was not see on Unix-based system. This could be due to how Windows handles changing file names.

This change fixes this issue by manually creating the temp file to write kube config to, and closing it before using it to overwrite the current Kube config.

## Checklist

## Related Issues

Resolves #2586

### DIFF
--- a/internal/cmd/kubernetes.go
+++ b/internal/cmd/kubernetes.go
@@ -2,10 +2,10 @@ package cmd
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/goware/urlx"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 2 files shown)


internal/cmd/kubernetes.go
```diff
@@ -193,26 +193,42 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
...
+
+// safelyWriteConfigToFile creates a temp file, then overwrites the target
+func safelyWriteConfigToFile(kubeConfig clientcmdapi.Config, fileToWrite string) error {
+	// get the directory of the file we're writing to avoid cross-filesystem moves
+	configDir := filepath.Dir(fileToWrite)
 	if err := os.MkdirAll(configDir, os.ModePerm); err != nil {
 		return err
 	}
 
+	temp, err := os.Create(fmt.Sprintf("%s/infra-kube-config-%d", configDir, time.Now().Unix()))
```
dnephin, line 209: Doesn't `ioutil.TempFile` return the same type (`os.File`) ?  Is there something about the temp file that doesn't work?

dnephin (approved):
LGTM, but I don't think I understand the change from `ioutil.TempFile` to `os.Create`.

Issue #2583, opened 2022-07-14
Replace remaining uses of go-playground/validator


dnephin:
When #1763 is complete all `api` request structs will be using `internal/validate` and we'll be able to remove the static `pgValidate` variable from the `internal/server` package.

This issue is to replace other uses of `go-playground/validator` with `internal/validate`.

* [x] `server/config.go` #2612
* [x] `data/migrations.go` - there was only one, so bundled it with #2611
* [x] `server/migrations.go` - removed in #2604
* [x] `models/*` - should be removed along with gorm #2659
* [x] `providers/oidc.go` - #2611

For any of these that only use `require`, we may be better off with a simple `if v.Field != ""` instead of using a validation library. I believe `models` should only use `require`.

PR #2611, opened 2022-07-15
Remove pgValidate from providers

dnephin:
## Summary

We can provide better error messages by validating the fields manually, so no need to use `internal/validate` here.

## Related Issues

Related to #2583


internal/server/providers/azure_test.go
```diff
@@ -125,7 +125,7 @@ func TestAzure_GetUserInfo(t *testing.T) {
 				"picture": "https://graph.microsoft.com/v1.0/me/photo/$value"
 			}`,
 			verifyFunc: func(t *testing.T, info *UserInfoClaims, err error) {
-				assert.ErrorContains(t, err, "failed on the 'required_without' tag")
+				assert.ErrorContains(t, err, "must include either a name or email")
```
jmorganca, line 128: Way better!

jmorganca (approved):

PR #2612, opened 2022-07-15
improve: replace server config validation

dnephin:
Branched from #2611

With internal/validate

Also required adding slice traversal to internal/validate

## Related Issues

Related to #2583

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -13,6 +13,7 @@ import (
 	"golang.org/x/crypto/bcrypt"
 	"gorm.io/gorm"
 
+	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data"
@@ -34,7 +35,7 @@ type Provider struct {
 
 func (p Provider) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
-		// TODO: api.ValidateName
+		api.ValidateName(p.Name),
 		validate.Required("name", p.Name),
 		validate.Required("url", p.URL),
 		validate.Required("clientID", p.ClientID),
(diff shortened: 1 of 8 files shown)


internal/server/config.go
```diff
@@ -19,55 +18,106 @@ import (
...
+	ClientID     string
+	ClientSecret string
 	Kind         string
 	AuthURL      string
 	Scopes       []string
 }
 
+func (p Provider) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		// TODO: api.ValidateName
```
jmorganca, line 37: Excited for this: consistent naming validation is so important for APIs

jmorganca (approved):

internal/server/config.go
dnephin: I pushed another commit which resolves this TODO

dnephin (commented):
Added some local test coverage in the `internal/validate` package for slice traversal. 

Also changed the default field name (when there are no tags) to use the field name with the first letter lowercased.

internal/server/config.go
```diff
@@ -9,65 +9,116 @@ import (
...
+}
+
 type Grant struct {
-	User     string `validate:"excluded_with=Group,excluded_with=Machine"`
-	Group    string `validate:"excluded_with=User,excluded_with=Machine"`
-	Machine  string `validate:"excluded_with=User,excluded_with=Group"` // deprecated
-	Resource string `validate:"required"`
+	User     string
+	Group    string
+	Machine  string // deprecated
```
mxyng, line 49: Can probably remove `machine` now

mxyng (approved):

internal/server/config.go
dnephin: I was wondering about that. Going to leave it for a follow up in case there are other implications to removing it.

Issue #2500, opened 2022-07-06
connector: support running the connector outside of a kube pod


dnephin:
Today the connector expects to be run from a kube pod. It looks for files that are setup by kube when the pod is deployed.

Those are great defaults, but it would also be really convenient to be able to run the connector locally and point it at a kubernetes cluster.  This would make it much easier to try out infra (quickstart), and easier to develop.

There may even be some operational advantages to deploying outside of a kube pod. If a kube cluster fails, then the connector auth proxy will not be reachable, so it may not be possible to recover the cluster.  If the connector is running somewhere else, then restoring access after an outage might be easier.

I'm not entirely sure what would need to change to support this. I'll add a list to this issue as I find things.

PR #2613, opened 2022-07-15
Allow the connector to run outside of a kube pod

dnephin:
## Summary

Branched from #3220. Best viewed by individual commit, and the third commit benefits from the  "Hide Whitespace" view.

This PR adds support for reading the kube API hostname and CA cert from kubectl config when reading the in-cluster files fails. This is the first step to allow us to run the connector outside of kube (primarily for testing).

Also extracts a new `type connector struct` for the many args being passed to `syncWithServer`, and extracts a new `syncDestination` from that function as well. This is the first step to split up the `syncWithServer` goroutine into its two responsibilities: updating the destination, and updating role bindings.

In the next follow up PR I'll be addressing the remaining in-cluster calls, to allow the connector to run outside of a kube pod.

Related to #2500 (but doesn't get us all the way there yet)

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -270,81 +270,90 @@ func httpTransportFromOptions(opts ServerOptions) *http.Transport {
 	return transport
 }
 
+func syncDestination(con connector) error {
+	host, port, err := con.k8s.Endpoint()
+	if err != nil {
+		return fmt.Errorf("failed to lookup endpoint: %w", err)
+	}
+
+	if ipv4 := net.ParseIP(host); ipv4 == nil {
+		// wait for DNS resolution if endpoint is not an IPv4 address
+		if _, err := net.LookupIP(host); err != nil {
+			return fmt.Errorf("host could not be resolved: %w", err)
+		}
+	}
+
+	// update certificates if the host changed
+	_, err = con.certCache.AddHost(host)
+	if err != nil {
+		return fmt.Errorf("could not update self-signed certificates")
+	}
+
+	endpoint := fmt.Sprintf("%s:%d", host, port)
+	logging.Debugf("connector serving on %s", endpoint)
+
+	namespaces, err := con.k8s.Namespaces()
+	if err != nil {
+		return fmt.Errorf("could not get kubernetes namespaces: %w", err)
+	}
+
+	clusterRoles, err := con.k8s.ClusterRoles()
+	if err != nil {
+		return fmt.Errorf("could not get kubernetes cluster-roles: %w", err)
+	}
+
+	switch {
+	case con.destination.ID == 0:
+		// TODO: move this warning somewhere earlier in startup
+		isClusterIP, err := con.k8s.IsServiceTypeClusterIP()
+		if err != nil {
+			logging.Debugf("could not determine service type: %v", err)
+		}
+
+		if isClusterIP {
+			logging.Warnf("registering Kubernetes connector with ClusterIP. it may not be externally accessible. if you are experiencing connectivity issues, consider switching to LoadBalancer or Ingress")
+		}
+
+		fallthrough
+
+	case !slicesEqual(con.destination.Resources, namespaces):
+		con.destination.Resources = namespaces
+		fallthrough
+
+	case !slicesEqual(con.destination.Roles, clusterRoles):
+		con.destination.Roles = clusterRoles
+		fallthrough
+
+	case string(con.destination.Connection.CA) != string(con.options.CACert):
+		con.destination.Connection.CA = api.PEM(con.options.CACert)
+		fallthrough
+
+	case con.destination.Connection.URL != endpoint:
+		con.destination.Connection.URL = endpoint
+
+		if err := createOrUpdateDestination(con.client, con.destination); err != nil {
+			return fmt.Errorf("create or update destination: %w", err)
+		}
+	}
+	return nil
+}
+
 func syncWithServer(con connector) func(context.Context) {
 	return func(context.Context) {
-		host, port, err := con.k8s.Endpoint()
-		if err != nil {
-			logging.Errorf("failed to lookup endpoint: %v", err)
+		if err := syncDestination(con); err != nil {
+			logging.Errorf("failed to update destination in infra: %v", err)
 			return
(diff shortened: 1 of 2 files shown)


internal/connector/connector.go
```diff
@@ -31,10 +32,12 @@ import (
 )
 
 type Options struct {
-	Server ServerOptions
-	Name   string
-	CACert string
-	CAKey  string
+	Server        ServerOptions
+	Name          string
+	CACert        string // TODO: types.StringOrFile
```
jmorganca, line 37: While just a comment, I absolutely love the idea of `types.StringOrFile`

mxyng (commented):
I'm not sure how I feel about this. Running the connector outside of Kubernetes is a great change but keeping support for Kubernetes while in this mode seems unnecessary. Perhaps what is needed is to decouple the connector from the Kubernetes connector so it can be run separately with possibly other integrations

dnephin:
> but keeping support for Kubernetes while in this mode seems unnecessary

The goal here isn't to support different kinds of connectors, but to allow the kubernetes connector to run in different environments. For "getting started" and while developing/contributing, being able to run the connector locally outside of a kube pod is a big improvement. It removes all the complications of helm, and makes it much easier to iterate and try things out.

Even in production environments, having the authentication proxy run inside the environment for which it is authenticating requests could cause problems, especially when trying to recover that environment.

> Perhaps what is needed is to decouple the connector from the Kubernetes connector so it can be run separately with possibly other integrations

I think that would also be a good change, but there's not much value in doing that until we are ready to add a different kind of destination.

internal/kubernetes/kubernetes.go
```diff
@@ -16,52 +16,41 @@ import (
...
 	}
 
-	clientset, err := kubernetes.NewForConfig(k.Config)
-	if err != nil {
-		return nil, err
+	if err := rest.LoadTLSFiles(config); err != nil {
+		return nil, fmt.Errorf("load TLS files: %w", err)
 	}
 
-	k.SecretReader = secrets.NewKubernetesSecretProvider(clientset, namespace)
```
dnephin, line 62: This field was not used anywhere, so I was able to remove it.

internal/connector/connector.go
```diff
@@ -172,14 +169,8 @@ func Run(ctx context.Context, options Options) error {
...
 
-	clusterCACert, err := kubernetes.CA()
-	if err != nil {
-		return fmt.Errorf("reading CA file: %w", err)
-	}
-
 	certPool := x509.NewCertPool()
-
-	if ok := certPool.AppendCertsFromPEM(clusterCACert); !ok {
+	if ok := certPool.AppendCertsFromPEM(k8s.Config.CAData); !ok {
```
dnephin, line 188: The CA data is populated by the `rest.InClusterConfig` as long as we call `rest.LoadTLSFiles` (which I added to `NewKubernetes`). This removes the need to look it up again, and helps keep most the in-cluster logic in the `NewKubernetes` constructor.

internal/connector/connector.go
```diff
@@ -55,17 +55,14 @@ type ListenerOptions struct {
...
+		return fmt.Errorf("failed to create kubernetes client: %w", err)
 	}
 
-	chksm, err := k8s.Checksum()
-	if err != nil {
-		logging.Errorf("k8s checksum error: %s", err)
-		return err
-	}
+	checkSum := k8s.Checksum()
+	logging.L.Debug().Str("uniqueID", checkSum).Msg("Cluster uniqueID")
```
dnephin, line 70: I've tested this manually for now, since we don't quite yet have an easy way of testing this outside of kube.

I tested this by running `infra connector` locally using my kubectl config, and again by running it in a pod. I compared the checksum in both cases to ensure it produced the same value.

internal/kubernetes/kubernetes.go
```diff
@@ -501,6 +487,8 @@ func (k *Kubernetes) Name(chksm string) (string, error) {
 	return name, nil
 }
 
+const podLabelsFilePath = "/etc/podinfo/labels"
```
dnephin, line 490: Moved this constant to next to where it is used. This helps made code more readable by reducing the need to scroll around the file when reading this function, or looking for functions that use the constant.

I know some styles encourage putting all the constants at the top of the file, but I think often that's not the best choice. It's fine if the constants are for something like an enum that will be used all over the place. For constants like this that are only used by a single function, and are unlikely to ever be used in another place, I think it works much better to keep it close to the function where it is used.

internal/connector/connector.go
```diff
@@ -52,6 +51,15 @@ type ListenerOptions struct {
 	Metrics string
 }
 
+// connector stores all the dependencies for the connector operations.
+type connector struct {
```
dnephin, line 55: I'm never sure what to call these, so I went with the same pattern we used for `Server` and called it `connector`. As the code changes maybe we'll find a better name for this, or find ways to split it into other types.

I'm doing this now to allow us to split up `syncWithServer` and removing the need to pass around 5+ args to the functions.

internal/connector/connector.go
```diff
@@ -270,76 +270,79 @@ func httpTransportFromOptions(opts ServerOptions) *http.Transport {
...
 
-func syncWithServer(con connector) func(context.Context) {
-	return func(context.Context) {
-		host, port, err := con.k8s.Endpoint()
-		if err != nil {
-			logging.Errorf("failed to lookup endpoint: %v", err)
-			return
-		}
+func syncDestination(con connector) error {
+	host, port, err := con.k8s.Endpoint()
```
dnephin, line 274: The next step for allowing this to run locally will be to add a config setting that allows us to skip the endpoint lookup. So before making that change I'm extracting this logic from `sycnWithServer` to give us some space to add that new logic.

dnephin (commented):
Marking this as ready for review. I've cleaned up the changes, and tested it manually. 

I cut out some of the changes required to run locally and I'll re-introduce those in a follow up PR.

internal/connector/connector.go
dnephin: Done in #3220 !

mxyng (approved):

Issue #2600, opened 2022-07-15
Certificate generated for new server doesn't get correct IP


BruceMacD:
## Describe the bug
When deploying the Infra server without a certificate one gets automatically generated, but it may not be generated for the correct host.

In my case I expect the IP to match the IP of my Kubernetes cluster, but it was automatically set to `127.0.0.1`.

```
$ helm install infra infrahq/infra
...
$ kubectl get service infra-server -o jsonpath="{.status.loadBalancer.ingress[*]['ip', 'hostname']}" -w
123.456.789.000
$ infra login 123.456.789.000
Error: Get "https://123.456.789.000": x509: certificate is valid for 127.0.0.1, ::1, 10.244.0.40, not 123.456.789.000
```

## Expected behavior
The certificate is generated for address/host of the ingress I use to login.

## Environment
- Kubernetes cluster on Digital Ocean
- Kubernetes 1.22.11-do.0

```
$ infra version

 Client: 0.13.7+dev
 Server: 0.13.6
```

## Relevant Infra Logs

```
{"level":"info","SHA256 fingerprint":"CB:71:71:FC:5B:A9:6B:55:E4:9C:4F:01:46:8F:25:2B:D3:5D:C4:7F:B5:70:6C:95:39:67:A4:BC:46:5B:1C:3B","time":1657895973917,"caller":"server/tls.go:53","message":"TLS CA"}
{"level":"info","time":1657895973918,"caller":"logging/logger.go:76","message":"starting infra server (0.13.6) - http:[::]:80 https:[::]:443 metrics:[::]:9090"}
{"level":"info","time":1657896490027,"caller":"logging/logger.go:76","message":"no cached TLS cert for 10.244.0.40"}
{"level":"info","time":1657896490027,"caller":"logging/logger.go:76","message":"no cached TLS cert for 10.244.0.40"}
{"level":"info","Server name":"10.244.0.40","SHA256 fingerprint":"CC:1B:2F:75:25:30:64:E7:57:31:01:DA:74:DC:AD:29:2A:D5:90:AD:00:76:B0:5F:8F:1E:64:7A:5F:7C:39:F9","time":1657896495882,"caller":"server/tls.go:175","message":"new server certificate"}
```

PR #2614, opened 2022-07-16
fix(cli): send SNI value with infra login for certificate generation

jmorganca:
This draft PR demonstrates sending an additional SNI value on `infra login` and subsequent requests

Fixes https://github.com/infrahq/infra/issues/2600

In the future we should consider leaving TLS & certificate management up to the user since it is invariably better to use a public certificate (or signed by an internal team-specific CA)


jmorganca:
Closing for #2638

PR #2631, opened 2022-07-18
ci: disable caching in lint, and restore cache in fuzz

dnephin:
## Summary

The caching in the lint action duplicates the cache we already have from `setup-go`. Using the second cache causes ~23000 lines of warnings in the lint step.  Disabling the cache should prevent that.

Also restore the caching of the fuzz corpus. I had removed the cache because I thought the setup-go cache would take care of it, but it turns out that setup-go only caches go/cache/mod, not all of go/cache.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -96,6 +96,9 @@ jobs:
         with:
           version: v1.46.2
           args: --verbose
+          # setup-go already restores the build and module cache
+          skip-build-cache: true
+          skip-pkg-cache: true
 
   helm-lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


dnephin:
I'm going to let the fuzz jobs finish, then run it again to see it use the cache. The first run was a cache miss because we don't run these jobs on `main`.

.github/workflows/pull_request.yml
```diff
@@ -96,6 +96,9 @@ jobs:
         with:
           version: v1.46.2
           args: --verbose
+          # setup-go already restores the build and module cache
+          skip-build-cache: true
+          skip-pkg-cache: true
```
dnephin, line 101: Seems like lint runs even faster with these disabled (17s vs ~30s)

mxyng (approved):

PR #2639, opened 2022-07-19
fix: remove direct dependency on aws-sdk-go

### LINKED ISSUE #2456: Get rid of dependency on aws-sdk-go
## Describe the bug

`aws-sdk-go` is used to discover the cluster name of a deployment in AWS. This can be accomplished by talking to the [metadata service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instance-metadata-ex-7) directly. In fact, the connector is already doing something similar to retrieve the instance ID which it passes to `aws-sdk-go`.

In place of `aws-sdk-go`, the connector can implement a simple metadata client the above service.

A side note, the currently cluster name discovery only looks for `kubernetes.io/cluster-name/` prefixed tag for the cluster name. There's another `eks:cluster-name` which may also hold the cluster name as tag value. The connector should look for both.

## Environment

```
$ infra version
```

```
$ kubectl version
```

mxyng:
## Summary

Use the AWS API directly to get AWS cluster name. Use a combination of
instance metadata and AWS API. The instance metadata tags is not
sufficient because 1) it requires the user to explicitly enable
exporting instance tags (it's not enabled by default) and 2) the format
of the tags percludes it from being exposed through instance metadata
(they contain `/`).

Unfortunately, cannot remove aws-sdk-go completely because
infrahq/secrets depends on it.

## Checklist

## Related Issues

Resolves #2456

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.5
-	github.com/aws/aws-sdk-go v1.44.56
+	github.com/aws/aws-sdk-go v1.44.56 // indirect
 	github.com/cli/browser v1.1.0
 	github.com/gin-gonic/gin v1.8.1
 	github.com/go-playground/validator/v10 v10.11.0
--- a/internal/kubernetes/kubernetes.go
+++ b/internal/kubernetes/kubernetes.go
@@ -2,20 +2,20 @@ package kubernetes
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


mxyng:
TBH I'm inclined to leave it as is. The additional complexity is not worth it. The dependency isn't even being removed.

mxyng:
Dropping this for now. If/when the indirect dependency on aws-sdk-go is removed, this PR can be revived.

PR #2641, opened 2022-07-19
fix: correctly set `anonymousId` and `userId` for heartbeat events

jmorganca:


internal/server/telemetry.go
```diff
@@ -90,7 +90,28 @@ func (t *Telemetry) EnqueueHeartbeat() {
...
 
-	t.Event("heartbeat", "", map[string]interface{}{
+	first, err := data.ListIdentities(t.db, &models.Pagination{Limit: 1}, data.NotName(models.InternalInfraConnectorIdentityName))
+	if err != nil {
+		logging.Debugf("%s", err.Error())
+	}
+
+	var admin models.Identity
+	if len(first) > 0 {
+		admin = first[0]
```
BruceMacD, line 100: Is this guaranteed to be the admin? What if Infra was setup using a config file?

dnephin: The list of `ListIdentities` is most likely alphabetical as well, so probably not the admin.

I think we likely need to query for grants and find the `infra/admin` grant with the earliest created time.

jmorganca: Thanks for the review. This should be ready for another quick look. @BruceMacD @dnephin

internal/server/telemetry.go
```diff
@@ -90,7 +90,48 @@ func (t *Telemetry) EnqueueHeartbeat() {
...
-	t.Event("heartbeat", "", map[string]interface{}{
+	admins, err := data.ListGrants(
+		t.db,
+		&models.Pagination{Limit: 1},
+		data.ByResource(access.ResourceInfraAPI),
+		data.ByPrivilege(models.InfraAdminRole),
+		data.ByIsIdentity(),
+	)
+	if err != nil {
+		logging.Debugf("%s", err.Error())
```
mxyng, line 101: Should try to use the zerolog interface directly when logging errors:

```go
logging.L.Debug().Err(err).Msg("")
```

mxyng (approved):
LGTM assuming it's ok to get one admin, not a specific admin. An improvement maybe to sort on created by to get the first admin user

dnephin (approved):

PR #2646, opened 2022-07-19
improve: replace request log host with localAddr

mxyng:
## Summary

Host may be misunderstood to mean the server the request is hosted
rather the address used to address the service. There's also a
nice symmetry between localAddr and remoteAddr.


BruceMacD (approved):
Makes sense 👍

PR #2649, opened 2022-07-19
improve: descriptive container names

mxyng:
## Summary

Container names are not always visible but when they are, they should
reflect the running container. .Chart.Name only produces infra, which
isn't very useful to differentiate between what _type_ of infra
deployment the container is.


jmorganca (approved):
Great edit

Issue #2024, opened 2022-05-23
enforce password requirements on server


yeshwanth1312:
## Is your feature request related to a problem? Please describe.

We would like to enforcece password requirements when users are changing their password for the first time 

## Describe the solution you'd like

We would like to see an option for having caps, special character,s and minimum password length (which is 8 now, We would like to have something configurable )

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

kimskimchi:
Progress tracker:
- [x] Design doc 
- [ ] Implement in Database #2563

Issue #2648, opened 2022-07-19
implement a way to retrieve password settings


kimskimchi:
Implement the feature to retrieve or set settings (which currently only has password settings).

PR #2651, opened 2022-07-19
feat: passwordrequirements for cli edit user

kimskimchi:
## Summary

This implements the password requirements check for the `users edit --password` command.

## Checklist

## Related Issues

Implements #2024

### DIFF
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -14,6 +14,7 @@ import (
 	"github.com/infrahq/infra/internal/generate"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/internal/validate"
 )
 
 func CreateCredential(c *gin.Context, user models.Identity) (string, error) {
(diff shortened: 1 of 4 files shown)


internal/access/credential.go
```diff
@@ -120,36 +121,34 @@ func hasMinimumCount(min int, password string, check func(rune) bool) bool {
 }
 
 func checkPasswordRequirements(db *gorm.DB, password string) error {
-	var errs []string
 	settings, err := data.GetSettings(db)
 	if err != nil {
 		return err
 	}
+	errs := make(validate.Error)
```
BruceMacD, line 128: I think it would be simpler to return an error from this function (similar to before), then wrap it in a validation error at the caller in a higher level.

internal/cmd/users.go
```diff
@@ -267,18 +286,23 @@ func updateUser(cli *CLI, name string) error {
...
+
+	PROMPT:
 		req.Password, err = promptSetPassword(cli, "")
 		if err != nil {
 			return err
 		}
 
+		logging.Debugf("call server: update user %s", req.ID)
 		if _, err := client.UpdateUser(req); err != nil {
+			if passwordError(cli, err) {
```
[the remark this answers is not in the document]
kimskimchi: 👍 do you mean this: https://github.com/infrahq/infra/issues/2648?

internal/cmd/users_test.go
```diff
@@ -21,7 +21,7 @@ func TestCheckPasswordRequirements(t *testing.T) {
 	assert.NilError(t, err)
 
 	err = checkPasswordRequirements("")("passwor")
```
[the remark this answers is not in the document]
kimskimchi: done

internal/access/credential.go
kimskimchi: resolved in person

internal/cmd/users.go
```diff
@@ -373,14 +397,15 @@ PROMPT:
...
 	return func(val interface{}) error {
 		newPassword, ok := val.(string)
 		if !ok {
 			return fmt.Errorf("unexpected type for password: %T", val)
 		}
 
-		if len(newPassword) < 8 {
-			return fmt.Errorf("input must be at least 8 characters long")
+		if len(newPassword) == 0 {
+			return fmt.Errorf("Value is required")
```
BruceMacD, line 408: nit:
```suggestion
			return fmt.Errorf("value is required")
```

BruceMacD (approved):
Looks good!

PR #2653, opened 2022-07-19
feat(ui): pagination

j-sneh:
## Summary

Adds interactive pagination links to the UI and adds pagination to API requests (for each of the main pages).

Screen Shot 2022-07-19 at 3 19 05 PM

## Checklist

### DIFF
--- a/ui/components/pagination.js
+++ b/ui/components/pagination.js
@@ -2,8 +2,8 @@ import { useRouter } from 'next/router'
 import { ChevronLeftIcon, ChevronRightIcon } from '@heroicons/react/solid'
 
 function getPageNums(selected, count, totalPages) {
-  let pageNums = []
-  let beginOffset = Math.max(3, 6 + selected - totalPages)
+  const pageNums = []
+  const beginOffset = Math.max(3, 6 + selected - totalPages)
   for (
     let i = Math.max(1, selected - beginOffset);
     pageNums.length <= count && i <= totalPages;
--- a/ui/components/pagination.js
+++ b/ui/components/pagination.js
@@ -43,7 +43,7 @@ function LeftArrow({ path, page }) {
   )
 }
 
-function RightArrow(path, page, maxPage) {
+function RightArrow({ path, page, maxPage }) {
   return (
     <a
       href={path + '?p=' + (page < maxPage ? page + 1 : Math.max(1, maxPage))}
--- a/ui/components/pagination.js
+++ b/ui/components/pagination.js
@@ -1,59 +1,83 @@
+import { useRouter } from 'next/router'
+
 import { ChevronLeftIcon, ChevronRightIcon } from '@heroicons/react/solid'
 
-export default function Pagination(curr) {
+function CurrentPage(path, page) {
   return (
-    <nav className="pb-4 px-4 flex justify-end">  
-        <a
-          href="#"
-          className=" py-2 pr-1 inline-flex items-center text-sm font-medium text-gray-500 hover:text-violet-300 "
-        >
-          <ChevronLeftIcon className=" h-5 w-5 text-gray-400 hover:text-violet-300" aria-hidden="true" />
-          
-        </a>
-        <a
-          href="#"
-          className="text-gray-500 hover:text-violet-300 py-2 px-4 inline-flex items-center text-sm font-medium"
-        >
-          1
-        </a>
-        {/* Current: "border-indigo-500 text-indigo-600", Default: "text-gray-500 hover:text-violet-300 " */}
-        <a
-          href="#"
-          className="text-white hover:text-violet-300 py-2 px-4 inline-flex items-center text-sm font-medium rounded-md bg-gray-700"
-          aria-current="page"
-        >
-          2
-        </a>
-        <a
-          href="#"
-          className="text-gray-500 hover:text-violet-300  py-2 px-4 inline-flex items-center text-sm font-medium"
-        >
-          3
-        </a>
-        <a
-          href="#"
-          className="text-gray-500 hover:text-violet-300   py-2 px-4 inline-flex items-center text-sm font-medium"
-        >
-          4
-        </a>
-        <a
-          href="#"
-          className="text-gray-500 hover:text-violet-300   py-2 px-4 inline-flex items-center text-sm font-medium"
-        >
-          5
-        </a>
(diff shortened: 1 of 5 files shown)


ui/components/pagination.js
```diff
@@ -0,0 +1,99 @@
...
+  const path = router.pathname
+  const limit = 13 // TODO: dynamic limit
+
+  curr = curr === undefined ? 1 : parseInt(curr)
+  totalPages = totalPages === undefined ? 1 : parseInt(totalPages)
+  totalCount = totalCount === undefined ? 0 : parseInt(totalCount)
+
+  let pages = [LeftArrow(path, curr)]
+
+  let beginOffset = Math.max(3, 6 + curr - totalPages)
```
hoyyeva, line 70: doesn't seem like `beginOffset`, `lowerItem` and `upperItem` are getting reassigned. So we can use `const` instead of `let`

ui/components/pagination.js
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+      pages.push(Page(path, page))
+    }
+  }
+  pages.push(RightArrow(path, curr, totalPages))
+
+  let lowerItem = totalCount === 0 ? 0 : 1 + (curr - 1) * limit
+  let upperItem = Math.min(1 + (curr - 1) * limit + limit, totalCount)
+
+  return [
```
hoyyeva, line 87: why are we returning this as an array?

ui/pages/destinations/index.js
```diff
@@ -386,6 +389,11 @@ export default function Destinations() {
                 )}
               </div>
             )}
+            <Pagination
```
hoyyeva, line 392: do we want to hide the pagination if the data is less than 13?

ui/components/pagination.js
```diff
@@ -0,0 +1,95 @@
+import { useRouter } from 'next/router'
+import { ChevronLeftIcon, ChevronRightIcon } from '@heroicons/react/solid'
+
+function getPageNums(selected, count, totalPages) {
```
hoyyeva, line 4: you can move this function in `Pages`  because no one else is using this function.

ui/components/pagination.js
```diff
@@ -0,0 +1,95 @@
...
+    >
+      <ChevronRightIcon
+        className='ml-3 h-5 w-5 text-gray-400 hover:text-violet-300'
+        aria-hidden='true'
+      />
+    </a>
+  )
+}
+
+export default function Pagination({ curr, totalPages, totalCount }) {
```
hoyyeva, line 60: to check if it is undefined - what you can do is you can set a default value for each props 
so something like this 
`Pagination({ curr = 1, totalPages = 1, totalCount = 0 }) { ... }` 
so then you don't need to do the check on line 65, 66, 67, and 73

ui/components/pagination.js
```diff
@@ -0,0 +1,95 @@
...
+  totalCount = totalCount === undefined ? 0 : parseInt(totalCount)
+
+  const lowerItem = totalCount === 0 ? 0 : 1 + (curr - 1) * limit
+  const upperItem = Math.min(1 + (curr - 1) * limit + limit, totalCount)
+
+  return (
+    totalPages !== undefined &&
+    totalPages > 1 && (
+      <div>
+        <nav key='paginator' className='flex justify-end px-4 pb-2'>
```
hoyyeva, line 76: with this, can we try match design we have in figma?

ui/pages/groups/index.js
```diff
@@ -181,6 +183,7 @@ function Details({ group, admin, onDelete }) {
               )}
               <IdentityList
                 list={users}
+                actionText='Remove'
```
hoyyeva, line 186: We don't need the `Remove` here - it is set in the `IdentityList` component as the default value

ui/components/pagination.js
```diff
@@ -0,0 +1,95 @@
...
+          <LeftArrow path={path} page={curr}></LeftArrow>
+          <Pages
+            path={path}
+            count={7}
```
hoyyeva, line 80: is the `count` not dynamic?

ui/pages/destinations/index.js
hoyyeva: so do we want to hide the pagination if there is no next page?

hoyyeva: Oh i see you put the check on the pagination component. What I would do is probably put it in each parent component - because if there is no next page - we don't need to render the pagination at all

ui/pages/users/index.js
```diff
@@ -158,6 +160,7 @@ function Details({ user, admin, onDelete }) {
               )}
               <IdentityList
                 list={groups}
+                actionText='Remove'
```
hoyyeva, line 163: you can remove this line - the `IdentityList` sets the default value

ui/components/pagination.js
```diff
@@ -0,0 +1,95 @@
...
+        selected === page ? 'rounded-md bg-gray-700' : ''
+      }`}
+      key={page}
+    >
+      {page}
+    </a>
+  ))
+}
+
+function LeftArrow({ path, page }) {
```
hoyyeva, line 32: can we try to put `LeftArrow` and `RightArrow` to the same component?
So it would be something like this 

``function Arrow({path, direction}) {return (<a
      href={path}
      className=' inline-flex items-center pr-1 text-sm font-medium text-gray-500 hover:text-violet-300 '
    >{direction === 'left' && <ChvronLeftIcon ... />}{direction === 'RIGHT' && <ChvronRightIcon ... />}</a>)}``

hoyyeva: And we should probably calculate the `path` outside of this component - so the component should be a dummy where it doesn't have any logic but just render and display what it is being told to

ui/components/pagination.js
j-sneh: maybe `maxCount` is a better name for it. Essentially, it'll try to display `count` with a (roughly) equal number on the left side and right side of the currently selected page, but if it can't (too small or too large) it will display less. In figma, there's 7, so that's what I chose. Do you have a different idea?

hoyyeva (approved):

hoyyeva:
lgtm! 💯 will you be able to work on the unit test as a follow up task? @j-sneh

PR #2654, opened 2022-07-19
improve: remove msg for api errors

mxyng:
## Summary

`Msg` and similar functions sets the `message` field in the log json.
It's not very useful in the case of API errors because the interesting
bits are in the other attributes. Having a message field causes the
`message` to be highlighted in aggregators like Datadog which makes it
harder to see the interesting bits.

### DIFF
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -97,7 +97,7 @@ func sendAPIError(c *gin.Context, err error) {
 		Str("path", c.Request.URL.Path).
 		Int32("statusCode", resp.Code).
 		Str("remoteAddr", c.Request.RemoteAddr).
-		Msg("api request error")
+		Msg("")
 
 	c.JSON(int(resp.Code), resp)
 	c.Abort()


dnephin (commented):
Interesting! What does datadog highlight if there is no message?

Generally I would expect a field like this to be useful when searching logs. It allows someone to find all the "api errors" by searching for this predictable string. Otherwise you have to search for fields which could appear in many log entries, making it hard to filter the logs to the ones you want. It could also easily cause you to miss important logs that you would care about because you don't have this predictable string to search for.

Could we include this `api request error` string in a different field to make it easier to find? Maybe something like `logType` or `logKind` ?

mxyng:
If `Msg` is omitted, Datadog will highlight the JSON structure, what it calls attributes, in its content column. If it's available, it'll populate contents with the message which in this particular case is much less interesting. 

Since the log line has error, a good alternative may be to wrap the error being passed into `Err()`, e.g.
```
{"error": "api error: <actual error>"
```

Alternatively, the key can be modified to `api.error` instead of just `error`.

In both cases, there's a common pattern for searching

dnephin:
Cool, that makes sense. I was thinking about this a bit more, and I think in other systems I've seen something like `logName` used. Some logging libraries have named loggers and the name is a special field that lets you find or filter to just one type of log.

The nice thing about a separate field (vs using `api error` in the error field) is that we could use that pattern in other places, and the user will know they can always filter using the same `logName` pattern.

BruceMacD (approved):

internal/server/errors.go
```diff
@@ -97,7 +97,7 @@ func sendAPIError(c *gin.Context, err error) {
 		Str("path", c.Request.URL.Path).
 		Int32("statusCode", resp.Code).
 		Str("remoteAddr", c.Request.RemoteAddr).
-		Msg("api request error")
+		Msg("")
```
dnephin, line 100: How about something like this?

```suggestion
		Str("logName", "api request error").
		Msg("")
```

mxyng:
Perhaps this can be addressed in datadog instead. Closing for now

Issue #2415, opened 2022-06-27
Isolate all references to gorm in the data package


pdevine:
In preparation for removing gorm and replacing it with `database/sql` (or possibly sqlx/sqlc) we need to make sure that the `data` package is the only thing that is aware of `gorm`.

Once that is done, it should be much easier to change the library we use for interacting with the database.

Tasks:

* [x] audit `database/sql`, [kyleconroy/sqlc](https://github.com/kyleconroy/sqlc), and [jmoiron/sqlx](https://github.com/jmoiron/sqlx), specially how they deal with transactions to make sure that they all fit well with the new proposed interface for the data package (described in proposal 18)
* [x] create a `RequestContext` to expose a transaction (proposal 19), and store it in `gin.Context` https://github.com/infrahq/infra/pull/2702
* [ ] convert a small number of `data` query functions to the new pattern, and call them from the handlers using the transaction from `RequestContext` #2657
* [ ] if everything looks good, repeat for all the other `data` query functions

Follow ups

* [x] ~decide on `database/sql`, [kyleconroy/sqlc](https://github.com/kyleconroy/sqlc), or [jmoiron/sqlx](https://github.com/jmoiron/sqlx)~ stdlib should be sufficient
* [ ] convert one of the more involved queries to use the new library
* [x] find a replacement for gormigrate - forked and inlined https://github.com/infrahq/infra/pull/2722

dnephin:
Notes from task 1

I looked at these:
* [database/sql#Tx](https://pkg.go.dev/database/sql#Tx) - queries are performed on a `Tx` type, that uses `Exec` for writes, and `Query` for selects
* [jmoiron/sqlx#Tx](https://pkg.go.dev/github.com/jmoiron/sqlx#Tx) - embeds `sql.Tx` so the transaction has roughly the same shape
* [kyleconroy/sqlc](https://github.com/kyleconroy/sqlc) - does not itself implement a `Tx`, but does generate transaction interfaces that are satisfied by the stdlib `Tx`
* [lib/pq](https://pkg.go.dev/github.com/lib/pq) - is used through the stdlib `database/sql`, so does not have its own concept of a transaction
* [jackc/pgx#Tx](https://pkg.go.dev/github.com/jackc/pgx/v4#Tx) - has two modes, one is a driver for `database/sql` (which would use the stdlib `Tx`), the other is its own library that makes use of postgres binary protocol. Its own mode has a `Tx` that is very similar to the stdlib `Tx`.

[...]

PR #2657, opened 2022-07-20
data: early draft of data layer with new interface

dnephin:
## Summary

This is an early draft of our data layer converted to using structs as inputs instead of selector funcs, and using regular sql queries instead of gorm helpers.

By using regular sql queries it should be a much smaller shift away from gorm when we are ready to do that. I attempted to use the stdlib right away, but transaction management is so completely different that it wasn't working out. So I think we need to break the problem down into smaller problems. This is the first pass.

## Related Issues

Related to #2415

### DIFF
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -81,8 +81,77 @@ func SaveAccessKey(db *gorm.DB, key *models.AccessKey) error {
 	return save(db, key)
 }
 
-func ListAccessKeys(db *gorm.DB, p *models.Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
-	return list[models.AccessKey](db, p, selectors...)
+// ListAccessKeysQuery defines the fields used to list access keys. All fields
+// are optional.  If multiple fields have non-zero values the conditions will be
+// combined with AND operators.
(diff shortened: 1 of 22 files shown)


internal/server/data/access_key.go
```diff
@@ -82,29 +81,106 @@ func SaveAccessKey(db *gorm.DB, key *models.AccessKey) error {
...
+	IssuedFor  uid.ID
+	ProviderID uid.ID
 }
 
-func DeleteAccessKeys(db *gorm.DB, selectors ...SelectorFunc) error {
-	toDelete, err := list[models.AccessKey](db, &models.Pagination{}, selectors...)
-	if err != nil {
-		return err
+func DeleteAccessKeys(db *gorm.DB, opt DeleteAccessKeysQuery) error {
+	q := Query(`DELETE FROM access_keys WHERE`)
```
mxyng, line 170: This will do a full delete instead of a soft delete. Is that expected?

dnephin: Oh, right! I guess this should do a soft delete

dnephin:
Most of the changes from this early draft are now part of #3029 and #3094, so going to close this.

PR #2658, opened 2022-07-20
WIP: fix: check grant exists before create

### LINKED ISSUE #2652: Server does not properly propogate duplicate grant error on postgres
## Describe the bug

<!-- A clear and concise description of what the bug is. -->

When a duplicate grant is added, the server responds with 500 Internal Server Error.

```
$ infra grants add admin@email.com docker-desktop --role exec                                                                                                                                                                                                                                                                                                                                                                                                                                                
Created grant to "docker-desktop" for "admin@email.com"                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
$ infra grants add admin@email.com docker-desktop --role exec                                                                                                                                                                                                                                                                                                                                                                                                                                                
Error: internal server error
```

Infra Server logs:
```
{"rows":0,"level":"error","caller":"data/data.go:162","elapsed":8.766984,"error":"ERROR: duplicate key value violates unique constraint \"idx_grant_srp\" (SQLSTATE 23505)","time":1658266217630,"query":"INSERT INTO \"grants\" (\"created_at\",\"updated_at\",\"deleted_at\",\"subject\",\"privilege\",\"resource\",\"created_by\",\"id\") VALUES ($1,$2,$3,$4,$5,$6,$7,$8) RETURNING \"id\""}
{"rows":0,"level":"error","caller":"data/data.go:131","elapsed":1.063427,"error":"ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)","time":1658266217632,"query":"SELECT count(*) FROM \"grants\" WHERE subject = $1 AND privilege = $2 AND resource = $3 AND \"grants\".\"deleted_at\" IS NULL"}
{"error": "has grants: ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)","level": "error","caller": "server/routes.go:158","remoteAddr": "192.168.9.49:56490","time": 1658266217632,"path": "/api/grants","method": "POST","statusCode": 500}
```

## Expected behavior

<!-- A clear and concise description of what you expected to happen. -->

Same result as SQLite

BruceMacD:
- the db transaction is closed when create errors, need to check if grant exists first

## Summary
Postgres was failing with an internal server error on duplicate grant creation. The reason for this is that Postgres closes the DB transaction on error, but we would then attempt to read the existing grant on the same transaction.

Fix is to check if a grant exists before creating it.

Started on writing some Postgres server tests here, but hit some issues with too many open connections. Will re-visit server Postgres tests.

## Checklist

## Related Issues

Resolves #2652

### DIFF
--- a/internal/server/grants.go
+++ b/internal/server/grants.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/gin-gonic/gin"
@@ -9,7 +8,6 @@ import (
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/access"
-	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
 )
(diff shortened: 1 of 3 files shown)


internal/server/grants.go
```diff
@@ -61,29 +59,25 @@ func (a *API) CreateGrant(c *gin.Context, r *api.CreateGrantRequest) (*api.Creat
...
-		}
-
-		if len(grants) == 0 {
-			return nil, fmt.Errorf("duplicate grant exists, but cannot be found")
-		}
+	// check if grant already exists
+	grants, err := access.ListGrants(c, grant.Subject, grant.Resource, grant.Privilege, false, &models.Pagination{})
+	if err != nil {
+		return nil, err
+	}
```
dnephin, line 66: I suspect  this should be in `access.CreateGrant` so that it's after the authorization

ssoroka:
wait, isn't it a problem that postgres closes the transaction on error? that's suspect

internal/server/grants.go
```diff
@@ -61,29 +59,25 @@ func (a *API) CreateGrant(c *gin.Context, r *api.CreateGrantRequest) (*api.Creat
...
 
+	if len(grants) == 1 {
+		// this grant already exists
 		return &api.CreateGrantResponse{Grant: grants[0].ToAPI()}, nil
+	} else if len(grants) > 1 {
+		return nil, fmt.Errorf("multiple duplicate grants exists")
```
ssoroka, line 72: isn't this impossible with the current constraints?

BruceMacD: Yeah it is impossible, but wanted to have a sanity check just in case

ssoroka:
cc from a conversation with Bruce, 
I'd prefer to close this and instead manually control the savepoint to avoid this problem.
eg 
```
db.SavePoint("pre-grant")
if err := data.CreateGrant(db, grant); err != nil {
  db.RollbackTo("pre-grant")
}
```

BruceMacD:
Closing this PR, going to open a separate PR that rolls back the transaction instead.

PR #2659, opened 2022-07-20
Remove last use of go-playground/validator in models

### LINKED ISSUE #2583: Replace remaining uses of go-playground/validator

dnephin:
## Summary

These validations are not of user input, but are safeguards for developers. They don't really benefit from the additional logic and indirection of internal/validate, so we can validate that fields are set by checking they are non-zero before a save or update.

## Related Issues

Resolves #2583

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/aws/aws-sdk-go v1.44.56
 	github.com/cli/browser v1.1.0
 	github.com/gin-gonic/gin v1.8.1
-	github.com/go-playground/validator/v10 v10.11.0
+	github.com/go-playground/validator/v10 v10.11.0 // indirect
 	github.com/goware/urlx v0.3.2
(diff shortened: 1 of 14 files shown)


BruceMacD (approved):
I'd still like data layer validation for field specific requirements (ie: the name is not an empty string) as an extra layer of defense. Not blocking for this PR though.

PR #2660, opened 2022-07-20
fix: wait for logout to complete

### LINKED ISSUE #2526: Logout logs out but fails to redirect to `/login` page

BruceMacD:
- await logout to complete before redirecting to login to prevent race condition

## Summary
Occasionally when you click "logout" the UI would still display the dashboard. This was due to redirecting to login before the logout response was received. Awaiting the logout response ensures we always redirect to login when a user clicks "logout".

## Checklist

## Related Issues

Resolves #2526

### DIFF
--- a/ui/components/layouts/dashboard.js
+++ b/ui/components/layouts/dashboard.js
@@ -19,7 +19,7 @@ function Layout({ children }) {
   }
 
   async function logout() {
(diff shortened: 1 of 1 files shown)


jmorganca (approved):
Thanks for fixing!

hoyyeva (approved):

PR #2661, opened 2022-07-20
feat: password reset

### LINKED ISSUE #2644: ui: `forget password` on login page

ssoroka:
## Summary

adds api endpoints for password reset with email support.

## Checklist

## Related Issues

Resolves https://github.com/infrahq/internal/issues/21
Resolves #2644

### DIFF
--- a/api/passwordreset.go
+++ b/api/passwordreset.go
@@ -1,0 +1,10 @@
+package api
+
+type PasswordResetRequest struct {
+	Email string `json:"email"`
+}
+
+type VerifiedResetPasswordRequest struct {
+	Token    string `json:"token"`
+	Password string `json:"password"`
+}
--- a/api/passwordreset.go
+++ b/api/passwordreset.go
@@ -1,10 +1,28 @@
 package api
 
+import "github.com/infrahq/infra/internal/validate"
+
 type PasswordResetRequest struct {
 	Email string `json:"email"`
+}
+
+func (r PasswordResetRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("email", r.Email),
+		validate.Email("email", r.Email),
+	}
 }
 
 type VerifiedResetPasswordRequest struct {
 	Token    string `json:"token"`
 	Password string `json:"password"`
+}
+
+func (r VerifiedResetPasswordRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("token", r.Token),
+		validate.String("token", r.Token, 10, 10, validate.AlphaNumeric...),
+		validate.Required("password", r.Password),
+		// todo: validate password meets org's criteria
+	}
 }
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -47,7 +47,7 @@ func CreateCredential(c *gin.Context, user models.Identity) (string, error) {
 }
 
 func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string) error {
-	db, err := hasAuthorization(c, user.ID, isIdentitySelf, models.InfraAdminRole)
+	_, err := hasAuthorization(c, user.ID, isIdentitySelf, models.InfraAdminRole)
 	if err != nil {
 		return HandleAuthErr(err, "user", "update", models.InfraAdminRole)
 	}
@@ -57,7 +57,13 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
 		return err
 	}
 
-	err = checkPasswordRequirements(db, newPassword)
+	return updateCredential(c, user, newPassword, isSelf)
+}
+
+func updateCredential(c *gin.Context, user *models.Identity, newPassword string, isSelf bool) error {
+	db := getDB(c)
+
+	err := checkPasswordRequirements(db, newPassword)
 	if err != nil {
 		return err
 	}
@@ -93,10 +99,11 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
 		// if we updated our own password, remove the password-reset scope from our access key.
 		if raw, ok := c.Get(RequestContextKey); ok {
 			if rCtx, ok := raw.(RequestContext); ok {
-				accessKey := rCtx.Authenticated.AccessKey
-				accessKey.Scopes = models.CommaSeparatedStrings{}
-				if err = data.SaveAccessKey(db, accessKey); err != nil {
-					return fmt.Errorf("updating access key: %w", err)
+				if accessKey := rCtx.Authenticated.AccessKey; accessKey != nil {
+					accessKey.Scopes = models.CommaSeparatedStrings{}
+					if err = data.SaveAccessKey(db, accessKey); err != nil {
+						return fmt.Errorf("updating access key: %w", err)
+					}
 				}
 			}
 		}
--- a/internal/access/passwordreset.go
+++ b/internal/access/passwordreset.go
@@ -1,0 +1,51 @@
+package access
+
+import (
+	"github.com/gin-gonic/gin"
+
+	"github.com/infrahq/infra/internal"
+	"github.com/infrahq/infra/internal/server/data"
+	"github.com/infrahq/infra/internal/server/models"
+)
+
+func PasswordResetRequest(c *gin.Context, email string) (token string, err error) {
+	// no auth required
+	db := getDB(c)
+
+	users, err := data.ListIdentities(db, &models.Pagination{Limit: 1}, data.ByName(email))
+	if err != nil {
+		return "", err
+	}
+
+	if len(users) != 1 {
+		return "", internal.ErrNotFound
+	}
+
+	prt, err := data.CreatePasswordResetToken(db, &users[0])
+	if err != nil {
+		return "", err
+	}
+
+	return prt.Token, nil
+}
+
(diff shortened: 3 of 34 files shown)


internal/access/credential.go
```diff
@@ -56,7 +56,13 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
 		return err
 	}
 
-	err = checkPasswordRequirements(db, newPassword)
+	return updateCredential(c, user, newPassword, isSelf)
+}
+
+func updateCredential(c *gin.Context, user *models.Identity, newPassword string, isSelf bool) error {
```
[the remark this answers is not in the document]
ssoroka: I had that but it seemed redundant; We don't do anything special to the db objects. the one thing about that pattern is it's ensuring something: you don't get a db object until you've checked authorization, however this function doesn't care about authorization, as it's been checked (or not) by the caller. Since the db is in the ctx it seemed odd to pass both.

internal/access/passwordreset.go
```diff
@@ -0,0 +1,51 @@
...
+	users, err := data.ListIdentities(db, &models.Pagination{Limit: 1}, data.ByName(email))
+	if err != nil {
+		return "", err
+	}
+
+	if len(users) != 1 {
+		return "", internal.ErrNotFound
+	}
+
+	prt, err := data.CreatePasswordResetToken(db, &users[0])
```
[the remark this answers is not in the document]
ssoroka: hmm. Token is a pretty generic term that I think will keep coming up.

internal/server/passwordreset_test.go
```diff
@@ -0,0 +1,83 @@
...
+	err := data.CreateIdentity(s.db, user)
+	assert.NilError(t, err)
+
+	err = data.CreateCredential(s.db, &models.Credential{
+		IdentityID:   user.ID,
+		PasswordHash: []byte("foo"),
+	})
+	assert.NilError(t, err)
+
+	// request password reset
```
dnephin, line 38: There's a `runStep` helper in this package that might be useful for these multi-step tests. It uses `t.Run`, but fails the parent test if any of the subtests fail.  It can be nice for multi-steps tests because the extra step name will be in the failure output, which gives the developer a little more information about which step failed.

internal/server/email/sendgrid.go
```diff
@@ -0,0 +1,84 @@
...
+
+var emailTemplateIDs = map[EmailTemplate]string{
+	EmailTemplateAccountCreated: "",
+	EmailTemplatePasswordReset:  "d-d87873d2f11b4055befbd7064cda44d6", // transactional-password-reset
+	EmailTemplateUserInvite:     "d-ac6bcb2a7f02463c8b7fc8caffc35f2d", // transactional-user-invite // TODO: Should this instead be a "granted access to x" email?
+}
+
+var (
+	AppDomain          = "https://infrahq.com"
+	FromAddress        = "noreply@infrahq.com"
```
dnephin, line 29: I guess both of these will need to be part of `server.Options` to support on-prem.

internal/server/email/sendgrid.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func SendTemplate(name, address string, template EmailTemplate, data map[string]interface{}) error {
+	if TestMode {
+		logging.Debugf("sent email to %q: %+v\n", address, data)
+		TestDataSent = append(TestDataSent, data)
+		return nil // quietly return
+	}
+
+	key := os.Getenv("SENDGRID_API_KEY")
```
dnephin, line 49: It'd be nice for this to come from `server.Options` as well, even if we read it from this environment variable. That way all config comes from the same place, and it should make it a bit easier to test the email package without having to set environment variables.

api/passwordreset.go
```diff
@@ -0,0 +1,10 @@
+package api
+
+type PasswordResetRequest struct {
+	Email string `json:"email"`
+}
+
+type VerifiedResetPasswordRequest struct {
+	Token    string `json:"token"`
+	Password string `json:"password"`
```
dnephin, line 9: Should all of these be required fields (`validate.Required`) ?  
Should `email` use `validate.Email` ?

Any min/max length requirements for token and password?

internal/server/routes.go
```diff
@@ -107,6 +107,8 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	post(a, noAuthn, "/api/signup", a.Signup)
 
 	post(a, noAuthn, "/api/login", a.Login)
+	post(a, noAuthn, "/api/password-reset-request", a.RequestPasswordReset)
+	post(a, noAuthn, "/api/password-reset", a.VerifiedPasswordReset)
```
dnephin, line 111: Is `/api/password-reset` a link from an email? Doesn't that have to be a `GET`? Or do email clients support `<form>` which allows us to make this a POST?

dnephin (commented):
Pretty cool!

The email template is stored by sendgrid, is that right?  Any way we can manage it in this git repo? How would we go about updating it, and how can we preview it?

I imagine using a provider like this makes it easier for us to ensure our email ends up in their inbox (and not in spam).  I know there are lots of options for sending email. It'd be great to document somewhere whey we chose sendgrid over other options.

internal/server/email/sendgrid.go
```diff
@@ -0,0 +1,84 @@
...
+	FromAddress        = "noreply@infrahq.com"
+	FromName           = "Infra"
+	TestMode           = false
+	TestDataSent       = []map[string]interface{}{}
+	ErrUnknownTemplate = errors.New("unknown template")
+	ErrNotConfigured   = errors.New("email sending not configured")
+)
+
+func IsConfigured() bool {
+	key := os.Getenv("SENDGRID_API_KEY")
```
mxyng, line 38: This should be a server option and also support loading from file

internal/server/models/passwordresettoken.go
```diff
@@ -0,0 +1,17 @@
+package models
+
+import (
+	"time"
+
+	"github.com/infrahq/infra/uid"
+)
+
+type PasswordResetToken struct {
```
mxyng, line 9: To be clear, this _isn't_ a database model correct?

ssoroka:
> The email template is stored by sendgrid, is that right? Any way we can manage it in this git repo? How would we go about updating it, and how can we preview it?

Storing this in the repo is better for portability, but worse for managing this ourselves. It's harder to test email templates as delivered emails when you're rendering them yourself. I'd rather move forward with this and deal with that added complexity later.

internal/server/email/sendgrid.go
ssoroka: We don't really want the test package hitting the sendgrid api anyway.

api/passwordreset.go
ssoroka: sounds good. Password is going to be more complex as it depends on the org's requirements. It'll be the first validator that requires db access to look up the requirements.

internal/server/routes.go
ssoroka: no. /password-reset is the link from email. /api/password-reset is the api call that page makes.

internal/server/models/passwordresettoken.go
ssoroka: it is. why?

api/passwordreset.go
dnephin: Oh right! Our password validation happens at a different layer. I don't think we need to mix those things. I think it's fine if this does static validation, and we do more involved validation later on once we have access to the DB.

BruceMacD (approved):

ui/pages/password-reset/index.js
```diff
@@ -0,0 +1,185 @@
...
+      })
+
+      if (!res.ok) {
+        throw await res.json()
+      }
+
+      await res.json()
+      setSubmitted(true)
+    } catch (e) {
+      console.error(e)
```
hoyyeva, line 62: do we want to put the the error `e` to setError? so it can be displayed on the UI?

ui/pages/password-reset/index.js
```diff
@@ -0,0 +1,185 @@
...
+                {error}
+              </p>
+            )}
+          </form>
+        </>
+      ) : (
+        <>
+          <h1 className='text-base font-bold leading-snug'>Password Reset</h1>
+          {submitted ? (
+            <p>Please check your email for the reset link</p>
```
hoyyeva, line 126: hmm can we give some styling for the text here? maybe something like ` namespace='my-3 max-w-[260px] text-center text-xs text-gray-300'`

ui/pages/password-reset/index.js
```diff
@@ -0,0 +1,185 @@
...
+    } catch (e) {
+      console.error(e)
+    }
+
+    return false
+  }
+
+  return (
+    <>
+      {token ? (
```
hoyyeva, line 70: overall i think this looks good! to improve this a bit is maybe we can refactor this a bit because I notice many duplicate code. 

so what we can do is to make line 71 to line 123 to a component - `password-reset-form` and pass in props like `{title, value, onSubmit, label, placeholder, onChange, buttonTitle, error }`
And then you can reuse the same component to replace line 71 to 123 and line 129 to 176

ssoroka: Hmm. these forms aren't the same. Not sure where the reuse is.

hoyyeva: oh?O.o I thought they are similar that they both have a form with a label, input and a button.

PR #2662, opened 2022-07-20
maintain: paginate in api by default

j-sneh:
## Summary

List requests in the API will be paginated by default (Page 1 and Limit 100 if not specified). If users are writing against the API directly, then they must switch through pages themselves to get all results.

Internal ListRequests can pass in `nil` to avoid pagination. Pre-existing calls now pass in `nil` instead of `&models.Pagination{}`.

## Checklist

### DIFF
--- a/internal/access/access.go
+++ b/internal/access/access.go
@@ -9,6 +9,7 @@ import (
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal/server/data"
+	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
 )
 
@@ -127,7 +128,7 @@ func HandleAuthErr(err error, resource, operation string, roles ...string) error
 
 // Can checks if an identity has a privilege that means it can perform an action on a resource
 func Can(db *gorm.DB, identity uid.PolymorphicID, privilege, resource string) (bool, error) {
-	grants, err := data.ListGrants(db, nil, data.BySubject(identity), data.ByPrivilege(privilege), data.ByResource(resource))
+	grants, err := data.ListGrants(db, &models.Pagination{Limit: 1}, data.BySubject(identity), data.ByPrivilege(privilege), data.ByResource(resource))
 	if err != nil {
 		return false, fmt.Errorf("has grants: %w", err)
 	}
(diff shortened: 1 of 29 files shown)


internal/access/access.go
```diff
@@ -128,7 +127,7 @@ func HandleAuthErr(err error, resource, operation string, roles ...string) error
 
 // Can checks if an identity has a privilege that means it can perform an action on a resource
 func Can(db *gorm.DB, identity uid.PolymorphicID, privilege, resource string) (bool, error) {
-	grants, err := data.ListGrants(db, &models.Pagination{}, data.BySubject(identity), data.ByPrivilege(privilege), data.ByResource(resource))
+	grants, err := data.ListGrants(db, nil, data.BySubject(identity), data.ByPrivilege(privilege), data.ByResource(resource))
```
mxyng, line 130: Can probably set `Limit: 1` here since the subject, privilege, resourece 3-tuple is unique

internal/access/grant.go
```diff
@@ -72,7 +72,7 @@ func ListGrants(c *gin.Context, subject uid.PolymorphicID, resource string, priv
 }
 
 func userInGroup(db *gorm.DB, authnUserID uid.ID, groupID uid.ID) bool {
-	groups, err := data.ListGroups(db, &models.Pagination{}, data.ByGroupMember(authnUserID), data.ByID(groupID))
+	groups, err := data.ListGroups(db, nil, data.ByGroupMember(authnUserID), data.ByID(groupID))
```
mxyng, line 75: Can set `Limit: 1` since `groupID` is unique

internal/server/middleware.go
```diff
@@ -53,7 +53,7 @@ func DestinationMiddleware() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		uniqueID := c.GetHeader("Infra-Destination")
 		if uniqueID != "" {
-			destinations, err := access.ListDestinations(c, uniqueID, "", &models.Pagination{})
```
mxyng, line 56: Can set `Limit: 1` since the `uniqueID` is unique

dnephin (approved):
Looks good!

We should also document that list API responses will only include 100 items by default and that users will need to set a limit or make multiple requests for pages. I think a new section in  `docs/api/making-requests.md` is probably a good place for that. Could be a follow up PR.

Issue #1783, opened 2022-04-29
Return per-field errors in API


hoyyeva:
<img width="464" alt="Untitled" src="https://user-images.githubusercontent.com/63033505/165868419-62f1c5df-4da1-475f-b5af-a778b2a83903.png">

hoyyeva:
this might need to fix from the server side. The client is displaying the error message that is coming from the server side.

pdevine:
Depends on #1686

pdevine:
Also #1763

kimskimchi:
Will sync with @dnephin & write a design doc, then once changes are in, consume messages from UI/CLI (sync with @hoyyeva needed here).

kimskimchi:
Validation errors are now returned. 400s will be returned with ValidationError, which will have a map of field names and its errors. 

See example in https://github.com/infrahq/infra/pull/2651, `users.go`

PR #2665, opened 2022-07-21
api: use field errors for unique constraint and provider errors

dnephin:
## Summary

This PR does a bit of cleanup around api errors, and uses the new structured `validate.Error` in two more places. As usual, best viewed by individual commit.

1. remove unused fields from `api.Error`.
2. removes some unused errors from `cmd`
3. use `internal/validate.Error` to provide structured errors for "unique constraint violation" errors (#1783). Also make the wording more human readable.
4. use `internal/validate.Error` to provide structured errors for provider validation

## Related Issues

Implements the API side of #1783


internal/server/providers/oidc.go
```diff
@@ -83,29 +84,35 @@ func (o *oidcClientImplementation) Validate(ctx context.Context) error {
...
-				return ErrInvalidProviderClientSecret
+				return newValidationError("clientSecret")
 			}
 		}
 		logging.Debugf("%s", err.Error())
 	}
 
+	// return nil for all other errors, because the request was made with an
+	// invalid code, which will always fail
 	return nil
```
dnephin, line 109: This confused me again for a second, until I noticed the comment above. So I moved the comment to the `return nil`.

internal/server/data/data.go
```diff
@@ -169,12 +169,22 @@ type UniqueConstraintError struct {
...
+		table = "user"
+	case "access_keys":
+		table = "access key"
+	default:
+		table = strings.TrimSuffix(table, "s")
 	}
-	return fmt.Sprintf("value for %v already exists for %v", e.Column, e.Table)
+
+	if e.Column == "" {
+		return fmt.Sprintf("a %v with that value already exists", table)
```
mxyng, line 185: Maybe `UniqueConstraintError` can have a default Column value of `value` so as to not need this branch?

mxyng (approved):

internal/server/data/data.go
dnephin: Ya, that could work. Right now we create the type in two places, so we'd have to set that default in both places.

I think my preference would be to keep this branch so that we don't have to remember to set a default in every place we create the error, and so that we don't need a constructor to set it.

BruceMacD (approved):

PR #2670, opened 2022-07-21
fix: change field name

hoyyeva:
## Summary
api removed the `email` field (https://github.com/infrahq/infra/pull/2645/files) - this PR modify the UI accordingly


ui/pages/signup/index.js
```diff
@@ -93,15 +93,15 @@ export default function Signup() {
             type='email'
```
jmorganca, line 93: `name=`name`` ?

jmorganca (approved):
Looks good. I would also edit the input label to be `name` for consistency

hoyyeva:
> Looks good. I would also edit the input label to be `name` for consistency

should we name it `username or email` to be consistent with login page?

PR #2671, opened 2022-07-21
fix(ui): signup uses new parameters; fix issue where grant would be deleted incorrectly in the access dropdown

jmorganca:


BruceMacD (approved):
lgtm

hoyyeva (approved):

PR #2676, opened 2022-07-21
fix: fixes isLoggedIn check to include expiry

### LINKED ISSUE #2420: Failing login shows "session active" message, when it isn't
## Describe the bug

1. Let your session expire. 

2. run `./infra login`
fail login to see the following:
`...Your session as admin to localhost:8443 is still active.`

but they are not logged in. Running `./infra info` shows the correct message:
`...Session expired; run .... to start a new session`

## Environment

```
$ infra version
```

```
$ kubectl version
```

kimskimchi:
## Checklist

## Related Issues

Resolves #2420

### DIFF
--- a/internal/cmd/logout_test.go
+++ b/internal/cmd/logout_test.go
@@ -7,8 +7,10 @@ import (
 	"path/filepath"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/infrahq/infra/api"
 	"gotest.tools/v3/assert"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
@@ -55,6 +57,7 @@ func TestLogout(t *testing.T) {
 					UserID:        1,
 					SkipTLSVerify: true,
 					Current:       true,
+					Expires:       api.Time(time.Now().Add(time.Hour * 2).UTC().Truncate(time.Second)),
 				},
 				{
 					Name:          "user2",
@@ -62,6 +65,7 @@ func TestLogout(t *testing.T) {
 					AccessKey:     "the-access-key",
 					UserID:        2,
 					SkipTLSVerify: true,
+					Expires:       api.Time(time.Now().Add(time.Hour * 2).UTC().Truncate(time.Second)),
 				},
 			},
 		}
@@ -117,6 +121,7 @@ func TestLogout(t *testing.T) {
 					UserID:        1,
 					SkipTLSVerify: true,
 					Current:       true,
+					Expires:       api.Time(time.Now().Add(time.Hour * 2).UTC().Truncate(time.Second)),
 				},
 			},
 		}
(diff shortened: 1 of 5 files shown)


internal/cmd/logout_test.go
```diff
@@ -7,8 +7,10 @@ import (
 	"path/filepath"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/infrahq/infra/api"
```
mxyng, line 13: First party imports should be in its own section after the 3rd party imports

kimskimchi: fixed!

internal/cmd/logout_test.go
```diff
@@ -316,4 +331,12 @@ func TestLogout(t *testing.T) {
 		assert.ErrorContains(t, err, `"infra logout" accepts at most 1 argument`)
 		assert.ErrorContains(t, err, `Usage:  infra logout [SERVER]`)
 	})
+
+	t.Run("expired session fails", func(t *testing.T) {
```
mxyng, line 335: this should check if the expired config is removed

internal/cmd/logout_test.go
```diff
@@ -117,6 +122,16 @@ func TestLogout(t *testing.T) {
...
 					Current:       true,
+					Expires:       api.Time(time.Now().Add(time.Hour * 2).UTC().Truncate(time.Second)),
+				},
+				{
+					Name:          "user2",
+					Host:          srv.Listener.Addr().String(),
+					AccessKey:     "the-access-key",
+					UserID:        2,
+					SkipTLSVerify: true,
+					Current:       true,
```
dnephin, line 133: I don't think logout calls `mustBeLoggedIn`, so the test case won't cover the change you are making.

dnephin (commented):
Fix looks good, but I'm pretty sure the test is not covering this case, because we should expect an error with `Session expired`, but the test case is asserting `NilError`.

I think we can do something like this to test it, maybe in a new test case `TestMustBeLoggedIn`:

```go
cfg := ClientConfig{
    ClientConfigVersion: clientConfigVersion,
    Hosts: []ClientHostConfig{
        // a ClientHostConfig like the "user2" case you added 
     }
}

err := writeConfig(&cfg)
assert.NilError(t, err)

err := Run(context.Background(), "destinations", "list")
assert.ErrorContains(t, err, "Session expired")
```

internal/cmd/logout_test.go
kimskimchi: logout calls mustbeLoggedIn, and only prints out a success message if logged in, and nothing if logged out (or expired). 

The test case `expired session fails` has two sessions - one is logged in but api call fails (so api count +1), the other is expired so it won't even call the api (hence api count stays as 1).

internal/cmd/logout_test.go
kimskimchi: done (at least I think that is what you meant for me to write up, the equateEmpty() parts).

dnephin (approved):
Looks good!

PR #2677, opened 2022-07-21
fix: remove group name validation

dnephin:
## Summary

Group names can come from providers, which may allow any characters. So we need to support all characters group names unless we do some kind of internal mapping between names. For now we'll need to remove our validation of group name.

Also improve the error message when the invalid character is a space


BruceMacD (approved):
looks good

PR #2679, opened 2022-07-21
improve: add validate name to provider name

dnephin:
## Summary

To be consistent with most other names.


mxyng (approved):

dnephin:
Pushed another commit which replaces some `validate` tags on the provider that slipped in while I was merging the validation changes. Also added test coverage of those validations.

BruceMacD (approved):
Updates look good 👍

Issue #2584, opened 2022-07-14
Investigate options for replacing sqlite with postgres


dnephin:
## Background

Today the Infra server supports both sqlite and postgres as database engines. Most of that compatibility is provided by Gorm, but some of it (`handleError`, and any queries we have to build ourselves, ex: #2578) is handled in the `data` package.

We'd like to only support postgres in production for a number of reasons (reduce our maintenance burden, we don't have any way to support multiple server replicas with sqlite). We'd also like to replace Gorm (#2415), which may slightly increase the maintenance cost of multiple databases.

## Problem

Today we run all `data` tests against both databases, but we run all other tests against only sqlite. We can see from the test ouput ([log](https://github.com/infrahq/infra/runs/7343470624?check_suite_focus=true), look for `/postgres` and `/sqlite` ) that running tests against postgres is 5x slower in CI with `-race` enabled (on my local setup its 30x slower).

Running all of our tests against only postgres would significantly slow down our test suite, which would decrease productivity.

It seems like we have three options:
1. Continue to pay the small maintenance cost of making queries work with sqlite, so that we have a convenient way of testing other layers of our application, and so that we have a convenient workflow for trying out the rest of Infra
2. Completely remove sqlite support, and find some other way to provide a faster data layer for testing other parts of the system (API, CLI, core business logic). For example, a full in-memory fake implementation.
3. Spent time investigating why the postgres version is so much slower, and find ways to optimize it. My guess is that it may have something to do with gorm and/or migrations.

It's also important to note that today we can't easily test all our of DB queries in the data layer because of the `SelectorFunc` pattern. We plan on replacing selector funcs with a struct of input values, so that all query construction happens in a single function in `data`. That change should make it much easier to test all queries against postgres.

dnephin:
Given these trade-offs, my preference would be to keep sqlite for dev and demo purposes. We may have to write a second very of some queries, but we don't really need to care much about the performance of those queries.

If we compare that to the alternatives, I think it's likely less work to write a similar sqlite compatible query (for the few rare cases where we can't use a single query for both), than to try and create a completely separate implementation of our data layer.

dnephin:
[...]

Issue #2564, opened 2022-07-12
DB migrations tests fail if additional fields are added to the base model


pdevine:
## Describe the bug

In `data/migrations.go` we call `initializeSchema()` which calls `db.Automigrate()` with each of the models. This works fine with our existing model because none of the older migrations need to call `ALTER TABLE` on newer tables (such as `provider_user`). If you add another field to the base model, it will end up trying to alter every table with the new field, which fails because not every table exists in our older migrations.

The failure (in `make test`) will look something like:
```
--- FAIL: TestMigration_202204111503 (0.01s)
    writer.go:25: {"level":"info","time":1657653584313,"caller":"logging/logger.go:76","message":"running migration 202204111503"}
    writer.go:25: {"level":"error","rows":0,"query":"ALTER TABLE `provider_users` ADD `organization_id` integer","elapsed":0.005542,"error":"no such table: provider_users","time":1657653584319,"caller":"migrator/migrator.go:328"}
    migrations_test.go:29: assertion failed: error is not nil: migration failed: no such table: provider_users
--- FAIL: TestMigration_202204211705 (0.01s)
    writer.go:25: {"level":"error","rows":0,"query":"ALTER TABLE `provider_users` ADD `organization_id` integer","elapsed":0.005125,"error":"no such table: provider_users","time":1657653584325,"caller":"migrator/migrator.go:328"}
    migrations_test.go:44: assertion failed: error is not nil: migration failed: no such table: provider_users
```

## Expected behavior

We need to account for the proper state of the database when we're performing the migrations.

## Additional context

You can add any field to `models.Model` to replicate this.

ssoroka:
don't quite understand this. initializeSchema is only called in two scenarios:
1. after the migrations to support adding new fields that don't require migrations.
2. on a fresh database that has never been initialized.

If I understand this, you're having problems with #2? a fresh database that creates these fields and then tries to apply the migration afterwards? Migrations should probably have guards that check if anything needs to happen for this case.

pdevine:
It's pretty easy to replicate. Just add another field to `models.Model` and run `make test`. I'm assuming I was running into scenario #2 as you mentioned.

pdevine:
I'm going to close this. What ended up happening was the `Provider User` table didn't end up having the base model and I was including it in the tables to migrate which caused things to blow up.

Issue #2066, opened 2022-05-25
data: why do migrations require calling initializeSchema multiple times?


dnephin:
This issue is to track a couple TODOs in the `data` package that I left while working on #1989 and #1990.

~In `migrations_test.go` I'd like to be able to call `NewDB` and run the migrations as if someone was re-starting a server after an upgrade.~
Edit: resolved this, see https://github.com/infrahq/infra/issues/2066#issuecomment-1158187070

That might be related to the second todo

https://github.com/infrahq/infra/blob/4a8b418d66be8cca2b80140c44d2f95a1a22ae07/internal/server/data/migrations.go#L469-L470

In `NewDB` we already call this `preMigrate` function: https://github.com/infrahq/infra/blob/4a8b418d66be8cca2b80140c44d2f95a1a22ae07/internal/server/data/data.go#L27-L33

So why does `migrate` have to call it again later? Maybe it's necessary? Maybe gorm migrate changes behaviour by simply setting an `InitSchema` function? If we can't remove it, we should comment about why it's necessary.

We then call `initializeSchema` unconditionally at the end (`preMigrate` only calls it if the table doesn't exist).

https://github.com/infrahq/infra/blob/4a8b418d66be8cca2b80140c44d2f95a1a22ae07/internal/server/data/migrations.go#L476-L479

I think I understand why we have to do this at the end, to handle any alter tables to add new columns or constraints. But I'm not sure why we need the other two `preMigrate` if we have this at the end.

dnephin:
Ok, I think I've figured at least the first TODO in #2314.

The tests were not working as I expected because the sqlite db was using `file::memory:`. I changed the migration tests to use a file on disk and now they work as expected. I was able to resolve that TODO, and use `NewDB` to simulate a real server starting.

PR #2685, opened 2022-07-21
demo: use sql for initSchema

dnephin:
## Summary

This is mostly exploration work. With these changes, test run against postgres in comparable time to sqlite. It turns out essentially all of the extra time was due to `gorm.DB.AutoMigrate`, which we were running twice for every call to `NewDB`.

This PR replaces initializeSchema with a load of an sql file (one for each supported DB). It replaces the "AutoMigrate every table twice" with explicit migrations. It was still easier to use `AutoMigrate` in those migrations, but now at least they only run when we're testing specific migrations.

Essentially every call to `AutoMigrate` adds ~150ms, and since we were calling it twice it added ~300ms.

See https://github.com/infrahq/infra/issues/2584#issuecomment-1189520277 for more details of how I found this.

Also related to #2066, and possibly #2564


ssoroka:
I think if you're going to do something like this you need to update the sql schema file every time a migration runs and commit it to the codebase.

dnephin:
Ya, for sure! `gormigrate` already mostly does that for us, I think we just need to adjust how we use it. I've started that in #2722, and #2723.

Instead of using `go:embed` for the entire schema, I think it's probably better to have each table define its' schema. 

That, in combination with a table test for migrations, should allow us to verify that the static schema matches the migrations, and that we aren't missing any migrations. It should also remove the need to dump the sql schema from postgres.

Going to close this now, since we have a better way to do this.

PR #2686, opened 2022-07-22
fix: groups and users remove improvement

### LINKED ISSUE #2647: ui: removing admin, user and group improvement
- [x] 1. User should be able to revoke themselves as an infra admin on the setting. Also if the user has the permission, should be able to revoke any groups (including the groups that user is belonging to) - this behaviour is the same as CLI. The UI should provide a conformation modal to warn whenever and admin tries to delete themselves or a group they belong to

- [x] 2. User should be able to remove themselves from a group but there should be a confirmation modal to warn user that this action might effect them directly and immediately (ie they might lose the permission that this group has) 

- [x] 3. Should prevent user to revoke the user / group if it is the only one has infra admin permission

- [x] 4. Should not be able to remove the group if it is the only one has infra admin permission

hoyyeva:
## Summary
This PR removes restrictions to remove users and groups in UI - anyone who has permission should be able to remove users (including itself) and groups in UI. However UI adds confirmation modals in different cases to ensure that user will not accidentally and unintentionally revoke infra permission that they have or a group that they are belonging to  

## Related Issues
Resolves #2647


ui/components/identity-item.js
```diff
@@ -0,0 +1,50 @@
+import { useState } from 'react'
+
+import DeleteModal from './delete-modal'
+
+export default function IdentityItem({
+  item,
```
jmorganca, line 6: instead of item we could name this (and elsewhere) userOrGroup

ui/components/identity-item.js
```diff
@@ -0,0 +1,50 @@
...
+  deleteModalInfo,
```
jmorganca, line 7: Two parameters would be better here to reduce coupling

ui/pages/groups/index.js
```diff
@@ -183,20 +208,22 @@ function Details({ group, admin, onDelete }) {
...
-                  list={users}
-                  onClick={async userId => {
-                    const usersToRemove = [userId]
-                    await fetch(`/api/groups/${id}/users`, {
-                      method: 'PATCH',
-                      body: JSON.stringify({ usersToRemove }),
-                    })
+                {users
+                  .map(user => {
+                    const showDeleteModal = user.id === auth.id
```
jmorganca, line 213: Would fold these two map functions into one

ui/pages/groups/index.js
```diff
@@ -96,26 +96,52 @@ function EmailsSelectInput({
...
+  const adminGroups = infraAdmin?.map(admin => admin.group)
+
   const metadata = [
     { title: 'ID', data: id },
     {
       title: 'Created',
       data: group?.created ? dayjs(group.created).fromNow() : '-',
     },
   ]
+  const deleteModalInfo = {
```
jmorganca, line 123: Should we move this down to the code it is used in below?

ui/pages/settings/index.js
```diff
@@ -12,9 +12,19 @@ import DeleteModal from '../../components/delete-modal'
 import Notification from '../../components/notification'
 import GrantForm from '../../components/grant-form'
 
-function AdminGrant({ name, showRemove, onRemove }) {
+function AdminGrant({ name, showRemove, onRemove, message = undefined }) {
   const [open, setOpen] = useState(false)
 
+  const deleteModalMessage =
+    message === undefined ? (
```
jmorganca, line 19: Could check !message here - I believe empty string will be skipped as well

ui/pages/settings/index.js
```diff
@@ -60,11 +65,22 @@ export default function Settings() {
...
+  const { data: { items: selfGroups } = {} } = useSWR(
+    `/api/groups?userID=${auth?.id}`
+  )
 
   const hasInfraProvider = auth?.providerNames.includes('infra')
+  const listSelfGroupsId = selfGroups?.map(g => g.id)
 
   const loading = [auth, users, groups, grants].some(x => !x)
 
+  const deleteModalMessage = {
```
jmorganca, line 77: Similarly would move this lower to where it is used - we may not even need an object with keys

ui/components/identity-item.js
hoyyeva: which two parameters? O.o

ui/pages/users/index.js
```diff
@@ -69,12 +69,16 @@ function Details({ user, admin, onDelete }) {
     () => (groups ? groups.map(g => `/api/grants?group=${g.id}`) : null),
     (...urls) => Promise.all(urls.map(url => fetch(url).then(r => r.json())))
   )
+  const { data: { items: infraAdmin } = {} } = useSWR(
```
jmorganca, line 72:
```suggestion
  const { data: { items: infraAdmins } = {} } = useSWR(
```

ui/pages/users/index.js
```diff
@@ -69,12 +69,16 @@ function Details({ user, admin, onDelete }) {
...
+    '/api/grants?resource=infra&privilege=admin'
+  )
 
   const grants = items?.filter(g => g.resource !== 'infra')
   const inherited = groupGrantDatas
     ?.map(g => g?.items || [])
     ?.flat()
     ?.filter(g => g.resource !== 'infra')
+  const adminGroups = infraAdmin?.map(admin => admin.group)
```
jmorganca, line 81:
```suggestion
  const adminGroups = infraAdmins?.map(admin => admin.group)
```

ui/pages/settings/index.js
```diff
@@ -12,7 +12,7 @@ import DeleteModal from '../../components/delete-modal'
 import Notification from '../../components/notification'
 import GrantForm from '../../components/grant-form'
 
-function AdminGrant({ name, showRemove, onRemove }) {
+function AdminGrant({ name, showRemove, onRemove, message = undefined }) {
```
jmorganca, line 15:
```suggestion
function AdminGrant({ name, showRemove, onRemove, message = '' }) {
```

jmorganca (approved):
Nice work!

PR #2692, opened 2022-07-22
feat: make infra server read-only at runtime

### LINKED ISSUE #2689: Read-only mode
To follow best practices for pod/container security we should have a way to run the Infra server so that it does not write anything to the filesystem.

- In-memory TLS caches with a Postgres database.

BruceMacD:
## Summary
Best practices for Kubernetes security suggest setting `readOnlyRootFilesystem` on pods to ensure they are always in the expected state. This change enables this flag by default and removes filesystem writes from the server runtime.

## Checklist

## Related Issues

Resolves #2689

### DIFF
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -120,7 +120,9 @@ server:
   #   fsGroup: 2000
 
   ## Security context for the server container
-  securityContext: {}
+  securityContext: {
+    readOnlyRootFilesystem: true
+  }
   #   runAsUser: 999
   #   runAsGroup: 999
   #   readOnlyRootFilesystem: true
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -120,8 +120,9 @@ server:
   #   fsGroup: 2000
 
   ## Security context for the server container
-  securityContext: 
+  securityContext:
     readOnlyRootFilesystem: true
+    # runAsUser: 1000
 
   ## Liveness probe for the default backend
   ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
@@ -613,8 +614,9 @@ connector:
   #   fsGroup: 2000
 
   ## Security context for the connector container
-  securityContext: 
+  securityContext:
     readOnlyRootFilesystem: true
+    # runAsUser: 1000
 
   ## Liveness probe for the default backend
   ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -122,7 +122,7 @@ server:
   ## Security context for the server container
   securityContext:
     readOnlyRootFilesystem: true
-    # runAsUser: 1000
+    # runAsUser: 999
 
   ## Liveness probe for the default backend
   ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
@@ -616,7 +616,7 @@ connector:
   ## Security context for the connector container
   securityContext:
     readOnlyRootFilesystem: true
-    # runAsUser: 1000
+    # runAsUser: 999
 
   ## Liveness probe for the default backend
   ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -120,12 +120,8 @@ server:
   #   fsGroup: 2000
 
   ## Security context for the server container
-  securityContext: {
+  securityContext: 
     readOnlyRootFilesystem: true
-  }
-  #   runAsUser: 999
-  #   runAsGroup: 999
-  #   readOnlyRootFilesystem: true
 
   ## Liveness probe for the default backend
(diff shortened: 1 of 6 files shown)


helm/charts/infra/values.yaml
```diff
@@ -120,7 +120,9 @@ server:
   #   fsGroup: 2000
 
   ## Security context for the server container
-  securityContext: {}
+  securityContext: {
+    readOnlyRootFilesystem: true
```
BruceMacD, line 124: Enabling `readOnlyRootFilesystem` by default. This will help us not break "read-only" in the future accidentally.

internal/cmd/config.go
```diff
@@ -57,6 +57,15 @@ func infraHomeDir() (string, error) {
 
 	infraDir := filepath.Join(homeDir, ".infra")
 
+	return infraDir, nil
+}
+
+func initInfraHomeDir() (string, error) {
```
BruceMacD, line 63: Renamed this to reflect what it actually does

internal/server/tls.go
```diff
@@ -17,20 +16,51 @@ import (
...
+func (m MapCache) getKeyPair(ctx context.Context, serverName string) (cert, key []byte) {
+	certBytes, _ := m.Get(ctx, serverName+".crt")
+	keyBytes, _ := m.Get(ctx, serverName+".key")
+	if certBytes == nil || keyBytes == nil {
+		logging.Infof("no cached TLS cert for %v", serverName)
+	}
+	return certBytes, keyBytes
+}
+
+var certCache = make(MapCache)
```
BruceMacD, line 49: Made these package level rather than function level variables so we can just share them and simplify the logic of `getCertificate`

helm/charts/infra/values.yaml
```diff
@@ -120,7 +120,9 @@ server:
   #   fsGroup: 2000
 
   ## Security context for the server container
-  securityContext: {}
+  securityContext: {
```
mxyng, line 123: Don't need the braces

helm/charts/infra/values.yaml
```diff
@@ -120,7 +120,9 @@ server:
...
+    readOnlyRootFilesystem: true
+  }
   #   runAsUser: 999
   #   runAsGroup: 999
   #   readOnlyRootFilesystem: true
```
mxyng, line 128: Can remove this now that's set by default

helm/charts/infra/values.yaml
mxyng: This sets the security context for the server container, not the pod. For that you'll need to set `podSecurityContext`. It also doesn't set the security context for the connector

internal/cmd/config.go
mxyng: maybe to minimize changes, create a new function called `infraHomeDirName` and have `infraHomeDir` call that to get the name?

internal/server/tls.go
```diff
@@ -17,20 +16,50 @@ import (
 	"github.com/infrahq/infra/internal/logging"
 )
 
+// MapCache is a simple in-memory caching mechanism, it is not thread safe
+type MapCache map[string][]byte
+
+func (m MapCache) Get(ctx context.Context, name string) ([]byte, error) {
+	data, cached := m[name]
+	if !cached {
+		return nil, autocert.ErrCacheMiss
```
dnephin, line 25: I think we can use our own error, since we aren't using this with `autocert.Manager` anymore. Could also be a second pass cleanup.


```suggestion
		return nil, fmt.Errorf("no cache entry for %v", name)
```

internal/server/tls.go
dnephin: We really need to avoid the use of package level variables (#1939).

I think keeping the closure that we had before removes the need for this package level variable.

internal/server/tls.go
```diff
@@ -17,20 +16,50 @@ import (
...
+		logging.Infof("no cached TLS cert for %v", serverName)
+	}
+	return certBytes, keyBytes
+}
+
+var certCache = make(MapCache)
+
+var ca keyPair
+
+var cacheLock sync.RWMutex
```
dnephin, line 53: Same with these. I think they are better off as local state in a closure. We could also create a struct if you prefer that over a closure.

dnephin (commented):
Looks good! But I really think a closure or a struct is a better option than more package level variables.

Package level variables make code harder to test, and harder to audit (the scope and flow of the variables is less obvious).

helm/charts/infra/values.yaml
mxyng: Oops my bad. `readOnlyRootFilesystem` is only applicable for `securityContext`, not `podSecurityContext`

mxyng: The only case where this would be a problem is if the user is using sqlite _and_ isn't using a PVC, i.e. `server.persistence.enabled=false`. That use case should be extremely rare since it recreates the database on pod restart

internal/server/tls.go
BruceMacD: Gonna leave this for now because I know that getCertificate is called within the cert manager, so Im worried they rely on this error type

internal/server/tls.go
```diff
@@ -17,20 +16,35 @@ import (
 	"github.com/infrahq/infra/internal/logging"
 )
 
+// MapCache is a simple in-memory caching mechanism, it is not thread safe
+type MapCache map[string][]byte
+
+func (m MapCache) Get(_ context.Context, name string) ([]byte, error) {
```
BruceMacD, line 22: Need these context parameters to match the autocert.Cache function definitions

internal/server/tls.go
BruceMacD: I brought back the closure 👍

dnephin (approved):
LGTM!

mxyng (approved):

PR #2693, opened 2022-07-22
fix: run as non-root user by default

### LINKED ISSUE #2688: Do not run as `root` in Docker container
## Describe the bug
By default the server & connector process runs as user root.

## Expected behavior
Add a user directive to the Dockerfile.

## Additional context
- Do not write to the root filesystem

mxyng:
## Summary

Create and run as a non-root user (infra) by default.

Changing the run-as user through Kubernetes pod security contexts requires #2689. The server currently creates a `.infra` directory in the running user's home directory. If the security context `runAsUser` does not exist or does not have a home directory, Infra will errorneously try to create the `.infra` directory under `/` and error out with `permission denied`.

## Checklist

## Related Issues

Resolves #2688

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,5 +48,9 @@ FROM alpine
 COPY --from=builder /go/src/github.com/infrahq/infra/infra /bin/infra
 EXPOSE 80
 EXPOSE 443
+ARG USER=infra
+ARG GROUP=infra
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):
Smoke tested and validated

dnephin (approved):
LGTM! I built the image and ran it with `--entrypoint id` and got the response I expected.

PR #2696, opened 2022-07-22
fix(revert): run as non-root user by default

mxyng:
Reverts infrahq/infra#2693

Reverting this change for now since there's implication when migrating from a previous version without default non-root. The scope of this change is significantly greater than initially expected. This revert will only affect the default runtime user. The user can still override the runtime user by specifying a security context or pod security context when deploying to Kubernetes.

```yaml
server:
  podSecurityContext:
    runAsUser: 999
    runAsGroup: 999
    runAsNonRoot: true
```

```yaml
connector:
  podSecurityContext:
    runAsUser: 999
    runAsGroup: 999
    runAsNonRoot: true
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,9 +48,5 @@ FROM alpine
 COPY --from=builder /go/src/github.com/infrahq/infra/infra /bin/infra
 EXPOSE 80
 EXPOSE 443
-ARG USER=infra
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):

mxyng:
Superseded by #2698

Issue #2697, opened 2022-07-22
api: failing to commit a transaction does not result in a 500 response


dnephin:
## Describe the bug

This bug report is based on code inspection. I haven't yet reproduced the issue with a test.

We commit a transaction from `DatabaseMiddleware`: https://github.com/infrahq/infra/blob/03cfb33c40b2f5c90e5d3486ece65fc27a76e983/internal/server/middleware.go#L41-L48

The commit error is only logged, and only at the debug level.  Simply changing this to `sendAPIError` won't work either, because it's too late in the flow of the request. Earlier in the flow we do this:
https://github.com/infrahq/infra/blob/03cfb33c40b2f5c90e5d3486ece65fc27a76e983/internal/server/routes.go#L198

`gin` does attempt to cache a status header for a bit before writing it, but it must write the headers before the body, and that call to `c.JSON` writes the response body. So by the time the `DatabaseMiddleware` calls `COMMIT`, we've already sent a success status header.

This bug would mislead users into thinking their operation was successful, when in reality the transaction commit failed, so the operation was not applied.

## Possible solution

I think we need to manage transactions differently. We should start one transaction for middlewares (`AuthenticationMiddleware` and `DestinationMiddelware`, which we can likely combine into one middleware that calls functions for the two operations , related: https://github.com/infrahq/internal/issues/19). That transaction will likely always be a write transaction (even for `GET` requests), because we must update the user `lastSeenAt`, and possibly the destination `lastSeenAt`. The middleware transaction should be committed before the handler starts.

In `wrappedHandler` we can start a transaction appropriate for the HTTP method. This gives us the opportunity to use different isolation levels based on the method or endpoint. Before writing the HTTP headers the transaction will be committed (for POST, PUT, PATCH, or DELETE), so that we can handle the error from commit before sending a successful status code. GET methods should probably always rollback.

PR #2700, opened 2022-07-22
api: combine the two authenticated middleware

dnephin:
## Summary

I've been trying to tackle a few large problems at the same time, but keep ending up with a change that is too big. This is my third attempt at this. The change in this PR is not really that interesting, it's just a refactor, but I believe it should help me make progress on some other problems (separate txn for middleware, passing DB into access functions, removing state from the API struct, etc).

This PR refactors the two middleware that were used for authenticated endpoints. It converts both to regular functions that return errors. Those two functions are then combined in a single smaller middleware.

Best viewed with [Hide Whitespace](https://github.com/infrahq/infra/pull/2700/files?w=1).

Related to https://github.com/infrahq/internal/issues/19, https://github.com/infrahq/internal/issues/20, and #2697

### DIFF
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/server/middleware.go
```diff
@@ -49,63 +50,80 @@ func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
...
-						return
-					}
-				}
-			default:
-				sendAPIError(c, fmt.Errorf("multiple destinations found for unique ID %q", uniqueID))
-				return
+	// TODO: use GetDestination(ByUniqueID())
+	destinations, err := access.ListDestinations(c, uniqueID, "", &models.Pagination{})
+	if err != nil {
+		return err
```
dnephin, line 62: Previously we would ignore this error, which is probably not a huge problem, but it seemed more correct to return the error now.

internal/server/middleware.go
```diff
@@ -49,63 +50,80 @@ func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
...
 			}
 		}
+		return nil
+	default:
+		return fmt.Errorf("multiple destinations found for unique ID %q", uniqueID)
+	}
+}
 
-		c.Next()
+// TODO: remove duplicate in access package
```
dnephin, line 84: Planning on addressing this in follow up PRs, but I wanted to avoid exporting it for now since that will change another 20+ places.

internal/server/middleware.go
```diff
@@ -49,63 +50,80 @@ func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
...
-		return errors.New("unknown db type in context")
-	}
+type authenticatedUser struct {
+	AccessKey *models.AccessKey
+	User      *models.Identity
+}
 
-	header := c.Request.Header.Get("Authorization")
+// requireAccessKey checks the bearer token is present and valid
+func requireAccessKey(db *gorm.DB, req *http.Request) (authenticatedUser, error) {
```
BruceMacD, line 124: What happens in this function when the db is nil?

internal/server/middleware.go
```diff
@@ -49,63 +50,80 @@ func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
...
+		if err != nil {
 			sendAPIError(c, err)
 			return
 		}
 
+		// TODO: save authnUser in a single key
+		c.Set("key", authnUser.AccessKey)
+		c.Set("identity", authnUser.User)
+
+		if err := updateDestinationLastSeenAt(c); err != nil {
```
BruceMacD, line 110: Bit strange to have this as an unconditional call even when users login, feels like a likely spot to create a bug.

dnephin: Do you mean the name of the function could be misleading?

The behaviour hasn't changed, we previously always looked for the `Infra-Destination` header on all authenticated requests.

I could move the `c.Request.Header.Get("Infra-Destination")` condition to this point, so that the `updateDestinationLastSeen` function is only for the update. Or I could rename the function to something like `handleInfraDestinationHeader`.  What do you think?

internal/server/middleware.go
dnephin: I expect it would panic if the `db` is nil. From the users perspective that wouldn't be any different than what we have today (500 status code).

internal/server/middleware.go
BruceMacD: I like `handleInfraDestinationHeader`, it conveys that it might not be updating anything

dnephin: Sounds good! I've made that change

BruceMacD (approved):
Looks good 👍

PR #2702, opened 2022-07-22
api: Introduce RequestContext, and reduce use of gin.Context

dnephin:
## Summary

This PR also introduces a new `RequestContext` struct. I'd like to use this new type to limit our exposure to `gin.Context` (https://github.com/infrahq/internal/issues/19). We should be able to store other things on this context that we currently get from the API struct (Telemetry, server.Options, the ID of the infra provider, Org, etc). The `RequestContext` is used in one endpoint.

This PR replaces the `currentAccessKey` function with data from the `RequestContext`, and starts to replace `AuthenticatedIdentity` as well.

At the same time I noticed that our `SetCookie` operations were using a `gin.Context` helper. This helper provides no additional functionality over the stdlib function, and actually makes the call site harder to read because of the lack of struct keys.  So I removed the use of `gin.Context.SetCookie` as well.

### DIFF
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -10,15 +10,6 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-func currentAccessKey(c *gin.Context) *models.AccessKey {
-	accessKey, ok := c.MustGet("key").(*models.AccessKey)
-	if !ok {
-		return nil
-	}
-
-	return accessKey
-}
-
 func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired bool, p *models.Pagination) ([]models.AccessKey, error) {
 	roles := []string{models.InfraAdminRole, models.InfraViewRole}
 	db, err := RequireInfraRole(c, roles...)
@@ -57,11 +48,7 @@ func DeleteAccessKey(c *gin.Context, id uid.ID) error {
 	return data.DeleteAccessKeys(db, data.ByID(id))
 }
 
-func DeleteRequestAccessKey(c *gin.Context) error {
+func DeleteRequestAccessKey(c RequestContext) error {
 	// does not need authorization check, this action is limited to the calling key
-	key := currentAccessKey(c)
-
-	db := getDB(c)
-
-	return data.DeleteAccessKey(db, key.ID)
+	return data.DeleteAccessKey(c.DBTxn, c.Authenticated.AccessKey.ID)
 }
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -24,6 +24,9 @@ func isIdentitySelf(c *gin.Context, userID uid.ID) (bool, error) {
 // that was used to authenticate the request.
 // Returns nil if there is no identity in the context, which likely means the
 // request was not authenticated.
+//
+// TODO: remove once all callers use RequestContext.Authenticated.User
+// See https://github.com/infrahq/infra/issues/2796
 func AuthenticatedIdentity(c *gin.Context) *models.Identity {
 	if raw, ok := c.Get("identity"); ok {
 		if identity, ok := raw.(*models.Identity); ok {
@@ -117,24 +120,19 @@ func ListIdentities(c *gin.Context, name string, groupID uid.ID, ids []uid.ID, p
 	return data.ListIdentities(db.Preload("Providers"), p, selectors...)
 }
 
-func GetContextProviderIdentity(c *gin.Context) (*models.Provider, string, error) {
-	// added by the authentication middleware
-	identity := AuthenticatedIdentity(c)
+func GetContextProviderIdentity(c RequestContext) (*models.Provider, string, error) {
+	// does not need authorization check, this action is limited to the calling user
+	identity := c.Authenticated.User
 	if identity == nil {
 		return nil, "", errors.New("user does not have session with an identity provider")
 	}
 
-	// does not need authorization check, this action is limited to the calling user
-	db := getDB(c)
-
-	accessKey := currentAccessKey(c)
-
-	providerUser, err := data.GetProviderUser(db, accessKey.ProviderID, identity.ID)
+	providerUser, err := data.GetProviderUser(c.DBTxn, c.Authenticated.AccessKey.ProviderID, identity.ID)
 	if err != nil {
 		return nil, "", err
 	}
 
-	provider, err := data.GetProvider(db, data.ByID(providerUser.ProviderID))
+	provider, err := data.GetProvider(c.DBTxn, data.ByID(providerUser.ProviderID))
 	if err != nil {
 		return nil, "", fmt.Errorf("user info provider: %w", err)
 	}
@@ -143,21 +141,18 @@ func GetContextProviderIdentity(c *gin.Context) (*models.Provider, string, error
 }
 
 // UpdateIdentityInfoFromProvider calls the identity provider used to authenticate this user session to update their current information
-func UpdateIdentityInfoFromProvider(c *gin.Context, oidc providers.OIDCClient) error {
+func UpdateIdentityInfoFromProvider(c RequestContext, oidc providers.OIDCClient) error {
+	// does not need authorization check, this action is limited to the calling user
 	ctx := c.Request.Context()
 
 	// added by the authentication middleware
-	identity := AuthenticatedIdentity(c)
+	identity := c.Authenticated.User
 	if identity == nil {
 		return errors.New("user does not have session with an identity provider")
 	}
 
-	// does not need authorization check, this action is limited to the calling user
-	db := getDB(c)
-
-	accessKey := currentAccessKey(c)
-
-	provider, err := data.GetProvider(db, data.ByID(accessKey.ProviderID))
+	db := c.DBTxn
+	provider, err := data.GetProvider(db, data.ByID(c.Authenticated.AccessKey.ProviderID))
 	if err != nil {
 		return fmt.Errorf("user info provider: %w", err)
 	}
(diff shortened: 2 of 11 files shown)


internal/server/cookie.go
```diff
@@ -2,42 +2,66 @@ package server
...
+		Path:     cookiePath,
+		Secure:   true, // only over https
+		HttpOnly: true, // not accessible by javascript
+	})
+
+	http.SetCookie(writer, &http.Cookie{
+		Name:     cookieLoginName,
+		MaxAge:   cookieMaxAgeDeleteImmediately,
+		Path:     cookiePath,
+		Secure:   true, // only over https
```
dnephin, line 64: It seems strange that we set `secure: true` for deleting a cookie, but we don't require it for setting a cookie. If we can set a cookie on non-https, don't we also need to delete it on non-https? Otherwise the cookie will still be there for non-https?

Should these changes to `secure: false`, since they are deleting a sensitive value (not saving one).

internal/server/middleware.go
```diff
@@ -36,6 +36,12 @@ func TimeoutMiddleware(timeout time.Duration) gin.HandlerFunc {
 	}
 }
 
+type RequestContext struct {
```
dnephin, line 39: Thoughts on this naming?

I'd like to avoid Context is possible, but I couldn't find another appropriate word for this.  It's easy enough to change later if we find a better name.

internal/access/access_key.go
```diff
@@ -56,12 +56,3 @@ func DeleteAccessKey(c *gin.Context, id uid.ID) error {
...
 }
-
-func DeleteRequestAccessKey(c *gin.Context) error {
-	// does not need authorization check, this action is limited to the calling key
-	key := currentAccessKey(c)
-
-	db := getDB(c)
-
-	return data.DeleteAccessKey(db, key.ID)
-}
```
ssoroka, line 67: I don't agree with this change. Once we allow `data` package calls in the handlers, it's really hard to tell if we're skipping `access` because there's no authorization requirements, or because the dev forgot.

internal/server/cookie.go
ssoroka: it probably needs to match the behavior of the setting function. I think delete calls set cookie under the hood

internal/server/middleware.go
ssoroka: Context or Ctx is fine. this appears everywhere so shorter is probably better.

internal/server/middleware.go
```diff
@@ -97,15 +102,22 @@ func getDB(c *gin.Context) *gorm.DB {
...
 
-		// TODO: save authnUser in a single key
-		c.Set("key", authnUser.AccessKey)
-		c.Set("identity", authnUser.User)
+		rCtx := RequestContext{
+			Request:       c.Request,
+			DBTxn:         db,
+			Authenticated: authned,
+		}
+		c.Set("requestContext", rCtx)
```
ssoroka, line 116: we should change all the existing middleware and handlers to use our own context. I assume that's coming later.

ssoroka (commented):
a nice start, but let's not skip the access layer.

internal/server/middleware.go
dnephin: Ya, that's going to be a big change. I've got a few next steps in #2716, and #2717

internal/access/access_key.go
dnephin: Let's continue to talk about this point. I think that because so many of the functions in `access` do not do access control, that signal of "this has correct authz because it calls access" is not very reliable. The changes in #2716 and #2717 show there are quite a few access functions that don't do any authorization.  Removing those functions makes the code a lot easier to trace, which should it easier to see how and where authz is implemented.

I think the guideline of "the first thing in the function is authorization" is a great idea, but hiding it in the access layer makes it less effective. If we bring it up into the API handlers it'll be much more obvious, and that will allow us to include required roles for an endpoint in the OpenAPI doc as well.

ssoroka: Yeah, I don't agree with your assessment here. It's a lot easier to see something's missing in an empty function than missing at the start of a handler whose job is business logic and request translation; adding a third responsibility there muddies it, and devs are going to miss looking for it in review. It's not being hidden in the access layer, it's being explicit by putting it into its own layer. Empty functions are by design, and intentional, as they communicate intent effectively.

Don't get too attached to the current role-based design as it's not staying around long-term.

internal/server/cookie.go
dnephin: Not sure what you mean. Are you confirming that we should?

This is clearly calling `http.SetCookie`, no question there.  Above this in  `setAuthCookie` is where we create the cookies. When we create them we don't always set `Secure: true`. We based it on if the request came in over http or https.

So my understanding is that a login request that comes in over `http` will receive the session cookie, but a logout over `http` won't remove it from the client, which seems strange and probably not right.

I'd expect these calls to `http.SetCookie` in `deleteAuthCookie` to always use `Secure: false`, since there is nothing sensitive in the values to delete a cookie from the client.

internal/server/middleware.go
dnephin: I don't want to call this `Context` because that is too easily confused with the stdlib context. I was also considering something like `RequestState`, but state is a pretty general term as well.

dnephin:
Ok, I've restores the `access` functions, and change this PR to integrate more of the `RequestContext`. I was able to remove the `"key"` from `gin.Context`. `identity` will be a bit more work and should be a follow up.

internal/server/cookie.go
ssoroka: I'm saying the setting and deleting of the cookie need to use the same `secure: true` or `secure: false` setting for both. I suspect deleting a cookie with secure: true won't remove a cookie set with secure: false, and/or vice versa

internal/server/middleware.go
ssoroka: yeah, but you'll see the package name anyway. so something like ctx.Context won't be easily confused.

internal/access/identity.go
```diff
@@ -143,21 +139,18 @@ func GetContextProviderIdentity(c *gin.Context) (*models.Provider, string, error
...
 		return errors.New("user does not have session with an identity provider")
 	}
 
-	// does not need authorization check, this action is limited to the calling user
-	db := getDB(c)
-
-	accessKey := currentAccessKey(c)
-
-	provider, err := data.GetProvider(db, data.ByID(accessKey.ProviderID))
+	db := c.DBTxn
```
ssoroka, line 152: c.DB or c.Tx might make this easier to use everywhere without setting it to a variable.

ssoroka: long term I'd like to just pass this ctx down to the data layer.

internal/access/request.go
```diff
@@ -0,0 +1,26 @@
...
+type RequestContext struct {
+	Request       *http.Request
+	DBTxn         *gorm.DB
+	Authenticated Authenticated
+}
+
+// Authenticated stores data about the authenticated user. If the AccessKey or
+// User are nil, it indicates that no user was authenticated.
+type Authenticated struct {
+	AccessKey *models.AccessKey
```
ssoroka, line 24: how do we know AccessKey is always going to be set? are these better off as functions?

internal/server/cookie.go
ssoroka, line 64: I believe this needs to be false when http

ssoroka (approved):

internal/server/cookie.go
dnephin: Reading https://datatracker.ietf.org/doc/html/rfc6265#section-5.2.5 it appears this `secure` setting enables the `secure-only-flag` in the cookie car, which does this ([ref](https://datatracker.ietf.org/doc/html/rfc6265#section-5.4)):

> If the cookie's secure-only-flag is true, then the request uri's scheme must denote a "secure" protocol (as defined by the user agent).

My understanding is this means that a cookie with `secure: true` would only be sent back to the server if the url starts with `https://`.

I didn't see anything about this flag being used to compare cookies in the cookie jar, so I'm not sure if it's considered, or if the cookie name is used exclusively.



--- 

I believe we allow http for users who want to front the service with their own LB or proxy that handles TLS for them. We should definitely warn if the `http` addr is not localhost.

dnephin: https://datatracker.ietf.org/doc/html/rfc6265#section-5.3 seems to cover storage, and doesn't mention considering the secure flag. So I think setting `secure: false` for delete of cookies might be correct.

internal/access/identity.go
dnephin: I only used a variable here to reduce the diff. Generally I would expect most code to use `c.DBTxn` directly. We could shorten it to something like `c.Tx`.

I would also use to pass something like this down to the data layer, but I also don't want the data layer to depend on things like `http.Request`, and other properties we might store on this.

So what I'm thinking is that the data layer functions will accept some kind of interface, probably `ReadTxn` and `WriteTxn`, and that interface will be implemented by either this `RequestContext`, or the request context will provide a method that returns the implementation.

I think that should allow the data layer access to things like the authenticated user, and org, without it having to depend on the `RequestContext` directly.

internal/access/request.go
dnephin: It won't always be set because `RequestContext` is used by unauthenticated endpoints as well. Right now it's doing a simple `!= nil` check, but in the future I think exposing more of this through methods would be good.

internal/server/cookie.go
dnephin: Let's discuss in the other comment thread: https://github.com/infrahq/infra/pull/2702#discussion_r932621182

I left some links there

PR #2704, opened 2022-07-23
Revert "improve: move UI to a standalone service"

jmorganca:
This reverts infrahq/infra#2512 so we can better handle the previous `server.config.ui.enabled` flag in the helm chart.


mchiang0610 (approved):
LGTM

PR #2713, opened 2022-07-25
fix: login page ui bug

hoyyeva:
## Summary
The current login page without provider looks like this

Screen Shot 2022-07-25 at 10 12 50 AM

which the OR should be hidden when there is no provider

## 📷 
Screen Shot 2022-07-25 at 10 12 58 AM
Screen Shot 2022-07-25 at 10 18 57 AM


jmorganca (approved):
Great fix!

PR #2717, opened 2022-07-25
Remove more access functions that don't include access control

dnephin:
## Summary

Branched from #2716


ssoroka (changes requested):
as per the other discussion in #2702, I don't think this is the right approach.

PR #2721, opened 2022-07-25
feat: api getPasswordSettings, generate passwords in cli and api obey to server settings 

### LINKED ISSUE #2648: implement a way to retrieve password settings

### LINKED ISSUE #2650: cli: generate temporary password based on current password requirements setting

**Current behaviours:** 
1. It will error out if the user changes the settings and the generated password does not meet it. 
2. when a new user is created, password is generated on the server side.
3. when admin resets a user's password, password is generated on the client-side. The temporary password generated by the cli in: 
   ```
   cmd/users.go/updateUser: 	
   tmpPassword, err := generate.CryptoRandom(12, generate.CharsetPassword)
    ```

**To do:** 
Move #3 client-side password logic to server-side. So, UpdatePassword API endpoint should:
a) generate if an admin is trying to update another user's password, so no password is passed in the request
b) all other cases requires the password to be in the request

~~Blocked by: #2648~~
Blocked by: #2797

kimskimchi:
## Summary

This finishes up everything related to password settings, except `set password settings` which will be its own pr. 

Includes: 
- api `GetPasswordSettings` - unauthenticated
- when a new user is created, server-side `generateTemporaryPassword` passes requirements
- when admin resets a user's password, client-side `generateTemporaryPassword` passes requirements
- rename `publicjwk.go` to just `settings.go`, any settings-related fns belong here
- 

## Checklist

- [ ] Wrote appropriate unit tests - *In progress*

## Related Issues

Resolves #2650 

Resolves #2648


internal/access/credential.go
```diff
@@ -23,11 +24,26 @@ func CreateCredential(c *gin.Context, user models.Identity) (string, error) {
...
 	}
 
-	tmpPassword, err := generate.CryptoRandom(12, generate.CharsetPassword)
+	settings, err := data.GetSettings(db)
+	if err != nil {
+		return "", err
+	}
+
+GENERATE:
+	tmpPassword, err := generate.CryptoRandom(settings.LengthMin, generate.CharsetPassword)
```
ssoroka, line 33: we can't use the min length as the tmp password length. what if the min is 4? it should be the `max(LengthMin, InfraDefaultSize)`

internal/access/credential.go
```diff
@@ -23,11 +24,26 @@ func CreateCredential(c *gin.Context, user models.Identity) (string, error) {
...
 	}
 
+	err = checkPasswordRequirements(db, tmpPassword)
+	var validateError validate.Error
+	if err != nil {
+		if errors.As(err, &validateError) {
+			goto GENERATE
+		}
+		return "", err
+	}
```
ssoroka, line 45: there's a non-zero chance that this loops forever. eg if the sum of the min counts is higher than the number of characters in the string. we should have a max number of tries and error out as a simple solution. a more complex solution would be to generate a string with exactly the characteristics requested.

internal/access/credential.go
```diff
@@ -123,27 +139,33 @@ func hasMinimumCount(min int, password string, check func(rune) bool) bool {
 func checkPasswordRequirements(db *gorm.DB, password string) error {
 	settings, err := data.GetSettings(db)
 	if err != nil {
+		logging.Debugf("error while getting settings")
```
ssoroka, line 142: usually we log or return, but not both. you would be better off wrapping some context into the next error line.. 
`  return fmt.Errorf("getting settings: %w", err)`

internal/access/credential.go
```diff
@@ -123,27 +139,33 @@ func hasMinimumCount(min int, password string, check func(rune) bool) bool {
...
 		return err
 	}
 	errs := make(validate.Error)
 
 	if !hasMinimumCount(settings.LowercaseMin, password, unicode.IsLower) {
+		logging.Errorf("error generating password: not enough lower case letters")
```
ssoroka, line 148: debug maybe. are these helpful?

internal/access/credential.go
```diff
@@ -23,11 +24,26 @@ func CreateCredential(c *gin.Context, user models.Identity) (string, error) {
 		return "", HandleAuthErr(err, "user", "create", models.InfraAdminRole)
 	}
 
-	tmpPassword, err := generate.CryptoRandom(12, generate.CharsetPassword)
+	settings, err := data.GetSettings(db)
+	if err != nil {
+		return "", err
+	}
+
+GENERATE:
```
mxyng, line 32: This is problematic if the user sets a password requirement that is hard for the generator to generate, e.g. high minimum of lowercase characters, and potentially create an infinite loop if the requirements are specific enough.

The problem is that the generator charset is not synced to the password requirement charset. You could synchronize the charsets but I feel like validating randomly generated values is unnecessary.

internal/server/routes.go
```diff
@@ -112,6 +112,7 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	get(a, noAuthn, "/api/providers/:id", a.GetProvider)
 
 	get(a, noAuthn, "/api/version", a.Version)
+	get(a, noAuthn, "/api/password", a.GetPasswordSettings)
```
mxyng, line 115: This seems like a no-so-good path for password requirements. `password` is not really a resource. In this case, it seems a better path could be `/api/settings` which can be the start of the settings API

internal/cmd/users.go
```diff
@@ -346,6 +348,25 @@ func updateUser(cli *CLI, name string) error {
...
 
+func generateTemporaryPassword(client *api.Client) (string, error) {
+	logging.Debugf("call server: get password settings")
+	settings, err := client.GetPasswordSettings()
+	if err != nil {
+		return "", err
+	}
+
+GENERATE:
+	tmpPassword, err := generate.CryptoRandom(settings.LengthMin, generate.CharsetPassword)
```
ssoroka, line 359: same problem as other one. can loop forever.

internal/cmd/users.go
```diff
@@ -452,3 +507,19 @@ func hasAccessToChangePasswordsForOtherUsers(client *api.Client, config *ClientH
...
+		if check(r) {
+			count++
+		}
+	}
+	return count >= min
+}
+
+// list of valid special chars is from OWASP, wikipedia
+func isValidSymbol(letter rune) bool {
+	match, _ := regexp.MatchString(fmt.Sprintf(`(.*[ !"#$%%&'()*+,-./\:;<=>?@^_{}|~%s%s]){1,}`, regexp.QuoteMeta(`/\[]`), "`"), string(letter))
```
ssoroka, line 523: weird that we limit valid symbols, no?

internal/cmd/users.go
```diff
@@ -416,6 +437,40 @@ func checkPasswordRequirements(oldPassword string) survey.Validator {
 	}
 }
 
+func checkServerPasswordRequirements(settings *api.PasswordResponse, password string) error {
+	var errs []string
```
mxyng, line 441: This should probably be `error` type

ssoroka (changes requested):
looking good. couple comments before merge

internal/cmd/users.go
```diff
@@ -416,6 +437,40 @@ func checkPasswordRequirements(oldPassword string) survey.Validator {
...
+
+	if !hasMinimumCount(settings.NumberMin, password, unicode.IsDigit) {
+		errs = append(errs, fmt.Sprintf("- needs minimum %d numbers", settings.NumberMin))
+	}
+
+	if !hasMinimumCount(settings.SymbolMin, password, isValidSymbol) {
+		errs = append(errs, fmt.Sprintf("- needs minimum %d symbols", settings.SymbolMin))
+	}
+
+	if len(password) < settings.LengthMin {
```
mxyng, line 459: This should probably be the first check especially since the output is ordered based on the placement of the check. An alternative is probably to error on first failed check but inform user of the full requirements. Omitting requirements can cause the user to replace valid parts of their input, making them no longer valid.

internal/cmd/users.go
kimskimchi: Yea, this list was based off of a bunch of sources that listed "allowed symbols for passwords"

ssoroka: this list is too American.

internal/server/routes.go
ssoroka: I'd be fine with /api/password-requirements or something, but I think you have a point about their being a lot of different settings clients will need on startup. it doesn't make a ton of sense to make them query them all separately

kimskimchi: @mxyng @ssoroka `/api/settings?name=passwordrequirements`?

dnephin: I think we need to consider how this endpoint will be authorized. There are already things in settings that should only be visible to admins (`PrivateJWK`), and I expect there to be more.

Password requirements on the other hand are not sensitive, and will likely need to be queried without authorization.

I think this means that either:
* we'll need the endpoint to filter out some values based on authorization (something I think is fine, but I know @ssoroka has expressed not wanting to do)
* we should have separate endpoints for public and private settings

I don't think we want an endpoint that returns different structures, so I don't think a `name=...` query param is the right solution. If we change the output it should be based on authorization, not query parameters.

Maybe `/api/settings/public` ?

mxyng: We don't currently have a settings API `/api/settings` though you make a great point. There will be public and private settings as well as user and organization settings. These will all need to be kept separate since they represent distinct audiences.

kimskimchi:
@dnephin the issue is that we need to display the password in plaintext onto the CLI, sent from the server. Is this safe?  (tagging @ssoroka @BruceMacD incase they are familiar with it)

dnephin:
It doesn't seem much different from sending the password from the CLI to the server.  The transport is encrypted the same way.  We would never store the password in plaintext.

PR #2723, opened 2022-07-25
Integrate loadKey with init schema

### LINKED ISSUE #2066: data: why do migrations require calling initializeSchema multiple times?

dnephin:
## Summary

Branched from #2722

Pass `loadKey` to migrator so that we don't have to duplicate some of the "has schema been intialized" logic.

## Related Issues

Resolves #2066

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -476,7 +476,6 @@ func migrations() []*migrator.Migration {
 		addAuthURLAndScopeToProviders(),
 		setDestinationLastSeenAt(),
 		deleteDuplicateGrants(),
-		addFieldsFor_0_14(),
 		dropDeletedProviderUsers(),
 		// next one here
 	}
@@ -636,20 +635,4 @@ func dropDeletedProviderUsers() *migrator.Migration {
 func deleteDuplicates(tx *gorm.DB, group string, model models.Modelable) error {
 	subQuery := tx.Select("min(id)").Group(group).Model(model)
 	return tx.Where("id NOT in (?)", subQuery).Delete(model).Error
-}
-
-// nolint:revive
-func addFieldsFor_0_14() *migrator.Migration {
-	return &migrator.Migration{
-		ID: "2022-07-21T18:28",
-		Migrate: func(tx *gorm.DB) error {
-			if err := tx.AutoMigrate(&models.Provider{}); err != nil {
-				return err
-			}
-			if err := tx.AutoMigrate(&models.Settings{}); err != nil {
-				return err
-			}
-			return nil
-		},
-	}
 }
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -17,10 +17,8 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-func migrate(db *gorm.DB) error {
-	opts := migrator.DefaultOptions
-	opts.InitSchema = preMigrate
-	m := migrator.New(db, opts, []*migrator.Migration{
+func migrations() []*migrator.Migration {
+	return []*migrator.Migration{
 		// rename grants.identity -> grants.subject
 		{
 			ID: "202203231621", // date the migration was created
@@ -79,7 +77,7 @@ func migrate(db *gorm.DB) error {
 
 				var keys []AccessKey
 
-				err := db.Find(&keys).Error
+				err := tx.Find(&keys).Error
 				if err != nil {
 					return fmt.Errorf("migrating access key identities: %w", err)
 				}
@@ -115,7 +113,7 @@ func migrate(db *gorm.DB) error {
 						SecretChecksum: key.SecretChecksum,
 					}
 
-					if err := SaveAccessKey(db, convertedKey); err != nil {
+					if err := SaveAccessKey(tx, convertedKey); err != nil {
 						return fmt.Errorf("save converted key: %w", err)
 					}
 				}
@@ -139,7 +137,7 @@ func migrate(db *gorm.DB) error {
 
 				var creds []Credential
 
-				if err := db.Find(&creds).Error; err != nil {
+				if err := tx.Find(&creds).Error; err != nil {
 					return fmt.Errorf("migrating creds: %w", err)
 				}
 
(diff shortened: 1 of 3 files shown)


internal/server/data/migrations.go
```diff
@@ -17,10 +17,8 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-func migrate(db *gorm.DB) error {
-	opts := migrator.DefaultOptions
-	opts.InitSchema = preMigrate
-	m := migrator.New(db, opts, []*migrator.Migration{
+func migrations() []*migrator.Migration {
```
dnephin, line 20: I removed the `migrate` function and created this `migrations` function instead. This should allow a test to compare the list of migrations against the list of tests in the table, to find when we've missing a migration test.

It also found some potential bugs. We were using the `db` variable instead of the `tx` variable in some of these migrations.

internal/server/data/migrations.go
```diff
@@ -637,3 +619,19 @@ func deleteDuplicates(tx *gorm.DB, group string, model models.Modelable) error {
 	subQuery := tx.Select("min(id)").Group(group).Model(model)
 	return tx.Where("id NOT in (?)", subQuery).Delete(model).Error
 }
+
+// nolint:revive
+func addFieldsFor_0_14() *migrator.Migration {
```
dnephin, line 624: Now that we don't call `AutoMigrate` after `migrate` we will need to explicit create migrations for adding columns.

I think this is much safer. It'll prompt us to write tests for them as well, and it will be more clear that we are making changes to the db schema.

The downside is that it's currently a lot of work to add a test for a database migration, so I'm going to be looking at making that easier next.

internal/server/data/migrations.go
```diff
@@ -637,3 +619,19 @@ func deleteDuplicates(tx *gorm.DB, group string, model models.Modelable) error {
...
+	return &migrator.Migration{
+		ID: "202207211828",
+		Migrate: func(tx *gorm.DB) error {
+			if err := tx.AutoMigrate(&models.Provider{}); err != nil {
+				return err
+			}
+			if err := tx.AutoMigrate(&models.Settings{}); err != nil {
+				return err
+			}
```
dnephin, line 633: TODO: I should replace these with sql `ALTER TABLE` statements if possible.

internal/server/data/migrator/migrator.go
```diff
@@ -50,6 +54,11 @@ var DefaultOptions = Options{
 
 // New returns a new Migrator.
 func New(db *gorm.DB, options Options, migrations []*Migration) *Migrator {
+	if options.LoadKey == nil {
```
BruceMacD, line 57: Are we going to make it possible to call specific migrations? Otherwise this will be required most of the time I imagine.

internal/server/data/migrator/migrator.go
```diff
@@ -177,7 +190,7 @@ func (g *Migrator) runInitSchema() error {
 			return err
 		}
 	}
-	return nil
+	return g.options.LoadKey(g.tx)
```
BruceMacD, line 193: Why does this get called twice?

internal/server/data/migrator/migrator.go
dnephin: Specific migrations, no, but it's pretty easy to pass in an arbitrary list of migrations. In the test I plan on running `[0:n]` to simulate applying each migration in order.

In our tests though we don't always pass a `LoadKey`, so I think it has to be optional unless we start using a consistent data key with all of our tests. Should be easy to make it required later if we need to.

internal/server/data/migrator/migrator.go
dnephin: There are 3 possible flows for migrations:
1. fresh install - call initialize schema and write migration IDs to the migrations table
2. run new migrations
3. don't run any migrations if they have all been applied

One call is for flow 1, and second call is for flow 2 and 3.

internal/server/data/migrations.go
BruceMacD: Just a heads up, sqlite's support for alter table is really limited.

BruceMacD (approved):
Looks good

internal/server/data/migrations.go
```diff
@@ -655,3 +637,19 @@ func deleteDuplicates(tx *gorm.DB, group string, model models.Modelable) error {
 	subQuery := tx.Select("min(id)").Group(group).Model(model)
 	return tx.Where("id NOT in (?)", subQuery).Delete(model).Error
 }
+
+// nolint:revive
+func addFieldsFor_0_14() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-07-21T18:28 add fields from 0.14.0",
```
dnephin, line 644: I amended the commit to use a bit more verbose name for the migration IDs. I think this reads a lot better and should work just as well as the old IDs.

dnephin:
I'm going to wait to merge this, because further testing shows that removing the second `AutoMigrate` will require a few more migrations be added.

dnephin:
I pushed another commit which reverts part of the original change. Removing the calls to `AutoMigrate` is a lot more work. So for now we'll keep them. When the test coverage for migrations is complete we can replace the `AutoMigrate` with explicit migrations.

PR #2727, opened 2022-07-26
feat: do not prompt list of login methods if no providers are connected

### LINKED ISSUE #2636: If there's only one login method (eg. username/pw), don't prompt list but just go straight to that signing method.

Current:
```
? Select a server: localhost
? Select a login method: [Use arrows to move, type to filter]
> Login with username and password
```

Should be:

A: Only Login with username and password
```
? Select a server: localhost
? Username: jjj
? Password: ***
```
B: Only `Login with okta`
```
? Select a server: localhost
# redirects immediately
```

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

kimskimchi:
## Summary

Before:
```
$ infra login
? Select a server: [Use arrows to move, type to filter]
> localhost


$ infra login
? Select a server: localhost
? Select a login method: [Use arrows to move, type to filter]
> Login with username and password
```

After:
```
$ infra login
? Select a server: [Use arrows to move, type to filter]
> localhost


$ infra login
? Select a server: localhost
? Username: 
```

## Checklist

## Related Issues

Resolves #2636

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -564,6 +564,10 @@ func promptLoginOptions(cli *CLI, client *api.Client) (loginMethod loginMethod, 
 		return 0, nil, err
 	}
 
+	if len(providers) == 0 {
+		return localLogin, nil, nil
+	}
+
 	var options []string
 	for _, p := range providers {
 		options = append(options, fmt.Sprintf("%s (%s)", p.Name, p.URL))


BruceMacD (approved):
Looks good 👍  I think we should add similar behavior for only one server (automatically selecting localhost in your example), but that is a different change

PR #2736, opened 2022-07-26
feat(api): add totalUsers to groups endpoint

### LINKED ISSUE #2672: add number of users in each group to `/api/groups`

j-sneh:
## Summary

Adds totalUsers (a count of users in a group) by counting from the identities_groups join table for use in the API/CLI.

## Checklist

## Related Issues

Resolves #2672

### DIFF
--- a/api/group.go
+++ b/api/group.go
@@ -6,10 +6,11 @@ import (
 )
 
 type Group struct {
-	ID      uid.ID `json:"id"`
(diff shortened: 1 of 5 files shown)


ssoroka (approved):
looks good. the loop with the group count is going to eventually be problematic, but it's fine for now.

PR #2738, opened 2022-07-26
feat(helm): bind services to non-privileged port ranges

### LINKED ISSUE #2737: Configure server/connector listening address/port through Helm
Expose configuration options to change the service listening addresses. Default to a high port number (> 1024)

mxyng:
## Summary

Allow the bind addresses for server and connector to be configured through Helm.

Set a default bind address such that the pod can be run without additional privileges for binding to sub-1024 ports.

## Checklist

## Related Issues

Resolves #2737

### DIFF
--- a/helm/charts/infra/templates/connector/configmap.yaml
+++ b/helm/charts/infra/templates/connector/configmap.yaml
@@ -12,6 +12,11 @@ data:
     name: {{ . }}
 {{- end }}
 
+    addr:
+{{- range $key, $val := .Values.connector.config.addr }}
+      {{ $key }}: ':{{ $val }}'
+{{- end }}
+
     server:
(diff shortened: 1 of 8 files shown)


dnephin (approved):
LGTM! 

The change in default values should be fine because everything connects to the service port, right?

The only potentially breaking change is if someone is using a `values.yaml` file that doesn't set values for these fields?  Should we use `| default` in the template to account for that?

mxyng:
Values are merged with `values.yaml` that comes with the chart so this shouldn't be an issue unless they use something like `--reuse-values`. Could add `| default` just in case but I don't think it's necessary

dnephin:
> Values are merged with values.yaml that comes with the chart

Ahh, I didn't know this. I assumed it was replaced. Good to know, and this seems safe.

Issue #2740, opened 2022-07-26
Fix default fsGroup in helm


fsdaniel:
## Describe the bug

0.14.1 introduces running process as infra (UID/GID) 1000.
As no podSecurityContext is passed by default mount is done as root making it unable for app to write data to it.

## Expected behavior

Server comes online.

## Screenshots
```
{"level":"error","time":1658864558395,"caller":"logging/database.go:29","message":"failed  to initialize database, got error unable to open database file: no such  file or directory"}
```

## Environment
```
infra: 0.14.1

```

```
kubernetes: v1.22.9
```

## Relevant Infra Logs
```
{"level":"error","time":1658864558395,"caller":"logging/database.go:29","message":"failed  to initialize database, got error unable to open database file: no such  file or directory"}
```

## Additional context

- Information on security context: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
- change in 0.14.1: https://github.com/infrahq/infra/commit/c21623d27875815b5d749a744912121ccb7a8486

mxyng:
Fixed in #2741

PR #2741, opened 2022-07-26
fix(helm): Change default podSecurityContext & securityContext

fsdaniel:
## Summary
- Update helm chart to 0.14.2
- change default podSecurityContext to set fsGroup
- set securityContext runAsUser & runAsGroup by default to match dockerfile.

## Checklist

## Related Issues

Resolves [#2740]

### DIFF
--- a/helm/charts/infra/Chart.yaml
+++ b/helm/charts/infra/Chart.yaml
@@ -3,6 +3,6 @@ name: infra
 description: A Helm chart for Infra
 type: application
 ## {x-release-please-start-version}
-version: 0.14.2
-appVersion: 0.14.2
+version: 0.14.1
+appVersion: 0.14.1
 ## {x-release-please-end}
--- a/helm/charts/infra/Chart.yaml
+++ b/helm/charts/infra/Chart.yaml
@@ -3,6 +3,6 @@ name: infra
 description: A Helm chart for Infra
 type: application
 ## {x-release-please-start-version}
-version: 0.14.1
-appVersion: 0.14.1
+version: 0.14.2
+appVersion: 0.14.2
 ## {x-release-please-end}
--- a/helm/charts/infra/values.yaml
+++ b/helm/charts/infra/values.yaml
@@ -116,13 +116,14 @@ server:
   podAnnotations: {}
 
   ## Security context for the server pod
-  podSecurityContext: {}
-  #   fsGroup: 2000
+  podSecurityContext:
+    fsGroup: 1000
 
   ## Security context for the server container
   securityContext:
     readOnlyRootFilesystem: true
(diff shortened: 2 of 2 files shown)


helm/charts/infra/Chart.yaml
```diff
@@ -3,6 +3,6 @@ name: infra
 description: A Helm chart for Infra
 type: application
 ## {x-release-please-start-version}
-version: 0.14.1
-appVersion: 0.14.1
+version: 0.14.2
+appVersion: 0.14.2
```
mxyng, line 7: There's no need to increment the version. The helm chart version will be bumped automatically once a release is tagged and built.

mxyng (commented):
Hey @fsdaniel, thanks for creating this PR and issue. Can you provide more context to the issue you're facing? I have not been able to reproduce the problem you described in #2740. In particular, the Kubernetes provider (minikube, EKS, GKE, etc.) and any relevant Helm values.

That being said, I see no reason not to accept this.

helm/charts/infra/Chart.yaml
fsdaniel: I do a clean helm install on a vanilla kubernetes cluster ( based on rancher RKE ). You should be able to replicate the issue by starting a clean helm install with a refresh PVC.

fsdaniel:
> Hey @fsdaniel, thanks for creating this PR and issue. Can you provide more context to the issue you're facing? I have not been able to reproduce the problem you described in #2740. In particular, the Kubernetes provider (minikube, EKS, GKE, etc.) and any relevant Helm values.
> 
> That being said, I see no reason not to accept this.

I do a clean helm install on a vanilla on-prem Kubernetes cluster ( based on rancher RKE ). You should be able to replicate the issue by starting a clean helm install with a refresh PVC.

```yaml
server:
  service:
    type: ClusterIP
  config:
    users:
      - name: admin
        password: xxxxxxxx
    grants:
      - user: admin
        role: admin
        resource: infra
  ingress:
    enabled: true
    hosts:
      - auth.xxxxxx.io
    servicePort: 80
    className: nginx
    annotations:
      nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
      nginx.ingress.kubernetes.io/backend-protocol: "HTTP"
      nginx.ingress.kubernetes.io/proxy-http-version: "1.0"
      cert-manager.io/cluster-issuer: "letsencrypt-prod"
    tls:
      - hosts:
          - auth.xxxx.io
        secretName: io-xxx-auth
```

mxyng:
Thanks for that info. I was able to replicate in EKS and also verify your fix. Thanks for the contribution!

mxyng (approved):

PR #2746, opened 2022-07-26
Initial Orgs Checkin

pdevine:
## Summary

Add CRUD for Organizations

## Checklist

### DIFF
--- a/api/organization.go
+++ b/api/organization.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"github.com/infrahq/infra/internal/validate"
 	"github.com/infrahq/infra/uid"
 )
 
@@ -11,6 +12,28 @@ type Organization struct {
 	Updated Time   `json:"updated"`
 }
 
+type ListOrganizationsRequest struct {
+	Name string `form:"name"`
+	PaginationRequest
+}
+
+func (r ListOrganizationsRequest) ValidationRules() []validate.ValidationRule {
+	// no-op ValidationRules implementation so that the rules from the
+	// embedded PaginationRequest struct are not applied twice.
+	return nil
+}
+
 type CreateOrganizationRequest struct {
 	Name string `json:"name" validate:"required"`
+}
+
+func (r CreateOrganizationRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("name", r.Name),
+	}
+}
+
+func (req ListOrganizationsRequest) SetPage(page int) Paginatable {
+	req.PaginationRequest.Page = page
+	return req
 }
--- a/api/organization.go
+++ b/api/organization.go
@@ -1,0 +1,16 @@
+package api
+
+import (
+	"github.com/infrahq/infra/uid"
+)
+
+type Organization struct {
+	ID      uid.ID `json:"id"`
+	Name    string `json:"name"`
+	Created Time   `json:"created"`
+	Updated Time   `json:"updated"`
+}
+
+type CreateOrganizationRequest struct {
+	Name string `json:"name" validate:"required"`
+}
--- a/api/organization.go
+++ b/api/organization.go
@@ -24,7 +24,7 @@ func (r ListOrganizationsRequest) ValidationRules() []validate.ValidationRule {
 }
 
 type CreateOrganizationRequest struct {
-	Name string `json:"name" validate:"required"`
+	Name string `json:"name"`
 }
 
 func (r CreateOrganizationRequest) ValidationRules() []validate.ValidationRule {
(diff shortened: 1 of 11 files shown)


api/organization.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type ListOrganizationsRequest struct {
+	Name string `form:"name"`
+	PaginationRequest
+}
+
+func (r ListOrganizationsRequest) ValidationRules() []validate.ValidationRule {
+	// no-op ValidationRules implementation so that the rules from the
+	// embedded PaginationRequest struct are not applied twice.
```
ssoroka, line 22: seems like something we should look into more.

api/organization.go
```diff
@@ -0,0 +1,39 @@
...
+	return nil
+}
+
+type CreateOrganizationRequest struct {
+	Name string `json:"name" validate:"required"`
```
ssoroka, line 27: I think the validate tag isn't needed anymore.

internal/access/organization.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+func ListOrganizations(c *gin.Context, name string, pg *models.Pagination) ([]models.Organization, error) {
+	var selectors = []data.SelectorFunc{}
+	if name != "" {
+		selectors = append(selectors, data.ByName(name))
+	}
+
+	roles := []string{models.InfraAdminRole}
+	db, err := RequireInfraRole(c, roles...)
```
ssoroka, line 18:
```suggestion
	db, err := RequireInfraRole(c, models.InfraAdminRole)
```

internal/access/organization.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return HandleAuthErr(err, "organizations", "delete", roles...)
+	}
+
+	selectors := []data.SelectorFunc{
+		data.ByID(id),
+	}
+
+	return data.DeleteOrganizations(db, selectors...)
```
ssoroka, line 61:
```suggestion
	return data.DeleteOrganizations(db, data.ByID(id))
```

internal/server/models/organization.go
```diff
@@ -0,0 +1,24 @@
...
+	CreatedBy uid.ID
+
+	Identities []Identity `gorm:"many2many:identities_organizations"`
+}
+
+func (o *Organization) ToAPI() *api.Organization {
+	return &api.Organization{
+		ID:      o.ID,
+		Created: api.Time(o.CreatedAt),
+		Updated: api.Time(o.UpdatedAt),
```
ssoroka, line 21: might not want to expose created/updated for orgs

internal/server/organizations.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (a *API) CreateOrganization(c *gin.Context, r *api.CreateOrganizationRequest) (*api.Organization, error) {
+	org := &models.Organization{
+		Name: r.Name,
+	}
+
+	authIdent := access.AuthenticatedIdentity(c)
+	if authIdent != nil {
+		org.CreatedBy = authIdent.ID
+	}
```
ssoroka, line 42: I thought this happens automatically.. it should probably.

internal/access/organization.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return nil, HandleAuthErr(err, "organizations", "get", roles...)
+	}
+
+	return data.GetOrganization(db, data.ByID(id))
+}
+
+func CreateOrganization(c *gin.Context, org *models.Organization) error {
+	roles := []string{models.InfraAdminRole}
+	db, err := RequireInfraRole(c, roles...)
```
ssoroka, line 42: probably anyone should be able to create an organization, but maybe that's through a specific signup flow and not through this function?

ssoroka (approved):

internal/access/organization.go
pdevine: Yes. There's another commit which adds the `InfraSupportRole` which this will be modified to use in the future, and most people will instead use `/api/signup` to create the org.

internal/server/organizations.go
pdevine: Unfortunately it's not. Grants sets it in the access layer, but nothing else sets it at all. I'll file a ticket to fix it.

internal/server/models/organization.go
pdevine: This will only be available to the support role in the future, so I think it's probably OK.

api/organization.go
```diff
@@ -0,0 +1,39 @@
...
+	return nil
+}
+
+type CreateOrganizationRequest struct {
+	Name string `json:"name"`
+}
+
+func (r CreateOrganizationRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("name", r.Name),
```
dnephin, line 32: We should use `ValidateName(r.Name)` on this as well I suspect

dnephin (commented):
Nice test coverage!

PR #2750, opened 2022-07-27
fix: hard delete provider users on removal

### LINKED ISSUE #2706: Error when logging in with provider
Upon logging in with Google, an error might occur where a duplicate `provider_users` record is created

```
{"level":"warn","time":1658718927121,"caller":"logging/logger.go:79","message":"no refresh token returned from oidc client for \"accounts.google.com\", session lifetime will be reduced"}
{"level":"error","rows":0,"query":"INSERT INTO \"provider_users\" (\"created_at\",\"updated_at\",\"deleted_at\",\"provider_id\",\"identity_id\",\"email\",\"groups\",\"last_update\",\"redirect_url\",\"access_token\",\"refresh_token\",\"expires_at\",\"id\") VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13) RETURNING \"id\"","elapsed":103.63991,"error":"ERROR: duplicate key value violates unique constraint \"provider_users_pkey\" (SQLSTATE 23505)","time":1658718927242,"caller":"data/data.go:162"}
{"level":"info","error":"unauthorized: login failed: failed to login: add user for provider login: value already exists for provider_users","method":"POST","path":"/api/login","statusCode":401,"remoteAddr":"192.168.205.197:42262","time":1658718927242,"caller":"server/routes.go:158","message":"api request error"}
```

BruceMacD:
## Summary
When's user that's previously had their session revoked tried to sign in again through an IDP the soft-delete causes a conflict when their provider-user is recreated.

1. User signs in through IDP
2. Their session expires and they try to perform another action
3. Since their session with the IdP is invalid it returns an error, and the provider user record is deleted
4. When the user tries to sign-in again it attempts to re-create the provider user, but it fails because the deleted record is still seen as existing due to our soft-delete not removing the record.

Fix is to not solf delete provider users, and remove them on session revocation. 

New provider_users table:
Screen Shot 2022-07-27 at 1 56 16 PM

## Checklist

## Related Issues

Resolves #2706


internal/server/data/migrations.go
```diff
@@ -634,6 +635,13 @@ func setDestinationLastSeenAt() *migrator.Migration {
 	}
 }
 
+// dropDeletedProviderUsers removes soft-deleted provider users so they do not cause conflicts
+func dropDeletedProviderUsers() *migrator.Migration {
```
BruceMacD, line 639: This cleans up any provider users that are in a bad state due to soft-deletion.

internal/access/identity.go
```diff
@@ -173,10 +173,6 @@ func UpdateIdentityInfoFromProvider(c *gin.Context, oidc providers.OIDCClient) e
 			logging.Errorf("failed to revoke invalid user session: %s", nestedErr)
 		}
 
-		if nestedErr := data.DeleteProviderUsers(db, data.ByIdentityID(identity.ID), data.ByProviderID(provider.ID)); nestedErr != nil {
```
BruceMacD, line 176: Sessions are still revoked, just the provider is still seen as a way that the user logged in. This matches previous behavior.

internal/server/data/data.go
```diff
@@ -100,6 +100,11 @@ func getDefaultSortFromType(t interface{}) string {
 		return "name ASC"
 	}
 
+	if _, ok := ty.FieldByName("Email"); ok {
```
BruceMacD, line 103: Now that provider users don't have IDs we need to check for a different field before defaulting to sorting by ID. `email` seems like a logical choice after `name`.

internal/server/models/provideruser.go
```diff
@@ -8,10 +8,8 @@ import (
 
 // ProviderUser is a cache of the provider's user and their groups, plus any authentication-specific information for that provider.
 type ProviderUser struct {
-	Model
```
mxyng, line 11: This may cause problems when deleting provider users since gorm updates `deleted_at` column rather than deleting the row. It's possible the gorm will hard delete the row if the `deleted_at` is not found but that means we will not be able to recover deleted rows. I'm not sure that's a good idea at this time.

internal/server/data/data.go
ssoroka: hmm. does this table even need sort? email might be ineffective as a sort since it should be the same on all providers.

BruceMacD: It doesn't need it publicly, but we still need it for the list function which always has pagination, so calling it internally (such as in tests) this is required.

mxyng (commented):
Should add a test for deleting ProviderUsers; I don't see one in `provideruser_test.go`

ssoroka (approved):
lgtm

internal/server/models/provideruser.go
BruceMacD: That's actually the intended behavior in this case. Gorm has some limitations around setting the unique index that make soft-delete here a pain. We can't set the "deleted at" to be part of the primary key since it is nullable and on a separate struct. Then if we try to set this manually, we hit issues with Gorm automatically resetting the primary key on migration.

PR #2754, opened 2022-07-27
feat(ui): add providers to user data

### LINKED ISSUE #2678: show which providers each user and group is provided by
Expanding up https://github.com/infrahq/infra/issues/2521, there's also currently no way to see which provider(s) each user is provided by like you can with the CLI. It should show up on the sidebar when a user is clicked on.

j-sneh:
## Summary

Adds providerNames to user metadata
Screen Shot 2022-07-27 at 1 58 49 PM

## Checklist

## Related Issues

Resolves #2678

### DIFF
--- a/ui/pages/users/index.js
+++ b/ui/pages/users/index.js
@@ -81,6 +81,7 @@ function Details({ user, admin, onDelete }) {
       title: 'Created',
       data: user?.created ? dayjs(user.created).fromNow() : '-',
     },
+    { title: 'Providers', data: user?.providerNames.join(', ') },
   ]
 
   const loading = [


ui/pages/users/index.js
```diff
@@ -81,6 +81,7 @@ function Details({ user, admin, onDelete }) {
       title: 'Created',
       data: user?.created ? dayjs(user.created).fromNow() : '-',
     },
+    { title: 'Providers', data: user?.providerNames.join(', ') },
```
BruceMacD, line 84: nit:

You could consider 'Provided By' rather than 'Providers' to match the CLI

BruceMacD (approved):

ui/pages/users/index.js
j-sneh: That was what I had initially, but changed it to Providers for consistency in the UI

PR #2755, opened 2022-07-27
feat: add infra support admin user

pdevine:
## Summary

Add a support admin role which is separate from the admin role. This role is specifically for:

* performing CRUD operations on organizations (a normal admin shouldn't be able to do this)
* granting support admin access

This change additionally assigns the org permission on signup, although that will change when the signup endpoint morphs to allow users to create new organizations. It does not add the role for existing admins, since it's difficult to know who to assign the permission to.

Also in this change is a cleanup of the CreateGrant unit test to make it more straight forward, and the Organizations unit tests to remove some unnecessary code.

## Checklist

### DIFF
--- a/internal/server/grants_test.go
+++ b/internal/server/grants_test.go
@@ -667,7 +667,7 @@ func TestAPI_CreateGrant(t *testing.T) {
 				assert.DeepEqual(t, actual, expected, cmpAPIGrantJSON)
 			},
 		},
-		"unprivileged support admin grant": {
+		"admin can not grant infra support admin role": {
 			setup: func(t *testing.T, req *http.Request) {
 				req.Header.Set("Authorization", "Bearer "+adminAccessKey(srv))
 			},
--- a/internal/server/grants_test.go
+++ b/internal/server/grants_test.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"sort"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


internal/server/grants_test.go
```diff
@@ -565,85 +564,161 @@ func TestAPI_CreateGrant(t *testing.T) {
...
+					accessKey.IssuedFor,
+					models.InfraAdminRole,
+					someUser.ID.String(),
+					time.Now().UTC().Format(time.RFC3339),
+				))
+				actual := jsonUnmarshal(t, resp.Body.String())
+				assert.DeepEqual(t, actual, expected, cmpAPIGrantJSON)
+			},
+		},
+		"unprivileged support admin grant": {
```
dnephin, line 670: Maybe something like:

```suggestion
		"admin can not grant infra support admin role": {
```

dnephin (approved):
Looks good.  Do we need to change `RequireInfraRole` to allow the infra admin role to do everything an admin can do. Or will we have to create separate grants for that?

internal/server/grants_test.go
pdevine: I've been thinking of the InfraAdminRole as being essentially an OrgAdminRole, and the InfraSupportAdminRole as only able to do high level org manipulation. I think we still want orgs to have to invite a support admin into their role (and grant them admin access) in order to change anything in an org (i.e. principle of least privilege)

PR #2756, opened 2022-07-28
fix: `infra info` handle error when session is invalid for the server (not expired or cleared, but just invalid)

### LINKED ISSUE #2587: cli: when you run `infra info` on a fresh install, error message needs to be user-friendly
## Describe the bug

% infra info
```
 Server: localhost
Error: GET "/api/users/avjSmN1C5G": Get "https://localhost/api/users/avjSmN1C5G": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "Infra Server")
```

## Environment

```
$ infra version
```

```
$ kubectl version
```

kimskimchi:
## Summary

When a server gets re-installed, it does not clear the config as that is client side. 
So, client will try to call the API, and if it fails with unauthorized (but is before expiry and is not cleared yet), it will suggest user to re-login due to the session being invalid for this server. 

The change is just move the api call earlier so it fails before printing anything + if 401, print a message. 

## Related Issues

Resolves #2587


internal/cmd/info.go
```diff
@@ -36,6 +36,14 @@ func info(cli *CLI) error {
 		return err
 	}
 
+	user, err := client.GetUser(config.UserID)
+	if err != nil {
+		if api.ErrorStatusCode(err) == 401 {
+			return Error{Message: "Session is not valid for this server; run 'infra login' to start a new session"}
```
BruceMacD, line 42: It seems like this would happen for any API call on a new server (ex: If I installed a new server and ran `infra list`). Any way we can check for this scenario in the CLI error handling logic?

kimskimchi: I thought about this too, but I don't see an easy way. 401s means different things for each cmd - EG)
1. Info - unauthorized means they don't have access to the server, otherwise they would have been able to query OR fail as not logged in/expired. 
2. Destinations list - unauthorized means they don't have enough access to query the list of destinations
3. users edit - unauthorized means they are trying to edit someone else and they are not admin

All three 401s means different things, so I wasn't able to centralize this in a catch-all, like NotLoggedIn (CLI) or BadRequest (API).

BruceMacD (approved):
Seems good to me, makes sense that we can't infer this everywhere.

PR #2759, opened 2022-07-28
fix: rollback postgres db transaction on fail

### LINKED ISSUE #2652: Server does not properly propogate duplicate grant error on postgres

BruceMacD:
- postgres db transaction needs to be rolled back on failure or it can't be re-used to make other database requests

## Summary
Postgres was failing with an internal server error on duplicate grant creation. The reason for this is that Postgres closes the DB transaction on error, but we would then attempt to read the existing grant on the same transaction.

Fix (as suggested by Steven) is to rollback the failed query in the case of a postgres database. 

This replaces the previous PR for the same issue: #2658 

## Checklist

## Related Issues

Resolves #2652

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -155,7 +155,18 @@ func save[T models.Modelable](db *gorm.DB, model *T) error {
 }
 
 func add[T models.Modelable](db *gorm.DB, model *T) error {
-	err := db.Create(model).Error
+	var err error
+	if db.Name() == "postgres" {
+		// failures on postgres need to be rolled back in order to
+		// continue using the same transaction
+		db.SavePoint("beforeCreate")
+		err = db.Create(model).Error
+		if err != nil {
+			db.RollbackTo("beforeCreate")
+		}
+	} else {
+		err = db.Create(model).Error
+	}
 	return handleError(err)
 }
 
--- a/internal/server/data/data_test.go
+++ b/internal/server/data/data_test.go
@@ -216,7 +216,7 @@ func TestCreateTransactionError(t *testing.T) {
 
 		// attempt to re-create, which results in a conflict
 		err = add(tx, g)
-		assert.ErrorContains(t, err, "a grant with that id already exists")
+		assert.ErrorContains(t, err, "already exists")
 
 		// the same transaction should still be usable
 		_, err = get[models.Grant](tx, ByID(g.ID))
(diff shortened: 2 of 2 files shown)


internal/server/data/data.go
```diff
@@ -155,7 +155,18 @@ func save[T models.Modelable](db *gorm.DB, model *T) error {
 }
 
 func add[T models.Modelable](db *gorm.DB, model *T) error {
-	err := db.Create(model).Error
+	var err error
```
BruceMacD, line 158: I didn't add this rollback logic to `save` because I couldn't think of a scenario we would expect an error when calling that function. If you can think of something let me know and I'll add it to `save` to.

dnephin (approved):
LGTM

PR #2761, opened 2022-07-28
data: generate methods for data models

dnephin:
## Summary

This PR adds code generation for the boilerplate model methods. `GenerateMethods` uses reflection to retrieve the list of struct fields, and the `schema.sql` to find the list of columns in the table. From those two pieces of data it's able to generate a `Columns`, `Values`, and `ScanFields` method, which we can use with generic query functions to implement the standard operations for most tables.

The code generation can be run with:

```
go test -run TestGenerateTableMethods ./internal/server/data -generate=all 
# or -generate=<type name>
```

The generated code is injected into the regular source file. 

This PR only adds the code generation. The next PR will start to use it.

## Related Issues

Related to #2415


internal/server/data/model/generate.go
```diff
@@ -0,0 +1,234 @@
...
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Recv == nil {
+			continue
+		}
+		if astIdentForNode(fn.Recv.List[0].Type).Name != rt.Name() {
+			continue
+		}
+
+		if names := fn.Recv.List[0].Names; len(names) > 0 && names[0].Name != "_" {
+			recvName = names[0].Name
```
BruceMacD, line 43: Any way to test this file? I find the code really hard to follow personally, without some examples I can run through the test suite

internal/server/data/model/model.go
```diff
@@ -0,0 +1,17 @@
+//go:build template
+
+package model
+
+type TemplateTable struct{}
```
BruceMacD, line 5: Maybe a comment in here that this is used for file generation would be good

BruceMacD (commented):
We should add a makefile command for this too, its pretty buried so I think that would make it easier for new contributors. We can then call it out in our contributing guide also as part of the workflow

internal/server/data/model/generate.go
```diff
@@ -0,0 +1,223 @@
...
+	for i, decl := range file.Decls {
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Recv == nil {
+			continue
+		}
+		if astIdentForNode(fn.Recv.List[0].Type).Name != rt.Name() {
+			continue
+		}
+
+		if names := fn.Recv.List[0].Names; len(names) > 0 && names[0].Name != "_" {
```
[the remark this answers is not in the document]
dnephin: I think this is the method receiver name, something like this:

```go
func (_ AccessKey) Table() string { ... }
```

Could be underscore in methods that don't use any struct fields, but probably also pretty uncommon.

internal/server/data/model/model.go
dnephin: Comments would be great! I tried to add comments, but the way `go/ast` handles them is super weird. I think it uses some kind of heuristics to decide where to place the comments, and those can get messed up when the target file doesn't have any comments. It was causing really strange output formatting.

I'll try to add some comments to the target file and try to restore comments to this template.

BruceMacD (approved):

dnephin:
I've rebased and moved the package from `internal/server/data/model` to `internal/server/data/table`, so that it's not confused with the `models` package. 

Right now this code only runs from a test, and I expect it to stay that way. By running only from a test we can always preview any changes it makes before the enter `main`.

PR #2764, opened 2022-07-28
data: remove migrations from pre-0.12.0

dnephin:
## Summary

Branched from #2723

This PR removes old database migrations that existed before version 0.12.0. This is being done now in advance of changes to our migration tests, to limit the number of migrations that need to be ported.

The consequence of this change is that someone would have to upgrade to infra 0.12.0, before they can upgrade to infra 0.14.3 (which I assume will be the first version that will include this change).  Since 0.12.0 was pre-launch, I don't think anyone will be impacted by this change.


BruceMacD (approved):
Looks good, but I'm wondering if we have any analytics to confirm no one is on this server version?

dnephin:
I'd like to add something to the release notes to inform people when they are upgrading that they would need to upgrade to a 0.13.x version first. But I think currently release-please makes that difficult without also messing with the release version. So I guess it has to be done post-release.

I'm not sure if we have analytics. I'll look into it.

dnephin:
Ok, I confirmed the oldest running version is 0.12.2, which will be able to upgrade fine with this change.

PR #2765, opened 2022-07-28
maintain: move expected error to trace level

BruceMacD:
## Summary
I just had a hard time debugging a problem with this error message in the debug logs. Moving the log to trace level with some extra context because an error here is expected and means things are working correctly.

## Checklist

### DIFF
--- a/internal/server/providers/oidc.go
+++ b/internal/server/providers/oidc.go
@@ -110,7 +110,7 @@ func (o *oidcClientImplementation) Validate(ctx context.Context) error {
 				return newValidationError("clientSecret")
 			}
 		}
-		logging.Debugf("%s", err.Error())
+		logging.Tracef("validating oidc provider resulted in: %s. this may be expected.", err.Error())
 	}
 
 	// return nil for all other errors, because the request was made with an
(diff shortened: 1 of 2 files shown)


internal/server/providers/oidc.go
```diff
@@ -110,7 +110,7 @@ func (o *oidcClientImplementation) Validate(ctx context.Context) error {
 				return newValidationError("clientSecret")
 			}
 		}
-		logging.Debugf("%s", err.Error())
+		logging.Tracef("validating oidc provider resulted in: %s. this may be expected.", err.Error())
```
dnephin, line 113: Not blocking , but it might be nice to limit the number of helpers we have in `logger.go`, and use the `zerolog` interface directly as much as possible. I think those helpers were mostly to migrate from the previous logger.

```suggestion
		logging.L.Trace().Err(err).Msg("validating oidc provider. An error may be expected.")
```

That allows users to find the error in a predictable place in the structured json output.

We should remove `Tracef` if we go with this approach.

dnephin (approved):

internal/server/providers/oidc.go
BruceMacD: Good suggestion, made this change.

PR #2767, opened 2022-07-28
fix: remove deleted identities from identities_groups

### LINKED ISSUE #2766: users aren't removed from `identities_groups` table when removed from infra
## Describe the bug

Users aren't removed from identities_groups when running `infra users rm user@example.com`, they're only removed if removed from the group itself. This causes `group.UserCount` to be off.

## Expected behavior

Users should be removed from identities_groups when removed from infra, probably through a call to `data.RemoveUsersFromGroup`

## Environment

```
$ infra version
0.14.3+dev
```

```
$ kubectl version
```

j-sneh:
## Summary

Deleted users were still showing up in the `identities_groups` table, as though they still had group membership. This caused `group.UserCount` to be off.

This change removes users from all groups on user deletion and adds a migration to remove previously deleted users from groups.

## Checklist

## Related Issues

Resolves #2766

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -83,6 +83,16 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return fmt.Errorf("delete identity access keys: %w", err)
 	}
 
+	groups, err := data.ListGroups(db, nil, data.ByGroupMember(id))
+	if err != nil {
+		return fmt.Errorf("delete identity from groups: %w", err)
+	}
+	for _, group := range groups {
+		err = data.RemoveUsersFromGroup(db, group.ID, []uid.ID{id})
+		if err != nil {
+			return fmt.Errorf("delete identity from groups: %w", err)
+		}
+	}
 	// if an identity does not have credentials in the Infra provider this won't be found, but we can proceed
 	credential, err := data.GetCredential(db, data.ByIdentityID(id))
 	if err != nil && !errors.Is(err, internal.ErrNotFound) {
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -85,12 +85,12 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 
 	groups, err := data.ListGroups(db, nil, data.ByGroupMember(id))
 	if err != nil {
-		return fmt.Errorf("delete identity from groups: %w", err)
+		return fmt.Errorf("list groups for identity: %w", err)
 	}
 	for _, group := range groups {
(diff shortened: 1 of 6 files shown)


internal/access/identity.go
```diff
@@ -80,6 +80,16 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return fmt.Errorf("delete identity access keys: %w", err)
 	}
 
+	groups, err := data.ListGroups(db, nil, data.ByGroupMember(id))
+	if err != nil {
+		return fmt.Errorf("delete identity from groups: %w", err)
```
BruceMacD, line 85:
```suggestion
		return fmt.Errorf("deleted identity group membership: %w", err)
```

I'd like to have a different message here from the actual error removing the user, so we can differentiate them.

BruceMacD (commented):
Thanks for fixing this, I think adding a check for the group removal to the identity_test would be good too

internal/access/identity_test.go
```diff
@@ -81,6 +82,13 @@ func TestDeleteIdentityCleansUpResources(t *testing.T) {
 	err = data.CreateGrant(db, grantDestination)
 	assert.NilError(t, err)
 
+	err = data.CreateGroup(db, &models.Group{Name: "Group"})
+	assert.NilError(t, err)
+	groups, err := data.ListGroups(db, nil)
+	assert.NilError(t, err)
+	err = data.AddUsersToGroup(db, groups[0].ID, []uid.ID{identity.ID})
```
dnephin, line 89: I think you can get the `group.ID` this way without needing to list:

```suggestion
	group := &models.Group{Name: "Group"}
	err = data.CreateGroup(db, group)
	assert.NilError(t, err)
	err = data.AddUsersToGroup(db, group.ID, []uid.ID{identity.ID})
```

This works because `CreateGroup` sets the ID on the passed in pointer, so we can access it after the call to `CreateGroup`.

internal/server/data/identity.go
```diff
@@ -124,6 +124,18 @@ func DeleteIdentities(db *gorm.DB, selectors ...SelectorFunc) error {
 		if err != nil {
 			return err
 		}
+
+		groups, err := ListGroups(db, nil, ByGroupMember(i.ID))
```
dnephin, line 128: We should add a test for this change as well

dnephin (commented):
Looks great! I think this is just missing test coverage for `access.DeleteIdentities`.

I opened #2838 to resolve the fact that we have two flows that handle this logic in different places. This PR doesn't introduce the problem, so I think we should merge this fix and address it later.

dnephin (approved):

Issue #2768, opened 2022-07-28
Replace gorm.AutoMigrate


dnephin:
In #2584 we identified `gorm.AutoMigrate` is the most significant source of time when running tests against postgreSQL. By replacing `AutoMigrate` we can more easily run all of our tests against postgreSQL.  Removing it will also give us a much clearer picture of the SQL that is being used to construct and modify our database tables.

However, there are two consequence to removing `AutoMigrate`:

1. we will need to write explicit migrations every time we add, or change a table or column
2. we will need to create and manage the initial schema used to populate a database

Units of work:

* [x] convert our migration tests to a [table driven test](https://github.com/gotestyourself/gotest.tools/wiki/Go-Testing-Patterns#table-driven-tests). See section below for more details https://github.com/infrahq/infra/pull/2770
* [x] remove old migrations to limit the number we need to add #2764
* [x] replace migration implementations with regular SQL, especially the ones that use `AutoMigrate`  #2862
* [x] replace the last remaining `data/migrator` use of gorm heleprs (`db.Migrator().HasTable`) #2846
* [x] replace `gorm.AutoMigrate` with an initial schema generated from a database dump that we embed with `go:embed`  #2843

### Table driven tests

A table driven test has a few advantages:

1. it lets us easily compare the list of migrations to the list of test cases for migrations. This gives us the opportunity to prompt the contributor to add a test case when they are adding a migration. If the list of migrations contains an ID that isn't in the list of test cases we can fail the test and indicate they should add a test case.
2. we can run each migration from a database initialized by running all previous migrations. This removes the need to dump the schema for a database every time we write a migration test. We'll only need to dump a new schema when we remove old migrations.
3. we'll also be able to leverage this table test to either dump the fully migrated schema (which we'll embed using `go:embed`) for initializing new databases, or if we decide to keep the schema in the Go source, we can use the test to assert that the migrated schema is identical to the schema stored in the Go source.

PR #2770, opened 2022-07-28
Replace migration tests with a single table drive test

dnephin:
## Summary

The diff is large because this involves adding a new database dump and removing a bunch of previous ones.

This PR replaces individual tests for migrations with a single table test. The table test allows us to check for missing test cases to ensure that all migrations have test coverage. It also allows us to generate the initial schema from the migrated schema (which will be done in a follow up PR).

Many of the tests had to be modified. Some had to change because the fixture data was part of the SQL dump, and others had to change because they relied on `gorm.AutoMigrate` being called at a particular point in time. Calling it now would produce a different schema.

I'll leave comments inline about what had to change.

## Related Issues

Related to #2768


internal/server/data/migrations.go
```diff
@@ -37,6 +37,16 @@ func migrations() []*migrator.Migration {
 				return nil
 			},
 		},
+		// Add roles and migrations to destination table
+		//{
+		//	ID: "2022-05-04T12:13",
```
dnephin, line 42: Previously this was handled by `AutoMigrate`

internal/server/data/migrations.go
```diff
@@ -50,8 +60,8 @@ func migrations() []*migrator.Migration {
 		addAuthURLAndScopeToProviders(),
 		setDestinationLastSeenAt(),
 		deleteDuplicateGrants(),
-		addFieldsFor_0_14(),
 		dropDeletedProviderUsers(),
+		addFieldsFor_0_14(),
```
dnephin, line 64: Normally re-ordering these is not good, but this was added in a not-yet-merged PR, so no risk here.

internal/server/models/types.go
```diff
@@ -37,6 +37,9 @@ func (s CommaSeparatedStrings) Value() (driver.Value, error) {
 }
 
 func (s *CommaSeparatedStrings) Scan(v interface{}) error {
+	if v == nil {
+		return nil
+	}
```
dnephin, line 42: Calling this out. This started failing because I'm using `db.Raw` in the migration queries instead of gorm. This allows us to accept null values from fields for this type.

internal/server/data/migrations.go
BruceMacD: This commented out section is something on your TODO?

internal/server/data/migrations_test.go
```diff
@@ -1,231 +1,336 @@
...
+		}
 
-// this test does an external call to example.okta.com, if it fails check your network connection
-func TestMigration_AddAuthURLAndScopesToProvider(t *testing.T) {
-	for _, driver := range dbDrivers(t) {
-		t.Run(driver.Name(), func(t *testing.T) {
-			db, err := newRawDB(driver)
-			assert.NilError(t, err)
+		opts := migrator.Options{
+			InitSchema: func(db *gorm.DB) error {
```
BruceMacD, line 62: What scenarios could cause an unexpected call to InitSchema?

internal/server/data/testdata/migrations/202204281130-sqlite.sql
```diff
@@ -15,25 +15,26 @@ INSERT INTO migrations VALUES('202203301648');
 INSERT INTO migrations VALUES('202204061643');
 INSERT INTO migrations VALUES('202204111503');
 INSERT INTO migrations VALUES('202204181613');
-CREATE TABLE `groups` (`id` integer,`created_at` datetime,`updated_at` datetime,`deleted_at` datetime,`name` text,`provider_id` integer,PRIMARY KEY (`id`));
-CREATE TABLE `identities_groups` (`group_id` integer,`identity_id` integer,PRIMARY KEY (`group_id`,`identity_id`),CONSTRAINT `fk_identities_groups_group` FOREIGN KEY (`group_id`) REFERENCES `groups`(`id`),CONSTRAINT `fk_identities_groups_identity` FOREIGN KEY (`identity_id`) REFERENCES `identities`(`id`));
+INSERT INTO migrations VALUES('202204211705');
```
BruceMacD, line 18: Why are these two migration files still needed?

internal/server/data/migrations.go
dnephin: Ya, this is what I need to get working

internal/server/data/migrations_test.go
dnephin: If the initial SQL file is not correct and doesn't include the migration IDs. It's unlikely, but we don't have anything else to do here, so I figured an error to indicate there is a bug in the test would be best.

internal/server/data/migrations.go
```diff
@@ -167,7 +166,25 @@ func addAuthURLAndScopeToProviders() *migrator.Migration {
 // #2360: delete duplicate grants (the same subject, resource, and privilege) to allow for unique constraint
 func deleteDuplicateGrants() *migrator.Migration {
 	return &migrator.Migration{ID: "202207081217", Migrate: func(tx *gorm.DB) error {
-		return deleteDuplicates(tx, "subject, resource, privilege", &models.Grant{})
```
dnephin, line 170: I'm not sure why, but this `deleteDuplicates` was not working. The test showed that duplicate records still existed.  I replaced it with raw SQL that I believe matched the original intent, and it works now. So I'm guessing gorm was building the wrong query.

internal/server/data/migrations.go
```diff
@@ -167,7 +166,25 @@ func addAuthURLAndScopeToProviders() *migrator.Migration {
...
+		if err := tx.Exec(stmt).Error; err != nil {
+			return err
+		}
+
+		stmt = `CREATE UNIQUE INDEX idx_grant_srp ON grants
+USING btree (subject, privilege, resource)
+WHERE (deleted_at IS NULL);`
+		if err := tx.Exec(stmt).Error; err != nil {
+			return err
+		}
```
dnephin, line 186: I added the index explicitly here in preparation for removing `gorm.AutoMigrate`. This statement was copied from a DB dump, so it should match exactly what `gorm.AutoMigrate` produces.

internal/server/data/migrations_test.go
```diff
@@ -1,231 +1,363 @@
...
 	"gotest.tools/v3/assert"
 	"k8s.io/utils/strings/slices"
 
 	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/internal/server/data/migrator"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/internal/testing/patch"
 )
 
-func TestMigration_SettingsPopulatePasswordDefaults(t *testing.T) {
```
dnephin, line 20: This test case can't be represented in the table test yet, because we don't have an explicit migration for this. I'll be restoring it in the follow up PR when we add an explicit migration.

dnephin (commented):
Ok, the TODOs are resolved, and tests are passing. I think this is ready for review.

It's unfortunately a big change, but thankfully it's almost exclusively a change to tests. There's only one change to a migration which I comment on below.

BruceMacD (approved):
Looks good to me

internal/server/data/migrations.go
j-sneh: This is the query that I had written out but there was some issues when translating it gorm I guess,

j-sneh: This is the query that I had written out but there was some issues when translating it gorm I guess,

dnephin:
I pushed one more commit which makes a couple small changes to allow the test to continue to run against sqlite. 

We'll be dropping support in the next follow up, but I wanted to make sure the new migration query worked for sqlite until we officially drop support.

PR #2777, opened 2022-07-29
feat: truncate users in `infra groups list`

### LINKED ISSUE #2762: Truncate users in `infra groups list`
The `infra groups list` command outputs each user for every group, but that can get unwieldy if there are a lot of users in a particular group.

We should:
* Automatically truncate on a fixed number of users
* Have a flag to show all users
* Have a configuration setting to allow changing the number of users to show

j-sneh:
## Summary

In order to prevent clutter when running infra groups list, this change limits the number of users displayed for each group (by default 8). For this reason, I also added a `COUNT` column so users can easily see how many users are in a group.

The change also adds two flags: `--no-truncate`, which keeps the old behavior and `--num-users` which changes the limit.

https://github.com/infrahq/infra/pull/2767 should be merged in the same release in order to fix the `COUNT` field. We also should consider adding an easy way to list all users in a group (i.e. `infra users list --group Engineering`)

## Checklist

## Related Issues

Resolves #2762

### DIFF
--- a/docs/reference/cli-reference.md
+++ b/docs/reference/cli-reference.md
@@ -438,6 +438,13 @@ List groups
 infra groups list [flags]
 ```
 
+#### Options
+
+```
+      --no-truncate     Do not truncate the list of users for each group
+      --num-users int   The number of users to display in each group (default 8)
+```
+
 #### Options inherited from parent commands
 
 ```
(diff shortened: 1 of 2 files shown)


BruceMacD (approved):
Looks good 👍  

Not blocking for this change at all, but we are missing test coverage for this command.

PR #2779, opened 2022-07-29
feat: domain support for organizations

ssoroka:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -80,6 +80,9 @@ func (u User) ValidationRules() []validate.ValidationRule {
 }
 
 type Config struct {
+	OrganizationName   string
+	OrganizationDomain string
+
 	Providers []Provider
 	Grants    []Grant
 	Users     []User
@@ -620,6 +623,15 @@ func (s Server) loadConfig(config Config) error {
 			Role:     models.InfraConnectorRole,
 			Resource: "infra",
 		})
+
+		org, err := data.GetOrganization(tx, data.ByName(config.OrganizationName))
+		if errors.Is(err, internal.ErrNotFound) {
+			org = &models.Organization{Name: config.OrganizationName, Domain: config.OrganizationDomain}
+			err = data.CreateOrganization(tx, org)
+		}
+		if err != nil {
+			return err
+		}
 
 		if err := s.loadProviders(tx, config.Providers); err != nil {
 			return fmt.Errorf("load providers: %w", err)
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -603,6 +603,12 @@ func getKindFromUnstructured(data interface{}) string {
 }
 
 func (s Server) loadConfig(config Config) error {
+	if config.OrganizationName == "" {
+		config.OrganizationName = "Default"
+	}
+	if config.OrganizationDomain == "" {
+		config.OrganizationDomain = "localhost"
+	}
 	if err := validate.Validate(config); err != nil {
 		return err
 	}
@@ -630,7 +636,7 @@ func (s Server) loadConfig(config Config) error {
 			err = data.CreateOrganization(tx, org)
 		}
 		if err != nil {
-			return err
+			return fmt.Errorf("loading org: %w", err)
 		}
 
 		if err := s.loadProviders(tx, config.Providers); err != nil {
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -182,3 +182,76 @@ func getCookie(req *http.Request, name string) (string, error) {
 	}
 	return url.QueryUnescape(cookie.Value)
 }
+
+func OrganizationFromDomain(defaultOrgName, defaultOrgDomain string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		db := getDB(c)
+		host := c.Request.Header.Get("Host")
+
+		// remove port
+		if i := strings.Index(host, ":"); i >= 0 {
+			host = host[0:i]
+		}
+
+		if len(host) == 0 {
+			// tests are lazy and don't set Host
+			host = defaultOrgDomain
+		}
+
+		if len(host) > 0 {
+			org, err := data.GetOrganization(db, data.ByDomain(host))
+			if err != nil {
+				if errors.Is(err, internal.ErrNotFound) {
+					org, err = getDefaultOrg(db, defaultOrgName, defaultOrgDomain)
+					if err != nil {
+						sendAPIError(c, fmt.Errorf("creating default organization: %w", err))
+						return
+					}
+					goto success
+				} else {
+					logging.Errorf("fetching org: %s", err)
+				}
+				// failed to find org
+				c.Next()
+				return
+			}
+		success:
+			c.Set("host", host)
+			c.Set("org", org)
+			logging.Debugf("organization set to %s for host %s", org.Name, host)
+		}
+		c.Next()
+	}
+}
+
+func OrganizationRequired() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		o, ok := c.Get("org")
+		if !ok {
+			sendAPIError(c, internal.ErrBadRequest)
+			return
+		}
+		if _, ok = o.(*models.Organization); !ok {
+			sendAPIError(c, internal.ErrBadRequest)
+			return
+		}
+		c.Next()
+	}
+}
+
+// check for configured default org
+func getDefaultOrg(db *gorm.DB, defaultOrgName, defaultOrgDomain string) (*models.Organization, error) {
+	if len(defaultOrgName) == 0 {
+		return nil, errors.New("organization not configured")
+	}
+	org, err := data.GetOrganization(db, data.ByName(defaultOrgName))
+	if err != nil {
+		if errors.Is(err, internal.ErrNotFound) {
+			org = &models.Organization{Name: defaultOrgName, Domain: defaultOrgDomain}
+			return org, data.CreateOrganization(db, org)
+		}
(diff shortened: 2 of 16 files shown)


internal/server/config.go
```diff
@@ -621,6 +624,15 @@ func (s Server) loadConfig(config Config) error {
 			Resource: "infra",
 		})
 
+		org, err := data.GetOrganization(tx, data.ByName(config.OrganizationName))
```
BruceMacD, line 627: This seems like it will act weirdly if the config doesnt specify and organization name or domain. This should be in a check that the fields aren't empty strings.

internal/server/config.go
```diff
@@ -621,6 +624,15 @@ func (s Server) loadConfig(config Config) error {
...
+		if errors.Is(err, internal.ErrNotFound) {
+			org = &models.Organization{Name: config.OrganizationName, Domain: config.OrganizationDomain}
```
BruceMacD, line 629: We should fail if the organization domain has a port (to match the middleware which strips it)

internal/server/middleware.go
```diff
@@ -182,3 +182,76 @@ func getCookie(req *http.Request, name string) (string, error) {
...
+		host := c.Request.Header.Get("Host")
+
+		// remove port
+		if i := strings.Index(host, ":"); i >= 0 {
+			host = host[0:i]
+		}
+
+		if len(host) == 0 {
+			// tests are lazy and don't set Host
+			host = defaultOrgDomain
```
BruceMacD, line 198: Seems like this could be unexpected if the client making the API request doesn't know to set this header value. Should we fail without the host?

internal/server/middleware.go
```diff
@@ -182,3 +182,76 @@ func getCookie(req *http.Request, name string) (string, error) {
...
+					if err != nil {
+						sendAPIError(c, fmt.Errorf("creating default organization: %w", err))
+						return
+					}
+					goto success
+				} else {
+					logging.Errorf("fetching org: %s", err)
+				}
+				// failed to find org
+				c.Next()
```
BruceMacD, line 215: Failing to find the org feels like it should fail, what would happen without the org specified?

internal/server/middleware.go
```diff
@@ -182,3 +182,76 @@ func getCookie(req *http.Request, name string) (string, error) {
...
+}
+
+// check for configured default org
+func getDefaultOrg(db *gorm.DB, defaultOrgName, defaultOrgDomain string) (*models.Organization, error) {
+	if len(defaultOrgName) == 0 {
+		return nil, errors.New("organization not configured")
+	}
+	org, err := data.GetOrganization(db, data.ByName(defaultOrgName))
+	if err != nil {
+		if errors.Is(err, internal.ErrNotFound) {
```
BruceMacD, line 249: Feels like we should be creating this when the server starts, and failing if we dont find it here

internal/server/routes.go
```diff
@@ -56,10 +56,11 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	apiGroup := router.Group("/",
 		metrics.Middleware(promRegistry),
 		DatabaseMiddleware(a.server.db), // must be after TimeoutMiddleware to time out db queries.
+		OrganizationFromDomain(s.options.Config.OrganizationName, s.options.Config.OrganizationDomain),
 	)
 	apiGroup.GET("/.well-known/jwks.json", a.wellKnownJWKsHandler)
 
-	authn := apiGroup.Group("/", authenticatedMiddleware())
+	authn := apiGroup.Group("/", authenticatedMiddleware(), OrganizationRequired())
```
BruceMacD, line 63: We will need to required orgs for some unauthenticated endpoints too (providers will be bound to orgs)

internal/validate/string.go
```diff
@@ -24,6 +24,16 @@ type StringRule struct {
 	CharacterRanges []CharRange
 }
 
+func String(value, name string, min, max int, charset ...CharRange) StringRule {
```
dnephin, line 27: This seems harder to use than the struct literal. Functions with more than 3 arguments generally are. I'd suggest using the struct literal directly.

I'm not seeing this used in this PR.

internal/server/routes.go
dnephin: It would be great to include this as a function call from `authenticatedMiddleware` and `unauthenticatedMiddleware`.  That reduces the number of things that depend on `gin.Context`.

internal/server/middleware.go
```diff
@@ -182,3 +182,76 @@ func getCookie(req *http.Request, name string) (string, error) {
...
 	return url.QueryUnescape(cookie.Value)
 }
+
+func OrganizationFromDomain(defaultOrgName, defaultOrgDomain string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		db := getDB(c)
+		host := c.Request.Header.Get("Host")
+
+		// remove port
+		if i := strings.Index(host, ":"); i >= 0 {
```
dnephin, line 192: This breaks on IPv6 addresses I suspect.

`net.SplitHostPort` seems close to what we want, but that seems to fail if there is no port. Maybe we could wrap it and ignore errors. Assuming an error means no port in the input string.

internal/server/middleware.go
```diff
@@ -182,3 +182,76 @@ func getCookie(req *http.Request, name string) (string, error) {
 	}
 	return url.QueryUnescape(cookie.Value)
 }
+
+func OrganizationFromDomain(defaultOrgName, defaultOrgDomain string) gin.HandlerFunc {
```
dnephin, line 186: I suspect writing this as a regular function (not a middleware) would remove the need for the `goto` and make it easier to follow:

```go
func organizationFromRequest(req *http.Request, defaultOrgName string) (*models.Organization, error) {
   ...
}
```

internal/server/config.go
```diff
@@ -80,6 +80,9 @@ func (u User) ValidationRules() []validate.ValidationRule {
 }
 
 type Config struct {
+	OrganizationName   string
```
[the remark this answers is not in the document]
ssoroka: not sure. It seemed to fit well here as it was needed for server config.

internal/server/config.go
ssoroka: I kind of wanted to make it a required field. Might cause some complications for migrating from older versions. maybe I'll set it to a default value if it's missing

internal/server/config.go
ssoroka: hmm! interesting question. what if you intend to test it with different ports. maybe I need to only strip it if there are no matches.

internal/server/middleware.go
ssoroka: this is set automatically by all clients to whatever the domain is, even if you don't know to set it.

internal/server/middleware.go
ssoroka: There are separate middleware for finding the organization and requiring the organization. I thought maybe some unauthenticated endpoints should work properly with or without an organization, but I'm not 100% true that's the case. Anyway the separation lets you load the org without failing when it's missing. If we don't need it we should combine the middleware

internal/server/routes.go
ssoroka: hmm. I don't really want to start combining middleware to save middleware functions.. We can still switch these to our own context easily enough

internal/validate/string.go
ssoroka: I ended up writing it for something I delayed. I don't find the struct literal very clean looking in use.

BruceMacD (approved):
This seems like a nice first pass to me

internal/server/config.go
```diff
@@ -600,6 +603,12 @@ func getKindFromUnstructured(data interface{}) string {
 }
 
 func (s Server) loadConfig(config Config) error {
+	if config.OrganizationName == "" {
+		config.OrganizationName = "Default"
+	}
+	if config.OrganizationDomain == "" {
+		config.OrganizationDomain = "localhost"
```
jmorganca, line 610: Will this force users to use `localhost` for this org?

internal/server/middleware.go
```diff
@@ -182,3 +183,89 @@ func getCookie(req *http.Request, name string) (string, error) {
...
+
+func OrganizationRequired() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		o, ok := c.Get("org")
+		if !ok {
+			sendAPIError(c, internal.ErrBadRequest)
+			return
+		}
+		if _, ok = o.(*models.Organization); !ok {
+			sendAPIError(c, internal.ErrBadRequest)
```
dnephin, line 247: We also need to check that the org matches the access key here.

I think that would be much easier to do in the `authenticatedMiddleware` rather than a separate one.

internal/server/middleware.go
```diff
@@ -182,3 +183,89 @@ func getCookie(req *http.Request, name string) (string, error) {
...
+			return
+		}
+		c.Next()
+	}
+}
+
+var defaultOrgCache *models.Organization
```
dnephin, line 254: I hope we can avoid this. #1939 is an issue to track removing exactly this pattern from other places we've done it. These package-level static vars are a real risk for testing. They require `Patch` functions to clear one the cache, and it's easy to miss and hide bugs.

Could we cache this on the `Server` type instead? Maybe the defaults and this cache could be part of some type stored as a field on `Server` ?

That would give us a nice pattern for fixing #1939.

We'll need to address some part of that for org support anyway, because currently those caches are not org-aware.

ssoroka:
this is merged into `org_collab` branch. closing

PR #2792, opened 2022-08-01
feat: `infra` shell completion

### LINKED ISSUE #252: `infra` shell completion
Shell completion is a great way to expose command line options and features to the end user.

j-sneh:
## Summary

Adds shell completion using the [default code provided by cobra](https://github.com/spf13/cobra/blob/main/shell_completions.md) which adds static completion of commands. Dynamic completion could be added if wanted.

Note, the command couldn't be auto-generated in order to add it to the `Other commands` group.
## Checklist

## Related Issues

Resolves #252

### DIFF
--- a/internal/cmd/completion.go
+++ b/internal/cmd/completion.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


internal/cmd/completion.go
```diff
@@ -0,0 +1,70 @@
...
+  $ %[1]s completion fish > ~/.config/fish/completions/%[1]s.fish
+
+PowerShell:
+
+  PS> %[1]s completion powershell | Out-String | Invoke-Expression
+
+  # To load completions for every new session, run:
+  PS> %[1]s completion powershell > %[1]s.ps1
+  # and source this file from your PowerShell profile.
+`, `infra`),
```
dnephin, line 52: Could use `os.Args[0]` here in case they have a different name for the binary:

```suggestion
`, os.Args[0]),
```

Probably not a big deal either way.

internal/cmd/completion.go
```diff
@@ -0,0 +1,70 @@
...
+		Short: "Generate completions",
+		Long: fmt.Sprintf(`To load completions:
+
+Bash:
+
+  $ source <(%[1]s completion bash)
+
+  # To load completions for each session, execute once:
+  # Linux:
+  $ %[1]s completion bash > /etc/bash_completion.d/%[1]s
```
dnephin, line 22: I generally put my bash completion at  `~/.bash_complete/` , but I'm not sure how standard that is.

Someone trying to run this command as non-root will get an error, because they don't have permission to write to `/etc`.

Maybe this should use a directory under their home dir, or:

```
$ %[1]s completion bash | sudo tee /etc/bash_completion.d/%[1]s
```

dnephin (commented):
Looks good once the lint errors are fixed. I left some small suggestions.

I'm not sure how standardized it is, but the `$SHELL` environment variable may be set in some environments. I wonder if we should read `$SHELL` by default if they haven't specified a shell.

internal/cmd/completion.go
```diff
@@ -0,0 +1,77 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+)
+
+func newCompletionsCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "completion",
+		Short: "Generate completions",
```
dnephin, line 14:
```suggestion
		Short: "Generate shell auto-completion for the CLI",
```

I had a look at the help text on this branch, and in the list of commands it might not be too obvious what completions means. A few extra words might help put it in context.

dnephin (approved):
LGTM!

PR #2793, opened 2022-08-01
feat: add default org prompt

pdevine:
## Summary

This changes the initial sign-in to ask for an Organization name. The sign-up flow for other organizations will still change after this, but this at least provides an initial org (and is backward compatible to single orgs).

## Checklist

### DIFF
--- a/api/signup.go
+++ b/api/signup.go
@@ -7,12 +7,14 @@ type SignupEnabledResponse struct {
 }
 
 type SignupRequest struct {
+	OrgName  string `json:"orgName"`
 	Name     string `json:"name"`
 	Password string `json:"password"`
 }
 
 func (r SignupRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
+		validate.Required("orgName", r.OrgName),
 		validate.Required("name", r.Name),
 		validate.Required("password", r.Password),
 		validate.Email("name", r.Name),
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -45,7 +45,7 @@ func SignupEnabled(c *gin.Context) (bool, error) {
 
 // Signup creates a user identity using the supplied name and password and
 // grants the identity "admin" access to Infra.
-func Signup(c *gin.Context, name, password string) (*models.Identity, error) {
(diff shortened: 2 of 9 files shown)


api/signup.go
```diff
@@ -7,12 +7,14 @@ type SignupEnabledResponse struct {
...
 
 type SignupRequest struct {
+	OrgName  string `json:"orgName"`
 	Name     string `json:"name"`
 	Password string `json:"password"`
 }
 
 func (r SignupRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
+		validate.Required("orgName", r.OrgName),
```
dnephin, line 17: I think we probably want to `ValidateName` for the org name. That function assumes a field name of `"name"` so we'd have to do something like this:

```
rule := ValidateName(r.OrgName)
rule.Name = "orgName"
```

or refactor `ValidateName` to accept the field name as well.

dnephin (approved):
LGTM!

We might want to add to `access.SignupEnabled` to also check there are no existing orgs. Not strictly necessary because this signup also creates an identity, but it's a nice extra safeguard.

api/signup.go
```diff
@@ -7,12 +7,14 @@ type SignupEnabledResponse struct {
 }
 
 type SignupRequest struct {
+	OrgName  string `json:"orgName"`
```
ssoroka, line 10: I wasn't sure if we should pollute the existing signup process or create a new one. I think we should also be asking for the domain?

internal/access/signup.go
```diff
@@ -54,9 +54,14 @@ func Signup(c *gin.Context, name, password string) (*models.Identity, error) {
 		return nil, err
 	}
 
-	identity := &models.Identity{Name: name}
+	organization := &models.Organization{Name: orgName}
+	if err := data.CreateOrganization(db, organization); err != nil {
+		return nil, err
+	}
```
ssoroka, line 60: I think we need to set the org in the context

ssoroka (approved):
need to set the org from the context in internal/server/authn/oidc.go Authenticate method?

pdevine:
This was sorted out a while ago and the `org_collab` branch isn't needed anymore, so going to close this.

PR #2804, opened 2022-08-02
Add a test case to show that initial schema matches a migrated schema

dnephin:
## Summary

This PR removes the second call to gorm.AutoMigrate, and adds a test to show that the initial schema matches the migrated schema.

From now on all migrations must be explicit, no more auto-creating of columns or constraints using AutoMigrate.

The migration test is updated to dump both the initial schema and the migrated schema, to ensure that they match.

Removing this extra call to AutoMigrate required making a few changes:

* Add migrations that were previously done implicitly by the calls to AutoMigrate.

* For some reason the schema file we use for migrations has composite primary keys and columns in a different order from the initial schema. I guess gorm.AutoMigrate changes the order based on something in the code. I attempted to re-order the list of tables in `initializeSchema` many times, but none of the orders were able to reproduce the exactly correct result. Some problems were fixed, and a couple were not. So for now I made some small changes to the saved schema which allows the test to pass.

* Also re-order the list in initializeSchema to have it match the migrated schema.

* Also re-order some fields in the models struct, so that they match the order in the migrated schema.

## Related Issues

Related to https://github.com/infrahq/infra/issues/2768

### DIFF
--- a/internal/server/data/data_test.go
+++ b/internal/server/data/data_test.go
@@ -52,7 +52,7 @@ func dbDrivers(t *testing.T) []gorm.Dialector {
 		db, err := gorm.Open(pgsql)
 		assert.NilError(t, err, "connect to postgresql")
 		t.Cleanup(func() {
-			assert.NilError(t, db.Exec("DROP SCHEMA testing CASCADE").Error)
+			assert.NilError(t, db.Exec("DROP SCHEMA IF EXISTS testing CASCADE").Error)
 		})
 		assert.NilError(t, db.Exec("CREATE SCHEMA testing").Error)
 
--- a/internal/server/data/data_test.go
+++ b/internal/server/data/data_test.go
@@ -31,50 +31,47 @@ func setupDB(t *testing.T, driver gorm.Dialector) *gorm.DB {
 	return db
 }
 
-// dbDrivers returns the list of database drivers to test.
-// Set POSTGRESQL_CONNECTION to a postgresql connection string to run tests
-// against postgresql.
-func dbDrivers(t *testing.T) []gorm.Dialector {
-	t.Helper()
-	var drivers []gorm.Dialector
+var isEnvironmentCI = os.Getenv("CI") != ""
 
-	tmp := t.TempDir()
-	sqlite, err := NewSQLiteDriver(filepath.Join(tmp, t.Name()))
-	assert.NilError(t, err, "sqlite driver")
-	drivers = append(drivers, sqlite)
-
+func postgresDriver(t *testing.T) gorm.Dialector {
 	pgConn, ok := os.LookupEnv("POSTGRESQL_CONNECTION")
 	switch {
-	case ok:
-		pgsql, err := NewPostgresDriver(pgConn)
-		assert.NilError(t, err, "postgresql driver")
-
-		db, err := gorm.Open(pgsql)
-		assert.NilError(t, err, "connect to postgresql")
-		t.Cleanup(func() {
-			assert.NilError(t, db.Exec("DROP SCHEMA IF EXISTS testing CASCADE").Error)
-		})
-		assert.NilError(t, db.Exec("CREATE SCHEMA testing").Error)
-
-		pgsql, err = NewPostgresDriver(pgConn + " search_path=testing")
-		assert.NilError(t, err, "postgresql driver")
-
-		drivers = append(drivers, pgsql)
-	case os.Getenv("CI") != "":
-		t.Fatalf("CI must test all drivers, set POSTGRESQL_CONNECTION")
+	case !ok && isEnvironmentCI:
+		t.Fatal("CI must test all drivers, set POSTGRESQL_CONNECTION")
+	case !ok:
+		t.Skip("Set POSTGRESQL_CONNECTION to test against postgresql")
 	}
 
-	return drivers
+	pgsql, err := NewPostgresDriver(pgConn)
+	assert.NilError(t, err, "postgresql driver")
+
+	db, err := gorm.Open(pgsql)
+	assert.NilError(t, err, "connect to postgresql")
+	t.Cleanup(func() {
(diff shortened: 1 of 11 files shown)


internal/server/data/migrations.go
```diff
@@ -40,8 +40,7 @@ func migrations() []*migrator.Migration {
 		{
 			ID: "202206081027",
 			Migrate: func(tx *gorm.DB) error {
-				_ = tx.Migrator().DropConstraint(&models.Group{}, "idx_groups_name_provider_id")
-				return nil
+				return tx.Migrator().DropIndex(&models.Group{}, "idx_groups_name_provider_id")
```
dnephin, line 43: This was a bug in one of our existing migrations. I guess we never noticed because we ignored the error.

I fixed it in place. This is an old enough migration that I guess not many people will have this old index. 

We could also create a new migration to ensure that it gets removed.

Maybe I should add `HasIndex` here.

internal/server/data/migrations.go
```diff
@@ -37,10 +37,14 @@ func migrations() []*migrator.Migration {
 				return nil
 			},
 		},
+		// drop old Groups index; new index will be created automatically
 		{
-			ID: "202206081027",
+			ID: "2022-06-08T10:27-fixed",
```
dnephin, line 42: See #2811

I've included the fix here as well to get the tests passing.

internal/server/data/migrations.go
```diff
@@ -222,3 +226,73 @@ func dropDeletedProviderUsers() *migrator.Migration {
...
+ALTER TABLE provider_users DROP COLUMN IF EXISTS id;
+ALTER TABLE provider_users DROP COLUMN IF EXISTS created_at;
+ALTER TABLE provider_users DROP COLUMN IF EXISTS updated_at;
+`).Error; err != nil {
+				return err
+			}
+
+			if !tx.Migrator().HasConstraint("provider_users", "provider_users_pkey") {
+				if err := tx.Exec(`
+ALTER TABLE ONLY provider_users
```
BruceMacD, line 282: I believe this doesn't work on sqlite as it doesn't allow many options for table alteration, if that's the case is that intentional?

internal/server/data/migrations_test.go
```diff
@@ -307,10 +341,28 @@ INSERT INTO provider_users (identity_id, provider_id, id, created_at, updated_at
...
+			initial, err := dumpSchema(os.Getenv("POSTGRESQL_CONNECTION"))
+			assert.NilError(t, err)
+			initialSchema = initial.String()
+
+			assert.NilError(t, db.Exec("DROP SCHEMA IF EXISTS testing CASCADE").Error)
+		})
+	}
+
+	for _, driver := range dbDrivers(t) {
+		if driver.Name() != "postgres" {
```
BruceMacD, line 362: Should we just load the postgres driver directly now instead?

internal/server/models/access_key.go
```diff
@@ -31,6 +30,8 @@ type AccessKey struct {
 	KeyID          string `gorm:"<-;uniqueIndex:idx_access_keys_key_id,where:deleted_at is NULL"`
 	Secret         string `gorm:"-"`
 	SecretChecksum []byte
+
+	Scopes CommaSeparatedStrings // if set, scopes limit what the key can be used for
```
BruceMacD, line 34: Any reason this was moved? Not a big deal in any case.

internal/server/models/group.go
```diff
@@ -12,8 +12,6 @@ type Group struct {
 	CreatedBy         uid.ID
 	CreatedByProvider uid.ID
 
-	Identities []Identity `gorm:"many2many:identities_groups"`
```
BruceMacD, line 15: This still feels like a relation we will want to know (same for identities in an org), whats the reason for removing this?

internal/server/data/migrations.go
dnephin: Ya, good call out. I believe I changed the migration test to only postgres because of this. I should check if we're ready to discontinue sqlite support for production use cases.

internal/server/data/migrations_test.go
dnephin: Ya, I think that would be better. I'll make that change

internal/server/models/access_key.go
dnephin: Good question! Yes these fields on models moved so that the order would match the order that migrations were applied. Otherwise the initial schema does not match the migrated schema because the fields are in different orders.

I had to change a few of the fields in models to move newly added fields to the end of the struct.

internal/server/models/group.go
dnephin: I was struggling to get the initial schema to match the migrated schema. I was trying lots of things, and I don't remember now if this is necessary or just one of the things I tried.

It should be safe to remove this column, it's not used. Pretty soon we'll be writing regular sql that won't need these fields unless we plan on populating them in a query response.

dnephin: I added this back and the test still passes. I'll restore this line for now and we can look to remove it in the future.

internal/server/data/data_test.go
```diff
@@ -31,50 +31,47 @@ func setupDB(t *testing.T, driver gorm.Dialector) *gorm.DB {
...
+// and all supported DBs in CI.
+// Set POSTGRESQL_CONNECTION to a postgresql connection string to run tests
+// against postgresql.
 func runDBTests(t *testing.T, run func(t *testing.T, db *gorm.DB)) {
-	for _, driver := range dbDrivers(t) {
-		t.Run(driver.Name(), func(t *testing.T) {
-			run(t, setupDB(t, driver))
-		})
-	}
+	t.Run("sqlite", func(t *testing.T) {
```
BruceMacD, line 65: Should we remove this sqlite test now that we know of some scenarios that won't work? (ex: table altering migrations).

Im ok with leaving it for convenience though.

dnephin: We still run our other tests (api tests, cli tests, etc) with sqlite, so probably still too early to remove it from the data package tests.

PR #2808, opened 2022-08-03
feat: implement api get/update settings

### LINKED ISSUE #2648: implement a way to retrieve password settings

kimskimchi:
## Summary

Implements:
   
- API:
  - `get settings`
  - `put settings` 

## Checklist

## Related Issues

Resolves #2648


internal/server/settings.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if r.PasswordRequirements {
+		requirements, err := access.GetPasswordRequirements(c)
+		if err != nil {
+			return nil, err
+		}
+		res.PasswordRequirements = requirements
+	}
+
+	return &res, nil
```
BruceMacD, line 22: Should we just return all the setting when something specific isn't specified, rather than an empty response

api/settings.go
```diff
@@ -0,0 +1,17 @@
+package api
+
+type GetSettingsRequest struct {
+	PasswordRequirements bool `form:"passwordRequirements,omitempty"`
```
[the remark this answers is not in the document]
ssoroka: no need to use this. Get should always return all settings.

internal/access/settings.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return api.PasswordRequirements{}, fmt.Errorf("could not get password settings: %w", err)
+	}
+
+	var requirements api.PasswordRequirements
+	requirements.LowercaseMin = settings.LowercaseMin
+	requirements.UppercaseMin = settings.UppercaseMin
+	requirements.NumberMin = settings.NumberMin
+	requirements.SymbolMin = settings.SymbolMin
+	requirements.LengthMin = settings.LengthMin
```
ssoroka, line 42: Combine this with the GetSettings endpoint and build this in a single object, 
```
requirements := &api.Settings{
  PasswordRequirements: api.PasswordRequirements{
    LowercaseMin: settings.LowercaseMin,
    ...
  }
}
```

internal/access/settings.go
```diff
@@ -0,0 +1,63 @@
...
+
+	settings, err := data.GetSettings(db)
+	if err != nil {
+		return err
+	}
+	settings.LengthMin = updatedSettings.LengthMin
+	settings.UppercaseMin = updatedSettings.UppercaseMin
+	settings.LowercaseMin = updatedSettings.LowercaseMin
+	settings.SymbolMin = updatedSettings.SymbolMin
+	settings.NumberMin = updatedSettings.NumberMin
```
ssoroka, line 60: see the ToAPI methods, this could easily be a `SetFromAPI(pr api.PasswordRequirements)` method

internal/server/routes.go
```diff
@@ -116,6 +116,9 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 
 	get(a, noAuthn, "/api/version", a.Version)
 
+	get(a, authn, "/api/settings", a.GetSettings)
```
ssoroka, line 119: getting the settings need to be an unauthenticated endpoint, otherwise you can't get them to log in.

internal/server/settings.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/access"
+	"github.com/infrahq/infra/internal/server/models"
+)
+
+func (a *API) GetSettings(c *gin.Context, r *api.GetSettingsRequest) (*api.Settings, error) {
+	res := api.Settings{}
+
+	if r.PasswordRequirements {
```
ssoroka, line 14: let's drop the conditional here. always return password settings.

internal/server/settings.go
```diff
@@ -0,0 +1,35 @@
...
+		requirements, err := access.GetPasswordRequirements(c)
```
ssoroka, line 15: why is this called GetPasswordRequirements when the underlying call is GetSettings? make these consistent

ssoroka (changes requested):
some consistency issues. see comments.

internal/server/routes.go
kimskimchi: I believe it only needs to be public for the first user during login (as this is the only time an un-authenticated user needs to set a new password, thus needing this endpoint unauthenticated). 

If all other cases are that user MUST be logged in to set a new password (hence authenticated), I thought we could handle the 'unauthenticated user' portion because they would always be the default settings (which only requires min length of 8). They would update these settings after they're setup, which by then, you no longer need to retrieve password settings unauthenticated (?).

api/client.go
```diff
@@ -339,6 +339,17 @@ func (c Client) GetServerVersion() (*Version, error) {
 	return get[Version](c, "/api/version", Query{})
 }
 
+func (c Client) GetSettings(req *GetSettingsRequest) (*Settings, error) {
+	return get[Settings](c, "/api/settings", Query{
+		"passwordRequirements": {strconv.FormatBool(req.PasswordRequirements)},
+	})
+}
+
+func (c Client) UpdateSettings(req *Settings) error {
```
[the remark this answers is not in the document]
kimskimchi: done

internal/server/settings.go
kimskimchi: done

internal/access/settings.go
kimskimchi: done

internal/access/settings.go
kimskimchi: done

internal/access/settings.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	settings.SetFromAPI(updatedSettings)
+	if err = data.SaveSettings(db, settings); err != nil {
+		return nil, err
+	}
+
+	settings, err = data.GetSettings(db)
```
ssoroka, line 55: no need to fetch settings after updating

internal/access/settings.go
```diff
@@ -0,0 +1,60 @@
...
+	return settings.ToAPI(), nil
+}
+
+func SaveSettings(c *gin.Context, updatedSettings *api.Settings) (*api.Settings, error) {
+	db, err := RequireInfraRole(c, models.InfraAdminRole)
+	if err != nil {
+		return nil, HandleAuthErr(err, "settings", "update", models.InfraAdminRole)
+	}
+
+	settings, err := data.GetSettings(db)
```
ssoroka, line 45: no need to fetch settings before updating

BruceMacD (approved):
This looks good to me now

internal/access/settings.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	settings.SetFromAPI(updatedSettings)
```
ssoroka, line 50: set from api belongs in the handler, not the access layer. access should not know about api objects

ssoroka (changes requested):

internal/access/settings.go
kimskimchi: Bruce mentioned to return the object that was updated, should it not be from the source rather than the request?

internal/server/routes.go
kimskimchi: resolved in chat

internal/access/settings.go
ssoroka: all to/from api methods are supposed to go in the handler. it's the responsibility of the handler to translate to/from the request to our internal models.

internal/access/settings.go
ssoroka: since PUT requests completely replace the object, you can just return the incoming object if you need to return something.

kimskimchi: got it, done

internal/access/settings.go
kimskimchi: resolved in chat as well

ssoroka (approved):
Good to go! thanks!

PR #2809, opened 2022-08-03
maintain: update a test to use a struct instead of map

dnephin:
## Summary

Noticed this a few times while reading over the code. Using a struct removes the need to cast to the function type.


BruceMacD (approved):
Thanks

PR #2811, opened 2022-08-03
fix: data migration of group index name

dnephin:
## Summary

I extracted this fix from #2804 so it can be reviewed directly.

This was a bug in one of our existing migrations. I guess we never noticed because we ignored the error. The test that was added in #2804 noticed the problem. The thing being removed is an index, not a constraint.

I fixed this in place, but changed the name of the migration. This will act like a new migration and run for everyone. But it should check if the index exists first, so it will be a no-op for anyone who doesn't have this index.

I'll cherry-pick this fix back into #2804 which has test coverage for it.

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -37,11 +37,11 @@ func migrations() []*migrator.Migration {
 				return nil
 			},
 		},
+		// drop old Groups index; new index will be created automatically
 		{
-			ID: "202206081027",
+			ID: "2022-06-08T10:27-fixed",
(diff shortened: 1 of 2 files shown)


mxyng (approved):

dnephin:
I rebased this PR. Functionally it has not changed, but I did make the following changes:
* fix the merge conflict in migrations.go
* fixed the test failure by updating the migration name in the table test
* replaced the use of `gorm.Migrator` with raw SQL, which removes the need for a second statement

PR #2815, opened 2022-08-03
api: consolidate DatabaseMiddleware into other middleware

### LINKED ISSUE #2697: api: failing to commit a transaction does not result in a 500 response

dnephin:
## Summary

This PR consolidates `DatabaseMiddleware` into the existing `authenticatedMiddleware` and `unauthenticatedMiddleware`. This removes the need for these middleware to fetch the txn from `gin.Context`, they can use the txn directly.

This is the first step to fixing #2697. We'll need to do one of the following next, both of which require this change. Either we
1. use one txn for middleware, and a separate one for handlers. The txn for handlers would be managed from `wrappedHandler`.  This is my preference because it allows us to use different read modes, and isolation modes for the middleware and the handler.
2. use a single txn for both middleware and handlers, which would require moving the `(un)authenticatedMiddleware` into `wrappedHandler`. I think integrating those two is nice either way, but I like keeping separate txn.

## Related Issues

This moves us closer to being able to fix issue #2697
Related to https://github.com/infrahq/internal/issues/19

### DIFF
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -36,20 +36,6 @@ func TimeoutMiddleware(timeout time.Duration) gin.HandlerFunc {
 	}
 }
 
-// DatabaseMiddleware injects a `db` object into the Gin context.
-func DatabaseMiddleware(db *gorm.DB) gin.HandlerFunc {
-	return func(c *gin.Context) {
-		err := db.WithContext(c.Request.Context()).Transaction(func(tx *gorm.DB) error {
-			c.Set("db", tx)
-			c.Next()
-			return nil
-		})
-		if err != nil {
-			logging.Debugf(err.Error())
-		}
-	}
-}
-
 func handleInfraDestinationHeader(c *gin.Context) error {
 	uniqueID := c.Request.Header.Get("Infra-Destination")
 	if uniqueID == "" {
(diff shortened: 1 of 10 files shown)


internal/server/middleware.go
```diff
@@ -81,42 +67,61 @@ func handleInfraDestinationHeader(c *gin.Context) error {
...
-func authenticatedMiddleware() gin.HandlerFunc {
+func authenticatedMiddleware(db *gorm.DB) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		db := getDB(c)
-		authned, err := requireAccessKey(db, c.Request)
-		if err != nil {
-			sendAPIError(c, err)
-			return
-		}
+		withDBTxn(c.Request.Context(), db, func(tx *gorm.DB) {
```
dnephin, line 75: It's unfortunate this adds another level of indentation.

In the future, when we remove gorm, we'll have to replace this with calls to `Begin`, `Commit`, and `Rollback`. I expect that to change this code a little bit, but probably not too much.

internal/server/routes.go
```diff
@@ -268,15 +265,14 @@ type WellKnownJWKResponse struct {
 }
 
 func (a *API) wellKnownJWKsHandler(c *gin.Context) {
-	keys, err := access.GetPublicJWK(c)
+	rCtx := getRequestContext(c)
```
dnephin, line 268: We could further integrate this `wellKnownJWKsHandler` and make it one of our `HandlerFunc`. That's going to be a prerequisite for other changes, so I might try and do that here.

I also need to check if we have test coverage for this endpoint.

dnephin: I did this in the latest commit.

BruceMacD (commented):
The changes seem ok but I'm a bit confused about the actual scenario this is an issue. We debug log the error in the database middleware, but I am not sure what kind of errors we would get that aren't handled by the authenticated middleware `sendAPIError`

dnephin:
> I am not sure what kind of errors we would get that aren't handled by the authenticated middleware

The transaction commit can return an error, and right now we hide it in debug logs, where as we should be returning it as an error to the user (https://github.com/infrahq/infra/issues/2697). If the transaction commit failed, then the entire operation failed because nothing was applied to the DB.

BruceMacD (approved):

Issue #2801, opened 2022-08-02
New Org Sign-Up 


BruceMacD:
## Is your feature request related to a problem? Please describe.
To sign-up a new organization in a multi-tenant deployment it should be possible to register a new organization and specify an admin user for that organization. 

## Describe the solution you'd like
Rather than the `/signup` endpoint only working for the initial Infra deployment and setup. It should be possible to use this endpoint to register more organizations when configured.

## Tasks
- [x] #2807
- [ ] Update quick-start docs for orgs
- [x] #2822

PR #2816, opened 2022-08-03
Org collab signup

### LINKED ISSUE #2807: Add orgs to sign-up endpoint

BruceMacD:
## Summary
Add orgs to Infra sign-up. This is an incremental change to unblock other orgs work (note that it is going into the `org_collab` branch, not main).

### In This Change
- 36fbd80c92394a58bf47e9ad631d925be2d1f887 Deprecate the GET `/signup` endpoint. It will always return false to allow old CLI client to continue to function (which rely on this endpoint's response).
- 103c9a3159bc5848eaf8c99cdbe5fbb3abadd3f0 Create an org on sign-up, assign an admin user, and create appropriate grants.

### Next steps 
#2801 
- Update the sign-up UI
- Update the quickstart to direct new self-hosted deployments to the "sign-up" UI
- Update ingress when a new org subdomain is created
- https://github.com/infrahq/infra/issues/2806

## Checklist

## Related Issues

Resolves #2807


internal/access/signup.go
```diff
@@ -11,50 +11,29 @@ import (
...
-
-	err := checkPasswordRequirements(db, password)
+	err = checkPasswordRequirements(db, password)
 	if err != nil {
 		return nil, err
 	}
 
-	identity := &models.Identity{Name: name}
+	identity := &models.Identity{
+		Model: models.Model{
```
BruceMacD, line 32: Is there a simpler way to set the org ID in the current implementation than this?

api/signup.go
```diff
@@ -2,19 +2,17 @@ package api
...
 	Name     string `json:"name"`
 	Password string `json:"password"`
+	Org      string `json:"org"`
 }
 
 func (r SignupRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
 		validate.Required("name", r.Name),
 		validate.Required("password", r.Password),
+		validate.Required("org", r.Org),
```
BruceMacD, line 15: I'll separate this out into orgName and orgDomain, it's just not in this branch yet from #2779.

pdevine: Yeah, originally I was thinking just Org Name, but this makes more sense.

internal/access/signup.go
```diff
@@ -11,50 +11,29 @@ import (
...
-		return false, err
-	}
-
-	grants, err := data.Count[models.Grant](db.Unscoped(), data.NotPrivilege(models.InfraConnectorRole))
-	if err != nil {
-		return false, err
-	}
-
-	accessKeys, err := data.Count[models.AccessKey](db.Unscoped())
+	org := &models.Organization{Name: orgName}
```
pdevine, line 20: We should add the domain to the org model as well.

pdevine (approved):

PR #2820, opened 2022-08-03
improve(ui): use axios for xhr requests

jmorganca:
This change proposes using `axios` instead of fetch in the UI for XHR requests. Axios has better built-in defaults, error handling, and overall requires writing less code.


jmorganca:
Will close, rebase and re-open this at a later date

PR #2843, opened 2022-08-05
Replace gorm.AutoMigrate with schema.sql

dnephin:
## Summary

Branched from #2804

This PR replaces `gorm.AutoMigrate` with a `schema.sql` file that is included using `go:embed`. We don't support `sqlite` migrations , so we are still using `gorm.AutoMigrate` to create the schema for testing against `sqlite`. In a follow up I'll explore options for removing that as well.

The `schema.sql` file is generated by the schema produced by running migrations. Similar to our existing golden tests, when a new migration is added the `schema.sql` can be updated by running:

```
go test -run TestMigrations ./internal/server/data -update
```

The test will inform the user of this workflow when it fails.

## Related Issues

Related to #2768


internal/server/data/schema.sql
```diff
@@ -0,0 +1,328 @@
+-- SQL generated by TestMigrations DO NOT EDIT.
+-- Instead of editing this file, add a migration to ./migrations.go and run:
+--
+--     go test -run TestMigrations ./internal/server/data -update
+--
+
+CREATE TABLE access_keys (
+    id bigint NOT NULL,
+    created_at timestamp with time zone,
```
BruceMacD, line 9: Are these timestamps in the current schema? I thought they were unix time, I could be wrong though.

dnephin: Ya, `gorm.AutoMigrate` produces this column type for `time.Time`. In sqlite it uses `datetime`.

internal/server/data/migrations.go
```diff
@@ -56,7 +57,21 @@ func migrations() []*migrator.Migration {
 	}
 }
 
+//go:embed schema.sql
+var schemaSQL string
+
 func initializeSchema(db *gorm.DB) error {
+	if db.Dialector.Name() == "sqlite" {
```
BruceMacD, line 64: Do we still need this sqlite automigrate where some of the migrations wont work on sqlite anyway?

BruceMacD (approved):

internal/server/data/migrations.go
dnephin: Probably not, but it wasn't much work and I'm trying to keep some support for sqlite where its easy.

PR #2845, opened 2022-08-05
Support running all tests against postgres

dnephin:
## Summary

Each test package needs to use a different schema name because `go test` will run tests for different packages in parallel. Since all packages can use the same PostgreSQL DB server they have to target different schemas. 

For some reason the `cmd` tests also seem to overlap, and I'm not sure why. It's possible the server processes keep running after the test ends, but logging suggests this isn't the case. For now I've worked around that by using a different schema name for each test function in `cmd`.

Best viewed by individual commit

Test failures included:
* [x] the database connection is not properly closed
* [x] tests don't call `data.InvalidateCache`
* [x] tests don't work with foreign key constraints
* [x] tests use different DB connections
* [x] infinite recursion in `TestDBTimeout`
* [x] `internal/cmd` tests still fail (but pass if a different schema name is used)
* [x] why do some test failures cause following tests to fail. It seems like cleanup is not done correctly somehow

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -178,6 +178,15 @@ func (s *Server) Run(ctx context.Context) error {
 
 	err := group.Wait()
 	s.tel.Close()
+
+	if sqlDB, err := s.db.DB(); err != nil {
+		logging.L.Warn().Err(err).Msg("failed to get database conn to close")
+	} else {
+		if err := sqlDB.Close(); err != nil {
+			logging.L.Warn().Err(err).Msg("failed to close database connection")
+		}
+	}
+
 	if errors.Is(err, context.Canceled) {
 		return nil
 	}
@@ -330,7 +339,7 @@ func (s *Server) getPostgresConnectionString() (string, error) {
 		}
 
 		if s.options.DBParameters != "" {
-			fmt.Fprintf(&pgConn, "%s", s.options.DBParameters)
+			fmt.Fprint(&pgConn, s.options.DBParameters)
 		}
 	}
 
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -14,6 +14,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/infrahq/secrets"
 	"golang.org/x/sync/errgroup"
+	"gorm.io/driver/postgres"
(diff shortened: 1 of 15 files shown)


dnephin:
The tests are passing now, so I guess this is ready for review. There is still some strange overlapping of tests I can't explain, so I'll probably wait to merge this for a bit to make sure it doesn't cause flaky test failures.

internal/server/server.go
```diff
@@ -178,6 +178,15 @@ func (s *Server) Run(ctx context.Context) error {
 
 	err := group.Wait()
 	s.tel.Close()
+
+	if sqlDB, err := s.db.DB(); err != nil {
+		logging.L.Warn().Err(err).Msg("failed to get database conn to close")
+	} else {
+		if err := sqlDB.Close(); err != nil {
+			logging.L.Warn().Err(err).Msg("failed to close database connection")
+		}
```
ssoroka, line 187: this might cause the server to hang. Looks like some of the docs around this say:
```
// Close closes the database and prevents new queries from starting.
// Close then waits for all queries that have started processing on the server
// to finish.
```
we might want a timeout around this so we can ensure the app closes in a reasonable amount of time.

ssoroka (approved):
seems good. too bad about the multiple schemas. not sure what's going on there.

internal/server/server.go
dnephin: Waiting for queries to finish seems like a nice behaviour. We should have timeouts on all queries since we set a deadline on the transaction, right? I'd expect queries to have a pretty short timeout.

I don't think we can set a timeout here because there's no `context.Context` being passed in. In the worst case, kubernetes is going to have its own "graceful shutdown timeout", and if the pod hands after SIGTERM, kubernetes should SIGKILL it, which gives us basically the preivous behaviour of no graceful shutdown.

Is there another documented way of setting a timeout to `Close`?

dnephin: Oh, I forgot we don't even have signal handlers yet, so this code will never be reached in production, it's only used in tests.

When we have signal handlers I think we'll want to make sure a second `SIGTERM` or `SIGINT` does a forced shutdown to avoid a potential hang anywhere in the app.

PR #2846, opened 2022-08-05
Replace gorm.Migrator calls with sql

dnephin:
## Summary

`gorm.Migrator` (not to be confused with `gormigrate` which we already replaced) provides helper functions for inspecting and changing the schema.

This PR replaces calls to those helpers with regular SQL. In a couple cases (`HasTable`, `HasColumn`) I added a helper function to the `data/migrator` package, because these two require inspecting the informational schema and are a bit more involved.

This PR keeps sqlite support by writing the equivalent sqlite queries (which are not much different), but it is important to note that the sqlite versions of a couple of these are not idempotent.

The only remaining use of `gorm.Migrator` is `AutoMigrate`, which will be replaced next.

## Related Issues

Related to #2768

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -122,7 +122,13 @@ func dropCertificateTables() *migrator.Migration {
 	return &migrator.Migration{
 		ID: "202206161733",
 		Migrate: func(tx *gorm.DB) error {
-			return tx.Migrator().DropTable("trusted_certificates", "root_certificates")
+			if err := tx.Exec(`DROP TABLE IF EXISTS trusted_certificates`).Error; err != nil {
+				return err
+			}
+			if err := tx.Exec(`DROP TABLE IF EXISTS root_certificates`).Error; err != nil {
+				return err
+			}
+			return nil
 		},
 	}
 }
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -22,20 +22,22 @@ func migrations() []*migrator.Migration {
 		{
 			ID: "202204281130",
 			Migrate: func(tx *gorm.DB) error {
-				return tx.Migrator().DropColumn(&models.Settings{}, "signup_enabled")
+				stmt := `ALTER TABLE settings DROP COLUMN IF EXISTS signup_enabled`
+				if tx.Dialector.Name() == "sqlite" {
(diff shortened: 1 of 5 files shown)


internal/server/data/migrator/helpers.go
```diff
@@ -0,0 +1,57 @@
...
+	`
+
+	if tx.Dialector.Name() == "sqlite" {
+		stmt = `
+			SELECT count(*)
+			FROM sqlite_master
+			WHERE type = 'table' AND name = ?
+			AND sql LIKE ?
+		`
+		column = "%`" + column + "`%"
```
dnephin, line 50: I based this on the gorm migrator implementation we're using now: https://github.com/go-gorm/sqlite/blob/v1.3.6/migrator.go#L70-L71

The `sqlite_master` table stores the `CREATE TABLE` statement as sql, so we have to use substring matching to find the column.

The gorm implementation has a lot more `sql LIKE ?` parts of the query. My guess is that different sqlite implementations might use different quoting for column names. This works for our existing use cases, and since we won't support sqlite in production, this seems sufficient for our use case.

internal/server/data/migrations.go
```diff
@@ -185,14 +194,17 @@ func setDestinationLastSeenAt() *migrator.Migration {
 	return &migrator.Migration{
 		ID: "202207041724",
 		Migrate: func(tx *gorm.DB) error {
-			if tx.Migrator().HasColumn(&models.Destination{}, "last_seen_at") {
+			if migrator.HasColumn(tx, "destinations", "last_seen_at") {
 				return nil
 			}
 
-			if err := tx.Migrator().AddColumn(&models.Destination{}, "last_seen_at"); err != nil {
+			stmt := `ALTER TABLE destinations ADD COLUMN last_seen_at timestamp with time zone`
```
ssoroka, line 201: This is good, I prefer the timezone in the db, but do you know if other tables in postgres have been getting `timestamp with time zone`? might cause subtle problems somewhere if they're different.

ssoroka (approved):
lgtm

internal/server/data/migrations.go
dnephin: Ya, this matches what `gorm.AutoMigrate` produces, you can see that here: https://github.com/infrahq/infra/pull/2843/files#r940287138

PR #2862, opened 2022-08-08
Replace the last of AutoMigrate

### LINKED ISSUE #2768: Replace gorm.AutoMigrate

dnephin:
## Summary

Branched from #2845

This PR replaces the last of the calls to `gorm.AutoMigrate` and `gorm.Migrator`. The only remaining calls to `gorm.AutoMigrate` are in a function we are using for sqlite compatibility, and to catch any new implicit migrations that might be added from other PRs.  I'm thinking of leaving those in until this merges (to catch problems during merge and rebase), and removing it immediately in a follow up.

## Related Issues

Resolves #2768

### DIFF
--- a/internal/server/data/migrator/helpers.go
+++ b/internal/server/data/migrator/helpers.go
@@ -1,6 +1,8 @@
 package migrator
 
-import "gorm.io/gorm"
+import (
+	"gorm.io/gorm"
+)
 
 type Tx interface {
 	Exec(stmt string, args ...any) *gorm.DB
@@ -47,7 +49,7 @@ func HasColumn(tx *gorm.DB, table string, column string) bool {
 			WHERE type = 'table' AND name = ?
 			AND sql LIKE ?
 		`
-		column = "%`" + column + "`%"
+		column = "% " + column + " %"
 	}
 
 	if err := tx.Raw(stmt, table, column).Scan(&count).Error; err != nil {
--- a/internal/server/data/migrator/helpers.go
+++ b/internal/server/data/migrator/helpers.go
@@ -55,3 +55,30 @@ func HasColumn(tx *gorm.DB, table string, column string) bool {
 	}
 	return count != 0
 }
+
+// HasConstraint returns true if the database table has the constraint. Returns
+// false if the database table does not have the constraint, or if there was a
+// failure querying the database.
+func HasConstraint(tx *gorm.DB, table string, constraint string) bool {
+	var count int
+	stmt := `
+		SELECT count(*)
+		FROM information_schema.table_constraints
+		WHERE table_schema = CURRENT_SCHEMA()
+		AND table_name = ? AND constraint_name = ?
+	`
+	if tx.Dialector.Name() == "sqlite" {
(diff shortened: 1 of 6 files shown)


internal/server/data/migrator/helpers.go
```diff
@@ -55,3 +55,30 @@ func HasColumn(tx *gorm.DB, table string, column string) bool {
...
+			SELECT count(*)
+			FROM sqlite_master
+			WHERE type = 'table' AND tbl_name = ?
+			AND sql LIKE ?
+		`
+		constraint = "%CONSTRAINT `" + constraint + "`%"
+	}
+
+	if err := tx.Raw(stmt, table, constraint).Scan(&count).Error; err != nil {
+		return false
```
BruceMacD, line 81: Is there a reason to suppress this error rather than fail?

dnephin: Good question! My reasoning was:

1. this is what gorm did, so matching that behaviour should be fine if it was working previously
2. the only way this could fail in production would be a runtime error because the database is unavailable. If the database is unavailable then it will return false, which will cause the migration to run, and that error should be the same connection error as we would have seen here.

I think that second statement would be a lot safer if we had tests specifically for these function. I can add those.

BruceMacD (approved):

PR #2863, opened 2022-08-08
feat: endpoint to check server configuration

### LINKED ISSUE #2818: backend: expose important config to frontend
some config needs to be exposed to the user. This includes org-specific settings like "what are the org's password requirements for users?", but also some server-specific settings like "Is email configured?"

Create an unauthenticated endpoint (perhaps `GET /api/configuration`) that exposes non-sensitive information the client will need. In this case, a setting such as `{isEmailConfigured: true}`

ssoroka:
## Summary

expose server configuration to the frontend so it can make intelligent UI choices. 

## Checklist

## Related Issues

Resolves #2818

### DIFF
--- a/internal/server/configuration.go
+++ b/internal/server/configuration.go
@@ -1,0 +1,14 @@
+package server
+
+import (
+	"github.com/gin-gonic/gin"
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/server/email"
+)
+
+func (a *API) GetServerConfiguration(c *gin.Context, _ *api.EmptyRequest) (*api.ServerConfiguration, error) {
+	return &api.ServerConfiguration{
+		IsEmailConfigured: email.IsConfigured(),
+	}, nil
+}
--- a/internal/server/testdata/openapi3.json
+++ b/internal/server/testdata/openapi3.json
@@ -912,6 +912,13 @@
           }
         }
       },
+      "ServerConfiguration": {
+        "properties": {
+          "isEmailConfigured": {
+            "type": "boolean"
+          }
+        }
+      },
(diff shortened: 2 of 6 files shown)


jmorganca:
Great stuff. This may become doubly useful for feature-flagging upcoming features.

internal/server/configuration.go
```diff
@@ -0,0 +1,14 @@
+package server
+
+import (
+	"github.com/gin-gonic/gin"
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/server/email"
+)
+
+func (a *API) GetServerConfiguration(c *gin.Context, _ *api.EmptyRequest) (*api.ServerConfiguration, error) {
```
BruceMacD, line 10: I feel like this should be part of the `/api/settings` endpoint, its a bit weird to have both `/api/settings` and `api/server-config`. They sound the same to me.

ssoroka: one is server-specific configuration, one is org-specific settings.

BruceMacD: `/settings` is only for password requirements at the moment, which will be org specific also

ssoroka: that's the org-specific one

internal/server/testdata/openapi3.json
```diff
@@ -4489,6 +4499,91 @@
         ]
       }
     },
+    "/api/server-config": {
```
pdevine, line 4502: Still need to update this route.

pdevine (approved):
Totally went back and forth on the API route, but I think `/api/server-configuration` probably is the best bet. Looks like the swagger still needs to be updated.

PR #2866, opened 2022-08-08
fix: make infra header optional for well-known jwks

### LINKED ISSUE #2865: regression: `/.well-known/jwks.json` requires `Infra-Version`
## Describe the bug

The endpoint for `/.well-known/jwks.json` now requires the header `Infra-Version` which it didn't previously. If the header is missing, the request fails. This is a problem because the connector doesn't send this header when querying the endpoint and so can't verify the tokens created by the server. The end result is the user is not able to access their cluster using infra.

```
$ http --verify no https://localhost:10443/.well-known/jwks.json
HTTP/1.1 400 Bad Request
Content-Length: 105
Content-Type: application/json; charset=utf-8
Date: Mon, 08 Aug 2022 20:26:10 GMT

{
    "code": 400,
    "message": "bad request: Infra-Version header is required. The current version is 0.14.3-dev"
}
```

```
$ http --verify no https://localhost:10443/.well-known/jwks.json Infra-Version:0.1.2
HTTP/1.1 200 OK
Content-Length: 171
Content-Type: application/json; charset=utf-8
Date: Mon, 08 Aug 2022 20:26:16 GMT

{
    "keys": [
        {
            "alg": "ED25519",
            "crv": "Ed25519",
            "kid": "0g6I3uLociKcmO7Z9N4_MLy9X81VqG-bIJTXlIZi3p8=",
            "kty": "OKP",
            "use": "sig",
            "x": "Y95aVyXoezIZYPcth6SQtWQ0X9_Gb7naao4V2V9ad1U"
        }
    ]
}
```

## Expected behavior

- `/.well-known/jwks.json` does not require `Infra-Version`
- The connector can verify server-created tokens
- The user can access their cluster using infra

## Environment

```
$ infra version

 Client: 0.14.3-dev
 Server: 0.14.3-dev

```

dnephin:
## Summary

Fixes a regression from #2815

Resolves #2865

### DIFF
--- a/internal/server/handlers_test.go
+++ b/internal/server/handlers_test.go
@@ -178,7 +178,6 @@ func TestWellKnownKWKs(t *testing.T) {
 	run := func(t *testing.T, tc testCase) {
 		req, err := http.NewRequest(http.MethodGet, "/.well-known/jwks.json", nil)
 		assert.NilError(t, err)
-		req.Header.Set("Infra-Version", apiVersionLatest)
 
 		resp := httptest.NewRecorder()
 		routes.ServeHTTP(resp, req)
--- a/internal/server/routes.go
+++ b/internal/server/routes.go
@@ -114,11 +114,12 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	get(a, noAuthn, "/api/settings", a.GetSettings)
 	put(a, authn, "/api/settings", a.UpdateSettings)
 	add(a, noAuthn, route[api.EmptyRequest, WellKnownJWKResponse]{
-		method:            http.MethodGet,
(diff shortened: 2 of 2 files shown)


internal/server/handlers_test.go
```diff
@@ -178,7 +178,6 @@ func TestWellKnownKWKs(t *testing.T) {
 	run := func(t *testing.T, tc testCase) {
 		req, err := http.NewRequest(http.MethodGet, "/.well-known/jwks.json", nil)
 		assert.NilError(t, err)
-		req.Header.Set("Infra-Version", apiVersionLatest)
```
dnephin, line 181: Removing this from the test reproduces the issue.

mxyng (approved):

PR #2867, opened 2022-08-08
improve: separate ci workflows

mxyng:
## Summary

update CI to only build when necessary

### DIFF
--- a/.github/workflows/ci-commit.yaml
+++ b/.github/workflows/ci-commit.yaml
@@ -1,0 +1,15 @@
+name: ci / commit
+
+on:
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+      - run: npm install -g @commitlint/{config-conventional,cli}
+      - run: npx commitlint --from origin/$GITHUB_BASE_REF --to @ --verbose --config .github/commitlint.config.js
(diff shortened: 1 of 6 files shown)


.github/workflows/ci/helm.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  pull_request:
+    paths:
+      - 'helm/**'
+  push:
+    branches:
+      - main
+    paths:
+      - 'helm/**'
```
BruceMacD, line 11: This means it only runs when the helm chart actually changes? Nice.

.github/workflows/ci/ui.yaml
```diff
@@ -0,0 +1,77 @@
...
+          node-version: 18
+          cache: npm
+          cache-dependency-path: ui/package-lock.json
+      - uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: ${{ runner.os }}-npm-
+      - name: Dependencies
+        run: npm install
```
BruceMacD, line 54: Not necessary to be part of this PR, but I wonder if we can run all these tests on the same image so that we don't have to `npm install` 3 separate times, which looks like it may re-install on the dependencies on a fresh image each time

BruceMacD (approved):

.github/workflows/ci/ui.yaml
mxyng: That's a valid point. The cache step above should help alleviate some of those concerns

.github/workflows/ci/helm.yaml
mxyng: That's right. One issue with the CI workflow previously is it unconditionally runs CI for all components regardless if there are changes to those components.

PR #2869, opened 2022-08-08
improve: add helm cd workflow

mxyng:
## Summary

Separate out Helm build/release from code build/release

### DIFF
--- a/.github/workflows/cd/helm.yaml
+++ b/.github/workflows/cd/helm.yaml
@@ -1,0 +1,52 @@
+name: cd/helm
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'helm/**'
+  workflow_dispatch:
+    inputs:
+      OVERRIDE_EXISTING:
+        type: boolean
+        default: false
+        required: false
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/cd/helm.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - run: |
+          for chart in *.tgz; do
+            if aws s3 ls s3://${{ env.HELM_BUCKET }}/$chart && ! ${{ inputs.OVERRIDE_EXISTING }} ; then
+              echo "Skipping $chart which already exists."
+              continue
+            fi
+
+            echo "Syncing $chart to ${{ env.HELM_BUCKET }}."
+            mkdir -p ${chart%%.tgz}
+            cp -a $chart ${chart%%.tgz}
```
mxyng, line 38: This is a hack. `helm repo index` requires a directory so if there are multiple charts in a directory, everything will be indexed.

BruceMacD, line 38:
```suggestion
            cp -a $chart ${chart%%.tgz} # hack to index multiple charts in one directory
```

BruceMacD (approved):

BruceMacD (approved):

PR #2873, opened 2022-08-09
improve: add core and ui cd workflows

mxyng:
## Summary

Build/publish UI and core services separately.


.github/workflows/cd/core.yaml
```diff
@@ -0,0 +1,128 @@
...
+          echo "::set-output name=release-name::${RELEASE_NAME#v}"
+
+  build:
+    runs-on: ubuntu-latest
+    needs: [release-name]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
BruceMacD, line 33:
```suggestion
          go-version: 1.18.5
```

Not necessary, but nice to update to the latest version while we are in the neighborhood

BruceMacD (approved):

.github/workflows/cd/core.yaml
j-sneh: Is there a reason not to change to `1.19`?

mxyng: This is pinning the minor version so it can freely upgrade the patch version. Right now, 1.8 and 1.8.5 are equivalent in this context but if/when 1.8.6 is release, using 1.8 will automatically update to 1.8.6 while 1.8.5 will not.

As for upgrading to 1.9, I haven't gone through the release in detail so not sure what changes, if any, are needed so it's out of scope of this PR.

jmorganca (approved):

PR #2882, opened 2022-08-09
fix: use unique org db context key

BruceMacD:
- appease the lint gods

## Summary
Should use a unique context key for the org db context rather than a string to ensure uniqueness. 

## Checklist

### DIFF
--- a/internal/access/access_test.go
+++ b/internal/access/access_test.go
@@ -33,7 +33,7 @@ func setupDB(t *testing.T) *gorm.DB {
 	err = data.CreateOrganization(db, org)
 	assert.NilError(t, err)
 
-	db.Statement.Context = context.WithValue(db.Statement.Context, "org", org)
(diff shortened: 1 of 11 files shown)


ssoroka (approved):
data is not the right package, but this will all change soon anyway.

internal/access/access_test.go
```diff
@@ -33,7 +33,7 @@ func setupDB(t *testing.T) *gorm.DB {
 	err = data.CreateOrganization(db, org)
 	assert.NilError(t, err)
 
-	db.Statement.Context = context.WithValue(db.Statement.Context, "org", org)
+	db.Statement.Context = context.WithValue(db.Statement.Context, data.OrgCtxKey{}, org)
```
dnephin, line 36: The goal of the unique context key is to prevent other packages from being able to set the field.

So I think we'd also want a `WithOrg`, and `OrgFromContext` functions in `data` to hide this.

PR #2888, opened 2022-08-09
Use SQL for access key queries

dnephin:
## Summary

Branched from #2761

Started converting queries to sql. Still a few access keys queries to complete before making this ready for review.

## Related Issues

Related to #2415


dnephin:
Replaced by #3029

PR #2889, opened 2022-08-09
improve: refactor release workflow

mxyng:
## Summary

Add a release workflow which will create a Github release based on some inputs. Release notes are populated from `docs/changelog/*.md`. This action uses `workflow_dispatch` which can be triggered manually from the Github Actions UI. The deploy itself is pretty simple. It uses `infra` to setup a Kubernetes context and ArgoCD Image Updater to update the image backing the deployment. ArgoCD Image Updater can piggyback on the context `infra` creates so no additional configurations is required. 

This also adds `workflow_run` which allows the configured workflows to trigger off other workflows. It's particularly useful for build and deploy which can trigger of CI tests and CD builds respectively.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          cache: true
+      - run: |
+          error() { echo ">>> $*"; exit 1; }
+
+          npm install -g semver
+          RELEASE_NAME=$(semver ${{ inputs.RELEASE_NAME }})
```
mxyng, line 25: Another way I've done this instead of specifying the new version directly, specify a bump level, `patch`, `minor`, or `major`, and use semver to increment the version based on a previous version. It's slightly less manual but interferes with the changelog slightly which looks for a specific version

.github/actions/setup-argocd-image-updater/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    required: true
+    default: latest
+  argocd-server:
+    description: ArgoCD server URL
+    required: false
+  argocd-token:
+    description: ArgoCD server authentication token
+    required: false
+outputs:
+  argocd-image-updater-version:
```
BruceMacD, line 15: This should never differ from the input should it? Is there a reason to output this?

.github/actions/setup-argocd-image-updater/index.js
```diff
@@ -0,0 +1,123 @@
+const path = require('path');
```
BruceMacD, line 1: This file is a copy/paste correct?

.gitignore
```diff
@@ -4,6 +4,6 @@
 .vscode
 
 # infra binary
-dist
-docs/api/openapi3.json
```
BruceMacD, line 8: This intentionally removed from the gitignore?

.github/actions/setup-argocd-image-updater/action.yml
mxyng: This will differ if the input is `latest` which is default. The output will be the real version of the installed binary

.github/actions/setup-argocd-image-updater/index.js
mxyng: Yes, this is mostly a copy of the `setup-infra` action which is itself a copy of the `setup-terragrunt` action in `infrahq/infrastructure`

.gitignore
mxyng: Yes. The ignore on `dist` was recursive which erroneously ignored the actions' `dist` directories. Removing the openapi spec is intentional as well because the file has been moved.

BruceMacD (approved):

Issue #2238, opened 2022-06-08
Revisit `connector` user


jmorganca:
Currently having a "system" `connector` user causes potential issues:

* Users can login as the `connector` user
* It shows up in `infra users list` and this is hard to understand for new users

PR #2892, opened 2022-08-09
feat(api): hide connector by default through showSystem query parameter

### LINKED ISSUE #2733: ui paginates when there are 12 users instead of 13
## Describe the bug

The ui paginates "prematurely" (when there are 12 users). This is likely because the connector is returned as a user in the API, but is hidden by default in the ui.

## Expected behavior

ui should either show the connector user or come up with some other solution, since pagination is done by the server.

## Environment

```
$ infra version
```

```
$ kubectl version
```

j-sneh:
## Summary

Currently the connector shows up in the user list. Although the connector is technically a user, it's not really useful for users to see it as it's mostly internal and we've considered changing the architecture altogether (see https://github.com/infrahq/infra/issues/2238). 

This also prevents pagination issues with the UI (as it was hiding the connector and displaying the wrong item count) and makes the UI and CLI consistent.

I checked for cases where we'd need to enable `showSystem`, but if I missed any use cases where the connector needs to be shown, please let me know.

## Checklist

## Related Issues

Resolves #2733

### DIFF
--- a/internal/server/users.go
+++ b/internal/server/users.go
@@ -14,7 +14,7 @@ import (
 
 func (a *API) ListUsers(c *gin.Context, r *api.ListUsersRequest) (*api.ListResponse[api.User], error) {
 	p := models.RequestToPagination(r.PaginationRequest)
-	users, err := access.ListIdentities(c, r.Name, r.Group, r.IDs, &p)
+	users, err := access.ListIdentities(c, r.Name, r.Group, r.IDs, r.ShowSystem, &p)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


internal/server/users.go
```diff
@@ -48,7 +48,7 @@ func (a *API) CreateUser(c *gin.Context, r *api.CreateUserRequest) (*api.CreateU
 	infraProvider := access.InfraProvider(c)
 
 	// infra identity creation should be attempted even if an identity is already known
-	identities, err := access.ListIdentities(c, user.Name, 0, nil, &models.Pagination{Limit: 2})
+	identities, err := access.ListIdentities(c, user.Name, 0, nil, false, &models.Pagination{Limit: 2})
```
BruceMacD, line 51: This will error when a user is created with the name "connector". 

That is actually a good thing though, because the connector shouldn't get assigned a password. Maybe we should add a check that the user name is not "connector" for a better error message though.

BruceMacD (approved):

PR #2896, opened 2022-08-09
Add organization ID, and remove FKs and sequences

dnephin:
## Summary

This PR cherry-picks 501cd0dd1e0b574c5aa1c8fa2ad9146b88543946,  the data parts of 02cd02c941d9cd016575cb05057285e958cdf9c3, and a bit of 0d1faf9fc40c5c0911fed950b80d18d1cd062178 for merging into main. Best viewed by individual commit.

I made a few changes as well:
* use interfaces instead of reflection for working with models
* use easier to read migration IDs, we started using these recently for new migraiotns
* use single statements for migrations
* remove double quotes from queries, and use caps consistently for sql keywords
* make the selector `ByOrgID` instead of `ByOrg`
* move the OrganizationID to an `OrganizationMember` so that tables that don't use it don't receive an org_id
* continue to require not-null for primary keys
* cascade the deletes of sequences to remove the other sequence data

I had to cherry-pick a couple of changes from #2862 so that AutoMigrate wouldn't create the wrong schema.

### DIFF
--- a/internal/access/access_test.go
+++ b/internal/access/access_test.go
@@ -41,13 +41,8 @@ func setupAccessTestContext(t *testing.T) (*gin.Context, *gorm.DB, *models.Provi
 	c, _ := gin.CreateTestContext(httptest.NewRecorder())
 	c.Set("db", db)
 
-	err := data.SaveSettings(db, &models.Settings{
-		LengthMin: 8,
-	})
-	assert.NilError(t, err)
-
 	admin := &models.Identity{Name: "admin@example.com"}
-	err = data.CreateIdentity(db, admin)
+	err := data.CreateIdentity(db, admin)
 	assert.NilError(t, err)
 
 	c.Set("identity", admin)
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -104,11 +104,11 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 		db = selector(db)
 	}
 
-	if hasOrgID(new(T)) {
-		db = ByOrgID(getOrg(db).ID)(db)
+	result := new(T)
+	if isOrgMember(result) {
+		db = ByOrgID(OrgFromContext(db.Statement.Context).ID)(db)
 	}
 
-	result := new(T)
 	if err := db.Model((*T)(nil)).First(result).Error; err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			return nil, internal.ErrNotFound
@@ -120,22 +120,13 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 	return result, nil
 }
 
-func getOrg(db *gorm.DB) *models.Organization {
-	org, ok := db.Statement.Context.Value("org").(*models.Organization)
-	if !ok {
-		// TODO(orgs): panic("no org") when not set
-		return &models.Organization{}
-	}
-	return org
-}
-
 func setOrg(db *gorm.DB, model any) {
 	member, ok := model.(orgMember)
 	if !ok {
 		return
 	}
 
-	org := getOrg(db)
+	org := OrgFromContext(db.Statement.Context)
 	member.SetOrganizationID(org.ID)
 }
 
@@ -144,7 +135,7 @@ type orgMember interface {
 	SetOrganizationID(id uid.ID)
 }
 
-func hasOrgID(model any) bool {
+func isOrgMember(model any) bool {
 	_, ok := model.(orgMember)
 	return ok
 }
@@ -154,8 +145,8 @@ func list[T models.Modelable](db *gorm.DB, p *models.Pagination, selectors ...Se
 	for _, selector := range selectors {
 		db = selector(db)
 	}
-	if hasOrgID(new(T)) {
-		db = ByOrgID(getOrg(db).ID)(db)
+	if isOrgMember(new(T)) {
+		db = ByOrgID(OrgFromContext(db.Statement.Context).ID)(db)
 	}
 
 	if p != nil {
@@ -286,9 +277,8 @@ func handleError(err error) error {
 }
 
 func delete[T models.Modelable](db *gorm.DB, id uid.ID) error {
-	if hasOrgID(new(T)) {
-		org := getOrg(db)
-		db = db.Where("organization_id = ?", org.ID)
+	if isOrgMember(new(T)) {
+		db = ByOrgID(OrgFromContext(db.Statement.Context).ID)(db)
 	}
 	return db.Delete(new(T), id).Error
 }
@@ -297,9 +287,8 @@ func deleteAll[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) error
 	for _, selector := range selectors {
 		db = selector(db)
 	}
-	if hasOrgID(new(T)) {
-		org := getOrg(db)
-		db = db.Where("organization_id = ?", org.ID)
+	if isOrgMember(new(T)) {
+		db = ByOrgID(OrgFromContext(db.Statement.Context).ID)(db)
 	}
 
 	return db.Delete(new(T)).Error
@@ -323,7 +312,7 @@ func GlobalCount[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (in
 // cache lasts for the entire lifetime of the process, so any test or test
 // helper that calls InfraProvider must call InvalidateCache to clean up.
 func InfraProvider(db *gorm.DB) *models.Provider {
-	org := getOrg(db)
+	org := OrgFromContext(db.Statement.Context)
 	infra, err := get[models.Provider](db, ByProviderKind(models.ProviderKindInfra), ByOrgID(org.ID))
 	if err != nil {
 		if errors.Is(err, internal.ErrNotFound) {
@@ -347,7 +336,7 @@ func InfraProvider(db *gorm.DB) *models.Provider {
 // The cache lasts for the entire lifetime of the process, so any test or test
 // helper that calls InfraConnectorIdentity must call InvalidateCache to clean up.
 func InfraConnectorIdentity(db *gorm.DB) *models.Identity {
-	org := getOrg(db)
+	org := OrgFromContext(db.Statement.Context)
 	connector, err := GetIdentity(db, ByName(models.InternalInfraConnectorIdentityName), ByOrgID(org.ID))
 	if err != nil {
 		logging.L.Panic().Err(err).Msg("failed to retrieve connector identity")
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -104,6 +104,10 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 		db = selector(db)
 	}
 
+	if hasOrgID(new(T)) {
+		db = ByOrgID(getOrg(db).ID)(db)
+	}
+
 	result := new(T)
 	if err := db.Model((*T)(nil)).First(result).Error; err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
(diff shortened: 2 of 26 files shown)


internal/server/data/data.go
```diff
@@ -116,11 +120,43 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
...
+		return &models.Organization{}
+	}
+	return org
+}
+
+func setOrg(db *gorm.DB, model any) {
+	member, ok := model.(orgMember)
+	if !ok {
+		return
+	}
```
dnephin, line 136: When we are ready to, we can make `SetOrganizationID` a required method, and implement no-op implementations on the other tables. That gives us the same safeguards as putting it in the base model, without adding unused organization_id fields.

internal/server/data/passwordresettoken.go
```diff
@@ -17,6 +18,7 @@ func CreatePasswordResetToken(db *gorm.DB, user *models.Identity) (*models.Passw
 	}
 
 	prt := &models.PasswordResetToken{
+		ID:         uid.New(),
```
dnephin, line 21: I forgot to say in the commit message, but this was to fix a test failure. On the org_collab branch we removed all the required not-null for primary keys. I did not include that part, so we need to set an primary key for this type before saving it.  It does not use `Model`, so it doesn't get set that way.

internal/server/data/data.go
```diff
@@ -104,6 +104,10 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 		db = selector(db)
 	}
 
+	if hasOrgID(new(T)) {
+		db = ByOrgID(OrgFromContext(db.Statement.Context).ID)(db)
```
BruceMacD, line 108: Should we just default to always setting this org ID for the time being? I can't think of an example of something that shouldn't have an org ID.

internal/server/data/data.go
```diff
@@ -116,11 +120,34 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
...
+	org := OrgFromContext(db.Statement.Context)
+	member.SetOrganizationID(org.ID)
+}
+
+type orgMember interface {
+	IsOrganizationMember()
+	SetOrganizationID(id uid.ID)
+}
+
+func hasOrgID(model any) bool {
```
BruceMacD, line 138:
```suggestion
func isOrgMember(model any) bool {
```

Seems like a more appropriate function name

internal/server/data/data.go
```diff
@@ -247,18 +277,27 @@ func handleError(err error) error {
 }
 
 func delete[T models.Modelable](db *gorm.DB, id uid.ID) error {
+	if hasOrgID(new(T)) {
+		org := OrgFromContext(db.Statement.Context)
+		db = db.Where("organization_id = ?", org.ID)
```
BruceMacD, line 282: This looks like it should use the selector too (like in `get` and `list`)
```suggestion
	if hasOrgID(new(T)) {
		db = ByOrgID(OrgFromContext(db.Statement.Context).ID)(db)
	}
```

internal/server/data/data.go
```diff
@@ -247,18 +277,27 @@ func handleError(err error) error {
...
+	}
 	return db.Delete(new(T), id).Error
 }
 
 func deleteAll[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) error {
 	for _, selector := range selectors {
 		db = selector(db)
 	}
+	if hasOrgID(new(T)) {
```
BruceMacD, line 291: Same comment as the other delete function, this looks like it should use the selector also.

internal/server/data/data.go
```diff
@@ -271,53 +310,44 @@ func Count[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (int64, e
...
-				p := &models.Provider{Name: models.InternalInfraProviderName, Kind: models.ProviderKindInfra}
-				if err := add(db, p); err != nil {
-					logging.L.Panic().Err(err).Msg("failed to create infra provider")
-				}
-				return p
+	org := OrgFromContext(db.Statement.Context)
+	infra, err := get[models.Provider](db, ByProviderKind(models.ProviderKindInfra), ByOrgID(org.ID))
+	if err != nil {
+		if errors.Is(err, internal.ErrNotFound) {
+			p := &models.Provider{
```
BruceMacD, line 321: nit: does not have to be in this PR

Should we rely on the create function for this Infra provider creation?

internal/server/data/data.go
```diff
@@ -271,53 +310,44 @@ func Count[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (int64, e
...
-		infra, err := get[models.Provider](db, ByProviderKind(models.ProviderKindInfra))
-		if err != nil {
-			if errors.Is(err, internal.ErrNotFound) {
-				p := &models.Provider{Name: models.InternalInfraProviderName, Kind: models.ProviderKindInfra}
-				if err := add(db, p); err != nil {
-					logging.L.Panic().Err(err).Msg("failed to create infra provider")
-				}
-				return p
+	org := OrgFromContext(db.Statement.Context)
+	infra, err := get[models.Provider](db, ByProviderKind(models.ProviderKindInfra), ByOrgID(org.ID))
```
BruceMacD, line 318: We will have to think about re-visiting this at some point to add the caching back imo, fine for now though

internal/server/data/organization.go
```diff
@@ -1,11 +1,32 @@
...
 	"github.com/infrahq/infra/internal/server/models"
 )
 
+type orgCtxKey struct{}
+
+// OrgFromContext returns the Organization stored using WithOrg.
+func OrgFromContext(ctx context.Context) *models.Organization {
+	org, ok := ctx.Value(orgCtxKey{}).(*models.Organization)
+	if !ok {
+		// TODO(orgs): panic("no org")
```
BruceMacD, line 17: I think we should merge this with the panic or create an issue to track adding the panic, this feels dangerous otherwise

internal/server/data/schema.sql
```diff
@@ -215,65 +141,27 @@ CREATE TABLE providers (
...
-CREATE SEQUENCE settings_id_seq
-    START WITH 1
-    INCREMENT BY 1
-    NO MINVALUE
-    NO MAXVALUE
-    CACHE 1;
-
-ALTER SEQUENCE settings_id_seq OWNED BY settings.id;
-
-ALTER TABLE ONLY access_keys ALTER COLUMN id SET DEFAULT nextval('access_keys_id_seq'::regclass);
```
BruceMacD, line 255: Should we still have these alter table statements? We manually set the ID by they seem ok to have.

internal/access/access_test.go
```diff
@@ -41,13 +41,8 @@ func setupAccessTestContext(t *testing.T) (*gin.Context, *gorm.DB, *models.Provi
 	c, _ := gin.CreateTestContext(httptest.NewRecorder())
 	c.Set("db", db)
 
-	err := data.SaveSettings(db, &models.Settings{
```
pdevine, line 44: My last change (commit 5f6492f0c9bcae58c59ab6917a0768d4371182d1) should fix this, and `credential_test.go` below this.

pdevine: Oh, also, the difference here is on org_collab we are creating a default org for the tests in `setupDB()` in `access.go`. That's covered in commit fe6121273.

internal/server/data/data.go
dnephin: Ya, may as well be consistent. All the selectors are going to be removed soon, but doesn't hurt to be consistent in the interim.

internal/server/data/data.go
dnephin: :100: https://github.com/infrahq/infra/pull/2902/commits/b1921154af3bd65039363123fafe72357ab31dfb (I see you found the TODO).

internal/server/data/data.go
dnephin: Ya, we should look at this again. I think we should try and cache these things and the default org on `Server` , so that we can pass it via `RequestContext`.

internal/server/data/data.go
BruceMacD: Ah, of course, organizations themselves. 👍

internal/server/data/organization.go
dnephin: Adding the panic is what causes tons of tests to break and makes the change difficult. I think an issue would be good. 

I created https://github.com/infrahq/infra/issues/2903 to track these kinds of things.

internal/server/data/schema.sql
dnephin: I'll let @ssoroka comment as well, because I think he removed them originally.

I like removing them because we don't use them, and I think having these sequences could hide bugs from us in testing. I think it's better off if the column is set to `NOT NULL` without a sequence to force us to set values.

internal/access/access_test.go
dnephin: I saw your change in https://github.com/infrahq/infra/commit/5f6492f0c9bcae58c59ab6917a0768d4371182d1 but I took a different approach here.

I think calling `SaveSettings` in this kind of helper ends up causing more test failures as we make code changes. We should either set this in `setupDB` for everything, or do it in specific tests that require it.  I opted for specific tests.

I think #2902 will remove the need to create a default org in tests. It should exist as soon as we load the database. Although saying it now, I realize there is something missing from that PR to do it as part of initial schema creation.

internal/server/data/data.go
```diff
@@ -104,6 +104,10 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
 		db = selector(db)
 	}
 
+	if hasOrgID(new(T)) {
```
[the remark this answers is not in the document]
dnephin: Good call!

internal/server/data/data.go
```diff
@@ -116,11 +120,34 @@ func get[T models.Modelable](db *gorm.DB, selectors ...SelectorFunc) (*T, error)
...
+	}
+
+	org := OrgFromContext(db.Statement.Context)
+	member.SetOrganizationID(org.ID)
+}
+
+type orgMember interface {
+	IsOrganizationMember()
+	SetOrganizationID(id uid.ID)
+}
```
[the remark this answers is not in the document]
dnephin: Interfaces in Go should be defined where they are used, not where they are implemented. This is definitely the right place for this interface because it is used in `data`.  This practices aligns nicely with the philosophy of "accept interfaces, return structs".

internal/server/data/data.go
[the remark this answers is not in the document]
dnephin: Same, but until we remove gorm I'm not sure what to do about that.   Did you have a solution for this in `org_collab` that I missed? 

My thinking is use this for now, and hopefully we have removed gorm before the next release.

internal/server/data/migrations_test.go
```diff
@@ -389,6 +389,18 @@ INSERT INTO provider_users (identity_id, provider_id, id, created_at, updated_at
...
+			expected: func(t *testing.T, db *gorm.DB) {
+				// column changes are tested with schema comparison
+			},
+		},
+		{
+			label: testCaseLine("2022-08-04T17:72"),
+			expected: func(t *testing.T, db *gorm.DB) {
+				// schema changes are tested with schema comparison
+			},
+		},
```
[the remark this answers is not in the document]
dnephin: Not sure what you mean. These are the migration tests, and for schema-only changes there isn't really anything to test.

BruceMacD (approved):
Looks good to me pending Steven's comments

internal/server/data/data.go
ssoroka: some tables don't have an OrganizationID column

internal/server/data/migrations_test.go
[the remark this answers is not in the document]
ssoroka: don't they need to be in the list to avoid failing a test, even if it's just schema?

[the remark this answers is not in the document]
dnephin: Ya, that's what this is. They have no-op `expected` functions to fulfill the requirement of having a test case.

internal/access/access_test.go
pdevine: OK. I'm fine with the changes. I think setting the settings in the specific tests ends up being slightly more confusing, because there's really not a case where you wouldn't already have an Org (and consequently settings attached to the org). For example, in `TestSettingsPasswordRequirements` I think getting the current settings and then saving them again is less confusing than pushing in a new set of settings just because that's the way we would expect developers to write their code.

This isn't a huge point though, so let's just move the ball forward.

pdevine (approved):

ssoroka (approved):

internal/access/access_test.go
dnephin: Yes, good point. `SaveSettings` does exactly that now. It updates the settings for the org.

dnephin: Done in #2921, now the only calls to SaveSettings are for testing something that changes settings.

PR #2899, opened 2022-08-10
improve: better error message for cert mismatch

### LINKED ISSUE #2898: cli: better error message if client's cached cert settings no longer are synced with the server
## Describe the bug

1. Install cli & infra server. 
2. Login successfully.
3. Re-install infra server. Now the client's cached session should no longer work the new server. 
4. Run any command that calls the api/server. Error message is:5. 
```
Error: GET "/api/users/bzeggdZVCu": Get "https://localhost/api/users/bzeggdZVCu": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "Infra Server")
```

## Expected behavior

User should try to re-login, as it should fix it (since it's probably the mismatch in old/new servers that share the same address). Error should advise user to re-login.

kimskimchi:
## Summary

Adds a formal error message (along with the actual error) for cert mismatching the server.  

## Related Issues

Resolves #2898

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,7 +26,7 @@ func main() {
 		case errors.As(err, &unknownAuthErr):
 			// Cert error is most likely caused by mismatch between the client session and its cached server
 			// eg) server was re-installed, but the client cache was not cleared
-			fmt.Fprintf(os.Stderr, "Error: %v\n\nRun `infra login` again OR `infra logout --clear` to see if a new session resolves the issue.\n", err)
+			fmt.Fprintf(os.Stderr, "Error: %v\n\nTLS certificate is no longer valid with this server; to see if a new session resolves the issue, run `infra login` again OR `infra logout --clear`\n", err)
 		default:
 			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 		}
--- a/main.go
+++ b/main.go
@@ -2,10 +2,10 @@ package main
 
 import (
 	"context"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/AlecAivazis/survey/v2/terminal"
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -20,6 +21,10 @@ func main() {
 			logging.Debugf("user interrupted the process")
 		case errors.As(err, &userErr):
 			fmt.Fprintln(os.Stderr, userErr.Error())
+		case strings.Contains(err.Error(), "x509: certificate signed by unknown authority"):
```
dnephin, line 24: I think you may be able to use this:

```go
unknownAuthErr := x509.UnknownAuthorityError{}
...
case errors.As(err, &unknownAuthErr):
...
```

main.go
```diff
@@ -20,6 +21,10 @@ func main() {
...
+			// Cert error is most likely caused by mismatch between the client session and its cached server
+			// eg) server was re-installed, but the client cache was not cleared
+			fmt.Fprintf(os.Stderr, "Session certificate is no longer valid for this server; run 'infra login' to start a new session:\n\n%v\n", err)
```
dnephin, line 27: Maybe "TLS certificate ..." ? In a lot of installs the server TLS certificate will not change, so it won't break on each restart. 

```suggestion
			fmt.Fprintf(os.Stderr, "TLS certificate is no longer valid for this server; run 'infra login' to start a new session:\n\n%v\n", err)
```

dnephin (approved):
Looks good! This new error message should be a lot more helpful to users.

Left a couple suggestions, but nothing blocking.

main.go
pdevine: I'm not 100% sure of when the error pops up, but I wouldn't think that as a user that `infra login` would sort out a TLS certificate problem. I think most people would think that a TLS certificate error is an issue with the server, and not the client.

dnephin: It is, but the client will prompt them to fix in on login.

This is also only an issue with self-signed certs, so maybe something we are less concerned about now?

main.go
kimskimchi: done

main.go
kimskimchi:
```> infra users list
Error: GET "/api/users": Get "https://localhost/api/users?group=&limit=0&name=&page=1": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "Infra Server")

To see if the issue is with the client, run `infra login` again OR `infra logout --clear` to clear session```

PR #2900, opened 2022-08-10
improve: better error message when user tries to delete last admin grant

### LINKED ISSUE #2673: api should block user / group to be removed
Today in the CLI, a user is able to revoke any user / groups Infra admin permission as long as they have the permission. However, user might revoke themselves infra admin permission accidentally and unintentionally. Same thing would happen if a user trying to remove themselves from a group, they might lose some permission that the group grants.

So in order to prevent that happens, there are couple checks in the client side to inform users (https://github.com/infrahq/infra/issues/2647) but we should block it in the server side to make sure there is always at least one user / one group has infra admin permission

kimskimchi:
## Related Issues

Resolves #2673

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -252,6 +252,12 @@ func removeGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 					Message: "Cannot revoke grants: missing privileges for DeleteGrant",
 				}
 			}
+			if api.ErrorStatusCode(err) == 400 && strings.Contains(err.Error(), "cannot remove the last infra admin") {
+				logging.Debugf("%s", err.Error())
+				return Error{
+					Message: "Cannot revoke grant: at least one admin grant must exist",
+				}
+			}
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


internal/cmd/grants.go
```diff
@@ -252,6 +252,12 @@ func removeGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
 					Message: "Cannot revoke grants: missing privileges for DeleteGrant",
 				}
 			}
+			if api.ErrorStatusCode(err) == 400 && strings.Contains(err.Error(), "cannot remove the last infra admin") {
+				logging.Debugf("%s", err.Error())
+				return Error{
+					Message: "Cannot revoke grant: at least one admin grant must exist",
```
BruceMacD, line 258:
```suggestion
					Message: "Cannot revoke grant: at least one Infra admin grant must exist",
```

nit: I think this could be misinterpreted otherwise

pdevine: We may want to change this to `Organization Admin` in the future, but I think `Infra admin` is fine for now.

pdevine (approved):

PR #2902, opened 2022-08-10
Add default organization

dnephin:
Branched from #2896

Adds the default org as a migration, and migrates existing data to the org.
For new installs and in tests the default org, settings, and infra provider are created by `NewDB`.

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -26,7 +26,7 @@ import (
 // NewDB creates a new database connection and runs any required database migrations
 // before returning the connection. The loadDBKey function is called after
 // initializing the schema, but before any migrations.
-func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*gorm.DB, error) {
+func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*DB, error) {
 	db, err := newRawDB(connection)
 	if err != nil {
 		return nil, fmt.Errorf("db conn: %w", err)
@@ -41,7 +41,23 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*gorm.
 		return nil, fmt.Errorf("migration failed: %w", err)
 	}
 
-	return db, nil
+	// TODO: initialize, and populate settings and org on DB
+
+	return &DB{DB: db}, nil
+}
+
+// DB wraps the underlying database and provides access to the default org,
+// and settings.
+type DB struct {
+	*gorm.DB // embedded for now to minimize the diff
+}
+
+func (db *DB) Close() error {
+	sqlDB, err := db.DB.DB()
+	if err != nil {
+		return fmt.Errorf("failed to get database conn to close: %w", err)
+	}
+	return sqlDB.Close()
 }
 
 // newRawDB creates a new database connection without running migrations.
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -91,23 +91,25 @@ func newRawDB(connection gorm.Dialector) (*gorm.DB, error) {
 }
 
 func initialize(db *DB) error {
-	p := &models.Provider{
-		Name: models.InternalInfraProviderName,
-		Kind: models.ProviderKindInfra,
-		// TODO: OrganizationMember: models.OrganizationMember{OrganizationID: org.ID},
-	}
-	if err := add(db.DB, p); err != nil {
-		logging.L.Panic().Err(err).Msg("failed to create infra provider")
+	org, err := GetOrganization(db.DB, ByName(models.DefaultOrganizationName))
+	switch {
+	case errors.Is(err, internal.ErrNotFound):
+		org = &models.Organization{
+			Name:      models.DefaultOrganizationName,
+			CreatedBy: models.CreatedBySystem,
+		}
+		if err := CreateOrganization(db.DB, org); err != nil {
+			return fmt.Errorf("failed to create default organization: %w", err)
+		}
+	case err != nil:
+		return fmt.Errorf("failed to get default organization: %w", err)
 	}
 
-	// TODO: set org
+	db.DefaultOrg = org
 
-	var err error
-	db.DefaultOrgSettings, err = initializeSettings(db.DB)
-	if err != nil {
-		return err
-	}
-	return nil
+	db.Statement.Context = WithOrg(db.Statement.Context, org)
+	db.DefaultOrgSettings, err = GetSettings(db.DB)
+	return err
 }
 
 func NewSQLiteDriver(connection string) (gorm.Dialector, error) {
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -42,14 +42,20 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*DB, e
 	}
 
 	// TODO: initialize, and populate settings and org on DB
+	dataDB := &DB{DB: db}
+	if err := initialize(dataDB); err != nil {
+		return nil, fmt.Errorf("initialize: %w", err)
+	}
 
-	return &DB{DB: db}, nil
+	return dataDB, nil
 }
 
 // DB wraps the underlying database and provides access to the default org,
 // and settings.
 type DB struct {
 	*gorm.DB // embedded for now to minimize the diff
+
+	DefaultOrgSettings *models.Settings
 }
 
 func (db *DB) Close() error {
@@ -81,6 +87,17 @@ func newRawDB(connection gorm.Dialector) (*gorm.DB, error) {
 	}
 
 	return db, nil
+}
+
+func initialize(db *DB) error {
+	// TODO: set org
+
+	var err error
+	db.DefaultOrgSettings, err = initializeSettings(db.DB)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 func NewSQLiteDriver(connection string) (gorm.Dialector, error) {
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -44,7 +44,7 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*DB, e
 	// TODO: initialize, and populate settings and org on DB
 	dataDB := &DB{DB: db}
 	if err := initialize(dataDB); err != nil {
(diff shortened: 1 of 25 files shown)


internal/server/data/migrations.go
```diff
@@ -460,3 +463,39 @@ DROP SEQUENCE IF EXISTS password_reset_tokens_id_seq CASCADE;
...
+
+func addDefaultOrganization() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-08-10T13:35",
+		Migrate: func(tx *gorm.DB) error {
+
+			stmt := `
+INSERT INTO organizations(id, name, created_at, updated_at)
+VALUES (? , ?, ?, ?);
+`
```
dnephin, line 475: We don't want to use `data.CreateOrganization` here, because that can change over time, which can break this migration.

internal/server/data/migrations.go
```diff
@@ -460,3 +463,39 @@ DROP SEQUENCE IF EXISTS password_reset_tokens_id_seq CASCADE;
...
 }
+
+func addDefaultOrganization() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-08-10T13:35",
+		Migrate: func(tx *gorm.DB) error {
+
+			stmt := `
+INSERT INTO organizations(id, name, created_at, updated_at)
+VALUES (? , ?, ?, ?);
```
BruceMacD, line 474:
```suggestion
VALUES (?, ?, ?, ?);
```

nit

internal/server/data/migrations.go
```diff
@@ -460,3 +463,39 @@ DROP SEQUENCE IF EXISTS password_reset_tokens_id_seq CASCADE;
...
+`
+			orgID := uid.New()
+			now := time.Now()
+			if err := tx.Exec(stmt, orgID, "Default", now, now).Error; err != nil {
```
BruceMacD, line 478: We should make this "Default" name a constant similar to `models.InternalInfraProviderName`

internal/server/data/organization.go
```diff
@@ -13,7 +14,7 @@ type orgCtxKey struct{}
 // OrgFromContext returns the Organization stored using WithOrg.
 func OrgFromContext(ctx context.Context) *models.Organization {
 	org, ok := ctx.Value(orgCtxKey{}).(*models.Organization)
-	if !ok {
+	if !ok || org == nil {
```
BruceMacD, line 17: Good catch

internal/server/data/organization.go
```diff
@@ -28,7 +29,20 @@ func WithOrg(ctx context.Context, org *models.Organization) context.Context {
...
+		return fmt.Errorf("creating org: %w", err)
+	}
+
+	db.Statement.Context = WithOrg(db.Statement.Context, org)
+	_, err = InitializeSettings(db)
+	if err != nil {
+		return fmt.Errorf("initializing org settings: %w", err)
+	}
+
+	// TODO: create infra provider here too?
```
BruceMacD, line 43: I was thinking that in the other PR too, we should do that in a separate change

internal/server/middleware.go
```diff
@@ -113,6 +116,27 @@ func withDBTxn(ctx context.Context, db *gorm.DB, fn func(tx *gorm.DB)) {
 func unauthenticatedMiddleware(db *gorm.DB) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		withDBTxn(c.Request.Context(), db, func(tx *gorm.DB) {
+			var orgID uid.ID
```
BruceMacD, line 119: I think this should set the default org in the case no Infra-Org header is found

internal/server/middleware.go
```diff
@@ -175,6 +199,13 @@ func requireAccessKey(db *gorm.DB, req *http.Request) (access.Authenticated, err
 		return u, fmt.Errorf("identity update fail: %w", err)
 	}
 
+	// TODO(orgs): make org ID required
+	if accessKey.OrganizationID != 0 {
```
BruceMacD, line 203: We should error out here if there's no organization on the access key. Better to play it safe. We can re-visit when we add support roles imo.

internal/server/data/migrations.go
```diff
@@ -460,3 +463,39 @@ DROP SEQUENCE IF EXISTS password_reset_tokens_id_seq CASCADE;
...
 	}
 }
+
+func addDefaultOrganization() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-08-10T13:35",
+		Migrate: func(tx *gorm.DB) error {
+
+			stmt := `
+INSERT INTO organizations(id, name, created_at, updated_at)
```
dnephin, line 473: This insert is missing from the original schema. I'll need to change the migration test to include this, or we could do a manual include for now.

internal/server/middleware.go
dnephin: I'll have to double check, but that may also cause a bunch of test failures. Either way, I'll add this to our list in #2903 so we don't forget.

internal/server/data/migrations.go
dnephin: We definitely should, but I don't think we'd want to use that constant here. Migrations need to be frozen in time from the point they are initially released. We could hypothetically change the constant, but this migration should never change.

internal/server/data/organization.go
```diff
@@ -13,7 +14,7 @@ type orgCtxKey struct{}
 // OrgFromContext returns the Organization stored using WithOrg.
 func OrgFromContext(ctx context.Context) *models.Organization {
 	org, ok := ctx.Value(orgCtxKey{}).(*models.Organization)
-	if !ok {
+	if !ok || org == nil {
 		// TODO(orgs): panic("no org")
 		return &models.Organization{}
```
ssoroka, line 19: I definitely don't like that this quietly passes.

ssoroka (approved):

internal/server/data/organization.go
dnephin: Ya, this won't stay this way. Tracking it in https://github.com/infrahq/infra/issues/2903. But it lets us make progress without having to fight test failure across multiple packages.

internal/server/data/migrations.go
```diff
@@ -460,3 +463,39 @@ DROP SEQUENCE IF EXISTS password_reset_tokens_id_seq CASCADE;
...
+`
+			orgID := uid.New()
+			now := time.Now()
+			if err := tx.Exec(stmt, orgID, "Default", now, now).Error; err != nil {
+				return err
+			}
+
+			// postgres only allows a single statement when using parameters
+			for _, stmt := range []string{
+				` UPDATE access_keys SET organization_id = ?;`,
```
pdevine, line 484: Looks like there's an extra space before the UPDATE statement here.

pdevine (approved):

BruceMacD (approved):

internal/server/data/settings.go
```diff
@@ -45,27 +47,38 @@ func InitializeSettings(db *gorm.DB) (*models.Settings, error) {
...
-	if err := db.First(&settings).Error; err != nil {
+	if err := db.Where("organization_id = ?", org.ID).First(&settings).Error; err != nil {
 		return nil, err
 	}
 
 	return &settings, nil
 }
 
 func SaveSettings(db *gorm.DB, settings *models.Settings) error {
+	// TODO: clean this up by having the query use the organization_id instead of the
```
dnephin, line 75: Should be easy to do this when we remove gorm.

internal/server/models/organization.go
```diff
@@ -21,6 +25,12 @@ func (o *Organization) ToAPI() *api.Organization {
 	}
 }
 
+// TODO(orgs): remove
+func (o *Organization) BeforeCreate(_ *gorm.DB) error {
+	// allow an org with id=0
+	return nil
+}
```
dnephin, line 32: To get around all the test failures when we create the default org, the last commit in this PR allows creating the default org with ID=0.

I'm tracking this cleanup in #2903, and all of the changes are done in a small commit, so it's going to be easy to revert when the tests are ready for it.

internal/server/middleware.go
dnephin: moved this to a separate PR

internal/server/data/data.go
```diff
@@ -41,7 +41,30 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*gorm.
...
+	if err := initialize(dataDB); err != nil {
+		return nil, fmt.Errorf("initialize database: %w", err)
+	}
+
+	return dataDB, nil
+}
+
+// DB wraps the underlying database and provides access to the default org,
+// and settings.
+type DB struct {
```
dnephin, line 55: I need this type for removing Gorm, and it helps us here as well by giving us a way to return the settings and default org.

pdevine (approved):

internal/server/data/data.go
```diff
@@ -67,6 +90,28 @@ func newRawDB(connection gorm.Dialector) (*gorm.DB, error) {
...
+		if err := CreateOrganization(db.DB, org); err != nil {
+			return fmt.Errorf("failed to create default organization: %w", err)
+		}
+	case err != nil:
+		return fmt.Errorf("failed to get default organization: %w", err)
+	}
+
+	db.DefaultOrg = org
+
+	db.Statement.Context = WithOrg(db.Statement.Context, org)
```
dnephin, line 110: Calling this out to review in our TODO list. #2903

internal/server/data/migrations_test.go
```diff
@@ -401,6 +401,35 @@ INSERT INTO provider_users (identity_id, provider_id, id, created_at, updated_at
...
+					{
+						Model: models.Model{
+							ID:        99,
+							CreatedAt: now,
+							UpdatedAt: now,
+						},
+						Name: "Default",
+					},
+				}
+				assert.DeepEqual(t, orgs, expected, cmpModel)
```
dnephin, line 430: TODO: should add  test coverage for existing data that now has the orgID set.

ssoroka (approved):

internal/server/middleware.go
```diff
@@ -73,6 +73,16 @@ func handleInfraDestinationHeader(c *gin.Context) error {
 func authenticatedMiddleware(db *gorm.DB) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		withDBTxn(c.Request.Context(), db, func(tx *gorm.DB) {
+
+			// TODO: get this from server
```
dnephin, line 77: This is done in #2907

PR #2906, opened 2022-08-10
improve: skip server selection and update docs for auto-login

### LINKED ISSUE #2758: Don't prompt in `infra login` when info is already known
This is related to #2557 but I think it's a distinct issue.

I would like for subsequent runs of `infra login` after the first setup to do the needful without prompting me.

For some background, I initially configured infra by running `infra login --provider okta infra.example.com`.  I only use the one server and the one login method, and I find it annoying that each subsequent time I need to run `infra login` that I either (a) need to provide the server and provider again or (b) hit enter through some prompts.

```
$ infra login
? Select a server:  [Use arrows to move, type to filter]
> infra.example.com
  Connect to a new server

? Select a login method:  [Use arrows to move, type to filter]
> okta (example.okta.com)
  Login with username and password
```

My preferred solution would be for `infra login` to use the saved config and do it without prompting if there is a previously configured server.  An also acceptable solution would be to have a flag to do this, like `infra login -y`.  My argument for the former is that the vast majority of the time people will only have a single login and want to use it (or log into their most recently used server if there's more than one).  Also this would match other tooling that performs similar things like `kubectl oulogin`.

## Environment Details

```
$ infra version
 Client: 0.14.2
 Server: 0.14.1
```

```
$ kubectl version
Client Version: v1.24.3
Kustomize Version: v4.5.4
Server Version: v1.22.8
WARNING: version difference between client (1.24) and server (1.22) exceeds the supported minor version skew of +/-1
```

j-sneh:
## Summary

Users have requested being able to run `infra login` and get logged in without the survey prompts. One could already set the `INFRA_SERVER` env var to skip server selection. This PR adds the `INFRA_PROVIDER` and cli docs to make it more clear. It also uses local login if the provider is `"infra"`

## Checklist

## Related Issues

Resolves #2758

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -26,7 +26,6 @@ import (
 	"github.com/infrahq/infra/internal/cmd/types"
 	"github.com/infrahq/infra/internal/generate"
 	"github.com/infrahq/infra/internal/logging"
-	"github.com/infrahq/infra/internal/server/models"
 )
 
 type loginCmdOptions struct {
@@ -163,7 +162,7 @@ func login(cli *CLI, options loginCmdOptions) error {
 			return Error{Message: "Non-interactive login only supports access keys, set the INFRA_ACCESS_KEY environment variable and try again"}
 		}
 
-		if options.Provider == models.InternalInfraProviderName {
+		if options.Provider == "infra" {
 			loginReq.PasswordCredentials, err = promptLocalLogin(cli)
 			if err != nil {
 				return err
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -66,6 +66,11 @@ $ infra login --provider okta
 
 # Login with an access key
 $ export INFRA_ACCESS_KEY=1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy
+$ infra login
+
+# Login with pre-set provider and server
+$ export INFRA_SERVER=example.infrahq.com
+$ export INFRA_PROVIDER=google
 $ infra login`,
 		Args:  MaxArgs(1),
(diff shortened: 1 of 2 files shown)


internal/cmd/login.go
```diff
@@ -78,6 +83,10 @@ $ infra login`,
 				options.Server = args[0]
 			}
 
+			if options.Provider == "infra" {
```
BruceMacD, line 86: The infra provider doesn't seem to work for me. Me guess is setting this blank doesnt see that the provider is set

j-sneh: This is the local login. The idea here is if someone has an `INFRA_PROVIDER` defined, if they want to log in to a local user then they specify `--provider infra`. Maybe `"infra"` shouldn't be a valid provider name?

BruceMacD: Infra as a provider name for this functionality seems good to me. Im just having issues with getting this behavior to work, I may be doing something wrong.

It still prompts me for the provider, but I expect it should just automatically prompt me for username/password:
```
$ export INFRA_PROVIDER=infra 

$ infra login localhost                 
? Select a login method:  [Use arrows to move, type to filter]
> google (accounts.google.com)
  okta (dev.okta.com)
  Login with username and password

 $ infra login localhost --provider infra
? Select a login method:  [Use arrows to move, type to filter]
> google (accounts.google.com)
  okta (dev-50966300.okta.com)
  Login with username and password
```

j-sneh: @BruceMacD That worked locally because my local infra instance only had local login. This should be fixed now.

BruceMacD (approved):

PR #2911, opened 2022-08-11
Add domains field to orgs, and use it to identify the org in unauthenticated requests

dnephin:
## Summary

### DIFF
--- a/api/organization.go
+++ b/api/organization.go
@@ -6,10 +6,9 @@ import (
 )
 
 type Organization struct {
-	ID      uid.ID `json:"id"`
-	Name    string `json:"name"`
-	Created Time   `json:"created"`
-	Updated Time   `json:"updated"`
+	ID     uid.ID `json:"id"`
+	Name   string `json:"name"`
+	Domain string `json:"domain"`
 }
 
 type ListOrganizationsRequest struct {
--- a/api/organization.go
+++ b/api/organization.go
@@ -6,9 +6,11 @@ import (
 )
 
 type Organization struct {
-	ID     uid.ID `json:"id"`
-	Name   string `json:"name"`
-	Domain string `json:"domain"`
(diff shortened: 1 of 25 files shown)


api/organization.go
```diff
@@ -6,10 +6,9 @@ import (
 )
 
 type Organization struct {
-	ID      uid.ID `json:"id"`
-	Name    string `json:"name"`
-	Created Time   `json:"created"`
-	Updated Time   `json:"updated"`
```
dnephin, line 12: What's the reason for removing these? Since orgs in the API are really only visible to support-admins, it seems like it would be pretty valuable to be able to see when an org was created or update.

internal/server/models/organization.go
```diff
@@ -11,16 +14,31 @@ type Organization struct {
...
+		ID:     o.ID,
+		Name:   o.Name,
+		Domain: o.Domain,
+	}
+}
+
+var domainNameReplacer = regexp.MustCompile(`[^\da-zA-Z-]`)
+
+// TODO: is the model the right place for this? Should we do this in CreateOrg ?
+func (o *Organization) SetDefaultDomain() {
```
dnephin, line 32: We don't have to do this now, but I think this would be better in `data.CreateOrganization`. The two callers to this call create immediately after, and generally the create function in the `data` package is where we set defaults.

internal/server/server.go
```diff
@@ -88,8 +88,11 @@ type TLSOptions struct {
 }
 
 type Server struct {
-	options  Options
-	db       *gorm.DB
+	options Options
+	// TODO: consolidate these two
+	db     *gorm.DB
+	dataDB *data.DB
```
dnephin, line 94: I am planning on doing this in a follow up.

ssoroka (approved):

PR #2912, opened 2022-08-11
fix: sort order tests on mac should be skipped

ssoroka:


dnephin (approved):

j-sneh:
This is odd. I'm on mac (`runtime.GOOS == darwin`) and the tests run fine on my machine.

dnephin:
I think you are still running tests against sqlite (which works ok). These tests break with macos postgres

j-sneh:
@dnephin That was the issue. I thought I had exported `POSTGRESQL_CONNECTION`, but not in my current session. Are we keeping support for sqlite?

dnephin:
It's not a requirement, and it's entirely possible we'll have to remove it when we remove gorm, but I'm trying to keep it around until it's necessary to remove it so we have a better idea of what the challenges are with supporting both.

If we do keep any support, it would only be for development purposes, not production.

ssoroka:
I believe it's darwin + Docker's new mac-native virtualization framework that triggers it.

BruceMacD (approved):

PR #2913, opened 2022-08-11
fix: hide connector grants

### LINKED ISSUE #2909: can't list grants in cli locally
## Describe the bug
When listing grants in the CLI, I get the error message `Error: unknown user for ID <uid.ID>`.

## Expected behavior

The user should see the list of grants and not encounter this error every time.
## Environment

```
$ infra version
Client: 0.14.5+dev
Server: 0.14.4
```
(server us also dev)

```
$ kubectl version
```

j-sneh:
## Summary

This is a natural continuation of this issue https://github.com/infrahq/infra/pull/2892, which was also causing a bug with the `infra grants list` command.

## Checklist

## Related Issues

Resolves #2909

### DIFF
--- a/api/grant.go
+++ b/api/grant.go
@@ -38,7 +38,7 @@ type ListGrantsRequest struct {
 	Resource      string `form:"resource" example:"production"`
 	Privilege     string `form:"privilege" example:"view"`
 	ShowInherited bool   `form:"showInherited" note:"if true, this field includes grants that the user inherits through groups"`
-	ShowSystem    bool   `form:"showSystem"`
+	ShowSystem    bool   `form:"showSystem" note:"if true, this shows the connector and other internal grants"`
 	PaginationRequest
 }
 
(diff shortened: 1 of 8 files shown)


internal/access/grant.go
```diff
@@ -68,6 +68,10 @@ func ListGrants(c *gin.Context, subject uid.PolymorphicID, resource string, priv
 		selectors = append(selectors, data.ByOptionalSubject(subject))
 	}
 
+	if !showSystem {
+		selectors = append(selectors, data.NotPrivilege(models.InfraConnectorRole))
```
BruceMacD, line 72: I think we will have to be more specific than this.

There could be a privilege called "connector" in other places outside of infra that its legit to create a grant for. We should hide system grants by something like `privilege != connector && destination != infra` together

api/grant.go
```diff
@@ -38,6 +38,7 @@ type ListGrantsRequest struct {
 	Resource      string `form:"resource" example:"production"`
 	Privilege     string `form:"privilege" example:"view"`
 	ShowInherited bool   `form:"showInherited" note:"if true, this field includes grants that the user inherits through groups"`
+	ShowSystem    bool   `form:"showSystem"`
```
ssoroka, line 41: add a note tag here so that the api docs know what this is.

ssoroka (approved):

PR #2914, opened 2022-08-11
fix: avoid annoying mac port warning running tests

ssoroka:
Avoids these annoying things by only binding to the local network device by default (tests use the default), not internet-facing ones.
Screen Shot 2022-08-11 at 3 05 00 PM

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -266,10 +266,14 @@ func (s *Server) listen() error {
 }
 
 func (s *Server) setupServer(server *http.Server) (net.Addr, error) {
+	if server.Addr == "" {
+		server.Addr = "127.0.0.1:"
+	}
 	l, err := net.Listen("tcp", server.Addr)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/server/server.go
```diff
@@ -266,10 +266,14 @@ func (s *Server) listen() error {
 }
 
 func (s *Server) setupServer(server *http.Server) (net.Addr, error) {
+	if server.Addr == "" {
+		server.Addr = "127.0.0.1:"
```
BruceMacD, line 270: Is the trailing semi-colon for the port intentional?

BruceMacD (approved):

internal/server/server.go
mxyng: I don't think this does what you expect @ssoroka. `server.Addr` should never be empty since there are defaults, e.g. `:8080` for the HTTP server, `:8443` for the HTTPS server

PR #2841, opened 2022-08-05
feat: org sign-up UI

BruceMacD:
## Summary
Add orgs to Infra sign-up UI (note that it is going into the org_collab branch, not main).

- Updates UI
- Updates sign-up for separate org and domain
- Does not require org name to be unique

image

image

### Next steps 
#2801 
- Update the quickstart to direct new self-hosted deployments to the "sign-up" UI
- Update ingress when a new org subdomain is created
- https://github.com/infrahq/infra/issues/2806

## Checklist

- [x] Wrote appropriate unit tests
- [x] Considered security implications of the change
- [ ] Updated associated docs where necessary
- [x] Updated associated configuration where necessary
- [x] Change is backwards compatible if it needs to be (user can upgrade without manual steps?)
- [x] Nothing sensitive logged
- [x] Considered data migrations for smooth upgrades

## Related Issues

Resolves #2822

PR #2915, opened 2022-08-11
improve: add component concurrency controls

mxyng:
## Summary

Minor change to the deploy workflow to add concurrency controls for individual components. A deploy for `core` should not override/block deploys for `ui`

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -7,16 +7,35 @@ on:
       - cd / core
     types:
       - completed
+    branches:
+      - main
   workflow_dispatch:
     inputs:
-      COMPONENT:
+      component-name:
         type: string
         default: core
         required: true
 
 jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
+    outputs:
+      component-name: ${{ steps.component-name.outputs.component-name }}
+    steps:
+      - id: component-name
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yaml
```diff
@@ -25,7 +42,8 @@ jobs:
           - name: Production
     environment: ${{ matrix.environment.name }}
     concurrency:
-      group: ${{ github.workflow }}-${{ github.job }}-${{ matrix.environment.name }}
+      group: ${{ github.workflow }}-synchronize-${{ matrix.environment.name }}-${{ needs.prepare.outputs.component-name }}
```
mxyng, line 45: `github.job` is only available in context of `steps` so hard code it here to `synchronize`

.github/workflows/deploy.yaml
```diff
@@ -25,7 +42,8 @@ jobs:
...
+      cancel-in-progress: true
```
mxyng, line 46: Workflow status is `waiting` when waiting for approvals and (presumably) during the wait timer so this is necessary. Otherwise new jobs will be pending until previous workflows have completed or expired. 

This may be a problem if multiple workflows are created in quick succession and cancel each other while actually running but we'll figure it out when it gets to that point.

BruceMacD (approved):

.github/workflows/deploy.yaml
BruceMacD: Does this seem like it will fix the "stuck waiting for workflows" problem I saw in #2841?

mxyng: I'm not sure why #2841 had waiting workflows. It's possible it erroneously created the deploy workflow so I've add a branch filter but there's no strong evidence of that from the actions history

PR #2916, opened 2022-08-11
fix: allow direct access key for idp user

### LINKED ISSUE #2895: api fails to update identity information from provider when using an access key directly
Create an access key for a user provided by Okta (or Google/Azure AD) and use it imperatively:

```
export INFRA_SERVER=infra.acme.com
export INFRA_ACCESS_KEY=x.y
infra login
infra use cluster
kubectl get pods
```

This will sometimes fail with 401 since when trying to sync with the identity provider

```
{"level":"info","error":"unauthorized: failed to update identity info from provider: record not found","method":"POST","path":"/api/tokens","statusCode":401,"remoteAddr":"192.168.205.197:33360","time":1660086393611,"caller":"server/routes.go:170","message":"api request error"}
```

BruceMacD:
## Summary
Moving this check for the provider sync fixes a bug where issuing an access key directly to a user from an identity provider did not work. Also added test coverage to the `/tokens` endpoint. 

## Checklist

## Related Issues

Resolves #2895

### DIFF
--- a/internal/server/tokens_test.go
+++ b/internal/server/tokens_test.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -10,6 +11,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"gotest.tools/v3/assert"
 
+	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal/access"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
@@ -19,9 +21,6 @@ import (
 func TestAPI_CreateToken(t *testing.T) {
 	srv := setupServer(t, withAdminUser)
 	routes := srv.GenerateRoutes(prometheus.NewRegistry())
-
-	_, err := data.InitializeSettings(srv.db)
-	assert.NilError(t, err)
 
 	type testCase struct {
 		setup    func(t *testing.T, req *http.Request)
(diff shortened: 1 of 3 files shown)


internal/server/tokens_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/infrahq/infra/internal/server/data"
+	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/internal/server/providers"
+)
+
+func TestAPI_CreateToken(t *testing.T) {
+	srv := setupServer(t, withAdminUser)
+	routes := srv.GenerateRoutes(prometheus.NewRegistry())
+
+	_, err := data.InitializeSettings(srv.db)
```
dnephin, line 23: This is unexported in `main`, and is done automatically as part of `setupServer`. Should hopefully be fine to just remove this line.

internal/server/tokens_test.go
```diff
@@ -0,0 +1,202 @@
...
+					ProviderID: data.InfraProvider(srv.db).ID,
+					ExpiresAt:  time.Now().Add(10 * time.Second),
+				}
+				accessKey, err := data.CreateAccessKey(srv.db, key)
+				assert.NilError(t, err)
+
+				req.Header.Set("Authorization", "Bearer "+accessKey)
+			},
+			expected: func(t *testing.T, resp *httptest.ResponseRecorder) {
+				assert.NilError(t, err)
```
dnephin, line 73: I think this is checking an error from outside the test case

```suggestion
```

internal/server/tokens_test.go
```diff
@@ -0,0 +1,202 @@
...
+				assert.Equal(t, resp.Code, http.StatusCreated)
```
dnephin, line 74: A check for the contents of `resp.Body` would be great too in one of these cases. Not blocking though, this is great additional test coverage.

dnephin (approved):
Looks great! There's a build failure, but should be an easy fix

Issue #2875, opened 2022-08-09
dynamic completion for commands (and flags)


j-sneh:
## Is your feature request related to a problem? Please describe.

Currently, the completions script only generates completions for commands/subcommands.

## Describe the solution you'd like

Generating the nouns dynamically by making an api call and getting the info.

## Describe alternatives you've considered

I don't think there's an alternative to making the corresponding list request. However, this should be enable-able/disable-able.

## Additional context

In order to have CLI settings, we would need to have some sort of cli config or expand the current config we have to cover things other than client - server connections.

PR #2918, opened 2022-08-11
feat: completions for `infra use`

j-sneh:
## Summary

As part of #2875, this pr gets the resources that the user has access to as dynamic completion nouns. I've tested this myself, and it's a little slow because it has to make an api call, but it's convenient and it works.

I also update the list command to use `listAll()`

## Checklist


BruceMacD (approved):
This is sweet 👍

PR #2919, opened 2022-08-11
feat: sign-up for orgs

### LINKED ISSUE #2822: Sign-up UI updates

BruceMacD:
## Summary
No changes from #2841, just re-opening to fix the actions which were stuck.

## Checklist

## Related Issues

Resolves #2822

### DIFF
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -2,6 +2,8 @@ package access
 
 import (
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"golang.org/x/crypto/bcrypt"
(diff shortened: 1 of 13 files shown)


internal/access/signup.go
```diff
@@ -13,38 +15,39 @@ import (
...
 
-	org := &models.Organization{Name: orgName}
+	err := checkPasswordRequirements(db, password)
+	if err != nil {
+		return nil, "", err
+	}
+
 	org.SetDefaultDomain()
+	// lower-case domains look better
+	org.Domain = strings.ToLower(org.Domain)
```
dnephin, line 29: I think it would be nice to do this in `SetDefaultDomain` so that the creating the domain values is all done in the same place.

internal/access/signup.go
```diff
@@ -13,38 +15,39 @@ import (
 
 // Signup creates a user identity using the supplied name and password and
 // grants the identity "admin" access to Infra.
-func Signup(c *gin.Context, name, password, orgName string) (*models.Identity, *models.Organization, error) {
+func Signup(c *gin.Context, keyExpiresAt time.Time, name, password string, org *models.Organization) (*models.Identity, string, error) {
```
dnephin, line 18: We'll probably want to create a struct to accept all these arguments pretty soon, but not a blocker

internal/access/signup.go
```diff
@@ -13,38 +15,39 @@ import (
...
 	// no authorization is setup yet
 	db := getDB(c)
 
-	org := &models.Organization{Name: orgName}
+	err := checkPasswordRequirements(db, password)
```
dnephin, line 22: Don't we need to do this after `CreateOrganization` , otherwise it's checking the password requirements for the default org, instead of the new org? Or is that by default? Maybe add a comment about that if it's by design?

I think this illustrates one of the disadvantages to just accepting org out of context. It makes behaviour like this less obvious. Although maybe we fix that later by not setting an org in the context for requests like this?

internal/access/signup.go
```diff
@@ -73,9 +76,22 @@ func Signup(c *gin.Context, name, password, orgName string) (*models.Identity, *
...
-			return nil, nil, err
+			return nil, "", fmt.Errorf("create grant on sign-up: %w", err)
 		}
 	}
 
-	return identity, org, nil
+	// grant the user a session on initial sign-up
+	accessKey := &models.AccessKey{
+		IssuedFor:         identity.ID,
+		IssuedForIdentity: identity,
```
dnephin, line 86: I think we only need to set `IssusedFor`, not `IssusedForIdentity`. Did you see failures when this was not set? Also not a blocker

dnephin (approved):
LGTM! 

Moving the domain lower casing to `SetDefaultDomain` would be good to do now I think, but otherwise just questions and ideas for future work.

I only reviewed the Go changes, not familiar with the UI changes

internal/access/signup.go
BruceMacD: I was bit torn on this part. I moved the check to after the org creation rather than some global minimum password requirements, that makes more sense.

internal/access/signup.go
BruceMacD: This part came from copy/pasting the login access key creation, the "IssuedForIdentity" is nil after save when you dont assign it so I assigned it to be safer.

PR #2920, opened 2022-08-11
fix: reuse connections with a properly-sized connection pool

ssoroka:
## Summary

- under load of more than one parallel connection, gorm is not reusing db connections and closing them immediately after each query, The `SetMaxIdleConns` value defaults to 2, so after 2 connections it starts closing idle connections immediately. This creates a lot of overhead. 
- also increases the connector poll time to 30s instead of 5s

tested manually and resolves the issue.

### DIFF
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -30,6 +30,7 @@ services:
 
   db:
     image: postgres:alpine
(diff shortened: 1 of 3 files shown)


dnephin (approved):
Nice, Looks good to me!

As Bruce would say, tracking these TODOs in a Github issue would be good so that we don't forget about them.

Issue #2903, opened 2022-08-10
Tracking issue for org TODOs


dnephin:
This issue is to track all the TODOs we add while working on adding organizations, so that we don't forget to re-enable some checks once all the code is setup to support orgs.

* [x] ~https://github.com/infrahq/infra/pull/2896#discussion_r942780752 - require an org from the context~ - #2945
* [ ]  - require all models to implement both `SetOrganizationID`, and `IsOrganizationMember() bool`.
* [x] ~https://github.com/infrahq/infra/pull/2902#discussion_r942792985 - require an orgID from the request~ This was fixed in a follow up .Replaced this item with another one below for the default behaviour when no org is specified by headers.
* [x] #2917
* [x] remove the "set default org" from `setupDB` in `server` (and other) packages. This causes tests to not match production.
* [ ] add tests cases for all handlers to show that an a request fails if the requestor (access key or domain) is for the wrong org
* [x] ~https://github.com/infrahq/infra/pull/2902/files#r943705460 - setting of the org in `data.initialize`.~ #2921
* [x] https://github.com/infrahq/infra/pull/2902/files#r943713595 - test coverage for org_id migration - #2961
* [x] error if domain from host header does not match the orgID of the access key #2911
* [x] ~reconsider behaviour of `unauthenticatedMiddleware` when no host header or infra-org header are set. Consider removing fallback to default org. Make this configurable so that we have default org fallback for on-prem.~  #2945
* [x] replace the regex in `SetDefaultDomain` with `strings.Map` #2957
* [x] add tests for `data.CreateOrganization`
* [ ] return fake data for unathenticated `GET` routes when in multi-tenancy mode and the org is unknown

dnephin:
I think we've either done all of these, or we have issues to track the remaining few, so going to close.

PR #2921, opened 2022-08-11
Make NewDB safer by not setting an org in the context

dnephin:
## Summary

Cleanup a couple things from #2902

* We don't need these calls to `SaveSettings` anymore, it's always done by `NewDB`
* It's risky to set an org in the context in `NewDB` because of how gorm clones the context. So add a new function that lets us pass the org explicitly in this case.

## Related Issues

Related to #2903

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -55,7 +55,8 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db *gorm.DB) error) (*DB, e
 type DB struct {
 	*gorm.DB // embedded for now to minimize the diff
 
-	DefaultOrg         *models.Organization
+	DefaultOrg *models.Organization
+	// DefaultOrgSettings are the settings for DefaultOrg
 	DefaultOrgSettings *models.Settings
 }
 
@@ -111,9 +112,7 @@ func initialize(db *DB) error {
 	}
 
 	db.DefaultOrg = org
-
-	db.Statement.Context = WithOrg(db.Statement.Context, org)
-	db.DefaultOrgSettings, err = GetSettings(db.DB)
+	db.DefaultOrgSettings, err = getSettingsForOrg(db.DB, org.ID)
 	return err
 }
 
--- a/internal/server/data/settings.go
+++ b/internal/server/data/settings.go
@@ -10,6 +10,7 @@ import (
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
 )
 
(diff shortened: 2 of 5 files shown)


internal/server/data/settings.go
```diff
@@ -62,9 +63,12 @@ func initializeSettings(db *gorm.DB) (*models.Settings, error) {
 
 func GetSettings(db *gorm.DB) (*models.Settings, error) {
 	org := OrgFromContext(db.Statement.Context)
+	return getSettingsForOrg(db, org.ID)
+}
 
+func getSettingsForOrg(db *gorm.DB, orgID uid.ID) (*models.Settings, error) {
 	var settings models.Settings
-	if err := db.Where("organization_id = ?", org.ID).First(&settings).Error; err != nil {
+	if err := db.Where("organization_id = ?", orgID).First(&settings).Error; err != nil {
```
BruceMacD, line 71: Are we moving away from using the `data.go` `get()` function? Trying to figure out why this is doing the call directly.

internal/server/data/data.go
```diff
@@ -106,9 +106,7 @@ func initialize(db *DB) error {
 	}
 
 	db.DefaultOrg = org
-
-	db.Statement.Context = WithOrg(db.Statement.Context, org)
-	db.DefaultOrgSettings, err = GetSettings(db.DB)
+	db.DefaultOrgSettings, err = getSettingsForOrg(db.DB, org.ID)
```
BruceMacD, line 109:
```suggestion
	db.OrgSettings, err = getSettingsForOrg(db.DB, org.ID)
```

This is a bit of a confusingly named field because these will no longer be the default settings once the default org modifies them

internal/server/data/settings.go
dnephin: Pretty soon we will be for sure, with the removal of gorm.  This I cherry-picked from `org_collab`, so I'm not sure exactly of the reason for doing it this way instead of the selector.  

Either way, I'll be replacing all of the SelectorFuncs and gorm stuff with raw sql, which will look more like this.

internal/server/data/data.go
dnephin: Ah, this is supposed to mean "the settings for the default org", not "the default settings". 

I could create a new struct , or a `Settings *models.Settings` field on `models.Organization` to store this, which would make this more obvious.

BruceMacD (approved):

internal/server/data/data.go
dnephin: I added a comment for now, hoping that helps.  This is only used for setting telemetry so it's possible we'll remove this soon.

PR #2928, opened 2022-08-12
Please review: error updates

kimskimchi:
## Summary

1. The Error object was not being used in a consistent manner
2. Failed due to logical/user errors were missing the `Error:` in the front, which might be confusing with successes that outputted that nothing changed
3. API errors used terms unknown on the CLI - sync terms with UI/CLI: `identity_id` is known as `user name` in the CLI, `credential` is a complete server-side term, so `create credential: a credential with that identity_id already exists` is not useful for a CLI user 
4. Lacking suggestions on how to fix the error 
5. Right now, the error messages regarding permission errors are generic (but still able to be viewed with log levels). Before updating it to be very specific, these are the things that would need to be addressed before we work with the api errors: 
- org changes (these will change auth errors more than just requiring specific type of roles
- which api call failed (will need to decide if these are done on the client side, as it might be misleading to have a different error (eg. 'listuser' permission error for 'createuser') as it looks like a bug, even though it is a feature: 
```
$ infra users **add**
Error: you do not have permission to **list** users, requires role admin, view, or connector
```
- documentation on which roles grant which permissions OR which api calls (it is not possible for a user to find out which api calls needs what roles through documentation (for infra roles)) 
- split up which server-logic failed with which role is required rather than a single string, so ui/cli can parse each as needed (eg. a list of roles that are accepted `[admin, connector, view]`, and a separate var for which backend action it is trying to do `[listUsers]` -- *we should also consider not suggesting connector role as a solution for missing grants on the cli* 

## Preview

1. Format API errors: 
```
$ infra users add a@g
Error: create credential: a credential with that identity_id already exists
--> 
DEBUG call server: create user named "a@g"
DEBUG create credential: a credential with that identity_id already exists
Error: user a@g already exists
```
2. Add `Error:` whenever the exit status is failure
```
$infra users rm c@g
No user named "c@g"
-->
Error: no user named "c@g"
```
3. Contain enough information about the api for server admins to debug, but concise messages for an average cli user 
```
$ infra users add d@g 
Error: you do not have permission to list users, requires role admin, view, or connector
--> 
$ infra users add d@g 
Error: missing permissions to run this command

Please make sure you have the correct grants; requires role view or admin
---or---
$ infra users add d@g --log-level=debug
DEBUG call server: create user named "d@g"
DEBUG you do not have permission to list users, requires role admin, view, or connector
Error: missing permissions to run this command

Please make sure you have the correct grants; requires role view or admin
```
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -63,12 +63,6 @@ $ infra users add johndoe@example.com`,
 
 			createResp, err := createUser(client, args[0])
 			if err != nil {
-				if api.ErrorStatusCode(err) == 403 {
-					logging.Debugf("%s", err.Error())
-					return Error{
-						Message: "Cannot add users: missing privileges for CreateUser",
-					}
-				}
 				return err
 			}
 
@@ -96,7 +90,7 @@ $ infra users edit janedoe@example.com --password`,
 		Args: ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if !editPassword {
-				return errors.New("Please specify a field to update. For options, run 'infra users edit --help'")
+				return Error{Suggestion: "Please specify a field to update. For options, run 'infra users edit --help'"}
 			}
 
 			return updateUser(cli, args[0])
@@ -152,6 +146,10 @@ func newUsersListCmd(cli *CLI) *cobra.Command {
 			logging.Debugf("call server: list users")
 			users, err := listAll(client.ListUsers, api.ListUsersRequest{})
 			if err != nil {
+				if api.ErrorStatusCode(err) == 403 {
+					logging.Debugf("%s", err.Error())
+					return unauthorizedError
+				}
 				return err
 			}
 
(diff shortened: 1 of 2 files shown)


internal/cmd/users.go
```diff
@@ -346,14 +330,19 @@ func updateUser(cli *CLI, name string) error {
 	return nil
 }
 
+// Throws error if none or multiple results are found
 func getUserByName(client *api.Client, name string) (*api.User, error) {
 	users, err := client.ListUsers(api.ListUsersRequest{Name: name})
 	if err != nil {
+		if api.ErrorStatusCode(err) == 403 {
+			logging.Debugf("%s", err.Error())
```
BruceMacD, line 338: Could we centralize this error debug logging somewhere?

BruceMacD (commented):
I really like the direction this is going 👍

internal/cmd/users.go
kimskimchi: Do you have any ideas? I was wondering about this too. Each client call /  response  to be logged everywhere, eg) 
instead of cli logging `call server: list users`, it would seem nicer if it was logged beside the call itself (eg. in the api/client.go itself?)

BruceMacD: I think that is a good idea. Looks like there is already a `checkError` function in `client.go`. The debug log could fit in there nicely.

PR #2930, opened 2022-08-12
Fix reading the domain from the request

dnephin:
## Summary

Best viewed by individual commit. I can remove any of these commits if there are concerns. I noticed a few other problems while working on this fix.

Fixes:
* permission required for the debug endpoint
* reading host header from the request, apparently some proxies remove the host header, so we have to use `req.Host`. The consequence of this change is that we no longer have a case where this is no host header. We plan on changing this later, and it is tracked in #2903 (second last item in the list currently).
* the postgres test helper so that `TestServerRun` uses postgres
* the wrong database connection was being used in the middleware

## Related Issues

Closes #2925

### DIFF
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -163,7 +163,8 @@ func orgRequired() gin.HandlerFunc {
 
 func setOrganizationInCtx(srv *Server) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		org, err := getOrgFromRequest(c.Request, srv.dataDB)
+		// TODO: use a transaction
+		org, err := getOrgFromRequest(c.Request, srv.dataDB.DB, srv.dataDB.DefaultOrg)
 		if err != nil {
 			c.Next()
 			return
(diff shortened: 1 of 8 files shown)


internal/server/middleware_test.go
```diff
@@ -359,36 +359,15 @@ func TestOrgLookupByDomain(t *testing.T) {
...
-	req.Header.Add("Host", "umbrella.infrahq.com")
+	req.Host = "umbrella.infrahq.com"
 
 	resp := httptest.NewRecorder()
 	router.ServeHTTP(resp, req)
 
 	assert.Equal(t, resp.Code, 200)
 }
 
-func TestMissingOrgErrors(t *testing.T) {
```
dnephin, line 370: This test was removed because this case is no longer possible now that we're using `req.Host` to get the host header.

In the future we'll restore this test case when we change the behaviour to be configurable, and  return fake data instead of the default org.

I'd also like to change these tests to use a real http server and client. That way the request is the server receives is actually parsed as it would be in production, instead of us attempting to create an `http.Request` that may or may not match production.

dnephin: I think we need these test changes to confirm this fix, so going to work on that now.

ssoroka: seems like this is still a valid use case for saas.

dnephin: Ya, we'll re-introduce this behaviour with the config setting

ssoroka (approved):

internal/server/middleware.go
```diff
@@ -163,7 +163,8 @@ func orgRequired() gin.HandlerFunc {
 
 func setOrganizationInCtx(srv *Server) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		org, err := getOrgFromRequest(c.Request, srv.dataDB)
+		// TODO: use a transaction
```
dnephin, line 166: The middleware changes that merged yesterday make this difficult once again. So I guess i'll need to address this in a follow up.

dnephin:
The original problem this was fixed has already been fixed in `main`, but there are important test fixes here, so I'm going to merge this for the test fixes and improvements.

PR #2931, opened 2022-08-12
fix: set key expiry during exchance

### LINKED ISSUE #2640: Default key TTL is too short to use with infra CLI
## Describe the bug

By default, infra CLI issues an access key for 720h (30d). The default session duration is also 720h (30d). There's a server-side check to ensure the new access key exchanged in a login does not expire after the original key. So if using all default values, all keys issued by infra CLI will be unusable except in raw HTTP requests

BruceMacD:
- set key expiry to match exchanged key if exchanged key is valid for less than default session
- return AuthenticatedIdentity struct from authn rather than many separate values

## Summary
Fixing this now because our ArgoCD automation is using this flow.

This change fixes the key exchange login flow to work as long as your access key is valid. When the expiry time of the access key being passed in is less than the server's default session duration, set the new key to match the expiry of the one being exchanged rather than failing.

## Checklist

## Related Issues

Resolves #2640

### DIFF
--- a/internal/server/authn/key_exchange.go
+++ b/internal/server/authn/key_exchange.go
@@ -7,9 +7,8 @@ import (
 
 	"gorm.io/gorm"
 
-	"github.com/infrahq/infra/internal"
+	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data"
-	"github.com/infrahq/infra/internal/server/models"
 )
 
 // keyExchangeAuthn allows exchanging a valid access key for new access key with a shorter lifetime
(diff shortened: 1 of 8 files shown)


internal/server/authn/key_exchange.go
```diff
@@ -25,22 +24,29 @@ func NewKeyExchangeAuthentication(requestingAccessKey string, requestedExpiry ti
 	}
 }
 
-func (a *keyExchangeAuthn) Authenticate(_ context.Context, db *gorm.DB) (*models.Identity, *models.Provider, AuthScope, error) {
+func (a *keyExchangeAuthn) Authenticate(_ context.Context, db *gorm.DB, requestedExpiry time.Time) (AuthenticatedIdentity, error) {
```
BruceMacD, line 27: Changes to this function should be the only actual logical change

dnephin (approved):
LGTM!

PR #2945, opened 2022-08-15
fix: checking hosts properly

ssoroka:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -24,12 +24,6 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, hostname string, details Sig
 	// no authorization is setup yet
 	db := getDB(c)
 
-	existingOrgs, err := data.ListOrganizations(db, &models.Pagination{Limit: 1}, data.NotName("Default"))
-	if err != nil {
-		return nil, "", err
-	}
-	isFirstOrg := len(existingOrgs) == 0
-
 	details.Org.GenerateDefaultDomain(hostname)
 
 	if err := data.CreateOrganization(db, details.Org); err != nil {
@@ -38,7 +32,7 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, hostname string, details Sig
 	db.Statement.Context = data.WithOrg(db.Statement.Context, details.Org)
 
 	// check the admin user's password requirements against our basic password requirements
-	err = checkPasswordRequirements(db, details.Password)
+	err := checkPasswordRequirements(db, details.Password)
 	if err != nil {
 		return nil, "", err
 	}
@@ -70,27 +64,14 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, hostname string, details Sig
 		return nil, "", fmt.Errorf("create credential on sign-up: %w", err)
 	}
 
-	grants := []*models.Grant{
-		{
-			Subject:   uid.NewIdentityPolymorphicID(identity.ID),
-			Privilege: models.InfraAdminRole,
-			Resource:  ResourceInfraAPI,
-			CreatedBy: identity.ID,
-		},
-	}
-	if isFirstOrg {
-		grants = append(grants, &models.Grant{
-			Subject:   uid.NewIdentityPolymorphicID(identity.ID),
-			Privilege: models.InfraSupportAdminRole,
-			Resource:  ResourceInfraAPI,
-			CreatedBy: identity.ID,
-		})
-	}
-
-	for _, grant := range grants {
-		if err := data.CreateGrant(db, grant); err != nil {
-			return nil, "", fmt.Errorf("create grant on sign-up: %w", err)
-		}
+	err = data.CreateGrant(db, &models.Grant{
+		Subject:   uid.NewIdentityPolymorphicID(identity.ID),
+		Privilege: models.InfraAdminRole,
+		Resource:  ResourceInfraAPI,
+		CreatedBy: identity.ID,
+	})
+	if err != nil {
+		return nil, "", fmt.Errorf("create grant on sign-up: %w", err)
 	}
 
 	// grant the user a session on initial sign-up
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -20,11 +20,17 @@ type SignupDetails struct {
 
 // Signup creates a user identity using the supplied name and password and
 // grants the identity "admin" access to Infra.
-func Signup(c *gin.Context, keyExpiresAt time.Time, details SignupDetails) (*models.Identity, string, error) {
+func Signup(c *gin.Context, keyExpiresAt time.Time, hostname string, details SignupDetails) (*models.Identity, string, error) {
 	// no authorization is setup yet
 	db := getDB(c)
 
-	details.Org.SetDefaultDomain()
+	existingOrgs, err := data.ListOrganizations(db, &models.Pagination{Limit: 1}, data.NotName("Default"))
+	if err != nil {
+		return nil, "", err
+	}
+	isFirstOrg := len(existingOrgs) == 0
+
+	details.Org.GenerateDefaultDomain(hostname)
 
 	if err := data.CreateOrganization(db, details.Org); err != nil {
 		return nil, "", fmt.Errorf("create org on sign-up: %w", err)
@@ -32,7 +38,7 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, details SignupDetails) (*mod
 	db.Statement.Context = data.WithOrg(db.Statement.Context, details.Org)
 
 	// check the admin user's password requirements against our basic password requirements
-	err := checkPasswordRequirements(db, details.Password)
+	err = checkPasswordRequirements(db, details.Password)
 	if err != nil {
 		return nil, "", err
 	}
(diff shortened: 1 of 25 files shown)


internal/access/signup.go
```diff
@@ -71,12 +77,14 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, details SignupDetails) (*mod
 			Resource:  ResourceInfraAPI,
 			CreatedBy: identity.ID,
 		},
-		{
+	}
+	if isFirstOrg {
```
ssoroka, line 81: don't give support-admin to all orgs!

internal/server/data/data.go
```diff
@@ -177,6 +177,9 @@ func setOrg(db *gorm.DB, model any) {
 	}
 
 	org := OrgFromContext(db.Statement.Context)
+	if org == nil {
+		panic("missing org id in context")
+	}
```
ssoroka, line 182: these panics are programmer errors, and should be obvious from tests. If you remove them, it will still panic, but with a worse message.

internal/server/middleware.go
```diff
@@ -138,19 +149,40 @@ func unauthenticatedMiddleware(srv *Server) gin.HandlerFunc {
...
+	return func(c *gin.Context) {
+		org := data.OrgFromContext(c.Request.Context())
+
+		if org == nil {
+			sendAPIError(c, internal.ErrBadRequest)
+		}
+	}
+}
+
+func setOrganizationInCtx(srv *Server) gin.HandlerFunc {
```
ssoroka, line 162: if you want more info on this workflow, see https://mermaid.live/edit#pako:eNp1k01u2zAQha8y4KoFnAsYRQJachJvGqDpppC8YMiRRVgiVf4gCGTfvUPJsdk40YoYfHzz5g01MmkVsiVrOvsqW-EC_C5rA_TxamOk7bXZwS_8G9GHLdzc3AIvajMTPr7snBha4DG0aIKWImhrQBgF1u1Atij3M5q-y4nzcW3UYLUJ-V1Ud8ccmtod_qA_AF9VjdAd6GZSNjaAx7C9pn_aA6xGfFd35Fw79OlWLr7KtIuxtESc7_g4DN0bNGKPoEQQ8OPF3b6SSSCnDkF76j_ZmEj9v-3ioryuEkS4HVIuoltAS9l06GCHBh1N7LM2NFuvvafAt5-FVpwH5KurQdLYZfV0imYKHtVJhgabDx-2luhn7FAmc1mSxbjxX47HLzaAl4n0QvjPiHlvD1X0eFZKsvnSyhx9rL5YbZm1vJ_0FDYiduGj3P38QDdZ6fG69HBd2nyTlt6gifg9r84gP8fIFqxH1wut6I8ZU7lm9CZ6rNmSjkq4fc1qcyQuDrRTXCsdrGPLRnQeF0zEYJ_fjGTL4CK-Q6UWtJD-RB3_AcFeE6I

internal/server/models/organization.go
```diff
@@ -28,6 +28,7 @@ func (o *Organization) ToAPI() *api.Organization {
 }
 
 var domainNameReplacer = regexp.MustCompile(`[^\da-zA-Z-]`)
+var HostDomain = "infrahq.com" // TODO: configure from settings
```
ssoroka, line 31: need to figure out what to do here before merge

internal/server/tls.go
```diff
@@ -62,12 +62,16 @@ func tlsConfigFromOptions(
 
 	if opts.CA != "" {
 		raw := pemDecode([]byte(opts.CA))
-		logging.L.Info().
-			Str("SHA256 fingerprint", certs.Fingerprint(raw)).
-			Msg("TLS CA")
-
-		if !roots.AppendCertsFromPEM([]byte(opts.CA)) {
-			logging.Warnf("failed to load TLS CA, invalid PEM")
+		if len(raw) == 0 {
```
ssoroka, line 65: added some checks in this file to avoid panics I ran into.

internal/server/debug_test.go
```diff
@@ -72,14 +71,13 @@ func TestAPI_PProfHandler(t *testing.T) {
 				key, user := createAccessKey(t, s.db, "user2@example.com")
 				err := data.CreateGrant(s.db, &models.Grant{
 					Subject:   user.PolyID(),
-					Privilege: models.InfraAdminRole,
+					Privilege: models.InfraSupportAdminRole,
```
pdevine, line 74: Do you want to create both roles?

internal/server/middleware.go
```diff
@@ -80,7 +80,25 @@ func authenticatedMiddleware(srv *Server) gin.HandlerFunc {
 				return
 			}
 
-			c.Request = c.Request.WithContext(data.WithOrg(c.Request.Context(), authned.Organization))
+			org := data.OrgFromContext(c.Request.Context())
+
+			if org == nil {
+				org = authned.Organization
```
pdevine, line 86: This reads a little weird. Maybe:
```
if org == nil {
  if authned.Organization != nil {
    org = authned.Organization
  } else {
    ... (Org is req'd)
    return
  }
}
```

internal/server/models/organization.go
pdevine, line 31: We should probably add this before releasing the next version.

internal/server/routes.go
```diff
@@ -95,24 +96,30 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
...
 
-	get(a, noAuthn, "/api/version", a.Version)
+	post(a, noAuthnWithOrg, "/api/login", a.Login)
+	post(a, noAuthnWithOrg, "/api/password-reset-request", a.RequestPasswordReset)
+	post(a, noAuthnWithOrg, "/api/password-reset", a.VerifiedPasswordReset)
 
-	get(a, noAuthn, "/api/settings", a.GetSettings)
-	put(a, authn, "/api/settings", a.UpdateSettings)
-	add(a, noAuthn, route[api.EmptyRequest, WellKnownJWKResponse]{
+	get(a, noAuthnWithOrg, "/api/providers/:id", a.GetProvider)
```
pdevine, line 120: There's probably no way for someone to guess a provider ID, but would we want to also return fake data here?

internal/server/migrations.go
```diff
@@ -16,3 +23,15 @@ func (a *API) addRewrites() {
 // addRedirects for API endpoints that have moved to a different path
 func (a *API) addRedirects() {
 }
+
+func (a *API) deprecatedRoutes(noAuthnNoOrg *gin.RouterGroup) {
```
pdevine, line 27: This feels weird since there would be no way to add deprecated routes to other RouterGroups.

pdevine (commented):
Looks good. This will help quite a bit with RLS because the current org in the DB transaction will get set up the same way we're setting it here.

internal/server/data/organization.go
```diff
@@ -14,9 +14,8 @@ type orgCtxKey struct{}
 // OrgFromContext returns the Organization stored using WithOrg.
 func OrgFromContext(ctx context.Context) *models.Organization {
 	org, ok := ctx.Value(orgCtxKey{}).(*models.Organization)
-	if !ok || org == nil {
-		// TODO(orgs): panic("no org")
-		return &models.Organization{}
+	if !ok {
```
BruceMacD, line 17: Should we panic in this central place rather than the callers?

internal/server/models/organization.go
BruceMacD: Any ideas other than a server setting? Feels like it should be required.

ssoroka: I think we should get it into the config, ideally required

internal/access/signup.go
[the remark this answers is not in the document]
ssoroka: 💯 agree with that

internal/server/debug_test.go
ssoroka: I think the test just requires the one

internal/server/middleware.go
ssoroka: is it possible that authned.Organization could ever be missing? what if the org is deleted after the key is issued?

internal/server/routes.go
ssoroka: we might need to. Not entirely sure yet. provider IDs are somewhat guessable if you have an unlimited number of guesses. Also I think there might be problems with generating reliably-convincing fake data. especially since we're open source.

internal/server/migrations.go
ssoroka: yeah, you'd have to pass in the other router group as an argument. I could pass in the base router group and let you redefine your own middleware etc, but that felt weird too

internal/server/data/data_test.go
```diff
@@ -50,15 +49,11 @@ func postgresDriver(t *testing.T) gorm.Dialector {
...
-		sqlite, err := NewSQLiteDriver(filepath.Join(tmp, t.Name()))
-		assert.NilError(t, err, "sqlite driver")
-		run(t, setupDB(t, sqlite))
-	})
 	t.Run("postgres", func(t *testing.T) {
 		pgsql := postgresDriver(t)
-		run(t, setupDB(t, pgsql))
+		db := setupDB(t, pgsql)
+		run(t, db)
+		db.Rollback()
```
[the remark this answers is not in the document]
ssoroka: afaict it shouldn't need to be closed.

internal/access/signup.go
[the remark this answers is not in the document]
ssoroka: Because we wanted to keep the configuration simple, it's currently not required to set up your first org in config. should we change this?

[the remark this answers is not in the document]
ssoroka: a reasonable suggestion

BruceMacD (approved):
looks good 👍

PR #2950, opened 2022-08-16
fix: do not make external API calls in tests

### LINKED ISSUE #2949: IDP test should not require external call to get OIDC info
## Describe the bug
Tests loading config are calling out to a demo okta endpoint and attempting to load info. 

## Expected behavior
It should mock the response from this endpoint.

See `internal/server/config_test.go`

BruceMacD:
## Summary
- Remove domains that may be called externally in the case of a bad test configuration
- Remove migration test that makes external call

## Checklist

## Related Issues

Resolves #2949

### DIFF
--- a/internal/server/config_test.go
+++ b/internal/server/config_test.go
@@ -316,9 +316,11 @@ func TestLoadConfigInvalid(t *testing.T) {
 		"MissingProviderName": {
 			Providers: []Provider{
 				{
-					URL:          "demo.okta.com",
+					URL:          "example.com",
 					ClientID:     "client-id",
 					ClientSecret: "client-secret",
+					AuthURL:      "example.com/auth",
+					Scopes:       []string{"openid", "email"},
 				},
 			},
 		},
@@ -335,8 +337,10 @@ func TestLoadConfigInvalid(t *testing.T) {
 			Providers: []Provider{
 				{
 					Name:         "okta",
-					URL:          "demo.okta.com",
+					URL:          "example.com",
 					ClientSecret: "client-secret",
+					AuthURL:      "example.com/auth",
+					Scopes:       []string{"openid", "email"},
 				},
 			},
 		},
@@ -344,8 +348,10 @@ func TestLoadConfigInvalid(t *testing.T) {
 			Providers: []Provider{
 				{
 					Name:     "okta",
-					URL:      "demo.okta.com",
+					URL:      "example.com",
 					ClientID: "client-id",
+					AuthURL:  "example.com/auth",
+					Scopes:   []string{"openid", "email"},
 				},
 			},
 		},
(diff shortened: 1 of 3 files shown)


internal/server/config_test.go
```diff
@@ -691,9 +699,11 @@ func TestLoadConfigPruneConfig(t *testing.T) {
 		Providers: []Provider{
 			{
 				Name:         "okta",
-				URL:          "new-demo.okta.com",
+				URL:          "new.example.com",
 				ClientID:     "new-client-id",
 				ClientSecret: "new-client-secret",
+				AuthURL:      "new.example.com/auth",
```
BruceMacD, line 705: These missing from this test were causing a failure due to an external call

internal/server/data/migrations_test.go
```diff
@@ -166,31 +163,28 @@ INSERT INTO providers (id, created_at, updated_at, deleted_at, name, url, client
...
+					actual = append(actual, p)
 				}
 
-				assert.Equal(t, len(providers), 2)
-				authUrls := make(map[string]string)
-				scopes := make(map[string][]string)
-				for _, p := range providers {
-					authUrls[p.Name] = p.AuthURL
-					scopes[p.Name] = p.Scopes
+				// updated := parseTime(t, "2022-07-05 17:13:14.172568+00")
```
dnephin, line 179: left over debugging ?
```suggestion
```

internal/server/data/migrations_test.go
```diff
@@ -151,12 +150,10 @@ func TestMigrations(t *testing.T) {
 			},
 		},
 		{
-			// this test does an external call to example.okta.com, if it fails check your network connection
 			label: testCaseLine("202206281027"),
 			setup: func(t *testing.T, db *gorm.DB) {
 				stmt := `
 INSERT INTO providers (id, created_at, updated_at, deleted_at, name, url, client_id, client_secret, kind, created_by) VALUES (67301777540980736, '2022-07-05 17:13:14.172568+00', '2022-07-05 17:13:14.172568+00', NULL, 'infra', '', '', 'AAAAEIRG2/PYF2erJG6cYHTybucGYWVzZ2NtBDjJTEEbL3Jvb3QvLmluZnJhL3NxbGl0ZTMuZGIua2V5DGt4MdtlZuxOUhZQTw', 'infra', 1);
-INSERT INTO providers (id, created_at, updated_at, deleted_at, name, url, client_id, client_secret, kind, created_by) VALUES (67301777540980737, '2022-07-05 17:13:14.172568+00', '2022-07-05 17:13:14.172568+00', NULL, 'okta', 'example.okta.com', 'client-id', 'AAAAEIRG2/PYF2erJG6cYHTybucGYWVzZ2NtBDjJTEEbL3Jvb3QvLmluZnJhL3NxbGl0ZTMuZGIua2V5DGt4MdtlZuxOUhZQTw', 'okta', 1);
```
dnephin, line 159: If we remove this, we are no longer testing the migration. We can't show that a non-infra provider has a kind set.  Can we keep it?

BruceMacD: Kind being set is tested separately in `202206151027`. I'll add this back for now though, this migration will probably be removed soon anyway 👍

internal/server/data/migrations_test.go
BruceMacD: it was, thanks

internal/server/data/migrations_test.go
BruceMacD: Brought it back

dnephin (approved):
Looks good!

PR #2952, opened 2022-08-17
maintain: clean up org db context setting

BruceMacD:
- rename CreateOrganization to reflect its functionality
- remove unneeded db statement context setting
- do not set db context in middleware test
- remove the "set default org" from setupDB in server (and other) packages

## Summary
Cleaning up some org creation and context setting. 

## Checklist

## Related Issues

Part of #2903

### DIFF
--- a/internal/server/data/organization_test.go
+++ b/internal/server/data/organization_test.go
@@ -1,0 +1,40 @@
+package data
+
+import (
+	"testing"
+
+	"github.com/infrahq/infra/internal/server/models"
+	"gotest.tools/v3/assert"
+)
+
+func TestCreateOrganizationAndSetContext(t *testing.T) {
+	pgsql := postgresDriver(t)
+	db := setupDB(t, pgsql)
+
+	org := &models.Organization{Name: "syndicate", Domain: "syndicate-123"}
+
+	err := CreateOrganizationAndSetContext(db, org)
+	assert.NilError(t, err)
+
+	// db context is set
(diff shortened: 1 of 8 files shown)


internal/server/data/organization_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"reflect"
+	"testing"
+
+	gocmp "github.com/google/go-cmp/cmp"
+	"gotest.tools/v3/assert"
+
+	"github.com/infrahq/infra/internal/server/models"
+)
+
+var cmpOrgShallow = gocmp.Comparer(func(x, y models.Organization) bool {
```
dnephin, line 13: Shallow comparisons are great for testing things like `List` where we don't care too much about every field, but for `Create` I think we need to make sure every field is checked. 

I think you can use `cmpModel` (in `settings_test.go`) to handle the fields we can't predict.

internal/server/middleware_test.go
```diff
@@ -43,7 +43,6 @@ func setupDB(t *testing.T) *data.DB {
 		assert.NilError(t, db.Close())
 	})
 
-	db.Statement.Context = data.WithOrg(db.Statement.Context, db.DefaultOrg)
```
dnephin, line 46: Nice! Now we only need to remove the similar call in `Initialize`

dnephin (approved):
LGTM!  We might need to change the signature of `CreateOrganization` when we remove gorm, but this is much more clear about how it works currently

PR #2953, opened 2022-08-17
fix: signup redirect should expect full domain

ssoroka:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -84,6 +84,7 @@ func newServerCmd() *cobra.Command {
 	cmd.Flags().Duration("session-duration", 0, "Maximum session duration per user login")
 	cmd.Flags().Duration("session-extension-deadline", 0, "A user must interact with Infra at least once within this amount of time for their session to remain valid")
 	cmd.Flags().Bool("enable-signup", false, "Enable one-time admin signup")
+	cmd.Flags().String("hostname", "example.com", "hostname for the server")
 
 	return cmd
 }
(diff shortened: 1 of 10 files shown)


BruceMacD (approved):

ui/pages/signup/index.js
```diff
@@ -39,9 +39,7 @@ export default function Signup() {
       // redirect to the new org subdomain
       let created = await res.json()
 
-      window.location = `${window.location.protocol}//${
-        created?.organization?.domain
-      }.${window.location.host.replace('www.', '')}`
+      window.location = `${window.location.protocol}//${created?.organization?.domain}`
```
BruceMacD, line 42: Just confirming, this returns the domain name with the base URL correct?

Ex:
`acme-123.app.infrahq.com`

ssoroka: it does. You have to set the base domain with `--hostname infrahq.com` or similar in the config

internal/cmd/server.go
```diff
@@ -84,7 +84,7 @@ func newServerCmd() *cobra.Command {
 	cmd.Flags().Duration("session-duration", 0, "Maximum session duration per user login")
 	cmd.Flags().Duration("session-extension-deadline", 0, "A user must interact with Infra at least once within this amount of time for their session to remain valid")
 	cmd.Flags().Bool("enable-signup", false, "Enable one-time admin signup")
-	cmd.Flags().String("hostname", "example.com", "hostname for the server")
+	cmd.Flags().String("hostname", "", "hostname for the server")
```
dnephin, line 87: I don't think we should be adding more flags for `infra server`. We should use env vars or config files for setting server options. Command line flags work well for interactive commands run by humans. Env vars and config files work well for long running processes that are generally run by other processes.

This can be set with `INFRA_SERVER_HOSTNAME`

ssoroka: my only concern there is it's pretty much a required field, as saas won't function without it. env variables always seem more like optional things to me.

PR #2955, opened 2022-08-17
fix: make linter happy

ssoroka:
linter is sad. make linter happy


dnephin (commented):
These changes are for Go1.19 gofmt. I made them over in #2954 while upgrading us to use go1.19.

PR #2957, opened 2022-08-17
Remove regex replacement from org domain generation

dnephin:
## Summary

With the changes in #3066, all we need to do here is lowercase the string. So we can remove the regex replacement.

### DIFF
--- a/internal/access/organization.go
+++ b/internal/access/organization.go
@@ -1,7 +1,6 @@
 package access
 
 import (
-	"regexp"
 	"strings"
 
 	"github.com/gin-gonic/gin"
@@ -56,14 +55,6 @@ func DeleteOrganization(c *gin.Context, id uid.ID) error {
 	return data.DeleteOrganizations(db, data.ByID(id))
 }
 
-var domainNameReplacer = regexp.MustCompile(`[^\da-zA-Z-]`)
-
 func SanitizedDomain(subDomain, serverBaseDomain string) string {
-	sanitizedDomain := domainNameReplacer.ReplaceAllStringFunc(subDomain, func(s string) string {
-		if s == " " {
-			return "-"
-		}
-		return ""
-	})
-	return strings.ToLower(sanitizedDomain) + "." + serverBaseDomain
+	return strings.ToLower(subDomain) + "." + serverBaseDomain
 }
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -57,6 +57,8 @@ type Options struct {
 	EmailFromName    string
 	SendgridApiKey   string
 
+	// BaseDomain of the server, which is appended to the organization slug to
+	// create a unique hostname for each organization.
 	BaseDomain string
 
 	Keys    []KeyProvider
@@ -113,9 +115,6 @@ type Addrs struct {
 
 // newServer creates a Server with base dependencies initialized to zero values.
 func newServer(options Options) *Server {
-	if options.BaseDomain == "" {
-		options.BaseDomain = "example.com"
-	}
 	return &Server{
 		options: options,
 		secrets: map[string]secrets.SecretStorage{},
--- a/internal/server/organizations_test.go
+++ b/internal/server/organizations_test.go
@@ -16,12 +16,7 @@ import (
 func createOrgs(t *testing.T, db data.GormTxn, orgs ...*models.Organization) {
 	t.Helper()
 	for i := range orgs {
-		o, err := data.GetOrganization(db, data.ByName(orgs[i].Name))
-		if err == nil {
-			*orgs[i] = *o
-			continue
-		}
-		err = data.CreateOrganization(db, orgs[i])
+		err := data.CreateOrganization(db, orgs[i])
 		assert.NilError(t, err, orgs[i].Name)
 	}
 }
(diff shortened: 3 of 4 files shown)


internal/access/organization.go
```diff
@@ -44,6 +48,25 @@ func CreateOrganization(c *gin.Context, org *models.Organization) error {
 	return data.CreateOrganization(db, org)
 }
 
+func GenerateOrganizationDomain(orgName string, serverHostname string) string {
+	replaceSymbols := func(r rune) rune {
+		switch {
+		case unicode.IsLetter(r), unicode.IsNumber(r):
```
BruceMacD, line 54: This will return true for some characters that may not be URL safe in all scenarios. 

Ex:
```
unicode.IsLetter('ð') // true
```

This may work but also presents some possible complications. I think its safer to follow [RFC2396](https://www.ietf.org/rfc/rfc2396.html#section-2.3).
```
   alpha    = lowalpha | upalpha

   lowalpha = "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" | "i" |
              "j" | "k" | "l" | "m" | "n" | "o" | "p" | "q" | "r" |
              "s" | "t" | "u" | "v" | "w" | "x" | "y" | "z"

   upalpha  = "A" | "B" | "C" | "D" | "E" | "F" | "G" | "H" | "I" |
              "J" | "K" | "L" | "M" | "N" | "O" | "P" | "Q" | "R" |
              "S" | "T" | "U" | "V" | "W" | "X" | "Y" | "Z"

   digit    = "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" |
              "8" | "9"

   alphanum = alpha | digit
```

internal/server/server.go
```diff
@@ -107,9 +109,6 @@ type Addrs struct {
 
 // newServer creates a Server with base dependencies initialized to zero values.
 func newServer(options Options) *Server {
-	if options.Hostname == "" {
```
BruceMacD, line 110: I think it is better to have some default here. Maybe localhost?

internal/access/organization.go
dnephin: True, I cut some corners here because our validation for this field should exclude all other characters. But that validation could change without someone realizing to update this function, so it would be good to be more strict.

I'll change this to `r >= 'a' && r <= 'z'`, etc

internal/server/server.go
dnephin: I think this is the wrong place to set defaults, they should go into `defaultServerOptions`. 

What purpose does this default serve? I was going to move it to `defaultServerOptions` until I realized no tests were failing, and I couldn't find any reason to set a default. 

If we have a reason for a default I can add a comment about it, but in general I think we are better off trying to make things work with a zero value default, or to at least know exactly why we can't use a zero value default.

internal/access/organization.go
dnephin: I made those changes

internal/server/server.go
BruceMacD: My guess was just that it's easier to see that you need to set a hostname when the base URL becomes example.com, but you're right that it doesn't match other places

BruceMacD (approved):

internal/access/organization.go
```diff
@@ -56,14 +55,17 @@ func DeleteOrganization(c *gin.Context, id uid.ID) error {
...
-
 func SanitizedDomain(subDomain, serverBaseDomain string) string {
-	sanitizedDomain := domainNameReplacer.ReplaceAllStringFunc(subDomain, func(s string) string {
-		if s == " " {
-			return "-"
+	replaceSymbols := func(r rune) rune {
+		switch {
+		case r >= 'a' && r <= 'z', r >= 'A' && r <= 'Z', r >= '0' && r <= '9':
+			return r
+		case r == ' ' || r == '-':
```
mxyng, line 63: Maybe `_` as well since it might be used as a separator?

mxyng (approved):

internal/access/organization.go
dnephin: Sounds like a good idea to me. It's different from the regex behaviour, but I'm ok to make that change if there are no concerns.

internal/server/server.go
dnephin: We discussed this a bit in slack recently. I think #3072 is a better option for setting a default org domain, instead of relying on a default in code that will never match production.

internal/access/organization.go
dnephin: We've updated the validation in #3066 to remove `_` from the valid character list, so I think we can omit it here.

dnephin:
I've updated this PR to lowercase the domain name.

internal/access/organization.go
```diff
@@ -56,14 +56,19 @@ func DeleteOrganization(c *gin.Context, id uid.ID) error {
...
+			return '-'
+		default:
+			return -1 // remove rune from the string
 		}
-		return ""
-	})
-	return strings.ToLower(sanitizedDomain) + "." + serverBaseDomain
+	}
+	slug := strings.Map(replaceSymbols, subDomain)
+	return strings.ToLower(slug) + "." + serverBaseDomain
```
dnephin, line 73: Oh, I forgot we did this unconditionally at the end... I guess we don't have to `ToLower` the individual runes

pdevine (approved):
LGTM.

PR #2959, opened 2022-08-17
fix: create connector identity for org

### LINKED ISSUE #2946: connector creation skipped in org signup

BruceMacD:
## Summary
When an org signs up they need a connector identity to start adding destinations. Create it by default.

Tests dependent on #2952 

## Checklist

## Related Issues

Resolves #2946

### DIFF
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -348,7 +348,7 @@ func updateUser(cli *CLI, name string) error {
 }
 
 func getUserByNameOrID(client *api.Client, name string) (*api.User, error) {
-	users, err := client.ListUsers(api.ListUsersRequest{Name: name})
+	users, err := client.ListUsers(api.ListUsersRequest{Name: name, ShowSystem: true})
 	if err != nil {
 		return nil, err
 	}
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -348,7 +348,13 @@ func updateUser(cli *CLI, name string) error {
 }
 
 func getUserByNameOrID(client *api.Client, name string) (*api.User, error) {
-	users, err := client.ListUsers(api.ListUsersRequest{Name: name, ShowSystem: true})
+	showSystem := false
+
+	if name == "connector" {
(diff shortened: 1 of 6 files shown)


internal/cmd/users.go
```diff
@@ -348,7 +348,13 @@ func updateUser(cli *CLI, name string) error {
 }
 
 func getUserByNameOrID(client *api.Client, name string) (*api.User, error) {
-	users, err := client.ListUsers(api.ListUsersRequest{Name: name})
+	showSystem := false
+
+	if name == "connector" {
```
pdevine, line 353: Did you want to use `models.InternalInfraConnectorIdentityName` here as well?

internal/server/data/data_test.go
```diff
@@ -138,44 +138,51 @@ func TestDatabaseSelectors(t *testing.T) {
...
-		letters := make([]string, 0, 26)
+		alphabeticalIdentities := []string{}
 		for r := 'a'; r < 'a'+26; r++ {
-			letters = append(letters, string(r))
+			alphabeticalIdentities = append(alphabeticalIdentities, string(r))
 			g := &models.Identity{Name: string(r)}
 			assert.NilError(t, CreateIdentity(db, g))
 		}
 
+		alphabeticalIdentities = append(alphabeticalIdentities, models.InternalInfraConnectorIdentityName)
```
pdevine, line 148: Why add the connector identity? Seems like a lot of trouble for just seeing it paginate correctly.

pdevine (approved):
Overall looks good. The `TestPaginationSelector` test seems a little strange, but looks like it should work fine.

internal/cmd/users.go
BruceMacD: We use the name directly in `cmd` to avoid importing the models.

internal/server/data/data_test.go
BruceMacD: Good call-out, it made more sense to remove the connector identity from the returned list of all identities. Made that change.

PR #2962, opened 2022-08-18
Invite email

ssoroka:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -30,7 +30,7 @@ services:
 
   db:
     image: postgres:alpine
-    command: postgres -c log_connections=on -c log_min_messages=INFO
+    command: postgres -c log_connections=off -c log_min_messages=INFO
     restart: always
     user: postgres
     environment:
@@ -43,6 +43,6 @@ services:
       - /var/lib/postgresql/data
     healthcheck:
       test: ["CMD-SHELL", "pg_isready"]
-      interval: 10s
+      interval: 15s
       timeout: 5s
       retries: 5
--- a/internal/server/config.go
+++ b/internal/server/config.go
@@ -603,8 +603,14 @@ func (s Server) loadConfig(config Config) error {
 	if err := validate.Validate(config); err != nil {
 		return err
 	}
+	org, err := data.GetOrganization(s.DB(), data.ByName(models.DefaultOrganizationName))
+	if err != nil {
+		return err
+	}
 
 	return s.db.Transaction(func(tx *gorm.DB) error {
+		tx.Statement.Context = data.WithOrg(tx.Statement.Context, org)
+
 		// inject internal infra provider
 		config.Providers = append(config.Providers, Provider{
 			Name: models.InternalInfraProviderName,
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -425,13 +425,13 @@ func scopeUniqueIndicesToOrganization() *migrator.Migration {
 		ID: "2022-08-04T17:72",
 		Migrate: func(tx *gorm.DB) error {
 			stmt := `
-DROP INDEX idx_access_keys_name;
-DROP INDEX idx_credentials_identity_id;
-DROP INDEX idx_destinations_unique_id;
-DROP INDEX idx_grant_srp;
-DROP INDEX idx_groups_name;
-DROP INDEX idx_identities_name;
-DROP INDEX idx_providers_name;
+DROP INDEX IF EXISTS idx_access_keys_name;
+DROP INDEX IF EXISTS idx_credentials_identity_id;
+DROP INDEX IF EXISTS idx_destinations_unique_id;
+DROP INDEX IF EXISTS idx_grant_srp;
+DROP INDEX IF EXISTS idx_groups_name;
+DROP INDEX IF EXISTS idx_identities_name;
+DROP INDEX IF EXISTS idx_providers_name;
 `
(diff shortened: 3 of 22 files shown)


ui/pages/password-reset/index.js
```diff
@@ -80,44 +50,7 @@ export default function PasswordReset() {
...
-            >
-              Set Password
-            </button>
-            {error && (
-              <p className='absolute -bottom-3.5 mx-auto w-full text-center text-2xs text-pink-400'>
-                {error}
-              </p>
-            )}
-          </form>
+          <PasswordResetForm />
```
hoyyeva, line 53: is the component not in this PR?

docker-compose.dev.yaml
```diff
@@ -43,6 +43,6 @@ services:
       - /var/lib/postgresql/data
     healthcheck:
       test: ["CMD-SHELL", "pg_isready"]
-      interval: 10s
+      interval: 15s
```
BruceMacD, line 46: This seems like a good change, but was there a specific reason you found that required this change?

internal/server/config.go
```diff
@@ -603,8 +603,14 @@ func (s Server) loadConfig(config Config) error {
 	if err := validate.Validate(config); err != nil {
 		return err
 	}
+	org, err := data.GetOrganization(s.DB(), data.ByName(models.DefaultOrganizationName))
```
BruceMacD, line 606: We got rid of the ability to set the default organization name in config, correct? Just validating.

internal/server/data/migrations.go
```diff
@@ -445,18 +445,18 @@ CREATE UNIQUE INDEX idx_grant_srp ON grants (organization_id,subject,privilege,r
 CREATE UNIQUE INDEX idx_groups_name ON groups (organization_id,name) where (deleted_at is null);
 CREATE UNIQUE INDEX idx_identities_name ON identities (organization_id,name) where (deleted_at is null);
 CREATE UNIQUE INDEX idx_providers_name ON providers (organization_id,name) where (deleted_at is null);
-CREATE UNIQUE INDEX settings_org_id ON settings (organization_id) where deleted_at is null;
+CREATE UNIQUE INDEX IF NOT EXISTS settings_org_id ON settings (organization_id) where deleted_at is null;
```
BruceMacD, line 448: Should the indexes about this also be `IF NOT EXISTS`?

internal/server/passwordreset.go
```diff
@@ -23,16 +24,11 @@ func (a *API) RequestPasswordReset(c *gin.Context, r *api.PasswordResetRequest)
 		return nil, err
 	}
 
-	// TODO: adjust domain for orgs when here
-	link, _ := url.Parse(email.AppDomain)
-	link.Path = "/password-reset"
-	query := link.Query()
-	query.Add("token", token)
-	link.RawQuery = query.Encode()
+	org := data.MustGetOrgFromContext(c)
```
BruceMacD, line 27: This isn't set in middleware?

ui/pages/password-reset/index.js
ssoroka: I missed two files, ui/pages/accept-invite/index.js and ui/components/password-reset-form.js. added them now

internal/server/config.go
ssoroka: I guess we did.

internal/server/data/migrations.go
ssoroka: not sure what you mean

ui/components/password-reset-form.js
```diff
@@ -0,0 +1,77 @@
...
+          }`}
+        />
+      </div>
+      <button
+        disabled={!password}
+        className='mt-6 mb-2 rounded-lg border border-violet-300 px-4 py-3 text-2xs text-violet-100 hover:border-violet-100 disabled:pointer-events-none disabled:opacity-30'
+      >
+        Set Password
+      </button>
+      {error && (
```
hoyyeva, line 70: you can use `<ErrorMessage>` component to display the error message - so the UI looks consistent throughout the app

ui/components/password-reset-form.js
```diff
@@ -0,0 +1,77 @@
...
+      if (!res.ok) {
+        throw await res.json()
+      }
+
+      await res.json()
+
+      await mutate('/api/users/self')
+      router.replace('/')
+    } catch (e) {
+      setError(e)
```
hoyyeva, line 34: are we sure the `e` is always a string?

ssoroka: no idea! This code was copied from ui/pages/login/index.js

internal/server/passwordreset.go
BruceMacD: That was confusing, my bad. I was thinking this was setting the org on the db, but this looks fine 👍

internal/server/data/migrations.go
BruceMacD: The other `CREATE UNIQUE INDEX` commands *above* this (typo now edited)

BruceMacD (approved):
Back-end code looks good

ui/pages/password-reset/index.js
jmorganca: I think the import is missing in this file

ui/components/password-reset-form.js
jmorganca: `setError(e.message)`

PR #2966, opened 2022-08-18
Please Review: improve: refactor delete user/identity in the api

### LINKED ISSUE #2838: Consolidate duplicate code paths for deleting users
@j-sneh noticed this while working on #2767

Today we have the following:
* `access.DeleteIdentity` - handles removing credentials, access keys, and grants associated with the identity, and #2767 adds removal of `identities_group` references
* `data.DeleteIdentity` - only deletes the one row from `identities` table
* `data.DeleteIdentities` - handles removing grants, and #2767 adds removal of `identities_group` references
* `Server.loadUsers` (from server config) - calls `data.DeleteIdentities` but does not remove credentials, or access keys for the users
* `access.DeleteProviders` - calls `data.DeleteIdentities` and handles deleting access keys (doesn't need to delete credentials)

With #2767 merged it seems like the only gap may be `loadUsers` not deleting  of credentials or access keys. I'm not sure if that is by design.

The problem now is that there are two ways to delete users, in one code path the logic to delete associations is handled in `access`, and in the other code path it's handled in `data`. 

This issue is to clean this up so that at the very least the two flows are consistent about where associations are deleted, either `access` or `data`. Ideally we would also remove the need to have separate code paths for deleting one identity and deleting many identities, but I'm not sure if that's possible.

kimskimchi:
Addresses #2838

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -59,6 +59,56 @@ func InfraConnectorIdentity(c *gin.Context) *models.Identity {
 	return data.InfraConnectorIdentity(getDB(c))
 }
 
+// This checks if the delete request is allowed
+func CanDeleteIdentity(c *gin.Context, id uid.ID) error {
+	self, err := isIdentitySelf(c, id)
+	if err != nil {
+		return err
+	}
+	if self {
+		return fmt.Errorf("cannot delete self: %w", internal.ErrBadRequest)
+	}
+
+	if InfraConnectorIdentity(c).ID == id {
+		return fmt.Errorf("%w: the connector user can not be deleted", internal.ErrBadRequest)
+	}
+
+	return nil
+}
+
+func DeleteIdentity2(c *gin.Context, id uid.ID) error {
+	if err := CanDeleteIdentity(c, id); err != nil {
+		return err
+	}
+
+	db, err := RequireInfraRole(c, models.InfraAdminRole)
+	if err != nil {
+		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
+	}
+
+	return data.DeleteIdentity(db, id)
+}
+
+func DeleteCredentialForUser(c *gin.Context, userID uid.ID) error {
+	db, err := RequireInfraRole(c, models.InfraAdminRole)
+	if err != nil {
+		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
+	}
+
+	credential, err := data.GetCredential(db, data.ByIdentityID(userID))
+	if err != nil && !errors.Is(err, internal.ErrNotFound) {
+		return fmt.Errorf("get delete identity creds: %w", err)
+	}
+
+	if credential != nil {
+		err := data.DeleteCredential(db, credential.ID)
+		if err != nil {
+			return fmt.Errorf("delete identity creds: %w", err)
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


internal/server/users.go
```diff
@@ -107,3 +107,27 @@ func (a *API) UpdateUser(c *gin.Context, r *api.UpdateUserRequest) (*api.User, e
 func (a *API) DeleteUser(c *gin.Context, r *api.Resource) (*api.EmptyResponse, error) {
 	return nil, access.DeleteIdentity(c, r.ID)
 }
+
+func (a *API) DeleteUser2(c *gin.Context, r *api.Resource) (*api.EmptyResponse, error) {
```
BruceMacD, line 111: Is this just a temporary function name showing how you would like to update `DeleteUser`?

internal/server/users.go
```diff
@@ -107,3 +107,27 @@ func (a *API) UpdateUser(c *gin.Context, r *api.UpdateUserRequest) (*api.User, e
...
+
+	if err := access.DeleteCredentialForUser(c, r.ID); err != nil {
+		return nil, err
+	}
+
+	if err := access.DeleteGrantsForUser(c, r.ID); err != nil {
+		return nil, err
+	}
+
+	return nil, access.DeleteIdentity2(c, r.ID)
```
BruceMacD, line 132: Functions in handler files (like this one) should just transform data if needed then call the access package. The exception to this rule is `login` logic.

So all you'll need to do here is call `return nil, access.DeleteIdentity2(c, r.ID)`

internal/server/users.go
```diff
@@ -107,3 +107,27 @@ func (a *API) UpdateUser(c *gin.Context, r *api.UpdateUserRequest) (*api.User, e
...
+
+	if err := access.RemoveUserFromAllGroups(c, r.ID); err != nil {
+		return nil, err
+	}
+
+	if err := access.DeleteCredentialForUser(c, r.ID); err != nil {
+		return nil, err
+	}
+
+	if err := access.DeleteGrantsForUser(c, r.ID); err != nil {
```
BruceMacD, line 128: You can unify all this functionality in one `access.DeleteIdentity` function, it will be simpler to read and follow since the permissions required are the same for all. That function can live in the `internal/server/access/identity.go` file and look something like this:

```
func DeleteUser(c *gin.Context, userID uid.ID) error {
 // check infra admin role
 // check CanDeleteIdentity if not admin
 // delete the access keys for the user (directly in data)
 // remove the user from the groups  (directly in data)
 // delete the credential for the user  (directly in data)
 // delete the grants for the user  (directly in data)
}
```

internal/access/identity.go
```diff
@@ -59,6 +59,56 @@ func InfraConnectorIdentity(c *gin.Context) *models.Identity {
...
+	return data.DeleteIdentity(db, id)
+}
+
+func DeleteCredentialForUser(c *gin.Context, userID uid.ID) error {
+	db, err := RequireInfraRole(c, models.InfraAdminRole)
+	if err != nil {
+		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
+	}
+
+	credential, err := data.GetCredential(db, data.ByIdentityID(userID))
```
BruceMacD, line 98: It would be simpler to have the as functions in the data package since they aren't called directly by an API call.

internal/server/users.go
kimskimchi: yes! so you can see both :)

PR #2967, opened 2022-08-18
Accept Org ID from access key for unauthenticated endpoints

dnephin:
## Summary

Implemented the linked proposal by allowing access key to specify the org ID for unauthenticated endpoints.

Also:
* removes the unused `Infra-Organization` header, which is no longer safe now that org name is not unique
* fixes a TODO about using a transaction for a db query
* adds some godoc for the `SignupEnabled` server option

Making this a draft until I have written tests to cover all the cases.

Resolves https://github.com/infrahq/internal/issues/23

### DIFF
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -141,7 +141,20 @@ func withDBTxn(ctx context.Context, db *gorm.DB, fn func(tx *gorm.DB)) {
 func unauthenticatedMiddleware(srv *Server) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		withDBTxn(c.Request.Context(), srv.DB(), func(tx *gorm.DB) {
-			tx.Statement.Context = c.Request.Context() // TODO: remove with gorm
+			org, err := getOrgFromRequest(c.Request, tx)
+			if err != nil {
+				sendAPIError(c, err)
+				return
+			}
+
+			// TODO: use an explicit setting for this, don't overload EnableSignup
+			if org == nil && !srv.options.EnableSignup { // is single tenant
+				org = srv.db.DefaultOrg
+			}
+			if org != nil {
+				c.Request = c.Request.WithContext(data.WithOrg(c.Request.Context(), org))
+				tx.Statement.Context = c.Request.Context() // TODO: remove with gorm
+			}
 
 			rCtx := access.RequestContext{
 				Request: c.Request,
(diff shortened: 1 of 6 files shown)


internal/server/middleware.go
```diff
@@ -157,37 +185,12 @@ func orgRequired() gin.HandlerFunc {
...
 		if org == nil {
 			sendAPIError(c, internal.ErrBadRequest)
-		}
-	}
-}
-
-func setOrganizationInCtx(srv *Server) gin.HandlerFunc {
-	return func(c *gin.Context) {
-		// TODO: use a transaction
-		org, err := getOrgFromRequest(c.Request, srv.DB())
```
dnephin, line 167: I have to remove this middleware to resolve this TODO because the middleware was called before we started a DB transaction.

This was originally a bug in the changes I made to merge in the `org_collab` branch.

internal/server/middleware_test.go
```diff
@@ -373,8 +372,9 @@ func TestGetOrgFromRequest_FromRequestHost(t *testing.T) {
 	assert.Equal(t, resp.StatusCode, 200)
 }
 
-func TestGetOrgFromRequest_MissingOrgErrors(t *testing.T) {
+func TestGetOrgFromRequest_MissingOrgErrorsInMultiTenantEnvironment(t *testing.T) {
 	srv := setupServer(t, withAdminUser)
+	srv.options.EnableSignup = true // test multi-tenancy environment
```
dnephin, line 377: This test was not correctly testing the case we wanted because it did not use the `setOrganizationInCtx` middleware. Now that the middleware is integrated the test failed and required this fix.

internal/server/middleware.go
```diff
@@ -80,27 +80,13 @@ func authenticatedMiddleware(srv *Server) gin.HandlerFunc {
 				return
 			}
 
-			// Before changing anything here, make sure you read the org-selection workflow diagram.
-			// https://github.com/infrahq/infra/blob/main/docs/dev/organization-request-flow.md
```
dnephin, line 84: I moved this link into `unauthenticatedMiddleware` where we check if org is unset, now that #2968 has updated the diagram for that case.

ssoroka (approved):

internal/server/routes.go
```diff
@@ -106,24 +106,16 @@ func (s *Server) GenerateRoutes(promRegistry prometheus.Registerer) Routes {
 	get(a, noAuthnNoOrg, "/api/version", a.Version)
 	get(a, noAuthnNoOrg, "/api/server-configuration", a.GetServerConfiguration)
 
-	// while an org is required for these endpoints, they return fake data when the org is not supplied
-	noAuthnOptOrg := apiGroup.Group("/",
```
dnephin, line 110: I combined this with `noAuthnWithOrg` with org for now. I had made that change before I brought in these changes from `main` to fix a panic in one of the test cases.

dnephin (commented):
I've added the test cases, so I think this is now ready for review again.

internal/server/routes.go
ssoroka: there's no panic anymore. I do prefer to leave them separate as it's clear what endpoints need to be updated later.

dnephin: Are you sure the current grouping is correct? It seems strange that get provider would be in one group while list providers is in a different one. Also the jwk route should probably be in the same group.

If anyone adds routes between now and when we add fake data they likely won't have any idea which group it goes into, and could put it in the wrong group.

It seems like either way we'll have to change these groups later, so why split them now?

ssoroka (approved):

PR #2976, opened 2022-08-18
feat(helm): bump infra helm chart to 0.20.0

mxyng:
## Summary

Add helm readme and upgrade instructions from <0.20.0.

### DIFF
--- a/helm/charts/infra/Chart.yaml
+++ b/helm/charts/infra/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: infra
 description: A Helm chart for Infra
 type: application
-version: 0.14.4
-appVersion: 0.14.4
+version: 0.20.0
+appVersion: 0.15.0
--- a/helm/charts/infra/README.md
+++ b/helm/charts/infra/README.md
@@ -1,0 +1,85 @@
+# Infra Chart
+
+A Helm chart for Infra, an infrastructure access management tool for Kubernetes.
+
+The default deployment installs `infra-server` and accompanying services including a frontend service and a PostgreSQL database. The full list of configuration values can be found with `helm show values infrahq/infra`.
+
+## Upgrade
+
+Run the commands below to update the Helm repository and perform the upgrade. Ensure the correct values are specified to avoid any unnecessary changes to the deployment. These commands assume the Helm release is called `infra`.
+
+```
+RELEASE_NAME=infra
+RELEASE_NAMESPACE=$(helm status $RELEASE_NAME -o json | jq -r .namespace)
+HELM_ARGS=  # additional arguments to helm, e.g. --set or --values
+
+helm repo update infrahq
+helm -n $RELEASE_NAMESPACE upgrade $RELEASE_NAME infrahq/infra $HELM_ARGS
+```
+
+## Rollback
+
+If the upgrade is unsuccessful, the deployment can be rolled back using Helm.
+
+```
+RELEASE_NAME=infra
+RELEASE_NAMESPACE=$(helm status $RELEASE_NAME -o json | jq -r .namespace)
+
+helm -n $RELEASE_NAMESPACE rollback $RELEASE_NAME
+```
+
+The upgrade can be attempted again or you can contact us through GitHub for additional assistance.
+
+## Version 0.20.0 and higher
+
+Infra chart version 0.20.0 introduces two backwards incompatible changes. The first is the migration to PostgreSQL as the default backend database from SQLite. The second is the migration of the database encryption key from within the server PVC to a Kubernetes secrets. This guide will detail the required steps to successfully migrate an Infra deploy to 0.20.0.
+
+### Before Upgrade
+
+#### PostgreSQL Database
+
+This change is automated and should not require manual migration. If your deployment uses an external database, i.e. if your Helm values defines `server.config.dbHost`, you can ignore this section.
+
+#### Database Encryption Key
+
+This change requires manual migration before performing the upgrade. If your deployment uses a database encryption key stored outside of Kubernetes, i.e. if your Helm values defines `server.config.dbEncryptionKey`, you can ignore this section.
+
+The following commands creates a Kubernetes secret with the necessary labels and annotations to masquerade as a Helm-managed resource. It seeds this secrets with the current encryption key such that it’s available to Infra after upgrade. 
+
+These commands assume the Helm release is called `infra`. Update the `RELEASE_NAME` variable if necessary with the correct release name. 
+
+```
+RELEASE_NAME=infra
+RELEASE_NAMESPACE=$(helm status $RELEASE_NAME -o json | jq -r .namespace)
+FULL_NAME=$(kubectl -n $RELEASE_NAMESPACE get deployment -l app.kubernetes.io/instance=$RELEASE_NAME -l app.infrahq.com/component=server -o name | cut -f2 -d/)
+
+ENCRYPTION_KEY=$(mktemp)
+kubectl -n $RELEASE_NAMESPACE exec -i deployment/$FULL_NAME -- cat /var/lib/infrahq/server/sqlite3.db.key >$ENCRYPTION_KEY
+kubectl -n $RELEASE_NAMESPACE create secret generic $FULL_NAME-encryption-key --from-file=key=$ENCRYPTION_KEY
+kubectl -n $RELEASE_NAMESPACE annotate secret $FULL_NAME-encryption-key meta.helm.sh/release-name=$RELEASE_NAME meta.helm.sh/release-namespace=$RELEASE_NAMESPACE 
+kubectl -n $RELEASE_NAMESPACE label secret $FULL_NAME-encryption-key app.kubernetes.io/managed-by=Helm
+rm $ENCRYPTION_KEY
+```
+
+### After Upgrade
+
+Once upgraded, verify the deployment with `infra login` or the Infra UI. Once verified, the server persistent volume claim should be disabled, `server.persistence.enabled=false`, to fully remove the SQLite database and old database encryption key.
+
+### Common Errors
+
+```
+$ kubectl logs deployment/infra-server
+Defaulted container "server" out of: server, postgres-ready (init)
+Error: creating server: db: migration failed: load key: unsealing: opening seal: cipher: message authentication failed
+```
+
+This error indicates an issue with the database encryption key migration; the encrypted data has been migrated but the key has not. Rollback and run the database encryption key migration before upgrading.
+
+```
+$ kubectl create secret generic infra-server-encryption-key --from-file=key=$ENCRYPTION_KEY
+error: failed to create secret secrets "infra-server-encryption-key" already exists
+error: --overwrite is false but found the following declared annotation(s): 'meta.helm.sh/release-name' already has a value (infra); 'meta.helm.sh/release-namespace' already has a value (default)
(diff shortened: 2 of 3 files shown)


helm/charts/infra/templates/NOTES.txt
```diff
@@ -37,6 +37,14 @@
 
 {{- end }}
 
+{{- if and .Values.server.persistence.enabled (not .Values.server.config.dbEncryptionKey) .Release.IsUpgrade }}
```
jmorganca, line 40: Adding a hint here is a great idea. I'd even recommend us linking to upgrade docs for _any_ install so users can look that up for manual migration instructions.

mxyng: I was thinking that too. Will make that change

mxyng: On second thought, it might not make the most sense since NOTES.txt is only outputted after an upgrade. This is more of a troubleshooting hint in case something goes wrong

jmorganca: Sounds good!

docs/install/install-infra-cli.md
```diff
@@ -8,6 +8,6 @@ position: 1
 
 {% partial file="../partials/cli-install.md" /%}
 
-## Upgrading the Infra CLI
+## Upgrade the Infra CLI
```
jmorganca, line 11: Good edit! This keeps it consistent with "install"

jmorganca (approved):

helm/charts/infra/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: infra
 description: A Helm chart for Infra
 type: application
-version: 0.14.4
+version: 0.20.0
 appVersion: 0.14.4
```
pdevine, line 6: Does the appVersion need to change too?

helm/charts/infra/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+### helm repo update infrahq
+helm -n $RELEASE_NAMESPACE upgrade $RELEASE_NAME infrahq/infra $HELM_ARGS
+```
+
+Once upgraded, verify the deployment with `infra login` or the Infra UI. Once verified, the server persistent volume claim should be disabled, `server.persistence.enabled=false`, to fully remove the SQLite database and old database encryption key.
+
+## Rollback
+
+If the upgrade is unsuccessful, the deployment can be rolled back using Helm. The upgrade can be attempted again or you can contact us through GitHub for additional assistance.
```
pdevine, line 24: I would move `The upgrade can be attempted again...` to the end of this section and put something like:

```To perform the rollback, run the following commands:```

File: helm/charts/infra/README.md [hunk 5]
```diff
@@ -0,0 +1,70 @@
...
+```bash
+RELEASE_NAME=infra
+RELEASE_NAMESPACE=$(helm status $RELEASE_NAME -o json | jq -r .namespace)
+
+helm -n $RELEASE_NAMESPACE rollback $RELEASE_NAME
+```
+
+### 0.14.4
+
+Infra chart version 0.20.0 introduces two backwards incompatible changes. The first is the migration to PostgreSQL as the default backend database from SQLite. The second is the migration of the database encryption key from within the server PVC to a Kubernetes secrets. This guide will detail the required steps to successfully migrate an Infra deploy to 0.20.0.
```
pdevine, line 35: This lists the PostgreSQL change before the db encryption key change, but then for the sub sections, they're listed in the reverse order. I would put the Postgres Database section first here.

helm/charts/infra/README.md
```diff
@@ -0,0 +1,70 @@
...
+kubectl -n $RELEASE_NAMESPACE exec -i deployment/$FULL_NAME -- cat /var/lib/infrahq/server/sqlite3.db.key >$ENCRYPTION_KEY
+kubectl -n $RELEASE_NAMESPACE create secret generic $FULL_NAME-encryption-key --from-file=key=$ENCRYPTION_KEY
+kubectl -n $RELEASE_NAMESPACE annotate secret $FULL_NAME-encryption-key meta.helm.sh/release-name=$RELEASE_NAME meta.helm.sh/release-namespace=$RELEASE_NAMESPACE 
+kubectl -n $RELEASE_NAMESPACE label secret $FULL_NAME-encryption-key app.kubernetes.io/managed-by=Helm
+rm $ENCRYPTION_KEY
+```
+
+#### PostgreSQL Database
+
+This change is automated and should not require manual migration. If your deployment uses an external database, i.e. if your Helm values defines `server.config.dbHost`, you can ignore this section.
```
pdevine, line 60: Why not put the part above for checking `server.persistence.enabled=false` here instead of above?

helm/charts/infra/README.md
pdevine, line 35: Maybe:

```Infra chart version 0.20.0 and greater introduces two backward incompatible changes from 0.14.4 and earlier.```
... and:
```This guide will show you the required steps to upgrade from earlier versions.```

I'm thinking we don't want to be specific about 0.20.0 because it will be out of date pretty quickly.

File: helm/charts/infra/README.md [hunk 7]
```diff
@@ -0,0 +1,70 @@
...
+
+Infra chart version 0.20.0 introduces two backwards incompatible changes. The first is the migration to PostgreSQL as the default backend database from SQLite. The second is the migration of the database encryption key from within the server PVC to a Kubernetes secrets. This guide will detail the required steps to successfully migrate an Infra deploy to 0.20.0.
+
+#### Database Encryption Key
+
+This change requires manual migration before performing the upgrade. If your deployment uses a database encryption key stored outside of Kubernetes, i.e. if your Helm values defines `server.config.dbEncryptionKey`, you can ignore this section.
+
+The following commands creates a Kubernetes secret with the necessary labels and annotations to masquerade as a Helm-managed resource. It seeds this secrets with the current encryption key such that it’s available to Infra after upgrade. 
+
+These commands assume the Helm release is called `infra`. Update the `RELEASE_NAME` variable if necessary with the correct release name.
```
pdevine, line 43: Does this get done before, or after the upgrade?

helm/charts/infra/templates/NOTES.txt
```diff
@@ -37,6 +37,14 @@
...
+
+  When upgrading from chart version <= 0.14.4, ensure the database encryption
```
pdevine, line 42: If the key is not migrated what ends up happening until they migrate it?

helm/charts/infra/README.md
```diff
@@ -0,0 +1,70 @@
+# Infra Chart
+
+A Helm chart for Infra, an infrastructure access management tool for Kubernetes.
+
+The default deployment installs `infra-server` and accompanying services including a frontend service and a PostgreSQL database. The full list of configuration values can be found with `helm show values infrahq/infra`.
+
+## Upgrade
+
+The upgrade is a standard Helm upgrade. Run the commands below to update the Helm repository and perform the upgrade. Ensure the correct values are specified to avoid any unnecessary changes to the deployment. As in the previous step, these commands assume the Helm release is called `infra`, update as needed.
```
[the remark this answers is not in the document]
mxyng: This is meant to say `update the release name if it's not infra`

helm/charts/infra/Chart.yaml
mxyng: It does not. The default Docker image for this Helm release is still 0.14.4

helm/charts/infra/README.md
```diff
@@ -0,0 +1,70 @@
...
+```bash
+RELEASE_NAME=infra
+RELEASE_NAMESPACE=$(helm status $RELEASE_NAME -o json | jq -r .namespace)
+HELM_ARGS=.. # additional arguments to helm, e.g. --set or --values
+
+### helm repo update infrahq
+helm -n $RELEASE_NAMESPACE upgrade $RELEASE_NAME infrahq/infra $HELM_ARGS
+```
+
+Once upgraded, verify the deployment with `infra login` or the Infra UI. Once verified, the server persistent volume claim should be disabled, `server.persistence.enabled=false`, to fully remove the SQLite database and old database encryption key.
```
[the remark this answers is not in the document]
mxyng: This is a product of some refactoring and originally only intended for upgrades from 0.14.4. Will move it to the 0.14.4 section

helm/charts/infra/README.md
mxyng: I think it's important to highlight that 0.20.0 introduces these changes. If we remove the migration at a later stage, these instructions will no longer be applicable so if a user upgrades, they will need to make a pit stop at 0.20.0. This section is intended to be upgrading from 0.14.4 but maybe the heading should be to `0.20.0`

helm/charts/infra/README.md
mxyng: These steps need to be run before the upgrade

helm/charts/infra/README.md
mxyng: The `server.persistence.enabled=false` step is a post-upgrade step to remove the server pvc and can only be set once upgraded and migrated. Doing so before will cause data loss. Added some more heading for clarity

helm/charts/infra/templates/NOTES.txt
mxyng: The new deployment will be unhealthy and the new server will not be usable

helm/charts/infra/Chart.yaml
pdevine: I'm still worried about us deviating here, since they've been in lockstep before. I'm just worried that someone will get confused.

helm/charts/infra/README.md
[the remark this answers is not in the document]
pdevine: I still think `The upgrade is a standard Helm upgrade.` and `As in the previous step, these commands assume the Helm release is called `infra`, update as needed.` are a little bit awkward.

For the first one this seems like nouning a verb (i.e. `the upgrade`), whereas it's the infra deployment which is being upgraded.

For the second sentence, maybe just: `As in the previous step, these commands assume the Helm release is called `infra``. I don't think `update as needed` adds anything here, and now I don't know if I'm supposed to update the values to `infra` or something else.

helm/charts/infra/templates/NOTES.txt
pdevine: Maybe we can call out what to do if they miss performing the manual upgrade steps?

helm/charts/infra/README.md
pdevine: I agree. We should definitely call out that 0.20.0 introduces the changes. My concern was we're going to move past 0.20.0 quickly with different releases and admins will not realize the changes are applicable to them if they skip 0.20.0.

I think you're right that the section should be called something like `### Version 0.20.0`, or maybe `### Version 0.20.0 and greater`. 

I'm not as worried about pulling the section out in the future. If we remove it in ~4-6 months I doubt there will still be anyone using the older versions.

helm/charts/infra/templates/NOTES.txt
mxyng: That's the intention with the link. The upgrade steps are too involved to include in the NOTES so a link to the README is the next best thing

helm/charts/infra/Chart.yaml
mxyng: Differing chart and app versions is common, probably more common than same chart and app versions. It provides us flexibility going forward to make changes to the helm chart independently of the app which may move at a faster or slower pace.

PR #2977, opened 2022-08-18
Refactor a test function and remove a test helper

dnephin:
## Summary

This PR:
* refactors the `TestRequireAccessKey` test case to use a `testCase` struct, removing the need to cast the `interface{}` to a function
* removes the `prometheus.Registry` arg from `Server.GenerateRoutes`. We call this in many tests, and it can receive the registry from a field on the server
* remove `data.InvalidateCache`, which was used in tests, but we removed the caches in a previous commit, so we no longer need to invalidate them. In the future if we want to cache these references we can use the `data.DB` struct which is different for each test, so we shouldn't need to invalidate the cache.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/infrahq/secrets"
+	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sync/errgroup"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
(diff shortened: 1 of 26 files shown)


internal/server/server.go
```diff
@@ -217,15 +220,14 @@ func registerUIRoutes(router *gin.Engine, opts UIOptions) {
...
-	router := s.GenerateRoutes(promRegistry)
+	router := s.GenerateRoutes()
 
 	httpErrorLog := log.New(logging.NewFilteredHTTPLogger(), "", 0)
 	metricsServer := &http.Server{
 		ReadHeaderTimeout: 30 * time.Second,
 		ReadTimeout:       60 * time.Second,
 		Addr:              s.options.Addr.Metrics,
-		Handler:           metrics.NewHandler(promRegistry),
+		Handler:           metrics.NewHandler(s.metricsRegistry),
```
BruceMacD, line 230: Seems like we should handle the possibility of this value being nil in the metrics functions. Feels like a likely nil pointer de-reference.

dnephin: In what scenario would we expect the metrics registry to be nil?

In production we set this to a value in `server.New`, so it can't ever be nil. I think for this to be nil it would have to be a programming error, which should panic.

ssoroka (approved):

PR #2985, opened 2022-08-19
feat: add orgs prom metrics

mxyng:
## Summary

Add prometheus metrics to monitor number of organizations


ssoroka (approved):
seems fine. how often is this run?

mxyng:
Metrics are collected when scraped by the upstream Prometheus instance so how often its run depends on the scrape interval of that instance and how many instances are scraping the service, etc. Usually scrape intervals are around once every 30 seconds.

PR #3000, opened 2022-08-22
maintain: update ui dev config

BruceMacD:
- add load to buildx to force local image cache
- include sample config file for ui development
- update ui dev to use sample config

## Summary
The server has gone out of sync with the UI development workflow. We dont support sqlite anymore and some other config is needed to emulate the multi-tenant deployment. Adding a local test config to the UI directory to make UI development easier. 

## Checklist


Makefile
```diff
@@ -12,8 +12,8 @@ test/update:
 	go test ./internal/cmd -test.update-golden
 
 dev:
-	docker buildx build . -t infrahq/infra:dev
-	docker buildx build ui -t infrahq/ui:dev
+	docker buildx build . --load -t infrahq/infra:dev
```
BruceMacD, line 15: The load flag explicitly loads the built image into the local store, which is something we always want in the case of local development. We had an issue where buildx was not adding the built image to the local store, this fixes it without manual dev configuration.

hoyyeva (approved):

ssoroka (approved):

PR #3008, opened 2022-08-22
Replace gorm.DB with an interface

dnephin:
## Summary

Branched from #3006, that will need to merge first.

This is a rather large PR because it was hard to split this change into smaller pieces. Introducing the interface should be the largest part of the change. After this everything should be easily broken up into much smaller changes.

This PR does the following:
* introduces a `GormTxn` interface that is accepted by all the `data` query functions
* adds methods to that interface to pass the Organization ID
* updates the middleware to set the orgID on this transaction, and all functions to read from this value.

This approach makes our code much safer, because we are no longer defaulting to the default orgID.  I'll call out some places where I had to explicitly do this in tests to prevent breaking many tests. I'll go through in a follow up to fix these places.

There are many commits, and a lot of them are very mechanism find/replace. I'll try to call out the most interesting bits with inline comments.

Related to https://github.com/infrahq/infra/issues/2415

### DIFF
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -17,12 +17,16 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired 
 		return nil, HandleAuthErr(err, "access keys", "list", roles...)
 	}
 
-	s := []data.SelectorFunc{data.ByOptionalIssuedFor(identityID), data.ByOptionalName(name)}
+	s := []data.SelectorFunc{
+		data.Preload("IssuedForIdentity"),
+		data.ByOptionalIssuedFor(identityID),
+		data.ByOptionalName(name),
+	}
 	if !showExpired {
 		s = append(s, data.ByNotExpiredOrExtended())
 	}
 
-	return data.ListAccessKeys(db.Preload("IssuedForIdentity"), p, s...)
+	return data.ListAccessKeys(db, p, s...)
 }
 
 func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string, err error) {
--- a/internal/access/organization.go
+++ b/internal/access/organization.go
@@ -44,7 +44,7 @@ func CreateOrganization(c *gin.Context, org *models.Organization) error {
 		return HandleAuthErr(err, "organizations", "create", models.InfraSupportAdminRole)
 	}
 
-	return data.CreateOrganizationAndSetContext(db, org)
+	return data.CreateOrganization(db, org)
 }
 
 func DeleteOrganization(c *gin.Context, id uid.ID) error {
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -27,7 +27,7 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, baseDomain string, details S
 
 	details.Org.Domain = SanitizedDomain(details.SubDomain, baseDomain)
 
-	if err := data.CreateOrganizationAndSetContext(db, details.Org); err != nil {
+	if err := data.CreateOrganization(db, details.Org); err != nil {
 		return nil, "", fmt.Errorf("create org on sign-up: %w", err)
 	}
 
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -31,6 +31,8 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, baseDomain string, details S
 		return nil, "", fmt.Errorf("create org on sign-up: %w", err)
 	}
 
+	db = data.NewTransaction(db.GormDB(), details.Org.ID)
+
 	// check the admin user's password requirements against our basic password requirements
 	err := checkPasswordRequirements(db, details.Password)
 	if err != nil {
--- a/internal/access/signup.go
+++ b/internal/access/signup.go
@@ -32,6 +32,10 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, baseDomain string, details S
 	}
 
 	db = data.NewTransaction(db.GormDB(), details.Org.ID)
+	c.Set("db", db)
+	rCtx := GetRequestContext(c)
+	rCtx.DBTxn = db
+	c.Set(RequestContextKey, rCtx)
 
 	// check the admin user's password requirements against our basic password requirements
 	err := checkPasswordRequirements(db, details.Password)
--- a/internal/server/authn/authn_method_test.go
+++ b/internal/server/authn/authn_method_test.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"golang.org/x/crypto/bcrypt"
-	"gorm.io/gorm"
 	"gotest.tools/v3/assert"
 
 	"github.com/infrahq/infra/internal/server/data"
@@ -15,7 +14,7 @@ import (
 	"github.com/infrahq/infra/internal/testing/patch"
 )
 
-func setupDB(t *testing.T) *gorm.DB {
+func setupDB(t *testing.T) *data.DB {
 	t.Helper()
 	driver := database.PostgresDriver(t, "_authn")
 	if driver == nil {
@@ -28,7 +27,7 @@ func setupDB(t *testing.T) *gorm.DB {
 	db, err := data.NewDB(driver.Dialector, nil)
 	assert.NilError(t, err)
 
-	return db.DB
+	return db
 }
 
 func TestLogin(t *testing.T) {
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -27,7 +27,7 @@ func secretChecksum(secret string) []byte {
 	return chksm[:]
 }
 
-func CreateAccessKey(db *gorm.DB, accessKey *models.AccessKey) (body string, err error) {
+func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err error) {
 	switch {
 	case accessKey.IssuedFor == 0:
 		return "", fmt.Errorf("issusedFor is required")
@@ -68,7 +68,7 @@ func CreateAccessKey(db *gorm.DB, accessKey *models.AccessKey) (body string, err
 			accessKey.ID = uid.New()
 		}
 
-		identityIssuedFor, err := GetIdentity(db, ByID(accessKey.IssuedFor))
+		identityIssuedFor, err := GetIdentity(db.GormDB(), ByID(accessKey.IssuedFor))
 		if err != nil {
 			return "", fmt.Errorf("key name from identity: %w", err)
 		}
@@ -83,7 +83,7 @@ func CreateAccessKey(db *gorm.DB, accessKey *models.AccessKey) (body string, err
 	return fmt.Sprintf("%s.%s", accessKey.KeyID, accessKey.Secret), nil
 }
 
-func SaveAccessKey(db *gorm.DB, key *models.AccessKey) error {
+func SaveAccessKey(db GormTxn, key *models.AccessKey) error {
 	if key.Secret != "" {
 		key.SecretChecksum = secretChecksum(key.Secret)
 	}
@@ -128,7 +128,8 @@ func DeleteAccessKeys(db *gorm.DB, selectors ...SelectorFunc) error {
 	return deleteAll[models.AccessKey](db, ByIDs(ids))
 }
 
-func ValidateAccessKey(db *gorm.DB, authnKey string) (*models.AccessKey, error) {
+func ValidateAccessKey(tx GormTxn, authnKey string) (*models.AccessKey, error) {
+	db := tx.GormDB()
 	keyID, secret, ok := strings.Cut(authnKey, ".")
 	if !ok {
 		return nil, fmt.Errorf("invalid access key format")
@@ -164,7 +165,7 @@ func ValidateAccessKey(db *gorm.DB, authnKey string) (*models.AccessKey, error) 
 			}
 			db.Statement.Context = WithOrg(db.Statement.Context, org)
 		}
-		if err := SaveAccessKey(db, t); err != nil {
+		if err := SaveAccessKey(tx, t); err != nil {
 			return nil, err
 		}
 	}
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -157,8 +157,7 @@ func ValidateAccessKey(tx GormTxn, authnKey string) (*models.AccessKey, error) {
 
 		t.ExtensionDeadline = time.Now().UTC().Add(t.Extension)
 
-		// TODO: constructor
-		tx = &Transaction{DB: tx.GormDB(), orgID: t.OrganizationID}
+		tx = NewTransaction(tx.GormDB(), t.OrganizationID)
 		if err := SaveAccessKey(tx, t); err != nil {
 			return nil, err
 		}
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -68,7 +68,7 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err 
 			accessKey.ID = uid.New()
 		}
 
-		identityIssuedFor, err := GetIdentity(db.GormDB(), ByID(accessKey.IssuedFor))
+		identityIssuedFor, err := GetIdentity(db, ByID(accessKey.IssuedFor))
 		if err != nil {
 			return "", fmt.Errorf("key name from identity: %w", err)
 		}
@@ -91,11 +91,12 @@ func SaveAccessKey(db GormTxn, key *models.AccessKey) error {
 	return save(db, key)
 }
 
-func ListAccessKeys(db *gorm.DB, p *models.Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
+func ListAccessKeys(db GormTxn, p *models.Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
 	return list[models.AccessKey](db, p, selectors...)
 }
 
-func GetAccessKey(db *gorm.DB, selectors ...SelectorFunc) (*models.AccessKey, error) {
+func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, error) {
+	db := tx.GormDB()
 	// GetAccessKey by keyID needs to not set an organization_id in the query.
 	// keyID should be globally unique.
 	for _, selector := range selectors {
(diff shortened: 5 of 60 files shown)


internal/access/access_key.go
```diff
@@ -17,12 +17,16 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired
...
+		data.Preload("IssuedForIdentity"),
+		data.ByOptionalIssuedFor(identityID),
+		data.ByOptionalName(name),
+	}
 	if !showExpired {
 		s = append(s, data.ByNotExpiredOrExtended())
 	}
 
-	return data.ListAccessKeys(db.Preload("IssuedForIdentity"), p, s...)
+	return data.ListAccessKeys(db, p, s...)
```
dnephin, line 29: In a few places I had to convert `Preload` to a selector so that I could accept a `data.GormTxn` as the first argument to the query function, and only selectors would continue to use `gorm.DB`. This is one of the places, there should be a few others.

internal/access/signup.go
```diff
@@ -27,10 +27,16 @@ func Signup(c *gin.Context, keyExpiresAt time.Time, baseDomain string, details S
...
-	if err := data.CreateOrganizationAndSetContext(db, details.Org); err != nil {
+	if err := data.CreateOrganization(db, details.Org); err != nil {
 		return nil, "", fmt.Errorf("create org on sign-up: %w", err)
 	}
 
+	db = data.NewTransaction(db.GormDB(), details.Org.ID)
+	c.Set("db", db)
+	rCtx := GetRequestContext(c)
+	rCtx.DBTxn = db
+	c.Set(RequestContextKey, rCtx)
```
dnephin, line 38: Now that we're passing orgID as part of the transaction the `data.CreateOrganization` function can't set the org ID in the context. I think this is a good thing, as previously it was surprising that creating an org mutated the DB to make all queries use that org. 

As a consequence of that change we have to set the tx here for the request of the request, but this is the only place where we need to do this.

I've also restored the original name for `CreateOrganization`, now that it's not doing the `SetContext` part.

internal/server/data/data.go
```diff
@@ -92,6 +92,16 @@ func (d *DB) QueryRow(query string, args ...any) *sql.Row {
 	return d.DB.Raw(query, args...).Row()
 }
 
+func (d *DB) OrganizationID() uid.ID {
+	// FIXME: this is a hack to keep our tests passing. The db should not
+	// be scoped to an org ID.
+	return d.DefaultOrg.ID
```
dnephin, line 98: Calling this out. This is not a regression, because it works this way on `main`, but it is a problem we should fix.

All of our test fixtures should be more explicit about which org they are using when creating things.

internal/server/data/data_test.go
```diff
@@ -76,66 +75,8 @@ func TestSnowflakeIDSerialization(t *testing.T) {
 	})
 }
 
-func TestDatabaseSelectors(t *testing.T) {
```
dnephin, line 79: Removed because this test is just a simulation of the selector behaviour, and we should be removing selector functions soon. Porting this test to the new interface would be a lot more work than other tests, and is not worth it.

internal/server/data/identity.go
```diff
@@ -94,24 +95,28 @@ func AssignIdentityToGroups(db *gorm.DB, user *models.Identity, provider *models
...
 }
 
-func DeleteIdentities(db *gorm.DB, selectors ...SelectorFunc) error {
-	toDelete, err := ListIdentities(db.Select("id"), nil, selectors...)
+func DeleteIdentities(tx GormTxn, selectors ...SelectorFunc) error {
+	selectID := func(db *gorm.DB) *gorm.DB {
+		return db.Select("id")
+	}
+	selectors = append([]SelectorFunc{selectID}, selectors...)
+	toDelete, err := ListIdentities(tx, nil, selectors...)
```
dnephin, line 119: Another place where I had to move a thing to the selectors. I think this is the only case that was not a `Preload`

internal/server/data/migrations_test.go
```diff
@@ -510,16 +510,18 @@ DELETE FROM settings WHERE id=24567;
...
+		db := &DB{DB: rawDB}
 		opts := migrator.Options{InitSchema: initializeSchema}
-		m := migrator.New(&DB{DB: db}, opts, nil)
+		m := migrator.New(db, opts, nil)
 		assert.NilError(t, m.Migrate())
 
 		initialSchema = dumpSchema(t, os.Getenv("POSTGRESQL_CONNECTION"))
 
-		assert.NilError(t, db.Exec("DROP SCHEMA IF EXISTS testing CASCADE").Error)
+		_, err = db.Exec("DROP SCHEMA IF EXISTS testing CASCADE")
```
dnephin, line 523: I converted the few remaining `Exec` calls to the new interface, which matches the stdlib. So you'll notice the `.Error` removed, and that it now returns two args.

internal/server/data/organization.go
```diff
@@ -1,50 +1,22 @@
...
-// OrgFromContext returns the Organization stored using WithOrg.
-func OrgFromContext(ctx context.Context) *models.Organization {
-	org, ok := ctx.Value(orgCtxKey{}).(*models.Organization)
-	if !ok {
-		return nil
-	}
-	return org
-}
-
-func MustGetOrgFromContext(ctx context.Context) *models.Organization {
```
dnephin, line 24: Now that we have a strongly typed interface we don't need to store the org in the context. This wasn't necessary to do all at once, but it was a very small change at the end (see the last commit).

We no longer panic, but an unset orgID will also be 0, which will never match a real org (not even the default), which will still error and should be caught by a test.

internal/server/data/organization_test.go
```diff
@@ -13,60 +13,58 @@ import (
 // PostgreSQL only has microsecond precision
 var cmpTimeWithDBPrecision = cmpopts.EquateApproxTime(time.Microsecond)
 
-func TestCreateOrganizationAndSetContext(t *testing.T) {
-	pgsql := postgresDriver(t)
-	db := setupDB(t, pgsql)
+func TestCreateOrganization(t *testing.T) {
+	runDBTests(t, func(t *testing.T, db *DB) {
```
dnephin, line 17: Updated this test to use `runDBTests` like all the others. We only really need to test against postgres right now, but I expect `runDBTests` to be useful in the future despite that.

One use case that comes to mind is testing postgres upgrades. We can run all of our `data` tests against multiple versions of postgres to catch problems.

internal/server/middleware.go
```diff
@@ -73,7 +73,9 @@ func handleInfraDestinationHeader(c *gin.Context) error {
 // possibly also of the destination.
 func authenticatedMiddleware(srv *Server) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		withDBTxn(c.Request.Context(), srv.DB(), func(tx *gorm.DB) {
+		withDBTxn(c.Request.Context(), srv.DB().GormDB(), func(db *gorm.DB) {
+
+			tx := data.NewTransaction(db, 0)
```
dnephin, line 78: As it was before, we now set the orgID in the transaction as part of these middleware. But until we have an org identified we query with orgID =0, which should ensure we never match an org.

internal/server/passwordreset.go
```diff
@@ -25,7 +24,7 @@ func (a *API) RequestPasswordReset(c *gin.Context, r *api.PasswordResetRequest)
 		return nil, err
 	}
 
-	org := data.MustGetOrgFromContext(c)
+	org := access.GetRequestContext(c).Authenticated.Organization
```
BruceMacD, line 27: This should fail if not set I think

internal/server/data/access_key.go
```diff
@@ -83,19 +83,20 @@ func CreateAccessKey(db *gorm.DB, accessKey *models.AccessKey) (body string, err
...
 }
 
-func ListAccessKeys(db *gorm.DB, p *models.Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
+func ListAccessKeys(db GormTxn, p *models.Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
 	return list[models.AccessKey](db, p, selectors...)
 }
 
-func GetAccessKey(db *gorm.DB, selectors ...SelectorFunc) (*models.AccessKey, error) {
+func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, error) {
+	db := tx.GormDB()
```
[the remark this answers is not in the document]
dnephin: Anything that builds a query itself using gorm methods will need to call `tx.GormDB`. Mostly that is our helpers like `add`, `save`, `get`, but also a few others that do more interesting queries.

This will give us an easy way of finding all the queries that need to be replaced with sql.

internal/server/data/access_key.go
```diff
@@ -155,16 +156,9 @@ func ValidateAccessKey(db *gorm.DB, authnKey string) (*models.AccessKey, error)
...
-		if org == nil {
-			org, err = GetOrganization(db, ByID(t.OrganizationID))
-			if err != nil {
-				return nil, fmt.Errorf("loading organization: %w", err)
-			}
-			db.Statement.Context = WithOrg(db.Statement.Context, org)
-		}
-		if err := SaveAccessKey(db, t); err != nil {
+
+		tx = NewTransaction(tx.GormDB(), t.OrganizationID)
```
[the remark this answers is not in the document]
dnephin: The naming is maybe a bit misleading. `NewTransaction` doesn't create a new DB transaction, it uses an existing reference to an `sql.Tx`, but sets a new orgID for the transaction.

internal/server/middleware.go
```diff
@@ -157,6 +158,7 @@ func unauthenticatedMiddleware(srv *Server) gin.HandlerFunc {
 				sendAPIError(c, internal.ErrBadRequest)
 				return
 			}
+			authned.Organization = org
```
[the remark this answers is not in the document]
dnephin: This is the unauthenticated route, so it may not be set yet.

I wasn't sure where to store it. Maybe we should split up the `Authenticated` struct into something like this:

```go
type RequestDetails struct {
    Authenticated Authenticated
    Organization *models.Organization
}
```

internal/server/passwordreset.go
dnephin: It will panic on the next line if it's not set.

internal/server/data/data.go
```diff
@@ -186,14 +239,15 @@ func getDefaultSortFromType(t interface{}) string {
...
+func get[T models.Modelable](tx GormTxn, selectors ...SelectorFunc) (*T, error) {
+	db := tx.GormDB()
 	for _, selector := range selectors {
 		db = selector(db)
 	}
 
 	result := new(T)
 	if isOrgMember(result) {
-		db = ByOrgID(MustGetOrgFromContext(db.Statement.Context).ID)(db)
+		db = ByOrgID(tx.OrganizationID())(db)
```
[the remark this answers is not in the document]
dnephin: I pushed another commit to add a panic to `OrganizationID` if the value is 0.

internal/server/data/organization.go
```diff
@@ -1,51 +1,20 @@
...
 	if err != nil {
 		return fmt.Errorf("creating org: %w", err)
 	}
 
-	db := tx.GormDB()
-	// TODO: remove
-	db.Statement.Context = WithOrg(db.Statement.Context, org)
-	// TODO: constructor?
-	tx = &Transaction{DB: db, orgID: org.ID}
+	tx = NewTransaction(tx.GormDB(), org.ID)
```
ssoroka, line 17: how does this affect the tx from the callers?

internal/access/organization.go
```diff
@@ -44,7 +44,7 @@ func CreateOrganization(c *gin.Context, org *models.Organization) error {
 		return HandleAuthErr(err, "organizations", "create", models.InfraSupportAdminRole)
 	}
 
-	return data.CreateOrganizationAndSetContext(db, org)
+	return data.CreateOrganization(db, org)
```
ssoroka, line 47: seems like the removal of CreateOrganizationAndSetContext would cause problems? How does the current context or tx get set to the new org? is this the responsibility of the caller now?

ssoroka (commented):
Would like to see the return of MustGetOrg type semantics, where things can't quietly fail when missing.

internal/server/data/access_key.go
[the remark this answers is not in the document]
pdevine: I found this a little confusing too when reading through the PR. Maybe a note around `GormDB()` which explains when to use it?

internal/server/data/access_key.go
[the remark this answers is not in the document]
pdevine: This also took me a moment to realize what was going on, and how `Exec()` works inside the transaction.

pdevine (commented):
Overall looks fine. We still need to somehow split this for privileged vs unprivileged requests, and I'm not 100% sure how to carve that up. We could hold two references inside of the transaction, although it would be preferable only to store the unprivileged connection just in case we mess something up and call something in the data layer which references the wrong connection.

Right now we need the privileged connection to:
 * Use the telemetry/metrics
 * Load the config (which needs to look up the default org, but we could drop this if we didn't have a default org at all)
 * DeleteIdentities() and DeleteGrants() also needs the work somehow in the config which gets rid of any users/grants which were set up previously. This gets really weird if the org changes.
 * Encryption keys (since they're not tied to an org -- the unprivileged connection can still look these up right now which is maybe not what we want)
 * Migrations
 * Sign-up (although this could also work if we don't turn on RLS for the organizations table)

internal/server/data/data.go
```diff
@@ -101,6 +106,41 @@ type ReadTxn interface {
...
+type GormTxn interface {
+	WriteTxn
+
+	GormDB() *gorm.DB
+}
+
+type Transaction struct {
+	*gorm.DB
+	orgID uid.ID
+}
```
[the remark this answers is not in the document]
dnephin: It could, but I don't think I understand the reason for the indirection. Why not have the type implement the transaction operations directly?

internal/server/data/data.go
dnephin: Exactly. This is why I prioritized this part of the `data` package work, because this is the behaviour we have now by setting the context when we create a database.

It "works" in production because requests use the transaction reference where we override the organization value. 

The only code that doesn't do that right now is:
* the config loading, where you fixed it to set this same default value
* tests in the `server` package where we do the same thing by setting the default org.

I plan on removing this in the first follow up PR.

internal/server/data/access_key.go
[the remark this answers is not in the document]
dnephin: We should be removing `GormDB` very soon, but I'll add a comment for the interim.

internal/server/data/access_key.go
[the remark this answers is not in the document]
dnephin: I just noticed this is done in the wrong place. I'll add a comment and move it to outside of this block.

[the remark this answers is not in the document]
dnephin: This `NewTransaction` wouldn't be necessary if we didn't have the `panic` thta was requested in these comments. It's kind of unfortunate that we're forcing unnecessary code because of safeguards that aren't correct in all cases.

internal/server/data/organization.go
dnephin: It doesn't, only changes the Tx in the scope of this function, by design. It's really confusing to have a function mutate the transaction you pass in.

internal/access/organization.go
dnephin: It is the responsibility of the caller now.

internal/server/middleware.go
```diff
@@ -147,7 +147,8 @@ func withDBTxn(ctx context.Context, db *gorm.DB, fn func(tx *gorm.DB)) {
 
 func unauthenticatedMiddleware(srv *Server) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		withDBTxn(c.Request.Context(), srv.DB(), func(tx *gorm.DB) {
+		withDBTxn(c.Request.Context(), srv.DB().GormDB(), func(db *gorm.DB) {
+			tx := data.NewTransaction(db, 0)
```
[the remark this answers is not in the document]
dnephin: We need to create the DB transaction before we know the org ID, so I think they have to be separate.

internal/server/authn/authn_method_test.go
```diff
@@ -15,7 +14,7 @@ import (
 	"github.com/infrahq/infra/internal/testing/patch"
 )
 
-func setupDB(t *testing.T) *gorm.DB {
+func setupDB(t *testing.T) *data.DB {
 	t.Helper()
 	driver := database.PostgresDriver(t, "_authn")
 	if driver == nil {
```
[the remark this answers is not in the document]
dnephin: No reason to yet. That should be a separate change either way.

dnephin:
> We still need to somehow split this for privileged vs unprivileged requests, and I'm not 100% sure how to carve that up. 

The `data.DB` struct should return two connection pools, one privileged, and one unprivileged.

When we choose which one to use when we create the transaction. 

I don't think the changes were should impact that much. The biggest change might be in tests.

dnephin:
I've restored the panic and added more godoc.

I don't like the panic, I think it's assuming too much about the caller, which maybe true for regular requests, but is not true for middleware and other callers of the `data` package. But I'm fine to add it for now , and we can revisit it in a smaller more focused PR.

There are some places where we have to create a transaction just to work around the panic. The `data.Transaction` wouldn't be necessary otherwise, because the org ID is already set on the object.

I've also added some godoc in a few places as requested.

This should be ready for another review.

internal/server/data/access_key.go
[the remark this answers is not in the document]
dnephin: I was able to remove this by moving where we panic:

https://github.com/infrahq/infra/compare/dnephin/data-interface-org-id-panic?expand=1

I'll open that as a follow up

internal/server/data/data.go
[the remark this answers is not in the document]
dnephin: I think a `data` function accepting a transaction is what's expect for this interface. If we split it up I think I'd rather move the `OrganizationID` to a different type, something like `Metadata`, that we can expose as part of the transaction. Then the calls to `NewTransaction` could be replaced with `SetTransactionMetadata` which only updates the metadata, not the transaction.

internal/server/middleware.go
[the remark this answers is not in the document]
BruceMacD: That seems like a good idea

BruceMacD (approved):
This looks good to me pending any further comments from Steven and Patrick

ssoroka (approved):

PR #3014, opened 2022-08-23
improve: better analytics

ssoroka:
## Summary

Add analytics around org membership and signup

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -140,7 +140,7 @@ func (a *API) Signup(c *gin.Context, r *api.SignupRequest) (*api.SignupResponse,
 	setCookie(c, cookie)
 
 	a.t.User(identity.ID.String(), r.Name)
-	a.t.Alias(identity.ID.String())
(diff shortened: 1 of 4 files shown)


internal/server/handlers.go
```diff
@@ -176,6 +178,15 @@ func (a *API) Login(c *gin.Context, r *api.LoginRequest) (*api.LoginResponse, er
...
 	setAuthCookie(c, bearer, key.ExpiresAt)
 
+	if name == "" {
+		user, err := access.GetIdentity(c, key.IssuedFor)
+		if err == nil {
+			name = user.Name
+		}
+	}
+	a.t.User(key.IssuedFor.String(), name)
+	a.t.OrgMembership(key.OrganizationID.String(), key.IssuedFor.String())
```
BruceMacD, line 188: Might make sense to combine these into one event where they are tracking similar things

BruceMacD (approved):

internal/server/handlers.go
ssoroka: Wasn't exactly sure how this shows up in segment, but I was thinking they're separate things.

PR #3018, opened 2022-08-24
maintain: set org cookie on signup redirect

### LINKED ISSUE #2980: Improve multi-org cookie experience
Currently a cookie created during sign-up applies to the wildcard base domain. This means going to any org sub domain will redirect you to your own org domain.

The cookie created on sign-up is removed and replaced with an org specific cookie.

BruceMacD:
- issue signup session cookie
- exchange signup session cookie for auth cookie
- add cookie auth middleware tests
- remove login cookie

## Summary
After sign-up we set a signup cookie that applies to all `.infrahq.com` subdomains, set a subdomain specific auth cookie when this signup cookie is seen. No redirect needed.

## Checklist

## Related Issues

Resolves #2980


ui/pages/signup/redirect/index.js
```diff
@@ -0,0 +1,39 @@
+import { useState } from 'react'
+
+export default function SignupRedirect() {
```
dnephin, line 3: If `/api/signup/session` returned a 3xx response code with a location header, could we use that to redirect instead of this extra page in the UI?

Not sure what the trade-offs are there

internal/server/handlers.go
```diff
@@ -124,7 +124,14 @@ func (a *API) Signup(c *gin.Context, r *api.SignupRequest) (*api.SignupResponse,
 		return nil, err
 	}
 
-	setAuthCookie(c, bearer, keyExpires)
+	// this cookie is set to send on all infra domains, make it expire quickly to prevent an unexpected org being set on requests to other orgs
```
dnephin, line 127: It'd be great to include a little more about why we do this. Something about us wanting to use separate cookies for each org, but since signup happens on a different domain from other requests we need this cookie to login the user after signup.

internal/server/middleware.go
```diff
@@ -170,6 +170,8 @@ func unauthenticatedMiddleware(srv *Server) gin.HandlerFunc {
 				c.Request = c.Request.WithContext(data.WithOrg(c.Request.Context(), org))
 				tx.Statement.Context = c.Request.Context() // TODO: remove with gorm
 			}
+			// if this is the first request after sign-up we must exchange the signup cookie for an auth cookie
+			exchangeSignupCookieForSession(c, srv.options.BaseDomain)
```
dnephin, line 174: This is currently only being called in the `unauthenticatedMiddleware`, I think we need it in `authenticatedMiddleware` as well. 

We could add this same call to both places, but another option might be to move this into `requireAccessKey`, which is called from both. If we did that, I guess `requireAccessKey` will need another argument, but `exchangeSignupCookieForSession` might be able to get the access key from an arg instead of from `RequestContext`.

internal/server/middleware.go
```diff
@@ -304,3 +290,33 @@ hostLookup:
...
+	} else {
+		// Fall back to checking cookies
+		cookie, err := getCookie(req, cookieAuthorizationName)
+		if err != nil {
+			logging.L.Trace().Msg("auth cookie not found, falling back to signup cookie")
+			cookie, err = getCookie(req, cookieSignupName)
+			if err != nil {
+				return "", fmt.Errorf("%w: valid token not found in request", internal.ErrUnauthorized)
+			}
+		}
```
dnephin, line 311: I think the only place we'd need to call `exchangeSignupCookieForSession` would be here, right? 

We've looked for the regular `auth` cookie, didn't find it, but we did find the signup cookie. So we could do the exchange at this point.

Although I'm not sure what happens if the first request returns a non-2xx code. Does the browser still save the cookie? If not, then I think the way you have it now is much safer.

dnephin (approved):
LGTM! I think the only thing this is missing is to call `exchangeSignupCookieForSession` from the `authenticatedMiddleware` as well. 

I left some other small suggestions, but we could explore those as a follow up.

It would also be great to add some test coverage of this for the signup API. I think we'd have to use `httptest.Server` instead of `httptest.ResponseRecorder` to make actual requests with cookies. It'd be great to show the whole cookie exchange flow in one test  case, although I'm not entirely sure how to do multiple domains. We might need to simulate some things. Could also be a follow up.

internal/server/middleware.go
BruceMacD: > Origin servers MAY send a Set-Cookie response header with any response. User agents MAY ignore Set-Cookie headers contained in responses with 100-level status codes but MUST process Set-Cookie headers contained in other responses (including responses with 400- and 500-level status codes). An origin server can include multiple Set-Cookie header fields in a single response. The presence of a Cookie or a Set-Cookie header field does not preclude HTTP caches from storing and reusing a response.

Looks like it should still be set, I'll move it here and test it

PR #3020, opened 2022-08-25
Migrations for RLS

pdevine:
## Summary

This change includes the migrations for setting up RLS, and also sets the current organization in the middleware. It's not complete because it doesn't set up the unprivileged/privileged connections because we've been changing so much in the data layer.

There are also a few tests that are missing here, and some more work that needs to happen to de-privilege places in the code which should use the unprivileged connection.

## Checklist

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -2,6 +2,7 @@ package data
 
 import (
 	"context"
+	"database/sql"
 	_ "embed"
 	"errors"
 	"fmt"
@@ -573,9 +574,10 @@ func addRLSUser() *migrator.Migration {
 			logging.Infof("adding RLS user")
 			var val int
 			err := tx.QueryRow("SELECT 1 FROM pg_roles WHERE rolname = 'infra_user'").Scan(&val)
-			if err != nil {
+			if err != nil && err != sql.ErrNoRows {
 				return err
 			}
+
 			if val == 0 {
 				if _, err := tx.Exec("CREATE USER infra_user"); err != nil {
 					logging.Infof("Couldn't create infra_user")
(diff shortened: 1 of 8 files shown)


internal/server/data/migrations.go
```diff
@@ -558,3 +566,85 @@ func dropOrganizationNameIndex() *migrator.Migration {
...
+				return err
+			}
+
+			if val == 0 {
+				if _, err := tx.Exec("CREATE USER infra_user"); err != nil {
+					logging.Infof("Couldn't create infra_user")
+					return err
+				}
+				// TODO: Properly store/set the user password
+				if _, err := tx.Exec("ALTER USER infra_user WITH PASSWORD 'changemetoanythingelse'"); err != nil {
```
dnephin, line 587: I believe the test failures are related to this.

We isolate tests from each other by using a [postgres schema](https://www.postgresql.org/docs/current/ddl-schemas.html) which is basically a namespace. But users are not scoped by schemas, they exist for the whole database.

So when we create this user locally it stays around. The tests pass because the user exists.

When we run this in CI the use does not exist, do we can't pass that connection the `data.NewDB`, and lots of tests fail.

This TODO is also a big thing to solve.

I think given these two problems, we should change how we setup the user.

Instead of creating to create a user in a migration, let's require the operator to specify the name of the user and password for the unprivileged user, the same way they specify it for the privileged connection.

We can create two users in `internal/testing/database.PostgresDriver`, and pass them both to our test.

I think managing users outside the application is what most operators would expect, and it makes this code much easier.

internal/server/data/migrations.go
```diff
@@ -558,3 +566,85 @@ func dropOrganizationNameIndex() *migrator.Migration {
...
 	}
 }
+
+func addRLSUser() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-08-15T15:19",
+		Migrate: func(tx migrator.DB) error {
+			logging.Infof("adding RLS user")
+			var val int
+			err := tx.QueryRow("SELECT 1 FROM pg_roles WHERE rolname = 'infra_user'").Scan(&val)
```
BruceMacD, line 576: Should we be messing with database privileges here? That feels unexpected in a migration, I'd rather move the process to something manual until we have a workflow to handle it through Infra personally.

PR #3022, opened 2022-08-25
Remove db from gin context

dnephin:
Related to #3008, although merging this may cause some small conflicts with that PR.

This PR removes the `gorm.DB` reference from `gin.Context`, and ensures we always get the DB from our `RequestContext`. This will make it possible for us to add metadata to a Transaction struct directly, without having to use the gorm.DB context.


BruceMacD (approved):
Looks good

PR #3026, opened 2022-08-25
maintain: reword org name placeholder

mxyng:
The organization name placeholder is isconsistent with other fields in the signup UI


BruceMacD (approved):
Thanks 👍

PR #3027, opened 2022-08-25
Move where we panic about a missing org ID to reduce the calls to data.NewTransaction

dnephin:
## Summary

Branched from #3008

This PR removes a few calls to `data.NewTransaction` which overall should make the code more obvious. To remove these extra calls the following changes were made:

1. moved the `panic` on missing org ID from `OrganizationID()` to `SetOrganizationID` and the `ByOrgID` selector. This allows us to only panic when there is no org ID at all, instead of when one is missing from the transaction/context. While the transaction will have an orgID set most of the time, there are a few places where it does not, for example when we are creating an org.
2. moved a `Signup` test from the `access` package to the `server` package. This removes the need to duplicate all of the setup we do in middleware, and also makes our code easier to maintain by ensuring our tests run against a public interface, not an internal one that could change at any time. Testing at this layer also improves the test coverage (more code is included in the unit-under-test), and makes the test more like a real production workflow (less simulation of the real environment).


internal/server/models/organization.go
```diff
@@ -30,8 +30,15 @@ type OrganizationMember struct {
 
 func (OrganizationMember) IsOrganizationMember() {}
 
-func (o *OrganizationMember) SetOrganizationID(id uid.ID) {
+type OrganizationIDSource interface {
+	OrganizationID() uid.ID
+}
+
+func (o *OrganizationMember) SetOrganizationID(source OrganizationIDSource) {
```
dnephin, line 37: Passing in the whole interface instead of just a value is a bit unfortunate, but I don't see any way to avoid this. 

We need to completely avoid the call to `OrganizationID` when an org ID is set, so we have to defer looking up the ID. Accepting a function seems like the easiest way to do that.

internal/server/middleware.go
```diff
@@ -228,6 +228,8 @@ func requireAccessKey(c *gin.Context, db data.GormTxn, srv *Server) (access.Auth
 	}
 
 	// now that the org is loaded scope all db calls to that org
+	// TODO: set the orgID explicitly in the options passed to GetIdentity to
+	// remove the need for this NewTransaction call.
 	db = data.NewTransaction(db.GormDB(), org.ID)
```
dnephin, line 233: This is the last remaining call to `data.NewTransaction` that I'd like to remove (the other ones are in middleware or in create organization, and those are necessary).

To do this I think we need to replace selector funcs with a struct. Then we can easily validate that the struct includes a non-zero OrganizationID.  Technically we could do this today by inspecting the `db.Statement.Clauses["WHERE"]`, but that code is going to be very-gorm specific, so will be removed soon. For now it seemed easier to keep this todo. I'll see it again when we remove the selector func to the `data.GetIdentity` call below.

internal/server/data/access_key.go
```diff
@@ -156,11 +156,6 @@ func ValidateAccessKey(tx GormTxn, authnKey string) (*models.AccessKey, error) {
 		}
 
 		t.ExtensionDeadline = time.Now().UTC().Add(t.Extension)
-
-		// Set the orgID in the tx. This is only necessary because our data
-		// layer requires an orgID be set in the transaction. If we remove
-		// that requirement, we can remove this line as well.
-		tx = NewTransaction(tx.GormDB(), t.OrganizationID)
```
BruceMacD, line 163: Does this mean that we are saving this access key without an org ID set or is that in the context at this point?

internal/server/data/organization.go
```diff
@@ -14,36 +14,37 @@ func CreateOrganization(tx GormTxn, org *models.Organization) error {
...
 		return fmt.Errorf("failed to create connector identity while creating org: %w", err)
 	}
 
 	err = CreateGrant(tx, &models.Grant{
-		Subject:   uid.NewIdentityPolymorphicID(connector.ID),
-		Privilege: models.InfraAdminRole,
-		Resource:  "infra",
-		CreatedBy: models.CreatedBySystem,
+		Subject:            uid.NewIdentityPolymorphicID(connector.ID),
+		Privilege:          models.InfraAdminRole,
```
BruceMacD, line 44:
```suggestion
		Privilege:          models.InfraConnectorRole,
```

This was fixed in my recent PR, but I want to make sure it doesn't regress here.

internal/server/data/selectors.go
```diff
@@ -51,6 +51,9 @@ func ByOptionalIDs(ids []uid.ID) SelectorFunc {
 }
 
 func ByOrgID(orgID uid.ID) SelectorFunc {
+	if orgID == 0 {
```
BruceMacD, line 54: Looks like we don't have any protection for `add` and `save` anymore if we only rely on this selector.

internal/server/data/settings.go
```diff
@@ -65,8 +65,9 @@ func GetSettings(db GormTxn) (*models.Settings, error) {
 
 func getSettingsForOrg(tx GormTxn, orgID uid.ID) (*models.Settings, error) {
 	db := tx.GormDB()
+	db = ByOrgID(orgID)(db)
```
BruceMacD, line 68: Maybe not related to this change, but looks like we should use the standard `get` function here rather than the `db.First`?

dnephin: Ya, seems like that might work. I'm not sure why it's different.  Either way we'll be moving away from `get` very soon when we replace gorm queries with sql. In the new sql version I don't think we'll have a `get` function because there's nothing really common to selecting from different tables other than error handling.

BruceMacD (approved):
looks good

PR #3028, opened 2022-08-25
Replace NewTransaction with a real db transaction

dnephin:
## Summary

The `data.NewTransaction` constructor is a bit misleading because it does not actually start a transaction. Instead it wraps a `gorm.DB` with our wrapper that implements the stdlib interface.

This PR removes the `NewTransaction` constructor and replaces it with two things:
* a `Transaction.WithOrgID` method for setting the OrgID metadata in the transaction
* a `DB.Begin` method for creating a new transaction.

This changes brings us much closer to the interface provided by the stdlib.

Also remove a couple places we were creating a `RequestContext` unnecessarily in tests.

### DIFF
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -73,8 +73,7 @@ func handleInfraDestinationHeader(c *gin.Context) error {
 // possibly also of the destination.
 func authenticatedMiddleware(srv *Server) gin.HandlerFunc {
 	return func(c *gin.Context) {
-		withDBTxn(c.Request.Context(), srv.DB().GormDB(), func(db *gorm.DB) {
-			tx := data.NewTransaction(db, 0)
+		withDBTxn(c.Request.Context(), srv.DB().GormDB(), func(tx *data.Transaction) {
 			authned, err := requireAccessKey(c, tx, srv)
 			if err != nil {
 				sendAPIError(c, err)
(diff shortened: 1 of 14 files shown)


internal/server/middleware.go
```diff
@@ -132,29 +131,44 @@ func validateOrgMatchesRequest(req *http.Request, tx data.GormTxn, accessKeyOrg
 	}
 }
 
-func withDBTxn(ctx context.Context, db *gorm.DB, fn func(tx *gorm.DB)) {
-	err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		fn(tx)
-		return nil
-	})
-	// TODO: https://github.com/infrahq/infra/issues/2697
+func withDBTxn(ctx context.Context, db *data.DB, fn func(tx *data.Transaction) error) error {
```
dnephin, line 134: This function implements roughly the same logic as `gorm.DB.Transaction`, but lets us call our transaction Commit and Rollback instead.

In the near future I expect us to remove this function, or at least integrate it into the route wrapper, so that we can resolve the issues linked in TODOs from the caller.

dnephin (commented):
I've changed the focus of this PR and marked it ready for review.

I think this now addresses the misleading calls to `data.NewTransaction`.

internal/server/middleware.go
dnephin, line 134: This gets removed in #3104

mxyng (approved):

PR #3029, opened 2022-08-25
Replace delete and insert access key functions with sql

dnephin:
## Summary

Related to #2415, this PR adds the building blocks for creating regular sql queries, and replaces insert and delete access keys with regular sql queries.

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -79,7 +79,7 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
 	}
 
-	if err := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByUserID: id}); err != nil {
+	if err := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByIssuedForID: id}); err != nil {
 		return fmt.Errorf("delete identity access keys: %w", err)
 	}
 
@@ -179,7 +179,7 @@ func UpdateIdentityInfoFromProvider(c RequestContext, oidc providers.OIDCClient)
 			return err
 		}
 
-		if nestedErr := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByUserID: identity.ID}); nestedErr != nil {
+		if nestedErr := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByIssuedForID: identity.ID}); nestedErr != nil {
 			logging.Errorf("failed to revoke invalid user session: %s", nestedErr)
 		}
 
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -79,7 +79,7 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
 	}
 
-	if err := data.DeleteAccessKeys(db, data.ByIssuedFor(id)); err != nil {
+	if err := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByUserID: id}); err != nil {
 		return fmt.Errorf("delete identity access keys: %w", err)
 	}
 
@@ -179,7 +179,7 @@ func UpdateIdentityInfoFromProvider(c RequestContext, oidc providers.OIDCClient)
 			return err
 		}
 
-		if nestedErr := data.DeleteAccessKeys(db, data.ByIssuedFor(identity.ID)); nestedErr != nil {
+		if nestedErr := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByUserID: identity.ID}); nestedErr != nil {
 			logging.Errorf("failed to revoke invalid user session: %s", nestedErr)
 		}
 
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -129,11 +129,9 @@ func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, err
 	return result, nil
 }
 
-func DeleteAccessKey(db GormTxn, id uid.ID) error {
-	return delete[models.AccessKey](db, id)
-}
-
 type DeleteAccessKeysOptions struct {
+	// ByID instructs DeleteAccessKeys to delete the key with this ID.
+	ByID uid.ID
 	// ByUserID instructs DeleteAccessKeys to delete keys issued for this user.
 	ByUserID uid.ID
 	// ByProviderID instructs DeleteAccessKeys to delete keys issued by this
@@ -145,6 +143,8 @@ func DeleteAccessKeys(tx WriteTxn, opts DeleteAccessKeysOptions) error {
 	query := Query("UPDATE access_keys")
 	query.B("SET deleted_at = ? WHERE", time.Now())
 	switch {
+	case opts.ByID != 0:
+		query.B("id = ?", opts.ByID)
 	case opts.ByUserID != 0:
 		query.B("issued_for = ?", opts.ByUserID)
 	case opts.ByProviderID != 0:
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -132,8 +132,8 @@ func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, err
 type DeleteAccessKeysOptions struct {
 	// ByID instructs DeleteAccessKeys to delete the key with this ID.
 	ByID uid.ID
-	// ByUserID instructs DeleteAccessKeys to delete keys issued for this user.
-	ByUserID uid.ID
+	// ByIssuedForID instructs DeleteAccessKeys to delete keys issued for this user.
+	ByIssuedForID uid.ID
 	// ByProviderID instructs DeleteAccessKeys to delete keys issued by this
 	// provider.
 	ByProviderID uid.ID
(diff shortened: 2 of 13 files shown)


internal/server/data/query.go
```diff
@@ -0,0 +1,120 @@
...
+	q := &queryBuilder{}
+	q.query.WriteString(stmt + " ")
+	return q
+}
+
+// B adds clause and args to the query. The clause must be a trusted string
+// literal. Any arguments must be passed as args so that they are properly
+// escaped by the database driver.
+func (q *queryBuilder) B(clause string, args ...interface{}) {
+	q.query.WriteString(clause + " ")
```
BruceMacD, line 29: Im really weary of adding this string clause directly to a query string, feels vulnerable

internal/server/data/query.go
```diff
@@ -0,0 +1,120 @@
...
+type Selectable interface {
+	Table
+	// ScanFields returns pointers to all the fields, which should be used in
+	// sql.Rows.Scan. The fields must be in the same order as the column names
+	// returned by Columns.
+	ScanFields() []any
+}
+
+func insert(tx WriteTxn, item Insertable) error {
+	if err := item.OnInsert(); err != nil {
```
BruceMacD, line 77: I'd like to see the org selector in one of these functions.

BruceMacD (commented):
Glad to see this coming along, this is a big milestone on all the work you've been doing

internal/access/identity.go
```diff
@@ -79,7 +79,7 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
 	}
 
-	if err := data.DeleteAccessKeys(db, data.ByIssuedFor(id)); err != nil {
+	if err := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByUserID: id}); err != nil {
```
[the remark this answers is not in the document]
dnephin: Ooops, good catch!

internal/server/models/access_key.go
```diff
@@ -21,15 +21,15 @@ type AccessKey struct {
...
+	IssuedForIdentity *Identity `gorm:"foreignKey:IssuedFor" db:"-"`
 	ProviderID        uid.ID
 
 	ExpiresAt         time.Time
 	Extension         time.Duration // how long to increase the lifetime extension deadline by
 	ExtensionDeadline time.Time
 
 	KeyID          string `gorm:"<-;uniqueIndex:idx_access_keys_key_id,where:deleted_at is NULL"`
-	Secret         string `gorm:"-"`
+	Secret         string `gorm:"-" db:"-"`
```
BruceMacD, line 32: What does the `db:"-"` tag do? It looks like this explicitly not added as a column in `data/access_key.go`

internal/access/identity.go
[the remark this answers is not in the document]
dnephin: Oh, I think I did change the name of this. I can change it back to `ByIssusedFor` to match the name of the field on the model.

internal/server/data/access_key.go
```diff
@@ -17,6 +16,24 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
+type accessKeyTable models.AccessKey
+
+func (a accessKeyTable) Table() string {
+	return "access_keys"
+}
+
+func (a accessKeyTable) Columns() []string {
```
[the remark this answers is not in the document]
dnephin: They are generated for now, but I don't think we want to force everyone to use the generated code to make small changes. It often ends up being more work to figure out the code generation than to simply add the new field to these 3 methods. It should be fine for someone to manually edit these when we add one or two fields, and use code gen when we make larger changes. It would be nice to support both workflows.

I think this is one of the last remaining open questions about this approach.

> If we create them maybe we should have a shared list of common columns for "created_at", "deleted_at", etc.

That's possible, but it will remove the option of using code gen for larger changes. We need these sorted alphabetically so that they match the list of fields in the other methods.

internal/server/data/access_key.go
```diff
@@ -76,7 +93,7 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err
 		accessKey.Name = fmt.Sprintf("%s-%s", identityIssuedFor.Name, accessKey.ID.String())
 	}
 
-	if err := add(db, accessKey); err != nil {
+	if err := insert(db, (*accessKeyTable)(accessKey)); err != nil {
```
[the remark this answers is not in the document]
dnephin: I don't think so. We could remove this type conversion by moving our models into the `data` package, then we could implement the methods on the model directly. That's a larger change, and I expect it to be controversial. So for now I think we have to type convert when calling these helpers.

Even if there was a way to infer it, I don't think we'd want to. The explicit types here make it very easy to go and find the methods that are being used. Things like reflection make that much more difficult. I think code is always better off explicit than allowing it to infer things implicitly.

internal/server/data/access_key.go
```diff
@@ -112,21 +129,33 @@ func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, err
 	return result, nil
 }
 
-func DeleteAccessKey(db GormTxn, id uid.ID) error {
-	return delete[models.AccessKey](db, id)
+type DeleteAccessKeysOptions struct {
+	// ByID instructs DeleteAccessKeys to delete the key with this ID.
+	ByID uid.ID
+	// ByUserID instructs DeleteAccessKeys to delete keys issued for this user.
+	ByUserID uid.ID
```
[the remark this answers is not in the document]
dnephin: Fail enough, I'll push a commit which changes this back to `ByIssusedForID`

internal/server/data/access_key.go
```diff
@@ -112,21 +129,33 @@ func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, err
...
-	ids := slice.Map[models.AccessKey, uid.ID](toDelete, func(k models.AccessKey) uid.ID {
-		return k.ID
-	})
-
-	return deleteAll[models.AccessKey](db, ByIDs(ids))
+func DeleteAccessKeys(tx WriteTxn, opts DeleteAccessKeysOptions) error {
+	query := Query("UPDATE access_keys")
+	query.B("SET deleted_at = ? WHERE", time.Now())
+	switch {
+	case opts.ByID != 0:
```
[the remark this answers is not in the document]
dnephin: That works, but it's a slightly different pattern. I don't expect to see that pattern in write operations (insert, update, delete). I believe generally write operations only use one identifier.

Read operations, especially `list`, will very commonly use multiple options. You can see an example of that here from a much earlier draft:  https://github.com/infrahq/infra/pull/2657/files#diff-4d74142dc90ac5e0880f7655a7c285b71a1c3b1b03569066cfeb460a805a7038R94-R125

Instead of a `switch/case` we use `if` conditions. Unlike that early draft, I think we'll use `query.B` and explicitly add the `ADD` or `OR` between the clauses. I'll do that next for the access key get and list queries.

internal/server/data/access_key.go
```diff
@@ -112,21 +129,33 @@ func GetAccessKey(tx GormTxn, selectors ...SelectorFunc) (*models.AccessKey, err
...
+		query.B("id = ?", opts.ByID)
+	case opts.ByUserID != 0:
+		query.B("issued_for = ?", opts.ByUserID)
+	case opts.ByProviderID != 0:
+		query.B("provider_id = ?", opts.ByProviderID)
+	default:
+		return fmt.Errorf("DeleteAccessKeys requires an ID to delete")
+	}
+	query.B("AND organization_id = ?", tx.OrganizationID())
```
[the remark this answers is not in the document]
dnephin: There isn't really any shared code anymore, which I see as advantage. The queries are very explicit and obvious now, I don't have to dig through hundreds of lines to guess at how the query is built.

I could add a panic here when `OrganizationID` is 0, but I think if we step back a bit we'll see even that is not adding anything.

This `AND organization_id` isn't necessary in this query. We're already using IDs in the where clause, and all of those IDs (primary key, user id, or provider id) can only refer to items in the same org (if they were different orgs we'd have a bug on insert).

So even removing this `AND organization_id` would be safe, and potentially makes this code easier to use (because we don't have to artificially set an organization ID in a transaction when we don't need one).

I added this extra clause as a safeguard, but I think safeguards on safeguards is going too far.

internal/server/data/query.go
dnephin: We definitely need a linter for this, so that we only allow single string literals (never concatenated strings, variables, or function return values) to be passed as the first argument to `B`.

That linter will prevent any possible injection because a string literal can never come from the user, so there's no possible way for them to inject something.

All parameters from the user have to be passed as `args`.

I can work on that linter next week. It should be pretty easy to implement. We'll need to allow some exceptions for `columnsForInsert` and `placeholderForColumns` and `Table.Table`. Those exceptions are likely the most involved.

I don't see any way to avoid accepting string literals here. Did you have something in mind?

internal/server/data/query.go
dnephin: Not sure what you mean. These is an insert, so there's no selector. I do call `setOrg` just below here, which has the panic I think you are looking for.

`update` has to operate on a primary ID, so there's nothing to really validate. Any already existing DB item must already have an OrgID.

internal/server/models/access_key.go
dnephin: Good question. These are left over from the code gen. I think if we don't use code gen I'll add tests that validate the methods on the tables, and these struct tags will be used by the tests. Maybe we'll add the validation tests either way.

This `db:"-"` indicates to the code gen and to the validation that the field is not mirrored by a database column. It's used for some other purpose.

internal/server/data/query.go
BruceMacD: I didn't have anything specific in mind, the linter sounds like it could work. I'll keep an eye out for that PR.

BruceMacD (approved):
Thanks for answering my questions, looking forward to trying this out

internal/server/data/query.go
dnephin: Jeff recently [shared this post](https://blog.fuzzbuzz.io/fuzzing-go-apis-for-sql-injection/) about using fuzz testing to detect possible injection attacks. I think that's another good safeguard. I'll see if we can get some fuzz tests setup for our APIs to catch anything that might sneak past the linter.

Issue #3025, opened 2022-08-25
Fix lookup of default org


dnephin:
We are looking up the default org using `name = 'Default'`. This worked when we had a unique constraint on the org name, but we have since removed that constraint.

We need a way to look up the default organization. One way to do that would be to add a new field `default bool` to the table, and query for the record where that is set to true.

Another option is to always create the default org with a hardcoded ID, something small like `1000` can never be assigned by `uid.New`. With a predictable ID we can look it up by ID.

PR #3030, opened 2022-08-26
maintain: update config tests for orgs

### LINKED ISSUE #2917: improve tests for `Server.loadConfig`, our current tests set a default org in the DB, so may be masking something in production code.

BruceMacD:
- specify default org in config tests to verify that expected results are returned
- connector was issued the wrong grant on org creation

## Summary
Update config test queries to query specifically for the default org ID we expect. This revealed a bug where the default org connector was getting the wrong grant, which I have also corrected in this PR.

## Checklist

## Related Issues

Resolves #2917

### DIFF
--- a/internal/server/config_test.go
+++ b/internal/server/config_test.go
@@ -298,16 +298,15 @@ func TestLoadConfigEmpty(t *testing.T) {
 	err := s.loadConfig(Config{})
 	assert.NilError(t, err)
 
-	org, err := data.GetOrganization(s.DB(), data.ByName(models.DefaultOrganizationName))
-	assert.NilError(t, err)
+	defaultOrg := s.db.DefaultOrg
 
 	var providers, grants int64
 
-	err = s.db.Raw("SELECT COUNT(*) FROM providers WHERE organization_id = ?;", org.ID).Scan(&providers).Error
+	err = s.db.Raw("SELECT COUNT(*) FROM providers WHERE organization_id = ?;", defaultOrg.ID).Scan(&providers).Error
 	assert.NilError(t, err)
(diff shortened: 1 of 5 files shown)


internal/server/data/organization.go
```diff
@@ -41,7 +41,7 @@ func CreateOrganization(tx GormTxn, org *models.Organization) error {
 
 	err = CreateGrant(tx, &models.Grant{
 		Subject:   uid.NewIdentityPolymorphicID(connector.ID),
-		Privilege: models.InfraAdminRole,
+		Privilege: models.InfraConnectorRole,
```
BruceMacD, line 44: This erroneous grant was revealed by the config tests.

internal/server/config_test.go
```diff
@@ -300,15 +298,18 @@ func TestLoadConfigEmpty(t *testing.T) {
 	err := s.loadConfig(Config{})
 	assert.NilError(t, err)
 
+	org, err := data.GetOrganization(s.DB(), data.ByName(models.DefaultOrganizationName))
+	assert.NilError(t, err)
```
dnephin, line 302: Minor, but related to #3025, I'm trying to avoid looking up the default org in other places.

You can get the default org without looking it here with this:

```go
org := s.db.DefaultOrg
```

dnephin (approved):
Nice, looks great!

I think the bug you fixed is a bug for all orgs, not just the default

internal/server/config_test.go
BruceMacD: I meant to remove all these, thanks for catching this one

PR #3032, opened 2022-08-26
demo: alternative to the access package

### LINKED ISSUE #2697: api: failing to commit a transaction does not result in a 500 response

dnephin:
Update: many of the changes from the original draft were merged as part of #3103

This PR is a demonstration / proof-of-concept, that shows a different approach to our API authorization. I am **not** looking to merge this any time soon. I am opening this to demonstrate an approach that I've mentioned a few times in a number of discussions. This PR should only act as a demonstration of the possible approach. It will need more work before it's anywhere close to mergeable.

By moving authorization checks to be immediately after authn in the request flow can allow us to remove the `access` package without losing the benefits that the `access` package is supposed to provide.  By checking permissions at the data layer (where the operation is actually being performed) we end up with better safeguards than the `access` package can provide, while also fixing the problems we have now (misleading error message, and unnecessary application layers that cause confusion and slow us down).

### DIFF
--- a/internal/server/grants.go
+++ b/internal/server/grants.go
@@ -11,6 +11,7 @@ import (
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/access"
+	ac "github.com/infrahq/infra/internal/server/access"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
@@ -22,6 +23,25 @@ func (r ListGrantsResponse) SetHeaders(h http.Header) {
 	if r.LastUpdateIndex.Index > 0 {
 		h.Set("Last-Update-Index", strconv.FormatInt(r.LastUpdateIndex.Index, 10))
 	}
+}
+
+type requestID interface {
+	RequestID() uid.ID
+}
+
+var listGrantsRoute = route[api.ListGrantsRequest, *api.ListResponse[api.Grant]]{
+	authorization: func(rCtx access.RequestContext, req any) (ac.PermissionSet, error) {
+		if reqID, ok := req.(requestID); ok {
+			if reqID.RequestID() == rCtx.Authenticated.User.ID {
+				// TODO: how does permission set work in this case?
+				return nil, nil
+			}
+
+			// TODO: user in group
+		}
+		// TODO: check ownership first
+		return access.IsAuthorized(rCtx, ac.RoleAdmin, ac.RoleView, ac.RoleConnector)
+	},
 }
 
 func (a *API) ListGrants(c *gin.Context, r *api.ListGrantsRequest) (*ListGrantsResponse, error) {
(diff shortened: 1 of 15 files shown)


internal/server/routes.go
```diff
@@ -140,41 +149,98 @@ func (s *Server) GenerateRoutes() Routes {
...
-	infraHeaderOptional bool
+	method                     string
+	path                       string
+	handler                    HandlerFunc[Req, Res]
+	omitFromDocs               bool
+	omitFromTelemetry          bool
+	infraVersionHeaderOptional bool
+	noAuthentication           bool
+	noOrgRequired              bool
+	requiredAuthorization      []string
```
dnephin, line 160: Not demonstrated in this PR, but to have routes set these values I think we'd declare these routes as static package-level variables. For example:

```
var createGrant = route{
    ...
    handler: createGrantHandler,
    requiredAuthorization: adminRole,
}

```

The method we call `API.CreateGrant` today would be renamed to `createGrantHandler`.

That way `GenerateRoutes` will continue to look almost identical to what it is today, but we can specify the authentication, authorization, and org requirement next to the handler.

internal/server/routes.go
```diff
@@ -140,41 +149,98 @@ func (s *Server) GenerateRoutes() Routes {
...
 
-	wrappedHandler := func(c *gin.Context) {
-		if !route.infraHeaderOptional {
+	// TODO: this is temporary for demo purposes
+	route.noAuthentication = r.noAuthentication
+	route.noOrgRequired = r.noOrgRequired
+
+	handler := func(c *gin.Context) {
+		if err := wrapRoute(a, route)(c); err != nil {
+			sendAPIError(c, err)
```
dnephin, line 184: With this change, we now long need to call `sendAPIError` in 2 places, here for normal requests, and in the `noRouteHandler` when we don't match a route. Previously we called it all over the middleware.

I think this cleans up the code a bunch, because now things read as regular functions that return an error, instead of:

```go
sendAPIError(c, err)
return
```

internal/server/grants.go
```diff
@@ -14,6 +14,25 @@ import (
...
+
+var listGrantsRoute = route[api.ListGrantsRequest, *api.ListResponse[api.Grant]]{
+	authorization: func(rCtx access.RequestContext, req any) (ac.PermissionSet, error) {
+		if reqID, ok := req.(requestID); ok {
+			if reqID.RequestID() == rCtx.Authenticated.User.ID {
+				// TODO: how does permission set work in this case?
+				return nil, nil
+			}
+
+			// TODO: user in group
```
dnephin, line 29: These are the last problems to solve for this approach I think. If the data layer (and other lower level packages) need to check the permissions then I think the `PermissionSet` needs to include an ID. That's easy enough for the "owned by user" case. The "owned by group" case is more involved. The permissions might need to be functions instead of simply an ID.

dnephin:
Closing this in favor of #3562.

This approach went further by using permission checks at the data layer. That's a safeguard we can add later if we want it.

Issue #2979, opened 2022-08-19
Connector deploys with a "postgres" stateful set


BruceMacD:
## Describe the bug
When deploying the connector a new "infra-connector-postgres" stateful set is deployed alongside it.

## Expected behavior
The connector shouldn't need persistence.

## Screenshots
<img width="1440" alt="Untitled" src="https://user-images.githubusercontent.com/5853428/185608561-f70e1786-0425-45cd-bf67-0cc9a5814c2f.png">

## Environment
mainline dev

## Additional context
```
helm install infra-connector helm/charts/infra --set connector.config.name=docker-desktop --set connector.config.accessKey=${CONNECTOR_KEY} --set connector.config.server=dev.acme.internal --set global.image.pullPolicy=Never --set global.image.tag=dev
```

mxyng:
Closed by #3034

PR #3034, opened 2022-08-26
fix(helm): use helper to check if server enabled

mxyng:
## Summary

This fixes the case where postgres is enabled when only connector is enabled


BruceMacD (approved):
Resolves #2979

mxyng:
I didn't see that. Thanks for the link!

PR #3055, opened 2022-08-29
feat: add welcome signup email

### LINKED ISSUE #2971: email invite formatting issues
- [ ] Email template is managed in code so we can audit + approve changes (consider plaintext for now for easy rendering)
- [ ] Off-centered text: Screen Shot 2022-08-18 at 3 08 06 PM
- [ ] Remove "unsubscribe" since it's a transactional email

ssoroka:
## Summary

- switches to SMTP for email delivery instead of sendgrid api
- renders email templates locally instead of depending on remote service
- switches existing emails over to new templates
- adds welcome/signup email

## Checklist

also..
Resolves #2971

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,7 +24,6 @@ issues:
 linters:
   enable:
     - asciicheck
-    - deadcode
     - durationcheck
     - errcheck
     - errorlint
@@ -47,10 +46,8 @@ linters:
     - rowserrcheck
     - sqlclosecheck
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
     - wastedassign
   disable:
     - scopelint
--- a/internal/generate/generate.go
+++ b/internal/generate/generate.go
@@ -25,6 +25,8 @@ func CryptoRandom(n int, charset string) (string, error) {
 
 	bytes := make([]byte, n)
 	for i := range bytes {
+		// linter is mistaken about which package this is
+		// nolint: gosec
 		bigint, err := rand.Int(rand.Reader, big.NewInt(int64(len(charset))))
 		if err != nil {
 			return "", fmt.Errorf("couldn't generate random string of len %d: %w", n, err)
--- a/internal/server/email/smtp.go
+++ b/internal/server/email/smtp.go
@@ -1,20 +1,17 @@
 package email
 
 import (
-	"bufio"
 	"crypto/tls"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"net/smtp"
-	"strings"
-	"sync"
 	"time"
 )
 
 type Client struct {
 	*tls.Conn
-	reader *bufio.Reader
-	sync.Mutex
+	Client        *smtp.Client
 	skipTLSVerify bool
 }
 
@@ -28,7 +25,7 @@ type Message struct {
 	HTMLBody    []byte
 }
 
-var client = &Client{}
+var testClient = &Client{}
 
 func (c *Client) connect() (err error) {
 	if c.Conn != nil {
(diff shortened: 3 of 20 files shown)


.golangci.yml
```diff
@@ -47,10 +46,8 @@ linters:
     - rowserrcheck
     - sqlclosecheck
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
```
ssoroka, line 53: these are deprecated in favor of unused, which we already use.

internal/generate/generate.go
```diff
@@ -25,6 +25,8 @@ func CryptoRandom(n int, charset string) (string, error) {
 
 	bytes := make([]byte, n)
 	for i := range bytes {
+		// linter is mistaken about which package this is
```
ssoroka, line 28: seems like a 1.49 linter bug

.golangci.yml
ssoroka, line 53: these are deprecated in favor of `unused`, which we already use

internal/server/email/smtp.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	c.Conn = conn
+
+	defer func() {
+		if err != nil {
+			c.Conn = nil
+		}
+	}()
+
+	c.reader = bufio.NewReader(c.Conn)
```
pdevine, line 63: Why not use use "net/smtp" here? I'm worried this code might be fragile.

I think you can do something like:
```
    auth := smtp.PlainAuth("","mail@infrahq.com", "password", host)
    
    client, err := smtp.NewClient(conn, host)
    if err != nil {
        log.Panic(err)
    }

    // Auth
    if err = client.Auth(auth); err != nil {
        log.Panic(err)
    }
```

internal/server/email/smtp.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+	}
+	if len(msg.Subject) > 0 {
+		if err := client.writeln(fmt.Sprintf("Subject: %s", msg.Subject)); err != nil {
+			return fmt.Errorf("write subject: %w", err)
+		}
+	}
+
+	// mime multipart
+	if err := client.writeln("MIME-Version: 1.0\r\nContent-Type: multipart/alternative; boundary=c3VwYWhpbmZyYQ\r\n"); err != nil {
```
BruceMacD, line 197: What does the `boundary` do?

ssoroka: part of how MIME formatting works, it tells the smtp server where each section starts and stops

internal/server/email/smtp.go
```diff
@@ -0,0 +1,222 @@
...
+// SendSMTP sends an email message
+func SendSMTP(msg Message) error {
+	client.Lock()
+	defer client.Unlock()
+
+	if err := client.connect(); err != nil {
+		return err
+	}
+
+	if err := client.writeln(fmt.Sprintf("MAIL FROM: %s", msg.FromAddress)); err != nil {
```
pdevine, line 142: Why not use `smtp.SendMail()` instead?

ssoroka: it was just so simple. :D I can change it though.

ssoroka: So one problem I'm running into with smtp.SendMail is that it assumes :smtp port (25) and then tries to upgrade to tls if the server supports it. This is different from what I'm doing; I'm connecting to a TLS-encrypted SMTP port and failing if it doesn't connect. Switching to this other code means there's a potential silent downgrade to non-TLS if the destination server doesn't support TLS

pdevine (approved):
LGTM

PR #3058, opened 2022-08-29
feat: update destination add to use api endpoint

mxyng:
## Summary

- Use `api.{baseDomain}` when templating destination add command
- Use values file instead of `--set`
- Use `upgrade --install` instead of `install`
- Make `baseDomain` required when signup is enabled 

image

### DIFF
--- a/ui/pages/destinations/add.js
+++ b/ui/pages/destinations/add.js
@@ -2,8 +2,23 @@ import Head from 'next/head'
 import Link from 'next/link'
 import { useState, useEffect } from 'react'
 import copy from 'copy-to-clipboard'
+import yaml from 'js-yaml'
 
 import Fullscreen from '../../components/layouts/fullscreen'
+import { useServerConfig } from '../../lib/serverconfig'
+
+function download(values) {
+  const blob = new Blob([values], { type: 'application/yaml' })
+  const href = URL.createObjectURL(blob)
+  const link = document.createElement('a')
+
+  link.href = href
+  link.setAttribute('download', 'values.yaml')
+  document.body.appendChild(link)
+  link.click()
+  document.body.removeChild(link)
+  URL.revokeObjectURL(href)
+}
 
 export default function DestinationsAdd() {
   const [name, setName] = useState('')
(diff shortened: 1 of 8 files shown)


ui/pages/destinations/add.js
```diff
@@ -2,8 +2,10 @@ import Head from 'next/head'
 import Link from 'next/link'
 import { useState, useEffect } from 'react'
 import copy from 'copy-to-clipboard'
+import yaml from 'js-yaml'
```
dnephin, line 5: JSON is valid yaml, so we could potentially avoid this dependency by using `JSON.stringify()` and writing it out as JSON.

dnephin (commented):
The Go changes LGTM, but I notice there is one test failing that needs to be updated for this new validation.

I guess using a heredoc with a `values.yaml` file should make the command easier to copy.

ui/pages/destinations/add.js
```diff
@@ -66,17 +69,24 @@ export default function DestinationsAdd() {
...
+  }
 
   if (window.location.protocol !== 'https:') {
-    command += ` \\
-    --set connector.config.skipTLSVerify=true`
+    values.connector.config.skipTLSVerify = true
   }
 
+  const valuesYaml = yaml.dump(values)
+  const command = `cat <<EOF | helm upgrade --install infra-connector infrahq/infra -f-\n${valuesYaml}EOF`
```
BruceMacD, line 88: Can we just do this all with flags? The `cat` piping to get multi-line input looks confusing imo

mxyng: We could. It would look like the previous command snippet. IMO it's actually harder to read and understand. Flags simulate a hierarchy which can easily be overlooked. This makes the relationship between the values clear while also making it easier to edit or append.

Additionally, the settings doesn't translate directly to a values file so if the user wants to adds the outputs to IAC, they'll need to transcribe it into YAML. This way, we can add an option to download the file instead of using a heredoc and the `helm` command will remain the same.

ui/pages/destinations/add.js
mxyng: We could do that but IMO JSON is less user friendly with its quotes and braces and brackets. I don't think it's worth using JSON solely to avoid this dependency.

BruceMacD (approved):

hoyyeva (approved):

PR #3059, opened 2022-08-29
fix: new signup session priority

BruceMacD:
## Summary
If a user has an `auth` cookie for the base Infra domain or an old org that had the same domain it will get sent to the server along with the `signup` cookie. The `signup` cookie should take priority over this `auth` cookie since it shows that the user is clearly being redirected after creating a new org and no previous auth should exist.

## Checklist


jmorganca (approved):
LGTM

PR #3060, opened 2022-08-29
Add a linter for the new SQL query builder

dnephin:
## Summary

Branched from #3029 (I need to merge that first).

PR #3029 introduced a new query builder for building SQL queries. The biggest risk with writing our own SQL is that we have to be careful that any strings added to the trusted side of the statement are indeed trusted and don't come from user input. This PR adds a linter to warn if we ever pass potentially untrusted strings to the query builder. 

We will guard against SQL injection using the following approach:
1. `queryBuilder` is moved to `querybuilder.Builder` so that only this small packages has direct access to the `query strings.Builder` field. This allows our linter to only check calls to exported methods. There is only one exported method (`Builder.B`) and one constructor (`NewQuery`), which are audited manually.
2. The linter checks that `NewQuery` and `Builder.B` are only called with a string literal as the first argument (with some explicit hard-coded exceptions, see 3)
3. There are 4 functions that are listed as exceptions, all 4 of these functions are documented as special cases in the `query.go` file, and they should only accept the `Table` type and use values from the `Table.Columns` method.
4. The linter checks that all implementations of `Table.Columns` return only a slice literal with values that are string literals.

Given these rules, the SQL statement produced by `querybuilder.Builder` should only every contain strings that came from string literals. String literals are always safe, because they are part of the binary and don't come from user input.  With this linter, the only risk is in the `querybuilder` package, and the 4 special functions. I've added godoc to all of these places to make it clear how to safely make changes to those functions. The implementation of all of these are very small (2 to 4 lines of code) and are unlikely to change in any significant way.


internal/server/data/query.go
```diff
@@ -3,42 +3,20 @@ package data
...
-// String returns the query string, which is used as the first parameter to
-// WriteTxn.Exec, or ReadTxn.Query. You must also pass q.Args as the varargs.
-func (q *queryBuilder) String() string {
-	return q.query.String()
-}
-
 type Table interface {
-	Table() string
-	// Columns returns the names of the tables columns.
+	Table(*querybuilder.Builder)
```
dnephin, line 11: By inverting the implementation of the `Table` method like this we can remove it as a special case. The `Table` method can call `Builder.B` like everything else.

I could not find a way to do that with the column functions without having to generate 4 separate methods for every type.

Makefile
```diff
@@ -47,6 +47,7 @@ postgres:
 
 
 lint:
+	(cd ./internal/tools/querylinter/cmd; go build -o ./querylinter.so -buildmode=plugin .)
```
BruceMacD, line 50: Any way we can get around building this every time we run linting, or is it pretty lightweight?

internal/server/data/query.go
```diff
@@ -79,22 +57,31 @@ func insert(tx WriteTxn, item Insertable) error {
...
 	_, err := tx.Exec(query.String(), query.Args...)
 	return err
 }
 
-func columnsForInsert(columns []string) string {
-	return strings.Join(columns, ", ")
+// columnsForInsert is a privileged function that is not checked by
+// internal/tools/querylinter. If the arguments to this function change
+// the linter will likely need to be updated.
+// The return value must only include trusted strings.
```
BruceMacD, line 74: What is a trusted string in this context?

dnephin: Good question, it should mean "not from a user", as user input is never trusted. I'll think about how to make this more clear.

dnephin: I've updated this to:

> // The return value must only include trusted strings from the source code,
> // never untrusted user input.

PR #3065, opened 2022-08-30
fix: correct article based on table name

### LINKED ISSUE #3064: Typo in error message
Screen Shot 2022-08-30 at 10 56 39 AM

BruceMacD:
## Summary
Add a check in the error string builder for table names that require the 'an' article rather than 'a'.

## Checklist

## Related Issues

Resolves #3064


dnephin (approved):
Looks good!

PR #3066, opened 2022-08-30
Add validation to org subdomain during signup, and add a list of reserved subdomains.

dnephin:
## Summary

This PR adds validation to the org subdomain, including min/max length, character restrictions, and a list of reserved words that we should not allow to be subdomains.

Best viewed by individual commit, more details in commit messages.

### DIFF
--- a/api/passwordreset.go
+++ b/api/passwordreset.go
@@ -21,8 +21,13 @@ type VerifiedResetPasswordRequest struct {
 func (r VerifiedResetPasswordRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
 		validate.Required("token", r.Token),
-		validate.String("token", r.Token, 10, 10, validate.AlphaNumeric...),
+		validate.StringRule{
+			Name:            "token",
+			Value:           r.Token,
+			MinLength:       10,
+			MaxLength:       10,
+			CharacterRanges: validate.AlphaNumeric,
+		},
 		validate.Required("password", r.Password),
-		// todo: validate password meets org's criteria
 	}
 }
--- a/api/signup.go
+++ b/api/signup.go
@@ -12,10 +12,30 @@ type SignupOrg struct {
 	Subdomain string `json:"subDomain"`
 }
 
+var reservedSubDomains = []string{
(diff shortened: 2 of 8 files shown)


api/signup.go
```diff
@@ -12,6 +12,33 @@ type SignupOrg struct {
 	Subdomain string `json:"subDomain"`
 }
 
+var reservedSubDomains = []string{
```
mxyng, line 15: @ssoroka also had a list of reserved subdomains. Perhaps you should work together on this

dnephin: I'm happy to add or edit this list! I should have mentioned that in the PR description. 

Please do suggest changes to this list and I can update it.

api/passwordreset.go
```diff
@@ -21,8 +21,13 @@ type VerifiedResetPasswordRequest struct {
...
 	return []validate.ValidationRule{
 		validate.Required("token", r.Token),
-		validate.String("token", r.Token, 10, 10, validate.AlphaNumeric...),
+		validate.StringRule{
+			Name:            "token",
+			Value:           r.Token,
+			MinLength:       10,
+			MaxLength:       10,
+			CharacterRanges: validate.AlphaNumeric,
+		},
```
ssoroka, line 30: way less elegant :P

api/signup.go
```diff
@@ -12,6 +12,33 @@ type SignupOrg struct {
...
+	"grants", "connector", "login", "signup",
+	"system", "admin", "email", "bastion",
+}
+
+func (r SignupOrg) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("name", r.Name),
+		validate.Required("subDomain", r.Subdomain),
+		validate.ReservedStrings("subDomain", r.Subdomain, reservedSubDomains),
+		validate.StringRule{
```
ssoroka, line 27: not necessarily a blocker, but it's missing the can't-start-with-dash case

ssoroka (approved):

api/signup.go
dnephin: I added first character validation to the `StringRule` and added it here as well

PR #3071, opened 2022-08-30
fix: do not use request for signup cookie exchange

BruceMacD:
## Summary
#3059 had an issue that it wouldn't set the new auth cookie after sign-up because moving around the cookie logic meant that the key wasn't set where I expected it to be previously. 

## Checklist

### DIFF
--- a/internal/server/cookie.go
+++ b/internal/server/cookie.go
@@ -61,11 +61,11 @@ func deleteCookie(c *gin.Context, name, domain string) {
 }
 
 // exchangeSignupCookieForSession sets the auth cookie on the current host making the request
-func exchangeSignupCookieForSession(c *gin.Context, opts Options) {
+func exchangeSignupCookieForSession(c *gin.Context, opts Options) string {
 	signupCookie, err := getCookie(c.Request, cookieSignupName)
 	if err != nil {
 		logging.L.Trace().Err(err).Msg("failed to find signup cookie, this may be expected")
-		return
+		return ""
 	}
 
 	exp := time.Now().UTC().Add(opts.SessionDuration)
@@ -78,4 +78,6 @@ func exchangeSignupCookieForSession(c *gin.Context, opts Options) {
 	}
 	setCookie(c, conf)
 	deleteCookie(c, cookieSignupName, opts.BaseDomain)
+
+	return signupCookie
 }
--- a/internal/server/cookie.go
+++ b/internal/server/cookie.go
@@ -51,7 +51,7 @@ func setCookie(c *gin.Context, config cookieConfig) {
 
 func deleteCookie(c *gin.Context, name, domain string) {
 	http.SetCookie(c.Writer, &http.Cookie{
-		Name:     cookieAuthorizationName,
+		Name:     name,
 		MaxAge:   cookieMaxAgeDeleteImmediately,
 		Path:     cookiePath,
 		Domain:   domain,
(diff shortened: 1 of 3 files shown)


internal/server/cookie.go
```diff
@@ -61,21 +61,21 @@ func deleteCookie(c *gin.Context, name, domain string) {
...
-		setCookie(c, conf)
-		deleteCookie(c, cookieSignupName, baseDomain)
+func exchangeSignupCookieForSession(c *gin.Context, opts Options) {
+	signupCookie, err := getCookie(c.Request, cookieSignupName)
+	if err != nil {
+		logging.L.Trace().Err(err).Msg("failed to find signup cookie, this may be expected")
+		return
 	}
+
+	exp := time.Now().UTC().Add(opts.SessionDuration)
```
BruceMacD, line 71: Approximating the auth cookie expiry based on the session duration. This should work fine since this will be exchanged within a few minutes.

internal/server/cookie_test.go
```diff
@@ -44,12 +45,13 @@ func TestResendAuthCookie(t *testing.T) {
...
 	assert.Equal(t, len(c.Writer.Header()["Set-Cookie"]), 2)
-	expectedCookies := []string{
-		"auth=aaa; Path=/; Domain=dev.example.com; Max-Age=299; HttpOnly; SameSite=Strict",
-		"auth=; Path=/; Domain=example.com; Max-Age=0; HttpOnly; Secure",
-	}
-	assert.DeepEqual(t, c.Writer.Header()["Set-Cookie"], expectedCookies)
+
+	matched, err := regexp.MatchString("auth=aaa; Path=/; Domain=dev.example.com; Max-Age=\\d\\d; HttpOnly; SameSite=Strict", c.Writer.Header()["Set-Cookie"][0])
+	assert.NilError(t, err)
+	assert.Assert(t, matched)
```
dnephin, line 54: One option to avoid the regex would be to use store the `httptest.NewRecorder()` that we passed into the `gin.Context` and call `Result().Cookies()` on it (https://pkg.go.dev/net/http#Response.Cookies). 

That should give you the parsed structs, and allow you to use `DeepEqual` with options to account for unpredictable max age.

Not a blocker, just an option if we need to make further changes some other time.

internal/server/cookie.go
dnephin: Was this longer duration for the signup cookie necessary for the fix, or is this change related to something else?

dnephin (approved):
LGTM, just one question about the change to expiry time

internal/server/cookie.go
dnephin: Ohh, this is for the new cookie. That makes sense! For some reason I thought this was for the signup cookie, but it's clearly not.

internal/server/middleware.go
```diff
@@ -305,7 +305,7 @@ func reqBearerToken(c *gin.Context, baseDomain string) (string, error) {
 		*/
 		cookie, err := getCookie(c.Request, cookieSignupName)
 		if err == nil {
-			exchangeSignupCookieForSession(c, baseDomain)
+			exchangeSignupCookieForSession(c, opts)
```
dnephin, line 308: Not a blocker, but I wonder if we could remove the duplicate lines:

```
cookie, err := getCookie(c.Request, cookieSignupName)
```

There's one just above here, and another in `exchangeSignupCookieForSession`.

I suspect either `exchangeSignupCookieForSession` could return a `handled bool` to indicate to the caller how to proceed. Or we can look up the cookie here and pass it into `exchangeSignupCookieForSession`

BruceMacD: Made this change, pretty clean.

PR #3073, opened 2022-08-30
Add userID and orgID to API request logs

### LINKED ISSUE #3031: Provide detailed logging for an user request

We would like to see clear logging when user logs in 

Example of sample log 

```userId=*** orgId=1 uname=admin@example.com t=2022-08-26T16:31:16.208480891Z level=info msg="Request Completed" method=POST path=***** ```

## Environment Details

```
$ infra version
```

```
$ kubectl version
```

dnephin:
## Summary

This PR moves the logging middleware to the `server` package.
Primarily this is being done so that we can add the userID and orgID to
the request. To fetch the user ID from the request we need to import the
access package, but the access package imports the logging package. So
attempting to get the userID caused an import cycle.

Since this middleware imports `gin.Context`, it's not really appropriate
for a logging package that gets imported by almost every package. This
middleware is specific to the API server, so we can keep it in the server
package.

The PR also adds the userID, and orgID to the API request log, and add
`API request completed` as the message for those logs, so that it's easy to filter
them from other log messages.

I also refactored the logic a bit to remove the need to switch between a logging event,
logging context, and back to a logger.

## Related Issues

Related to #3031, but we'll also need another change that sets the user in the RequestContext after login or signup is complete. Otherwise this will only work for authenticated routes.


dnephin:
Ah, this is missing one thing, which is to set the user in the request context as part of login.

BruceMacD (approved):
looks good

PR #3077, opened 2022-08-31
fix: clients should stop requesting when told their token is invalid

ssoroka:
## Summary

Clients (incl connectors) should respect the 401 Unauthorized received from an access token; it's not a transient error and requesting with it in the future will never work.

If the client knows the token is expired it shouldn't bother requesting.

## Checklist

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -35,7 +35,8 @@ type Client struct {
 	AccessKey string
 	HTTP      http.Client
 	// Headers are HTTP headers that will be added to every request made by the Client.
-	Headers http.Header
+	Headers        http.Header
+	OnUnauthorized func()
 }
 
 // checkError checks the resp for an error code, and returns an api.Error with
@@ -108,6 +109,9 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 	}
 
 	resp, err := client.HTTP.Do(req)
+	if resp != nil && resp.StatusCode == 401 && client.OnUnauthorized != nil {
+		defer client.OnUnauthorized()
+	}
 	if err != nil {
 		if connError := HandleConnError(err); connError != nil {
 			return nil, connError
--- a/api/client.go
+++ b/api/client.go
@@ -35,7 +35,9 @@ type Client struct {
 	AccessKey string
 	HTTP      http.Client
 	// Headers are HTTP headers that will be added to every request made by the Client.
-	Headers        http.Header
+	Headers http.Header
+	// OnUnauthorized is a callback hook for the client to get notified of a 401 Unauthorized response to any query.
+	// This is useful as clients often need to discard expired access keys.
 	OnUnauthorized func()
 }
 
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -97,7 +97,7 @@ func defaultAPIClient() (*api.Client, error) {
 		if config.isExpired() {
 			return nil, Error{Message: "Access key is expired, please login again", OriginalError: ErrAccessKeyExpired}
 		}
-		return nil, Error{Message: "Missing access key, must login or provide an INFRA_ACCESS_KEY", OriginalError: ErrAccessKeyMissing}
+		return nil, Error{Message: "Missing access key, must login or set INFRA_ACCESS_KEY in your environment", OriginalError: ErrAccessKeyMissing}
 	}
 
 	if envServer, ok := os.LookupEnv("INFRA_SERVER"); ok {
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -84,10 +84,20 @@ func defaultAPIClient() (*api.Client, error) {
 	}
 
 	server := config.Host
-	accessKey := config.AccessKey
+	var accessKey string
+	if !config.isExpired() {
+		accessKey = config.AccessKey
+	}
 
 	if envAccessKey, ok := os.LookupEnv("INFRA_ACCESS_KEY"); ok {
 		accessKey = envAccessKey
+	}
+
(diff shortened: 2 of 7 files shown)


internal/cmd/logout.go
```diff
@@ -63,25 +63,25 @@ $ infra logout --all --clear`,
 }
 
 func logoutOfServer(hostConfig *ClientHostConfig) (success bool) {
-	if !hostConfig.isLoggedIn() {
-		logging.Debugf("requested but not logged in to server [%s]", hostConfig.Host)
-		return false
-	}
```
ssoroka, line 69: this never made sense; a user can be logged out due to expiry, but still need the settings unset.

internal/connector/connector.go
```diff
@@ -211,6 +215,9 @@ func Run(ctx context.Context, options Options) error {
 		TLSConfig:         tlsConfig,
 		Handler:           router,
 		ErrorLog:          httpErrorLog,
+		BaseContext: func(l net.Listener) context.Context {
+			return ctx
+		},
```
ssoroka, line 220: so the http server exits when the main context is cancelled.

internal/cmd/cmd.go
```diff
@@ -84,12 +84,22 @@ func defaultAPIClient() (*api.Client, error) {
...
 
 	if envAccessKey, ok := os.LookupEnv("INFRA_ACCESS_KEY"); ok {
 		accessKey = envAccessKey
 	}
 
+	if len(accessKey) == 0 {
+		if config.isExpired() {
+			return nil, Error{Message: "Access key is expired, please login again", OriginalError: ErrAccessKeyExpired}
+		}
+		return nil, Error{Message: "Missing access key, must login or provide an INFRA_ACCESS_KEY", OriginalError: ErrAccessKeyMissing}
```
BruceMacD, line 100:
```suggestion
		return nil, Error{Message: "Missing access key, must login or set INFRA_ACCESS_KEY in your environment", OriginalError: ErrAccessKeyMissing}
```

internal/cmd/cmd.go
```diff
@@ -107,6 +117,42 @@ func apiClient(host string, accessKey string, transport *http.Transport) *api.Cl
 			Timeout:   60 * time.Second,
 			Transport: transport,
 		},
+		OnUnauthorized: logoutCurrent,
```
BruceMacD, line 120: I think we only want to logout in the case that a 401 is returned on `/tokens` for now, otherwise running a command you're not authorized for (ex: `infra users list`) will log you out.

dnephin: I think `OnUnauthorized` is only called when the response code is 401, right?

dnephin (commented):
Looks good! Mostly just a few questions about user experience

internal/connector/connector.go
dnephin: Ah, but then we still have the issue of many 401 server requests. Right. I think even if we shutdown the connector we still have that problem because the pod will be restarted and the requests will start again.

Maybe we should use exponential backoff when there are failures? Or we could stop making requests and keep logging an error that says to fix the access key?

ssoroka: Nice catch on the shutdown. I'll improve the code around server shutdown.
At least if the pod is failing it's likely to trigger some attention from the admin.

internal/cmd/cmd.go
ssoroka: yes. Unauthorized is specific to authn problems, unrelated to authz and access.

internal/connector/connector.go
dnephin: Thinking about this more, is the intent here to block requests to the connector so that a connector that is not in contact with the infra server can't continue to proxy requests to kube? I think that's an important thing to do, but I wonder:
1. should allow some grace period between the first failure and when we start refusing connections?
2. should we also do this for other errors, not just 401. If the connector has failed to query the API for any reason, it should have the same behaviour after some period of time
3. should we keep the http server running, but return a clear error message like "infra connector has failed to communicate with the infra server for x seconds. Please contact your administrator to fix the problem."

internal/cmd/cmd.go
```diff
@@ -107,6 +117,42 @@ func apiClient(host string, accessKey string, transport *http.Transport) *api.Cl
...
+	if host == nil {
+		return
+	}
+
+	host.AccessKey = ""
+	host.Expires = api.Time{}
+	host.UserID = 0
+	host.Name = ""
+
+	if err := clearKubeconfig(); err != nil {
```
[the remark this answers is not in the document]
mxyng: IIRC, we only sync the kubeconfig entries from the current infra context so removing all infra entries is right behaviour

internal/connector/connector.go
dnephin: I created #3079 to track further improvements

internal/cmd/cmd.go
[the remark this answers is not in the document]
ssoroka: I was wondering about that too. didn't dig into it, but it seems like weird behavior. I think your suggestion is a good one.

api/client.go
```diff
@@ -35,7 +35,8 @@ type Client struct {
 	AccessKey string
 	HTTP      http.Client
 	// Headers are HTTP headers that will be added to every request made by the Client.
-	Headers http.Header
+	Headers        http.Header
+	OnUnauthorized func()
```
dnephin, line 39: minor and not blocking: some godoc on this field would be great. Something about when it gets called so that someone using the `Client` knows how to use it.

dnephin (approved):
LGTM!

Issue #3076, opened 2022-08-31
Rollback transactions in middleware does not have an effect


BruceMacD:
## Describe the bug
The way we currently handle runtime errors means that errors are never fully bubbled up to our middleware database transaction, so errors would not cause a database transaction rollback.

## Expected behavior
Database transactions are handled in a way that can receive the error and rollback appropriately. 

## Environment
- Infra version <=0.14.5

## Additional context
```
// middleware.go
func withDBTxn(ctx context.Context, db *gorm.DB, fn func(tx *gorm.DB)) {
	err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
		fn(tx)
                // need to return an error here if the function fails
		return nil
	})
	// TODO: https://github.com/infrahq/infra/issues/2697
	if err != nil {
		logging.L.Error().Err(err).Msg("failed to commit database transaction")
	}
}
```

BruceMacD:
Related: #2697

dnephin:
I think I have a reasonable fix for this

PR #3081, opened 2022-08-31
fix: check signup password requirements in api

BruceMacD:
- update string length api error to be more ui friendly
- validate signup password between 8 and 253 characters

## Summary
Due to #3076 it was possible it sign-up and fail the password check after your org was created. Leaving the org in a bad state. Move the password requirement check to the API to prevent this.

Also updating the string error message to be more UI friendly.

Screen Shot 2022-08-31 at 3 42 48 PM

## Checklist

### DIFF
--- a/internal/server/signup_test.go
+++ b/internal/server/signup_test.go
@@ -9,6 +9,7 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/server/data"
 )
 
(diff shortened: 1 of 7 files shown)


internal/server/signup_test.go
```diff
@@ -59,13 +60,44 @@ func TestAPI_Signup(t *testing.T) {
...
+
+				expected := []api.FieldError{
+					{FieldName: "password", Errors: []string{
+						"password must be at least 8 characters",
+					}},
+				}
+				assert.DeepEqual(t, respBody.FieldErrors, expected)
+
+				// the org should have been rolled back
+				_, err = data.GetOrganization(srv.DB(), data.ByDomain("hello.example.com"))
```
dnephin, line 97: There's a change in `main` that updates this test to use a baseDomain of `exampledomain.com` instead of the old default of `example.com`. I think that changes won't fail the test, but it does mean the test isn't covering the behaviour we want.  

It'd be good to rebase and either use `hello.exampledomain.com` here, or change the `BaseDomain` in this test to `example.com`.

internal/validate/string.go
```diff
@@ -84,11 +84,11 @@ func (s StringRule) Validate() *Failure {
 		problems = append(problems, fmt.Sprintf(format, args...))
 	}
 	if s.MinLength > 0 && len(value) < s.MinLength {
-		add("length of string is %d, must be at least %d", len(value), s.MinLength)
+		add("%s must be at least %d characters", s.Name, s.MinLength)
```
dnephin, line 87: Repeating the field name in the message I think will end up duplicating the field name in the consolidated error message here: https://github.com/infrahq/infra/blob/main/internal/validate/validate.go#L121

We could do make a similar change for all the validators and remove it from `Error.Error`, but I kind of like the way it works now. It gives the caller the option to prepend the field name or group them differently if they choose.

What do you think about removing the field name from here, and instead prepending it in the UI ?

Otherwise the rewording of the message sounds good to me.

dnephin (commented):
Looks good! I think this is a good way to mitigate the larger problem for now, without making a big change to how transactions are managed.

Just the one concern about repeating the field names in the failure messages.

internal/validate/string.go
BruceMacD: My reason for rewording this was less about the field name and more about the error message being user-friendly to people without programming experience. This message is actually displayed in the UI.

`length of string is X, must be at least Y` is confusing to the average person, I'll move the name of the field up though 👍

dnephin (approved):
LGTM!

PR #3083, opened 2022-08-31
Send auth header from connector when requesting the jwk

dnephin:
## Summary

This allows the connector to use any server URL instead of having to use the org specific url.

The test is a bit artificial, but I don't think we have a better way of testing the connector and server interaction at this time.


BruceMacD (approved):
thanks for adding this

PR #3084, opened 2022-08-31
Remove sqlite support

dnephin:
## Summary

This PR removes sqlite support from both the server and tests.  The server `--dbFile` flag, and config file options will no longer do anything. From now on a postgreSQL database is required to run the server.

In tests, if the `POSTGRESQL_CONNECTION` environment variable is set, the tests will run using that postgres instance. If it's not set in CI the tests will fail. If it's not set outside of CI (local development) any tests that require the database will be skipped.

### DIFF
--- a/internal/server/data/migrations_test.go
+++ b/internal/server/data/migrations_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/infrahq/infra/internal/server/data/migrator"
 	"github.com/infrahq/infra/internal/server/data/schema"
 	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/internal/testing/database"
 	"github.com/infrahq/infra/internal/testing/patch"
 	"github.com/infrahq/infra/uid"
 )
@@ -575,7 +576,7 @@ DELETE FROM settings WHERE id=24567;
 	var initialSchema string
 	runStep(t, "initial schema", func(t *testing.T) {
 		patch.ModelsSymmetricKey(t)
-		rawDB, err := newRawDB(postgresDriver(t))
+		rawDB, err := newRawDB(database.PostgresDriver(t, "").Dialector)
 		assert.NilError(t, err)
 
 		db := &DB{DB: rawDB}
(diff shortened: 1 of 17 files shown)


internal/testing/database/postgres.go
```diff
@@ -25,8 +29,11 @@ type TestingT interface {
 func PostgresDriver(t TestingT, schemaSuffix string) *Driver {
 	t.Helper()
 	pgConn, ok := os.LookupEnv("POSTGRESQL_CONNECTION")
-	if !ok {
-		return nil
+	switch {
+	case !ok && isEnvironmentCI:
+		t.Fatal("CI must test all drivers, set POSTGRESQL_CONNECTION")
+	case !ok:
+		t.Skip("Set POSTGRESQL_CONNECTION to test against postgresql")
```
BruceMacD, line 36: So many of our tests rely on this I wonder if we should always fail without it

internal/server/data/migrations_test.go
```diff
@@ -646,6 +647,8 @@ type testCaseLabel struct {
 	Line string
 }
 
+var isEnvironmentCI = os.Getenv("CI") != ""
```
BruceMacD, line 650: Should this variable live closer to there it is used? I dont see it in `migrations_test.go`

BruceMacD (approved):
Great to see this step getting to the finish line

internal/server/data/migrations_test.go
dnephin: It's used 6 lines below here. The diff looks weird because I removed it from a different place in the `data` package.

internal/testing/database/postgres.go
dnephin: The tests for our main interfaces (CLI, API, config) do all rely on a database, but there are plenty of tests for internal things that do not. It's easy to change, but I think it can be nice to let people run tests with minimal setup.

dnephin:
I pushed one more commit that removed a couple things I had missed:
* removes `data.NewSQLiteDriver`
* one test in `models` was using that call, so it was updated to use `database.PostgresDriver`
* removed sqlite testing of migrator

Those changes allow us to remove our dependency on `gorm.io/driver/sqlite` and `mattn/go-sqlite3`.

PR #3085, opened 2022-08-31
improve server db connection string options

dnephin:
Previously the db connection string could either be entirely set by the
`dbConnectionString` option, or the user had to set `dbHost` if they wanted
to use the other options.

With this change the user can set `dbConnectionString` for most options,
and use only `dbPassword` to pull the secret value from a safe place.

This should allow us to deprecate most of the other db options at some point.
It also makes `dbParameters` effectively a duplicate of `dbConnectionString`, so
it would probably make sense to deprecate and remove at least one of those.

Branched from #3084, that will need to merge first.


BruceMacD (approved):

mxyng:
Good change overall though I'm in favour of keeping the individual options since it's easier to set from a Helm point of view.

mxyng (approved):

PR #3088, opened 2022-08-31
Add user to context after login and signup

### LINKED ISSUE #3031: Provide detailed logging for an user request

dnephin:
## Summary

Primarily this PR sets `RequestContext.Authenticated.User` after login and signup. Once added to the request context it will appear in the API request log. All other requests already have this value set.  We don't have any way to test our logs right now, so I tested this manually on a branch that included #3073, and I saw the userId set on the log line for login.

To accomplish this goal I did a bit of refactoring to expose the user we had already fetched from the database. I tried to split that refactor into smaller commits, so view by individual commit if the full diff doesn't read well.

As part of this change I removed the `LoginMethod.RequiresUpdate` method, which had a TODO on it to be removed. The issue the TODO links to is already closed.

## Related Issues

Along with the changes in #3073, this PR resolves #3031


internal/access/login.go
```diff
@@ -3,27 +3,10 @@ package access
...
-	// the login may fail if the password presented was a one-time password that has been used.
-	// This can be removed when #1441 is resolved
-	requiresUpdate, err := loginMethod.RequiresUpdate(db)
-	if err != nil {
-		return nil, "", false, err
-	}
-
-	return key, bearer, requiresUpdate, nil
+func Login(c RequestContext, loginMethod authn.LoginMethod, keyExpiresAt time.Time, keyExtension time.Duration) (authn.LoginResult, error) {
+	return authn.Login(c.Request.Context(), c.DBTxn, loginMethod, keyExpiresAt, keyExtension)
```
dnephin, line 11: I would really like to remove this `access.Login` function, but I suspect that is a controversial change.

If we remove this function we'd have the api handler calling the `authn.Login` directly, which removes one hop when tracing the code. The `access.Login` function does nothing except pull two values out of RequestContext, which could easily be done by the caller without adding additional lines.

BruceMacD (approved):
Looks good, feel free to cut the login access layer before merging

dnephin:
I pushed one more commit that adds API test coverage for login with a temp password, to make sure I did not mess up anything with the refactor.

PR #3094, opened 2022-09-01
Replace get, list, and update access key queries with sql

dnephin:
## Summary

Branched from #3029, and including the commit from #3093

This PR replaces gorm with sql in the query functions for get, list, and update access keys, and also adds more test coverage for those functions.

The biggest difference is in the pagination, which I'll comment on inline.

## Related Issues

Related to https://github.com/infrahq/infra/issues/2415

### DIFF
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -100,12 +100,13 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err 
 	return fmt.Sprintf("%s.%s", accessKey.KeyID, accessKey.Secret), nil
 }
 
-func SaveAccessKey(db GormTxn, key *models.AccessKey) error {
+func UpdateAccessKey(tx WriteTxn, key *models.AccessKey) error {
 	if key.Secret != "" {
 		key.SecretChecksum = secretChecksum(key.Secret)
 	}
+	// TODO: we should perform the same validation as Create
 
-	return save(db, key)
+	return update(tx, (*accessKeyTable)(key))
 }
 
 func ListAccessKeys(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
@@ -185,7 +186,7 @@ func ValidateAccessKey(tx GormTxn, authnKey string) (*models.AccessKey, error) {
 		}
 
 		t.ExtensionDeadline = time.Now().UTC().Add(t.Extension)
-		if err := SaveAccessKey(tx, t); err != nil {
+		if err := UpdateAccessKey(tx, t); err != nil {
 			return nil, err
 		}
 	}
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -105,8 +105,61 @@ func UpdateAccessKey(tx WriteTxn, key *models.AccessKey) error {
 	return update(tx, (*accessKeyTable)(key))
 }
 
-func ListAccessKeys(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.AccessKey, error) {
-	return list[models.AccessKey](db, p, selectors...)
+type ListAccessKeyOptions struct {
+	IncludeExpired bool
+	ByIssuedForID  uid.ID
+	ByName         string
+	Pagination     *Pagination
+}
+
+func ListAccessKeys(tx ReadTxn, opts ListAccessKeyOptions) ([]models.AccessKey, error) {
+	table := &accessKeyTable{}
+	query := Query("SELECT")
+	query.B(columnsForSelect("k", table.Columns()))
+	query.B(", u.name")
+	if opts.Pagination != nil {
+		query.B(", count(*) OVER()")
+	}
+	query.B("FROM access_keys AS k INNER JOIN identities AS u")
+	query.B("ON k.issued_for = u.id")
+	query.B("WHERE k.deleted_at is null AND u.deleted_at is null")
+
+	if !opts.IncludeExpired {
+		// TODO: can we remove the need to check for both the zero value and nil?
+		now, zero := time.Now(), time.Time{}
+		query.B("AND (expires_at > ? OR expires_at = ? OR expires_at is null)", now, zero)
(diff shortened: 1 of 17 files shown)


internal/server/data/access_key.go
```diff
@@ -113,21 +109,22 @@ func ListAccessKeys(db GormTxn, p *models.Pagination, selectors ...SelectorFunc)
...
-	}
-	result := new(models.AccessKey)
-	if err := db.Model(result).First(result).Error; err != nil {
-		if errors.Is(err, gorm.ErrRecordNotFound) {
-			return nil, internal.ErrNotFound
-		}
-		return nil, err
+// GetAccessKey using the keyID. Note that the keyID is globally unique, so
+// this query is not scoped by an organization_id.
+func GetAccessKey(tx ReadTxn, keyID string) (*models.AccessKey, error) {
```
dnephin, line 167: In production code we only ever queried by `keyID`, which I think is expected. In tests we can also just as easily query by this ID, so I changed this function to only accept this one argument.

In the future we could also make this an options struct, but I think it would be good to keep everything using keyID if possible.  We may also want to consider making the keyID the primary key and removing ID.

internal/server/models/access_key.go
```diff
@@ -20,9 +20,9 @@ type AccessKey struct {
 	OrganizationMember
 	Name string `gorm:"uniqueIndex:idx_access_keys_name,where:deleted_at is NULL"`
 	// IssuedFor is the ID of the user that this access key was created for
-	IssuedFor         uid.ID
-	IssuedForIdentity *Identity `gorm:"foreignKey:IssuedFor" db:"-"`
-	ProviderID        uid.ID
+	IssuedFor     uid.ID
+	IssuedForName string
```
dnephin, line 24: We only use the name for this join, so at least for now I changed this field to only the name of the identity.

When we've finished converting the identity queries over we could go back to the full struct, but I don't see a reason to do that until we make use of more of the fields.

internal/server/pagination.go
```diff
@@ -31,6 +33,6 @@ func PaginationToResponse(p data.Pagination) api.PaginationResponse {
 		Page:       p.Page,
 		Limit:      p.Limit,
 		TotalCount: p.TotalCount,
-		TotalPages: p.TotalPages,
+		TotalPages: int(math.Ceil(float64(p.TotalCount) / float64(p.Limit))),
```
dnephin, line 36: The `TotalPages` field is only used in the API response. We don't really have any use for it internally. So I removed the field from the internal struct, and moved the calculation here. I think the api handler code is the right place for it, because the value is only used in the API.

internal/server/data/access_key.go
```diff
@@ -105,8 +105,61 @@ func UpdateAccessKey(tx WriteTxn, key *models.AccessKey) error {
...
+	Pagination     *Pagination
+}
+
+func ListAccessKeys(tx ReadTxn, opts ListAccessKeyOptions) ([]models.AccessKey, error) {
+	table := &accessKeyTable{}
+	query := Query("SELECT")
+	query.B(columnsForSelect("k", table.Columns()))
+	query.B(", u.name")
+	if opts.Pagination != nil {
+		query.B(", count(*) OVER()")
```
dnephin, line 121: I think this is the biggest difference in the list query. Previously we would repeat the same query a second time, but with the select columns replaced by `Count`. That's a bit harder to do with regular SQL, and I think will make the query much harder to read.

Instead of a second query, I found a couple other options on stack overflow:
1. (the option that is implemented here) is to use a PostgreSQL window function https://www.postgresql.org/docs/current/tutorial-window.html which returns the count of all the rows as part of the regular query. I have no idea of the performance implications of this approach, but using an offset is going to be expensive anyway, so maybe this is not any worse.
2. another option was to use `GET DIAGNOSTICS integer_var = ROW_COUNT;` (https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-STATEMENTS-DIAGNOSTICS), but it's not clear to me exactly what this number represents. I'm not sure if it's always the full count, or if it depends on how the query is executed. We could explore that option further if this one doesn't work out.

internal/server/data/access_key.go
```diff
@@ -100,33 +96,88 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err
...
+	}
+	query.B("FROM access_keys AS k INNER JOIN identities AS u")
+	query.B("ON k.issued_for = u.id")
+	query.B("WHERE k.deleted_at is null AND u.deleted_at is null")
+
+	if !opts.IncludeExpired {
+		// TODO: can we remove the need to check for both the zero value and nil?
```
BruceMacD, line 128: we could remove this by setting the extension deadline to the same time as the expiry when the extension deadline isnt specified

internal/server/data/access_key.go
```diff
@@ -100,33 +96,88 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err
...
+		now, zero := time.Now(), time.Time{}
+		query.B("AND (expires_at > ? OR expires_at = ? OR expires_at is null)", now, zero)
```
BruceMacD, line 130: I dont think we need the zero time or null for the token expiry, we should require this value is not null when we create the table

internal/server/data/access_key.go
```diff
@@ -100,33 +96,88 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err
...
+
+		if err := rows.Scan(fields...); err != nil {
+			return nil, err
+		}
+		result = append(result, key)
+	}
+	return result, rows.Err()
+}
+
+// GetAccessKey using the keyID. Note that the keyID is globally unique, so
```
BruceMacD, line 165: General observation, we will have to be careful with this one

BruceMacD (approved):
My only feedback is on logic that already existed, so it doesnt have to be addressed in this PR specifically. I'll open new issues for it if you'd like

internal/server/data/access_key.go
dnephin: Good idea, I created #3118

PR #3098, opened 2022-09-01
fix(build): fix workflows for release

mxyng:
## Summary

fix container build to build both `infrahq/infra` and `infrahq/ui` components when not explicitly set. this can happen when triggered by a tag

checkout in binary publish job so `gh` is aware of the repo

### DIFF
--- a/.github/workflows/cd-binaries.yaml
+++ b/.github/workflows/cd-binaries.yaml
@@ -55,6 +55,7 @@ jobs:
     needs: [prepare, build]
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
     steps:
+      - uses: actions/checkout@v3
       - uses: actions/download-artifact@v3
         with:
           name: binaries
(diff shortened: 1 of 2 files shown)


.github/workflows/cd-binaries.yaml
```diff
@@ -55,6 +55,7 @@ jobs:
     needs: [prepare, build]
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
     steps:
+      - uses: actions/checkout@v3
```
BruceMacD, line 58: where is this used?

.github/workflows/cd-containers.yaml
```diff
@@ -24,36 +24,29 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
     outputs:
-      component-name: ${{ steps.component-name.outputs.component-name }}
-      container-image-name: ${{ steps.component-name.outputs.container-image-name }}
-      container-image-context: ${{ steps.component-name.outputs.container-image-context }}
+      components: ${{ steps.component-name.outputs.components }}
```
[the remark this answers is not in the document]
mxyng: This is set in this job in the `component-name` step with `echo ::set-output`

.github/workflows/cd-binaries.yaml
mxyng: This is used by `gh` in a subsequent step and was missed in refactoring

BruceMacD (approved):

PR #3100, opened 2022-09-01
maintain: update Helm upgrade docs

mxyng:
## Summary

Point server and connector upgrade to Helm README.md


pdevine (approved):
Seems fine. I think in the future as SaaS is more prevalent we may want to include the upgrade commands for the connector.

PR #3101, opened 2022-09-01
maintain: update docs for self-hosted quickstart

BruceMacD:
## Summary
Update docs for self-hosted initial setup.

## Checklist


jmorganca (approved):

technovangelist (commented):
Replicating the same content in multiple places means it will get out of sync and we have 3 ways to do something. Also, using a heredoc (is that the right name) means you lose the benefits of terseness of a command as well as losing the repeatability of a file.

mxyng:
In this case, it's probably better to refactor the install/upgrade code block as a doc fragment and include it with `{% partial file="partial.md" /%}`. This will reduce a lot of the duplication and make the docs less brittle

README.md
```diff
@@ -34,7 +34,18 @@ Install Infra via `helm`:
 ```
 helm repo add infrahq https://helm.infrahq.com
 helm repo update
-helm install infra infrahq/infra
+cat <<EOF | helm install infra infrahq/infra -f-
```
mxyng, line 37: Please use `upgrade --install` instead of `install` which is more reusable

BruceMacD:
Closed in favor of #3107

PR #3103, opened 2022-09-01
Fix request scoped transactions

### LINKED ISSUE #2697: api: failing to commit a transaction does not result in a 500 response

### LINKED ISSUE #3076: Rollback transactions in middleware does not have an effect

dnephin:
## Summary

Branched from #3028 this PR cleans up the commits from #3032 to fix both #2697 and #3076  

We had two problems with our request scoped transactions:
1. the transaction was never rolled back on error, it was always comitted
2. we never reported a failure to commit a transaction to the user

This PR addresses the problems by:
* moving the transaction to `wrapRoute`, which has access to the error from both the middleware and the handlers
* in `wrapRoute` add a defer which will rollback if commit has not happened
* and respond with a 500 response code if the transaction fails to commit

## Related Issues

Resolves #2697
Resolves #3076

### DIFF
--- a/internal/server/debug.go
+++ b/internal/server/debug.go
@@ -6,14 +6,15 @@ import (
 
 	"github.com/gin-gonic/gin"
 
+	"github.com/infrahq/infra/api"
+
 	"github.com/infrahq/infra/internal/access"
 	"github.com/infrahq/infra/internal/server/models"
 )
 
-func pprofHandler(c *gin.Context) {
+func pprofHandler(c *gin.Context, _ *api.EmptyRequest) (*api.EmptyResponse, error) {
 	if _, err := access.RequireInfraRole(c, models.InfraSupportAdminRole); err != nil {
-		sendAPIError(c, access.HandleAuthErr(err, "debug", "run", models.InfraSupportAdminRole))
-		return
+		return nil, access.HandleAuthErr(err, "debug", "run", models.InfraSupportAdminRole)
 	}
 
 	switch c.Param("profile") {
@@ -25,4 +26,5 @@ func pprofHandler(c *gin.Context) {
 		// All other types of profiles are served from Index
 		http.StripPrefix("/api", http.HandlerFunc(pprof.Index)).ServeHTTP(c.Writer, c.Request)
 	}
+	return nil, nil
 }
--- a/internal/server/debug.go
+++ b/internal/server/debug.go
@@ -7,7 +7,6 @@ import (
 	"github.com/gin-gonic/gin"
 
 	"github.com/infrahq/infra/api"
-
 	"github.com/infrahq/infra/internal/access"
 	"github.com/infrahq/infra/internal/server/models"
 )
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -67,10 +67,15 @@ func handleInfraDestinationHeader(c *gin.Context) error {
 	}
 }
 
-// authenticatedMiddleware is applied to all routes that require authentication.
-// It validates the access key, and updates the lastSeenAt of the user, and
-// possibly also of the destination.
-func authenticatedMiddleware(c *gin.Context, tx *data.Transaction, srv *Server) error {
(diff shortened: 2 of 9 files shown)


internal/server/debug.go
```diff
@@ -6,14 +6,14 @@ import (
 
 	"github.com/gin-gonic/gin"
 
+	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal/access"
 	"github.com/infrahq/infra/internal/server/models"
 )
 
-func pprofHandler(c *gin.Context) {
+func pprofHandler(c *gin.Context, _ *api.EmptyRequest) (*api.EmptyResponse, error) {
```
dnephin, line 14: Now that the database transaction comes from `wrapRoute` this handler had to be updated to fit the interface required by `wrapRoute`.

internal/server/routes_test.go
```diff
@@ -108,30 +108,6 @@ func TestBindsEmptyRequest(t *testing.T) {
 	assert.NilError(t, err)
 }
 
-func TestGetRoute(t *testing.T) {
```
dnephin, line 111: I think this test was probably useful before, but now it does not really do much. We have good test coverage for all of our real production routes, so we don't need to test an artificial route.

internal/server/middleware.go
```diff
@@ -70,40 +70,31 @@ func handleInfraDestinationHeader(c *gin.Context) error {
 // authenticatedMiddleware is applied to all routes that require authentication.
 // It validates the access key, and updates the lastSeenAt of the user, and
 // possibly also of the destination.
-func authenticatedMiddleware(srv *Server) gin.HandlerFunc {
```
BruceMacD, line 73: I think the name "authenticatedMiddleware" is confusing here where this is just an auth check and scope now, it doesnt actually pass the request through any middleware

internal/server/routes.go
```diff
@@ -140,43 +147,97 @@ func (s *Server) GenerateRoutes() Routes {
...
+
+	handler := func(c *gin.Context) {
+		if err := wrapRoute(a, route)(c); err != nil {
+			sendAPIError(c, err)
+		}
+	}
+	bindRoute(a, group.RouterGroup, route.method, route.path, handler)
+}
+
+func wrapRoute[Req, Res any](a *API, route route[Req, Res]) func(*gin.Context) error {
```
BruceMacD, line 186: I would like to see a comment explaining that `wrapRoute` effectively functions as a middleware for requests, there is a lot going on in here

BruceMacD (approved):

internal/server/middleware.go
dnephin: Very true, I'll rename this to remove "middlware", and I should also update the godoc at the same time.

internal/server/routes.go
dnephin: Good call, there is a lot going on here. #3104 is a follow up PR that reduces it a bit.

More godoc here would be better. I'll add that now.

internal/server/middleware.go
dnephin: I pushed a commit that renames these to `authenticateRequest`, and `validateRequestOrganization`. What do you think of those names?

PR #3104, opened 2022-09-01
Make data.Transaction.Rollback a noop if the transaction was already committed

dnephin:
## Summary

Branched from #3103

This PR does some cleanup around transactions. Best viewed by individual commit.

1. make `data.Transaction.Rollback` a noop if the transaction was already committed. This makes it possible to defer the rollback, and simplifies the logic where we use transactions.
2. remove `withDBTxn` since it is no longer necessary with the above change.
3. extracts a few small things from `wrapRoute`. This function has grown a lot, moving a few small things out of it helps keep it focused and easy to read.

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -167,7 +167,7 @@ func (t *Transaction) GormDB() *gorm.DB {
 	return t.DB
 }
 
-// Rollback the transaction. If the transaction was already committed than do
+// Rollback the transaction. If the transaction was already committed then do
 // nothing.
 func (t *Transaction) Rollback() error {
 	if t.committed.Load() {
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"reflect"
 	"strings"
+	"sync/atomic"
 	"time"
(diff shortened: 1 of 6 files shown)


internal/server/routes.go
```diff
@@ -229,12 +224,10 @@ func wrapRoute[Req, Res any](a *API, route route[Req, Res]) func(*gin.Context) e
 		}
 
 		req := new(Req)
-		if err := bind(c, req); err != nil {
+		if err := readRequest(c, req); err != nil {
 			return err
 		}
 
-		trimWhitespace(req)
```
dnephin, line 236: This moved into `readRequest` since they are both related to reading the request into the struct.

internal/server/data/data.go
```diff
@@ -165,16 +167,26 @@ func (t *Transaction) GormDB() *gorm.DB {
 	return t.DB
 }
 
+// Rollback the transaction. If the transaction was already committed than do
```
BruceMacD, line 170:
```suggestion
// Rollback the transaction. If the transaction was already committed then do
```

BruceMacD (approved):
elegant solution 👍

PR #3106, opened 2022-09-01
Forgot Domains

pdevine:
## Summary

Provide a UI screen + API endpoint to allow users to recover any domain/org sign-ins that they have previously created. 

It's possible that users may forget how to sign-in to a given org given that we don't support 'at-large' users. In order for a user to sign in, they *must* know the correct domain name. This change allows the user to specify an email address which we will send an email which contains each of the URLs which they can use to sign-in.

## Checklist

### DIFF
--- a/internal/server/email/sendgrid.go
+++ b/internal/server/email/sendgrid.go
@@ -19,6 +19,7 @@ const (
 	EmailTemplateSignup EmailTemplate = iota
 	EmailTemplatePasswordReset
 	EmailTemplateUserInvite
+	EmailTemplateForgottenDomains
 )
 
 type TemplateDetail struct {
@@ -39,6 +40,10 @@ var emailTemplates = map[EmailTemplate]TemplateDetail{
 		TemplateName: "user-invite",
 		Subject:      "{{.FromUserName}} has invited you to Infra",
 	},
+	EmailTemplateForgottenDomains: {
+		TemplateName: "forgot-domain",
+		Subject:      "Your sign-in links",
+	},
 }
 
 var (
@@ -58,21 +63,6 @@ func IsConfigured() bool {
 }
 
 func SendTemplate(name, address string, template EmailTemplate, data any) error {
-	if TestMode {
-		logging.Debugf("sent email to %q: %+v\n", address, data)
-		TestDataSent = append(TestDataSent, data)
-		return nil // quietly return
-	}
-
-	if len(SendgridAPIKey) == 0 {
-		return ErrNotConfigured
-	}
-
-	if name == "" {
-		// until we have real user names
-		name = BuildNameFromEmail(address)
-	}
-
 	details, ok := emailTemplates[template]
 	if !ok {
 		return ErrUnknownTemplate
@@ -110,6 +100,23 @@ func SendTemplate(name, address string, template EmailTemplate, data any) error 
 		return err
 	}
 	msg.HTMLBody = w.Bytes()
+
+	if TestMode {
+		logging.Debugf("sent email to %q: %+v\n", address, data)
+		logging.Debugf("plain: %s", string(msg.PlainBody))
+		logging.Debugf("html: %s", string(msg.HTMLBody))
+		TestDataSent = append(TestDataSent, data)
+		return nil // quietly return
+	}
+
+	if len(SendgridAPIKey) == 0 {
+		return ErrNotConfigured
+	}
+
+	if name == "" {
+		// until we have real user names
+		name = BuildNameFromEmail(address)
+	}
 
 	// TODO: handle rate limiting, retries, understanding which errors are retryable, send queues, whatever
 	if err := SendSMTP(msg); err != nil {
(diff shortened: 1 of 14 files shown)


ui/pages/forgot-domain/index.js
```diff
@@ -0,0 +1,100 @@
...
+                  className={`w-full border-b border-gray-800 bg-transparent px-px py-2 text-2xs placeholder:italic focus:border-b focus:border-gray-200 focus:outline-none ${
+                    error ? 'border-pink-500/60' : ''
+                  }`}
+                />
+              </div>
+              <button
+                disabled={!email}
+                className='mt-6 mb-2 rounded-lg border border-violet-300 px-4 py-3 text-2xs text-violet-100 hover:border-violet-100 disabled:pointer-events-none disabled:opacity-30'
+              >
+                Submit
```
BruceMacD, line 85: We should disable this submit button after the user submits a request also

internal/server/models/forgotdomain.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+import (
+	"time"
+)
+
+type ForgottenDomain struct {
+	OrganizationName   string
+	OrganizationDomain string
+	LastSeenAt         time.Time
```
BruceMacD, line 10: Seems like `LastSeenAt` isn't used anywhere, let's leave it out for now

internal/server/models/forgotdomain.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+import (
+	"time"
+)
+
+type ForgottenDomain struct {
+	OrganizationName   string
```
BruceMacD, line 8: This doesnt seem to be used

ui/pages/forgot-domain/index.js
mxyng: Password reset has the same problem: #3114

internal/server/models/forgotdomain.go
pdevine: I ended up using this in the latest change as I think it's somewhat important to give you an idea about when you last signed-in.

internal/server/models/forgotdomain.go
pdevine: Included this as well.

ui/pages/forgot-domain/index.js
```diff
@@ -0,0 +1,100 @@
...
+          email,
+        }),
+      })
+
+      if (!res.ok) {
+        throw await res.json()
+      }
+
+      await res.json()
+      setSubmitted(true)
```
BruceMacD, line 26: You'll want to move this `setSubmitted` above the API call or it means that the UI must wait for a response from the API before it disables the submit button. This could be a decent amount of time.

ui/pages/forgot-domain/index.js
```diff
@@ -0,0 +1,100 @@
...
+    try {
+      const res = await fetch('/api/forgot-domain-request', {
+        method: 'post',
+        body: JSON.stringify({
+          email,
+        }),
+      })
+
+      if (!res.ok) {
+        throw await res.json()
```
BruceMacD, line 23: I think throwing this here may cause the UI to crash. Better to handle it

BruceMacD (approved):
Minor UI comment, otherwise looks good to go

internal/server/email/sendgrid.go
```diff
@@ -111,6 +106,18 @@ func SendTemplate(name, address string, template EmailTemplate, data any) error
...
+		logging.Debugf("sent email to %q: %+v\n", address, data)
+		logging.Debugf("plain: %s", string(msg.PlainBody))
+		logging.Debugf("html: %s", string(msg.HTMLBody))
+		TestDataSent = append(TestDataSent, data)
+		return nil // quietly return
+	}
+
+	if len(SendgridAPIKey) == 0 {
+		return ErrNotConfigured
+	}
```
ssoroka, line 119: seems like this is too far down in the function now

ssoroka (approved):
looks good. some minor comments

internal/server/email/templates/forgot-domain.text.html
```diff
@@ -0,0 +1,5 @@
+
+<p>Someone has requested links to each of the organizations associated with your Infra account. If this was not you, you can safely ignore this email.</p>
+
+{{range $v := .Domains}}<p>{{$v.OrganizationName}} <a href="https://{{$v.OrganizationDomain}}/login">https://{{$v.OrganizationDomain}}/login</a> (last seen {{$v.LastSeenAt}})</p>
```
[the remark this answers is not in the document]
pdevine: Really? What would we use instead? I don't want to lump everything (including the HTML) into one big variable.

internal/server/email/sendgrid.go
pdevine: I did this so that the email still gets constructed (i.e. you can tell what _would_ have been sent), otherwise it short circuits too quickly (and makes it harder to test).

ui/pages/forgot-domain/index.js
pdevine: I patterned this off of how it's working in `signup/index.js`, `password-reset/index.js`, and `login/index.js` so I would assume it's broken in those places too? It looks like we're also doing it in `login/finish.js`. Is there a pattern I can follow to do it correctly?

PR #3113, opened 2022-09-02
fix: email domain and missing rcpt to

ssoroka:


internal/server/email/sendgrid.go
```diff
@@ -65,6 +68,11 @@ func SendTemplate(name, address string, template EmailTemplate, data any) error
 		return ErrNotConfigured
 	}
 
+	if name == "" {
+		// until we have real user names
+		name = BuildNameFromEmail(address)
```
mxyng, line 73: I don't think we need to over rotate on this. Instead, just set the name to the email until we have actual user names

ssoroka: the experience is pretty nice vs email addresses. I created this function for another spot that uses the name in the mail body and looked a lot better than "joe.smith@example.com invited you to ...."

internal/server/handlers.go
```diff
@@ -145,7 +145,7 @@ func (a *API) Signup(c *gin.Context, r *api.SignupRequest) (*api.SignupResponse,
 	a.t.Event("signup", identity.ID.String(), Properties{})
 
 	err = email.SendSignupEmail("", r.Name, email.SignupData{
-		Link: fmt.Sprintf("https://%s/login", r.Org.Subdomain),
+		Link: fmt.Sprintf("https://%s/login", suDetails.Org.Domain),
```
mxyng, line 148: If I'm reading this correctly, `r.Org.Subdomain` only contains the subdomain part but the details contains a FQDN?

ssoroka: yes

mxyng (approved):

internal/server/handlers.go
mxyng: Can be in a follow up PR but should add a test case for this

PR #3122, opened 2022-09-03
improve: replace quickstart plaintext password with secret - combine quickstart and install

jmorganca:


docs/getting-started/install.md
```diff
@@ -1,39 +1,48 @@
 ---
-title: Quickstart
+title: Install Infra
```
BruceMacD, line 2: Should we make this title more specific, maybe `Deploying Infra`? I think a very common scenario when opening the docs is a user looking to install the CLI. This makes it look like an end-user has to install an entire Infra server based on the "install infra" title.

BruceMacD (approved):
Minor title feedback, but looks good

ssoroka (approved):

PR #3139, opened 2022-09-06
fix: password reset sending multiple requests

### LINKED ISSUE #3114: Password reset button can be clicked multiple times producing multiple emails
## Describe the bug

The password reset button takes a non-zero amount of time to disappear which gives the user ample opportunity to click it more. This results in multiple emails being sent to the recipient. 

image

mxyng:
## Summary

Set `submitted` state as soon as the submit button is pressed rather than after the request returns.

Also some minor UI issues like
* punctuation
* divider on password reset
* margin between forgot password and login button

Resolves #3114

### DIFF
--- a/ui/pages/password-reset/index.js
+++ b/ui/pages/password-reset/index.js
@@ -9,7 +9,6 @@ export default function PasswordReset() {
   const { token } = router.query
 
   const [email, setEmail] = useState('')
-  const [error, setError] = useState('')
   const [submitted, setSubmitted] = useState(false)
 
   async function onSubmit(e) {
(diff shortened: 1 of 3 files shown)


ui/pages/password-reset/index.js
```diff
@@ -99,7 +91,10 @@ export default function PasswordReset() {
                   />
                 </div>
                 <button
-                  disabled={!email}
+                  disabled={submitted}
+                  onClick={() => {
+                    setSubmitted(true)
```
jmorganca, line 96: Thanks for fixing!

This should be set `onSubmit` handler above (vs writing this new handler)

ui/pages/password-reset/index.js
```diff
@@ -13,6 +13,7 @@ export default function PasswordReset() {
   const [submitted, setSubmitted] = useState(false)
 
   async function onSubmit(e) {
+    setSubmitted(true)
```
BruceMacD, line 16: You may want to set this back to false on error so the user can try again.

mxyng: Good call out though but the UI doesn't currently branch on errors. The error state is never set to a non-empty value so I've removed it

BruceMacD (approved):

PR #3140, opened 2022-09-06
maintain: use the same encrypted at rest type for both strings and bytes

dnephin:
## Summary

We had two copies of this type that were almost identical. One for []byte and one for string. Since `string` and `[]byte` are compatible, we can use the same type for both `bytea` and `text` database columns.

I added a test to show this type working with a `bytea` type column. The test was written first, so it also showed that a value serialzied using `EncryptedAtRestBytes` could be later read using `EncryptedAtRest`. This works because the `Value()` method uses the same function to encode the data going into the database.


dnephin:
That's correct. I didn't expect it to because it uses the same function to encode the data that goes into the database, and the test confirmed that.

PR #3141, opened 2022-09-06
Move CreateProviderUser into credential operations

dnephin:
## Summary

While working on another change I noticed this `access.CreateProviderUser` function. Unlike most of the other `access` functions this is not a function that performs authorization for an API operation. It's a helper function for creating the necessary join table when an admin performs an operation that creates an infra user (either by creating a new user, or giving an IDP user a password).

By moving this create-if-not-exists into the `CreateCredential` and `UpdateCredential` functions we remove the need for this `access` function, and we make the `{Create,Update}Credential` functions safer (because other callers will no longer need to perform this operation).

`data.CreateProviderUser` already checks for an existing user, so it's safe to call in any case. We should probably rename that function as it behaves differently from other `Create` functions in the `data` package, but i've left that rename for a different time. I'm also not sure what the new name should be.


internal/access/credential.go
```diff
@@ -56,7 +62,14 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
...
 	}
 
-	return updateCredential(c, user, newPassword, isSelf)
+	if err := updateCredential(c, user, newPassword, isSelf); err != nil {
+		return err
+	}
+
+	// if the request is from an admin, the infra user may not exist yet, so create the
+	// provider_user if it's missing.
+	_, _ = data.CreateProviderUser(rCtx.DBTxn, data.InfraProvider(rCtx.DBTxn), user)
```
dnephin, line 71: We ignored the error here previously. It's probably ok, but I wonder if we should handle it for consistency. The `data.CreateProviderUser` function already returns no error if the entry already exists.

BruceMacD (approved):
thanks for cleaning this up

PR #3144, opened 2022-09-07
maintain: remove two access helper functions

dnephin:
## Summary

Branched from #3141, this PR makes a similar change.
These functions `InfraProvider`, and `InfraConnectorIdentity` are not providing access. They exist because of the rules we have established for how the data package can be accessed. We can refactor the code slightly to remove the need for these functions by calling the functions in the `data` package directly from `access`.

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -32,7 +32,7 @@ func (a *API) CreateToken(c *gin.Context, r *api.EmptyRequest) (*api.CreateToken
 	rCtx := getRequestContext(c)
 
 	if rCtx.Authenticated.User != nil {
-		err := a.UpdateIdentityInfoFromProvider(c)
+		err := a.UpdateIdentityInfoFromProvider(rCtx)
 		if err != nil {
 			// this will fail if the user was removed from the IDP, which means they no longer are a valid user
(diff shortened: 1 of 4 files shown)


internal/server/handlers.go
```diff
@@ -85,7 +85,6 @@ func (a *API) CreateAccessKey(c *gin.Context, r *api.CreateAccessKeyRequest) (*a
 	accessKey := &models.AccessKey{
 		IssuedFor:         r.UserID,
 		Name:              r.Name,
-		ProviderID:        access.InfraProvider(c).ID,
```
BruceMacD, line 88: We still need this or running `infra tokens add` will fail when used with directly created access keys

dnephin: The behaviour is still the same, it's being set in `access.CreateAccessKey` which is called right after this.  Sorry I didn't call this out in the description.

`data.CreateAccessKey` validates that `ProviderID` is not 0, and we have tests for the `CreateAccessKey` endpoint which should cover this behaviour.

BruceMacD (approved):
Thanks for explaining the new behavior, looks good in that case

ssoroka (approved):

PR #3149, opened 2022-09-07
Use SQL for CreateGrant

### LINKED ISSUE #2415: Isolate all references to gorm in the data package

dnephin:
## Summary

Primarily this PR converts `data.CreateGrant` from using gorm to using sql. It includes a few other small changes to reduce conflicts, so best viewed by individual commit.

This PR also:
* moves the query builder into its own package. This was cherry-picked out of #3060 to prevent conflicts. As part of that change the constructor and type name for the query builder have changed.
* adds a transaction around the `NewDB` initialize operations. Not strictly necessary, but a nice to have to ensure that any problems setting up the db are rolled back. 
* adds an exclude rule for `tx.Rollback` so that we can ignore the rollback error. I think generally that error is not going to be relevant.
* adds `handleErr` in `insert` and `update` helpers, which I missed when I added those functions.

Related to #2415

### DIFF
--- a/internal/server/data/grant.go
+++ b/internal/server/data/grant.go
@@ -10,6 +10,24 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
+type grantTable models.Grant
+
+func (g grantTable) Table() string {
+	return "grants"
+}
+
+func (g grantTable) Columns() []string {
+	return []string{"created_at", "created_by", "deleted_at", "id", "organization_id", "privilege", "resource", "subject", "updated_at"}
+}
+
+func (g grantTable) Values() []any {
+	return []any{g.CreatedAt, g.CreatedBy, g.DeletedAt, g.ID, g.OrganizationID, g.Privilege, g.Resource, g.Subject, g.UpdatedAt}
+}
+
+func (g *grantTable) ScanFields() []any {
+	return []any{&g.CreatedAt, &g.CreatedBy, &g.DeletedAt, &g.ID, &g.OrganizationID, &g.Privilege, &g.Resource, &g.Subject, &g.UpdatedAt}
+}
+
 func CreateGrant(db GormTxn, grant *models.Grant) error {
 	switch {
 	case grant.Subject == "":
--- a/internal/server/data/grant.go
+++ b/internal/server/data/grant.go
@@ -1,8 +1,11 @@
 package data
 
 import (
+	"errors"
 	"fmt"
 
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgerrcode"
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal/logging"
@@ -38,7 +41,25 @@ func CreateGrant(tx WriteTxn, grant *models.Grant) error {
 		return fmt.Errorf("resource is required")
 	}
 
-	return insert(tx, (*grantsTable)(grant))
+	// Use a savepoint so that we can query for the duplicate grant on conflict
+	if _, err := tx.Exec("SAVEPOINT beforeCreate"); err != nil {
+		// ignore "not in a transaction" error, because outside of a transaction
+		// the db conn can continue to be used after the conflict error.
+		if !isPgErrorCode(err, pgerrcode.NoActiveSQLTransaction) {
+			return err
+		}
+	}
+	if err := insert(tx, (*grantsTable)(grant)); err != nil {
+		_, _ = tx.Exec("ROLLBACK TO SAVEPOINT beforeCreate")
+		return handleError(err)
+	}
+	_, _ = tx.Exec("RELEASE SAVEPOINT beforeCreate")
+	return nil
+}
+
+func isPgErrorCode(err error, code string) bool {
+	pgError := &pgconn.PgError{}
+	return errors.As(err, &pgError) && pgError.Code == code
 }
 
 func GetGrant(db GormTxn, selectors ...SelectorFunc) (*models.Grant, error) {
--- a/internal/server/data/grant.go
+++ b/internal/server/data/grant.go
@@ -10,25 +10,25 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-type grantTable models.Grant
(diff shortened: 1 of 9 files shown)


internal/server/data/grant.go
```diff
@@ -19,7 +40,26 @@ func CreateGrant(db GormTxn, grant *models.Grant) error {
...
 	}
-	return add(db, grant)
+
+	// Use a savepoint so that we can query for the duplicate grant on conflict
+	if _, err := tx.Exec("SAVEPOINT beforeCreate"); err != nil {
+		// ignore "not in a transaction" error, because outside of a transaction
+		// the db conn can continue to be used after the conflict error.
+		if !isPgErrorCode(err, pgerrcode.NoActiveSQLTransaction) {
+			return err
+		}
```
dnephin, line 50: Previously we were ignoring all errors. We have to ignore this one error code because we have plenty of tests that don't use a transaction to setup the fixture data. In the future as we add more transactions to tests maybe we'll be able to remove this.

internal/server/data/grant.go
```diff
@@ -19,7 +40,26 @@ func CreateGrant(db GormTxn, grant *models.Grant) error {
 	case grant.Resource == "":
 		return fmt.Errorf("resource is required")
 	}
-	return add(db, grant)
+
+	// Use a savepoint so that we can query for the duplicate grant on conflict
+	if _, err := tx.Exec("SAVEPOINT beforeCreate"); err != nil {
```
dnephin, line 45: For now I'd like to only do this savepoint/rollback in the specific create functions that require it. We had added it to the shared `add` function, but we only ever saw a problem with grants. I think grants are unique here because they are immutable and have a composite unique constraint on all the interesting fields. I don't think any of our other tables have that property.

In the future we can easily move this savepoint into insert if we find a reason to.

I tried using `ON CONFLICT DO NOTHING`, but unfortunately that doesn't give any indication if the query was on conflict or if it inserted (which we need for the response code).  Maybe there's a way to do this by looking at rows modified, and then doing the select. I kept our existing approach for now to keep this change small.

internal/server/data/grant.go
```diff
@@ -19,7 +40,26 @@ func CreateGrant(db GormTxn, grant *models.Grant) error {
...
+		// the db conn can continue to be used after the conflict error.
+		if !isPgErrorCode(err, pgerrcode.NoActiveSQLTransaction) {
+			return err
+		}
+	}
+	if err := insert(tx, (*grantsTable)(grant)); err != nil {
+		_, _ = tx.Exec("ROLLBACK TO SAVEPOINT beforeCreate")
+		return handleError(err)
+	}
+	_, _ = tx.Exec("RELEASE SAVEPOINT beforeCreate")
```
dnephin, line 56: We weren't doing this previously. It's not strictly required, but I think it's a good idea so that we don't end up clobbering earlier savepoints with the same name.

I've also ignored the error for both the release and the rollback, since both of those operations will fail if we did not start the savepoint, and I think the error from the insert statement is going to be more relevant. Neither of these operations are required for functional correctness.

BruceMacD (approved):
looks good

internal/server/data/grant.go
```diff
@@ -1,16 +1,37 @@
...
+	return []string{"created_at", "created_by", "deleted_at", "id", "organization_id", "privilege", "resource", "subject", "updated_at"}
+}
+
+func (g grantsTable) Values() []any {
+	return []any{g.CreatedAt, g.CreatedBy, g.DeletedAt, g.ID, g.OrganizationID, g.Privilege, g.Resource, g.Subject, g.UpdatedAt}
+}
+
+func (g *grantsTable) ScanFields() []any {
+	return []any{&g.CreatedAt, &g.CreatedBy, &g.DeletedAt, &g.ID, &g.OrganizationID, &g.Privilege, &g.Resource, &g.Subject, &g.UpdatedAt}
+}
```
ssoroka, line 32: seems like we could find a better way to do this. Is there reason why it's a composite currently?

internal/server/data/grant.go
ssoroka: do you remove the savepoint from add somewhere?

ssoroka (approved):

internal/server/data/grant.go
dnephin: Nope, it's still in `add`, but `add` is no longer used with the SQL queries, so it'll be removed at the end.

internal/server/data/grant.go
dnephin: Sorry, I don't follow. What do you mean by composite?

dnephin: We talked about this a bit offline. We could move these methods to the model struct, but then everything would have access to them. By creating a new type using the same data we can hide these methods and only expose them to the data package.

PR #3151, opened 2022-09-07
fix: list access keys query

dnephin:
## Summary

Set OrganizationID and order by in the query. Update the tests to catch the bug.

### DIFF
--- a/internal/server/data/access_key_test.go
+++ b/internal/server/data/access_key_test.go
@@ -291,7 +291,7 @@ func TestListAccessKeys(t *testing.T) {
 		createIdentities(t, db, user, otherUser)
 
 		first := &models.AccessKey{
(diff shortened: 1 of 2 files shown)


internal/server/data/access_key_test.go
```diff
@@ -291,23 +291,23 @@ func TestListAccessKeys(t *testing.T) {
 		createIdentities(t, db, user, otherUser)
 
 		first := &models.AccessKey{
-			Name:       "first",
+			Name:       "alpha",
```
dnephin, line 294: I changed these names so that the existing tests sort correctly. I changed the insertion order below so that the tests would fail without this change.

mxyng (approved):
LGTM

PR #3152, opened 2022-09-07
fix: extend list of reserved domains in signup

ssoroka:
## Summary

- extend list of reserved domains
- don't allow domains of 5 char or less (would have to be created manually by us if we want it)
- add test
- one change we might want to make after this (or in this pr) is to have different error messages for the reserved list vs the reject list. the reserve list should say to sign up with a different domain and/or contact us or something along those lines.

## Checklist

### DIFF
--- a/api/restricted_subdomains.yaml
+++ b/api/restricted_subdomains.yaml
@@ -9,12 +9,16 @@ reject: # these are always rejected
     - admin
     - api
     - app
+    - beta
     - blog
     - cache
+    - chat
     - connector
+    - data
     - docs
     - dns
     - db
+    - demo
     - dev
     - devel
     - download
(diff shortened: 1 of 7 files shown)


BruceMacD (approved):
looks good once lint/tests pass

dnephin (commented):
The list of words looks great! @pdevine had originally requested a minimum of 3, but personally I think it should be at least 4 or 5. This PR changes it to 6. I'm fine with that, it's easier to relax this later than to make it more strict.

What's the reason for storing this as yaml? Do you think we could store it as a go literal to remove the need for decoding?

Or even a simple "word per line" that would let us parse it with https://pkg.go.dev/bufio#NewScanner ?

ssoroka:
Yaml was just easy and the source list was already yaml. Alternatives are okay. We already use the yaml parsing library so it's no new dependencies. The comments are nice

dnephin (approved):
Maybe worth double checking with others about the min length change, but LGTM

PR #3153, opened 2022-09-08
helm chart: fix configmap generation with config.tls not empty

cr1cr1:
## Summary

- Providing explicit `config.tls` settings will not be rendered in the `infra-server` configmap because "tls" is ommited: https://github.com/infrahq/infra/blob/d87cc531bfa8b3ff3e9bd09a4be54a432a7dd35c/helm/charts/infra/templates/server/configmap.yaml#L10
- I think "ui" should be skipped from generic rendering, as it is already handled some lines below
- To avoid duplicate sections and potentially invalid yaml, added a test for non empty value for any key in `config`

### DIFF
--- a/helm/charts/infra/Chart.yaml
+++ b/helm/charts/infra/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: infra
 description: A Helm chart for Infra
 type: application
-version: 0.20.1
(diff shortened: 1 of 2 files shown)


mxyng (approved):
LGTM! Thanks for finding and fixing this!

PR #3155, opened 2022-09-08
Use SQL for GetGrant and DeleteGrants

dnephin:
## Summary

Converts `GetGrant` and `DeleteGrants` to use sql.

Also changes `columnsForSelect` to accept only the table type. This makes linting easier (no string arg to check), and will make our queries more consistent by always using the table name instead of an alias to qualify field names.

Adds a bunch of tests for both functions.

## Related Issues

Related to #2415

### DIFF
--- a/internal/server/data/grant.go
+++ b/internal/server/data/grant.go
@@ -3,12 +3,14 @@ package data
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgerrcode"
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/internal/server/data/querybuilder"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
 )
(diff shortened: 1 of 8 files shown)


internal/server/data/grant.go
```diff
@@ -62,26 +64,92 @@ func isPgErrorCode(err error, code string) bool {
...
+
+	switch {
+	case opts.ByID != 0:
+		query.B("AND id = ?", opts.ByID)
+	case opts.BySubject != "":
+		query.B("AND subject = ?", opts.BySubject)
+		query.B("AND privilege = ?", opts.ByPrivilege)
+		query.B("AND resource = ?", opts.ByResource)
+	default:
+		return nil, fmt.Errorf("GetGrant requires an ID or subject")
```
dnephin, line 99: @ssoroka good catch on this missing default case! 

I added the error here, and a test to cover this case.  I also went back and reviewed the changes in #3029 and #3094, and I believe all of those queries are safe.

`Insert`, and `Update` queries don't accept selectors (or option query structs), and we validate the input has required fields set.

`Get` and `Delete` queries should always require some ID, and will error if none are provided.

`List` queries seem to generally only use optional selectors. We may want to limit exactly which query parameters can be used with others, but we don't do that much yet.

BruceMacD (approved):

PR #3157, opened 2022-09-08
Use a transaction to run data migrations

dnephin:
## Summary

Also run each migration twice in the migration test to check for idempotence. Includes a couple fixes to migrations to make them idempotent.

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -37,6 +37,10 @@ func NewDB(connection gorm.Dialector, dbOpts NewDBOptions) (*DB, error) {
 		return nil, fmt.Errorf("db conn: %w", err)
 	}
 	dataDB := &DB{DB: db}
+	tx, err := dataDB.Begin(context.TODO())
+	if err != nil {
+		return nil, err
+	}
 
 	opts := migrator.Options{
 		InitSchema: initializeSchema,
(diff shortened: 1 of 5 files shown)


internal/server/data/data.go
```diff
@@ -32,22 +32,31 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db GormTxn) error) (*DB, er
...
 	dataDB := &DB{DB: db}
+	tx, err := dataDB.Begin(context.TODO())
+	if err != nil {
+		return nil, err
+	}
 
 	opts := migrator.Options{
 		InitSchema: initializeSchema,
-		LoadKey: func(tx migrator.DB) error {
+		LoadKey: func(_ migrator.DB) error {
```
BruceMacD, line 42: Why not take this as a `tx` arg rather than using the transaction from the outside scope? I believe we control this migrator library so we could change what migrator.Options LoadKey expects.

internal/server/data/data.go
```diff
@@ -32,22 +32,31 @@ func NewDB(connection gorm.Dialector, loadDBKey func(db GormTxn) error) (*DB, er
...
 			// TODO: use the passed in tx instead of dataDB once the queries
 			// used by loadDBKey are ported to sql
-			return loadDBKey(dataDB)
+			return loadDBKey(tx)
 		},
 	}
-	m := migrator.New(dataDB, opts, migrations())
+	m := migrator.New(tx, opts, migrations())
 	if err := m.Migrate(); err != nil {
+		_ = tx.Rollback()
```
BruceMacD, line 53: couldn't hurt to trace log this nested error if it exists

internal/server/data/migrations_test.go
```diff
@@ -79,6 +80,10 @@ func TestMigrations(t *testing.T) {
 		err := m.Migrate()
 		assert.NilError(t, err)
 
+		t.Run("run again to check idempotency", func(t *testing.T) {
```
BruceMacD, line 83: good idea 👍

internal/server/data/data.go
dnephin: Good question! The problem is that gorm has already been removed from the migrator package. It now accepts only this `DB` interface, which matches the stdlib `database/sql.DB`. We can't use the `data.GormTxn` interface in the migrator package because that would cause an import cycle, and I don't really want to move the `GormTxn` into the migrator pacakge.

I think to fix this all we need is to convert the two queries that are used by `loadDBKey` to the stdlib interface. I'll do that today, and that should allow us to resolve this TODO and use the passed in transaction.

dnephin: I've got a fix for that here: https://github.com/infrahq/infra/pull/3168/commits/b2c4a68e56de236a609e19a391876fc923439692

BruceMacD: Thanks, I'll re-review this PR when #3168 gets in

dnephin (commented):
Ok, the other PRs have merged, so the TODO has been resolved.

This should be ready for a review

BruceMacD (approved):
looks good to me

PR #3162, opened 2022-09-08
Use SQL for list grants

dnephin:
## Summary

Branched from #3155

This PR converts list grants to SQL, and adds a bunch of tests.

It also changes `access.RequireInfraRole` to use `IncludeInheritedFromGroups`, instead of iterating over all the groups a user belongs to.

## Related Issues

Related to #2415

### DIFF
--- a/api/grant.go
+++ b/api/grant.go
@@ -47,6 +47,7 @@ func (r ListGrantsRequest) ValidationRules() []validate.ValidationRule {
 		validate.MutuallyExclusive(
 			validate.Field{Name: "user", Value: r.User},
 			validate.Field{Name: "group", Value: r.Group},
+			// TODO: validate showInherited only allowed with User
 		),
 	}
 }
--- a/internal/access/access.go
+++ b/internal/access/access.go
@@ -43,7 +43,6 @@ func RequireInfraRole(c *gin.Context, oneOfRoles ...string) (data.GormTxn, error
 	}
 
 	// TODO: use privilege in (...) instead of iterating over roles
-	// TODO: check for test coverage that uses permission from a group
 	for _, role := range oneOfRoles {
 		ok, err := Can(db, identity.PolyID(), role, ResourceInfraAPI)
 		if err != nil {
--- a/internal/access/access.go
+++ b/internal/access/access.go
@@ -42,6 +42,8 @@ func RequireInfraRole(c *gin.Context, oneOfRoles ...string) (data.GormTxn, error
 		return nil, fmt.Errorf("no active identity")
 	}
 
+	// TODO: use privilege in (...) instead of iterating over roles
+	// TODO: check for test coverage that uses permission from a group
 	for _, role := range oneOfRoles {
 		ok, err := Can(db, identity.PolyID(), role, ResourceInfraAPI)
 		if err != nil {
@@ -52,26 +54,6 @@ func RequireInfraRole(c *gin.Context, oneOfRoles ...string) (data.GormTxn, error
 			return db, nil
 		}
 	}
-
-	// check if they belong to a group that is authorized
-	groups, err := data.ListGroups(db, nil, data.ByGroupMember(identity.ID))
-	if err != nil {
-		return nil, fmt.Errorf("auth user groups: %w", err)
-	}
-
-	for _, group := range groups {
-		for _, role := range oneOfRoles {
-			ok, err := Can(db, group.PolyID(), role, ResourceInfraAPI)
-			if err != nil {
-				return nil, err
-			}
-
-			if ok {
-				return db, nil
-			}
-		}
-	}
-
 	return nil, ErrNotAuthorized
 }
 
@@ -123,7 +105,13 @@ func HandleAuthErr(err error, resource, operation string, roles ...string) error
 
 // Can checks if an identity has a privilege that means it can perform an action on a resource
 func Can(db data.GormTxn, identity uid.PolymorphicID, privilege, resource string) (bool, error) {
-	grants, err := data.ListGrants(db, &data.Pagination{Limit: 1}, data.BySubject(identity), data.ByPrivilege(privilege), data.ByResource(resource))
+	grants, err := data.ListGrants(db, data.ListGrantsOptions{
+		Pagination:                 &data.Pagination{Limit: 1},
+		BySubject:                  identity,
+		ByPrivilege:                privilege,
+		ByResource:                 resource,
+		IncludeInheritedFromGroups: true,
+	})
 	if err != nil {
 		return false, fmt.Errorf("has grants: %w", err)
 	}
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -20,12 +20,8 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired 
 	opts := data.ListAccessKeyOptions{
 		Pagination:     p,
 		IncludeExpired: showExpired,
-	}
-	if identityID != 0 {
-		opts.ByIssuedForID = identityID
-	}
-	if name != "" {
-		opts.ByName = name
+		ByIssuedForID:  identityID,
+		ByName:         name,
 	}
 	return data.ListAccessKeys(db, opts)
 }
--- a/internal/access/access_test.go
+++ b/internal/access/access_test.go
@@ -86,11 +86,11 @@ func TestBasicGrant(t *testing.T) {
 	cant(t, db, "i:bob", "read", "infra.groups.1") // currently we check for exact grant match, this may change as grants evolve
 	cant(t, db, "i:bob", "write", "infra.groups")
 
-	grant(t, db, tom, "i:alice", "read", "infra.machines")
-	can(t, db, "i:alice", "read", "infra.machines")
-	cant(t, db, "i:alice", "read", "infra")
-	cant(t, db, "i:alice", "read", "infra.machines.1")
-	cant(t, db, "i:alice", "write", "infra.machines")
+	grant(t, db, tom, "i:a11ce", "read", "infra.machines")
+	can(t, db, "i:a11ce", "read", "infra.machines")
+	cant(t, db, "i:a11ce", "read", "infra")
+	cant(t, db, "i:a11ce", "read", "infra.machines.1")
+	cant(t, db, "i:a11ce", "write", "infra.machines")
 }
 
 func TestUsersGroupGrant(t *testing.T) {
@@ -192,6 +192,7 @@ func grant(t *testing.T, db data.GormTxn, currentUser *models.Identity, subject 
 }
 
 func can(t *testing.T, db *data.DB, subject uid.PolymorphicID, privilege, resource string) {
+	t.Helper()
 	canAccess, err := Can(db, subject, privilege, resource)
 	assert.NilError(t, err)
 	assert.Assert(t, canAccess)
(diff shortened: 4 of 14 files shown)


api/grant.go
```diff
@@ -47,6 +47,7 @@ func (r ListGrantsRequest) ValidationRules() []validate.ValidationRule {
 		validate.MutuallyExclusive(
 			validate.Field{Name: "user", Value: r.User},
 			validate.Field{Name: "group", Value: r.Group},
+			// TODO: validate showInherited only allowed with User
```
dnephin, line 50: I'll address this TODO in a follow up PR, since this one is already a good size.

internal/access/access.go
```diff
@@ -42,6 +42,8 @@ func RequireInfraRole(c *gin.Context, oneOfRoles ...string) (data.GormTxn, error
 		return nil, fmt.Errorf("no active identity")
 	}
 
+	// TODO: use privilege in (...) instead of iterating over roles
+	// TODO: check for test coverage that uses permission from a group
```
dnephin, line 46: The first TODO I think is for a separate PR. The second one might be good to do in this one. I think we do have some test coverage from API tests, but I did not find anything for this function itself.

internal/access/access.go
```diff
@@ -52,26 +54,6 @@ func RequireInfraRole(c *gin.Context, oneOfRoles ...string) (data.GormTxn, error
 			return db, nil
 		}
 	}
-
-	// check if they belong to a group that is authorized
-	groups, err := data.ListGroups(db, nil, data.ByGroupMember(identity.ID))
```
dnephin, line 57: I was able to remove this by using `IncludeInheritedFromGroups: true` in `Can` (below this).

internal/access/access_test.go
```diff
@@ -86,11 +86,11 @@ func TestBasicGrant(t *testing.T) {
 	cant(t, db, "i:bob", "read", "infra.groups.1") // currently we check for exact grant match, this may change as grants evolve
 	cant(t, db, "i:bob", "write", "infra.groups")
 
-	grant(t, db, tom, "i:alice", "read", "infra.machines")
-	can(t, db, "i:alice", "read", "infra.machines")
-	cant(t, db, "i:alice", "read", "infra")
-	cant(t, db, "i:alice", "read", "infra.machines.1")
-	cant(t, db, "i:alice", "write", "infra.machines")
+	grant(t, db, tom, "i:a11ce", "read", "infra.machines")
```
dnephin, line 89: These IDs had to change now that `data.ListGrants` checks if the ID is a valid userID.

internal/access/grant.go
```diff
@@ -20,58 +20,36 @@ func GetGrant(c *gin.Context, id uid.ID) (*models.Grant, error) {
...
-			if inherited {
-				selectors = append(selectors, data.GrantsInheritedBySubject(subject))
-			} else {
-				selectors = append(selectors, data.BySubject(subject))
-			}
-			return data.ListGrants(db, p, selectors...)
+		case subject.IsIdentity() && rCtx.Authenticated.User.ID == subjectID:
+			// authorized because the request is for their own grants
+		case subject.IsGroup() && userInGroup(rCtx.DBTxn, rCtx.Authenticated.User.ID, subjectID):
+			// authorized because the request is for grants of a group they belong to
```
dnephin, line 36: This function did not need to perform the query in different places. Once the authorization check is complete we can perform the query once.

These checks here ensure that if the user does not have the necessary role, the query contains their own userID or a groupID they belong to. The `subject` here is the query parameter.

internal/server/data/grant.go
```diff
@@ -106,8 +104,91 @@ func GetGrant(tx ReadTxn, opts GetGrantOptions) (*models.Grant, error) {
...
+		} else {
+			subjects := []string{opts.BySubject.String()}
+
+			userID, err := opts.BySubject.ID()
+			if err != nil || !opts.BySubject.IsIdentity() {
+				return nil, fmt.Errorf("IncludeInheritedFromGroups requires a userId subject")
+			}
+			// FIXME: store userID and groupID as a field on the grants table so
+			// that we can replace this with a sub-select or join.
+			groupIDs, err := groupIDsForUser(tx, userID)
```
dnephin, line 147: This is the same implementation as we had before (a query for group membership, encoding the string ID in code, and then using those IDs in the main query). 

I think we're going to need to solve this problem for long-polling, so it should be a follow up.

internal/server/data/grant.go
```diff
@@ -106,8 +104,91 @@ func GetGrant(tx ReadTxn, opts GetGrantOptions) (*models.Grant, error) {
...
+
+	Pagination *Pagination
+}
+
+func ListGrants(tx ReadTxn, opts ListGrantsOptions) ([]models.Grant, error) {
+	table := grantsTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(table))
+	if opts.Pagination != nil {
+		query.B(", count(*) OVER()")
```
dnephin, line 129: This is a change to how we get the total count of results. It's the same approach I used for access keys, but different from what we had with gorm.

internal/server/data/grant.go
```diff
@@ -106,8 +104,91 @@ func GetGrant(tx ReadTxn, opts GetGrantOptions) (*models.Grant, error) {
...
+	}
+	if opts.ByPrivilege != "" {
+		query.B("AND privilege = ?", opts.ByPrivilege)
+	}
+	if opts.ByResource != "" {
+		query.B("AND resource = ?", opts.ByResource)
+	}
+	if !opts.IncludeConnectorGrant {
+		query.B("AND NOT (privilege = 'connector' AND resource = 'infra')")
+	}
```
dnephin, line 165: In some of the intermediate commits I tried to change this to `IncludeCreatedBySystem`, and use `CreatedBy <> 1`. Unfortunately that did not work out because we create grants from the config using the same `CreatedBy` ID. It would be nice if we could exclude system grants in a more general way.

internal/access/access.go
dnephin: Found the test coverage, and renamed the test to `TestRequireInfraRole_GrantsFromGroupMembership` in the latest commit.

api/grant.go
dnephin: Addressed this in #3176

internal/access/access.go
dnephin: The first one is addressed in #3176

internal/access/access_key.go
```diff
@@ -20,12 +20,8 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired
...
 		Pagination:     p,
 		IncludeExpired: showExpired,
-	}
-	if identityID != 0 {
-		opts.ByIssuedForID = identityID
-	}
-	if name != "" {
-		opts.ByName = name
+		ByIssuedForID:  identityID,
+		ByName:         name,
```
ssoroka, line 24: I think this implies ByIssuedForID and ByName are no longer required. We're definitely going to break the security in a bad way if all required fields become optional

internal/access/grant.go
```diff
@@ -20,58 +20,36 @@ func GetGrant(c *gin.Context, id uid.ID) (*models.Grant, error) {
...
-	} else {
-		selectors = append(selectors, data.ByOptionalSubject(subject))
-	}
-
-	if !showSystem {
-		selectors = append(selectors, data.NotInfraConnector())
+	opts := data.ListGrantsOptions{
+		ByResource:                 resource,
+		ByPrivilege:                privilege,
+		BySubject:                  subject,
```
ssoroka, line 47: this drops the distinction of required subject vs optional subject. Subject is required for querying your own grants and optional for admins

internal/server/data/grant.go
ssoroka: We discussed changing the grant's polymorphic fields from a single field into table + id fields; `subject` becomes `subject_type text` and `subject_id int`. Then you can rewrite this as a join through the membership table.

internal/server/data/grant.go
```diff
@@ -106,8 +104,91 @@ func GetGrant(tx ReadTxn, opts GetGrantOptions) (*models.Grant, error) {
...
+		query.B("AND resource = ?", opts.ByResource)
+	}
+	if opts.ExcludeConnectorGrant {
+		query.B("AND NOT (privilege = 'connector' AND resource = 'infra')")
+	}
+
+	query.B("ORDER BY id ASC")
+	if opts.Pagination != nil {
+		opts.Pagination.PaginateQuery(query)
+	}
```
ssoroka, line 170: doesn't pagination decide ordering?

internal/server/data/grant.go
```diff
@@ -151,65 +232,3 @@ func DeleteGrants(tx WriteTxn, opts DeleteGrantsOptions) error {
...
-			return db.Where("subject in (?)", subjects)
-		case subjectID.IsGroup():
-			return BySubject(subjectID)(db)
-		default:
-			panic("unhandled subject type")
-		}
-	}
-}
-
-func ByPrivilege(s string) SelectorFunc {
```
ssoroka, line 195: reminder that any selector that wasn't prefixed with Optional was required, and not to change the functionality to optional

internal/server/data/pagination.go
```diff
@@ -16,10 +16,12 @@ func (p *Pagination) SetTotalCount(count int) {
 }
 
 func (p *Pagination) PaginateQuery(query *querybuilder.Query) {
-	if p.Page == 0 && p.Limit == 0 {
+	if p.Limit == 0 {
```
ssoroka, line 19: should we be setting a default limit instead of returning?

ssoroka (changes requested):
biggest concern is that required selectors are being treated as optional (searching for '' or 0 is not the same as omitting the condition from the query), which potentially opens security holes.

internal/server/data/grant.go
dnephin: Not in the current implementation, it's set separately: https://github.com/infrahq/infra/blob/main/internal/server/data/data.go#L260-L272

Pagination is generic and applied the same way for all types, where as ordering is occasionally different for some types.

internal/server/data/pagination.go
dnephin: We do set a default limit in the API if the query comes in without a limit (`server.PaginationFromRequest`), but I don't think we want to set one here. Internal queries need to be allowed without a limit.

This is mostly just a safeguard, but I think we don't need it anymore. Since internal calls can use a `nil` to indicate they do not want pagination. I could remove this and rely on the defaults set in the API.

internal/access/access_key.go
dnephin: These were previously optional: https://github.com/infrahq/infra/blob/v0.15.1/internal/access/access_key.go#L22-L23 

So I don't think this changes behaviour.

It's has always been up to the caller to decide if the parameter should be optional or not. Now we're doing that with an `if` statement in the caller, instead of a helper function that hides the `if`.

There aren't a lot of examples of that because generally the IDs are already required by API request validation.

internal/access/grant.go
dnephin: I'm confident that it does not, because we have test coverage for those cases: https://github.com/infrahq/infra/blob/main/internal/server/grants_test.go#L148-L153

Related to my comment just above this, on line 33 and 35 (in the new code) we check that the subject query parameter from the request either exactly matches the userID, or is a member of the group identified by the subject ID.

ssoroka (approved):

PR #3164, opened 2022-09-09
fix(helm): server configmap incorrectly omits falsy values

mxyng:
## Summary

https://github.com/infrahq/infra/pull/3153 introduced a change that templates out server config values only if they're truthy but there are cases where a falsy value is valid and should be templated.

Update configmap templates removing some duplication and simplifying some of the templates. The server config contains values that need specially handling. This was previously done by omitting them from the generic templating but this requires coordination between the `omit` and the special handling. Now, the template is structured such that the special handling will remove the key from the map. Due to limitations of Helm templates, this is automatic and still requires some coordination but at least they're closer together.

### DIFF
--- a/helm/charts/infra/Chart.yaml
+++ b/helm/charts/infra/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: infra
 description: A Helm chart for Infra
 type: application
-version: 0.20.2
+version: 0.20.3
 appVersion: 0.15.1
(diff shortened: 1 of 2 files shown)


dnephin (commented):
Does this still allow someone to configure the `tls.certificate` and `tls.privateKey` instead of the CA? Seems like it makes the helm chart only work with one setup. 

Do we use a CA in prod, or do we use a TLS certificate directly?

mxyng:
You're right, it locks it into using a CA. We don't explicit configure TLS in our deployments so it's using the default. The Ingress does configure a certificate though

mxyng:
Updated the template to only template out the default TLS CA configurations when they're not defined by the user. This preserves the previous behaviour of passing in user created certificates and private keys

dnephin (approved):
Nice! I think this approach works well for allowing nested config

PR #3166, opened 2022-09-09
fix: transactional emails should skip marketing list unsubscribes

ssoroka:
## Checklist

### DIFF
--- a/internal/server/email/smtp.go
+++ b/internal/server/email/smtp.go
@@ -83,8 +83,16 @@ func writeln(w io.WriteCloser, s string, args ...any) error {
 	return err
 }
 
+// named arguments for the bypassListManagement boolean positional argument.
+// use BypassListManagement if you want the email to ignore any unsubscribe lists, spam lists, and bounce lists.
+// use RespectListManagement if you want the email to not deliver if the recipient is on one of those lists.
+// Unfortunately Sendgrid has no option of specifying the lists individually, which is problematic.
+// I opened a support issue about this with sendgrid.
+const BypassListManagement bool = true
+const RespectListManagement bool = false
+
 // SendSMTP sends an email message
-func SendSMTP(msg Message) error {
+func SendSMTP(msg Message, bypassListManagement bool) error {
 	client := &Client{} // setup a new client for each smtp send
 
 	if testClient != nil {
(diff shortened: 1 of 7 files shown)


internal/server/email/smtp.go
```diff
@@ -106,6 +106,9 @@ func SendSMTP(msg Message) error {
 		return err
 	}
 
+	if err := writeln(w, `X-SMTPAPI: {"filters":{"bypass_list_management":{"settings":{"enable":1}}}}`); err != nil {
```
dnephin, line 109: This is currently in a function called `SendSMTP`, but I think this setting is only appropriate for transactional emails. I think it would be easy for us to forget about this and accidentally use this function for emails that should not bypass these lists.

Could we have the caller enabled this with a setting on `Message`, so that the default is to use all the lists? Could be something like `Transactional bool`, or more specific like `ByPassListManagement bool`.

It'd also be great to link to the docs for this setting. If you add a field to `Message` to enable this, that'd be a great place for the comment with the link.

internal/server/email/smtp_test.go
```diff
@@ -104,6 +104,7 @@ func successCase(t *testing.T, c net.Conn) {
 	write("354 Ready")
 	for s := read(""); s != "."; s = read("") {
 		switch {
+		case strings.HasPrefix(s, "X-SMTPAPI:"):
```
dnephin, line 107: How can we test this change? I guess you can manually run the code against a dev sendgrid account? 

Could we also include some automated test coverage here to at least check that the JSON literal is well formed and parses into the expected structure? That should allow someone to make minor changes to this code without needing to setup a dev sendgrid account.

dnephin (commented):
Reading over the docs in https://docs.sendgrid.com/for-developers/sending-email/smtp-filters#filter-bypass_list_management and https://docs.sendgrid.com/ui/sending-email/index-suppressions this seems like a good way to solve the problem.

internal/server/email/smtp_test.go
ssoroka: if you change this code you're not going to get around testing it manually against sendgrid, as they're the only ones who can say it's valid/working, valid json or not.

dnephin: Similar to how we use golden files to test the CLI output, I think asserting that the value is what is expected is still valuable. It catches problems where the value has changed accidentally without us noticing.

Right now we could completely remove this header and no tests would fail.

dnephin (commented):
Change looks good, but I think we would benefit from some automated tests that fail when this is removed or changed.

dnephin (approved):
LGTM with an assert added to that test case to check the value.

I missed that the `default` does error, so we have coverage for the header existing, just missing coverage for the value.

PR #3167, opened 2022-09-09
Use SQL for encryption keys table

dnephin:
## Summary

This PR updates the implementation of `CreateEncryptionKey` and `GetEncryptionKey` to use SQL instead of gorm.

## Related Issues

Related to #2415

### DIFF
--- a/internal/server/data/encryption_key.go
+++ b/internal/server/data/encryption_key.go
@@ -8,6 +8,24 @@ import (
 	"github.com/infrahq/infra/internal/server/models"
 )
 
+type encryptionKeysTable models.EncryptionKey
+
+func (e encryptionKeysTable) Table() string {
+	return "encryption_keys"
+}
+
+func (e encryptionKeysTable) Columns() []string {
+	return []string{"algorithm", "created_at", "deleted_at", "encrypted", "id", "key_id", "name", "root_key_id", "updated_at"}
+}
+
+func (e encryptionKeysTable) Values() []any {
+	return []any{e.Algorithm, e.CreatedAt, e.DeletedAt, e.Encrypted, e.ID, e.KeyID, e.Name, e.RootKeyID, e.UpdatedAt}
+}
+
+func (e *encryptionKeysTable) ScanFields() []any {
+	return []any{&e.Algorithm, &e.CreatedAt, &e.DeletedAt, &e.Encrypted, &e.ID, &e.KeyID, &e.Name, &e.RootKeyID, &e.UpdatedAt}
+}
+
 func CreateEncryptionKey(db GormTxn, key *models.EncryptionKey) (*models.EncryptionKey, error) {
 	if key.KeyID == 0 {
 		// not a security issue; just an identifier
--- a/internal/server/data/encryption_key.go
+++ b/internal/server/data/encryption_key.go
@@ -1,10 +1,10 @@
 package data
 
 import (
+	"fmt"
 	mathrand "math/rand"
 
-	"gorm.io/gorm"
-
+	"github.com/infrahq/infra/internal/server/data/querybuilder"
 	"github.com/infrahq/infra/internal/server/models"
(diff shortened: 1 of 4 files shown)


internal/server/models/encryption_key.go
```diff
@@ -3,7 +3,11 @@ package models
 type EncryptionKey struct {
 	Model
 
-	KeyID     int32 `gorm:"uniqueIndex:idx_encryption_keys_key_id"` // a short identifier for the key that can be embedded with the encrypted payload
+	// KeyID is a short identifier for the key that can be embedded with the
+	// encrypted payload.
+	// TODO: remove this field? it's not used anywhere
+	KeyID int32
```
dnephin, line 9: I think this is supposed to allow us to support data key rotation, but it is not currently being used from what I can tell.  Nothing reads this field.

There's a keyID in the encrypted payload, but that comes from first 4 bytes of the checksum of the encrypted value.

internal/server/data/encryption_key.go
```diff
@@ -1,32 +1,59 @@
...
+}
+
+func (e *encryptionKeysTable) ScanFields() []any {
+	return []any{&e.Algorithm, &e.CreatedAt, &e.DeletedAt, &e.Encrypted, &e.ID, &e.KeyID, &e.Name, &e.RootKeyID, &e.UpdatedAt}
+}
+
+func CreateEncryptionKey(tx WriteTxn, key *models.EncryptionKey) error {
+	switch {
+	case key.Name == "":
+		return fmt.Errorf("a name is required")
```
BruceMacD, line 32: can we rely on this being logged in a way that it is clear the name being required refers to an encryption key?

BruceMacD (approved):
looks good

internal/server/data/encryption_key.go
dnephin: This error should be only a safeguard for us, not something ever seen by a user. It should never end up in logs.

It doesn't look like we wrap it in any more context. I'll add "EncryptionKey" to the message to make it more clear.

internal/server/data/encryption_key.go
```diff
@@ -1,32 +1,59 @@
...
 }
 
-func ByEncryptionKeyID(keyID int32) SelectorFunc {
-	return func(db *gorm.DB) *gorm.DB {
-		return db.Where("key_id = ?", keyID)
+func GetEncryptionKeyByName(tx ReadTxn, name string) (*models.EncryptionKey, error) {
+	table := &encryptionKeysTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(table))
+	query.B("FROM encryption_keys")
```
ssoroka, line 50: I think it's an oversight that we never moved encryption_keys to org-specific. Not much to do about it here.

internal/server/models/encryption_key.go
ssoroka: I would like to keep this as you're probably not going to be able to rotate keys without it. Would remove the comment.

ssoroka (approved):
lgtm but would remove the comment about removing the KeyID

internal/server/models/encryption_key.go
dnephin: Right now this field is not used, so I think it's important we at least document that.

When we looked at the code last week, I think we found that we are using a checksum of the encrypted data key to identify the key (not this field). https://github.com/infrahq/secrets/blob/1b971b9e367bb9d8e9837fc488028495a27885ce/secrets.go#L181-L187

So if we want to rotate keys, we'd use that checksum, not this value. If we wanted to use this keyID we'd have to migrate the existing encrypted payloads to use this KeyID first.

Alternatively we could continue to use the checksum of the encrypted data key as the identifier and remove this field.  

I think both are fine options, but we should clarify this in the code so that we don't continue to assume this is the keyID being used.

PR #3168, opened 2022-09-09
Move db encryption key loading to data package

dnephin:
## Summary

Branched from #3167

This PR introduces `NewDBOptions` as a struct that lets us pass options to `NewDB`. We'll be able to use this to set connection pool settings, and we'll very likely have other settings to pass along as well.

The PR also moves the `loadDBKey` function to the data package. This encryption key is only used by models as part of `Scan` and `Value` methods when reading and writing to the database, so it's primarily a concern on the data package.  
Moving it to the data package should make it easier to support separate keys per organization (if that's something we want to do), and should make it easier to handle key rotation. It may also give us a way to remove the package-level symmetric key, which will be necessary if we make either of those changes.

Also resolves a TODO about using the passed in transaction, instead of one from the closure.

### DIFF
--- a/internal/server/data/encryption_key.go
+++ b/internal/server/data/encryption_key.go
@@ -8,6 +8,7 @@ import (
 	"github.com/infrahq/secrets"
 
 	"github.com/infrahq/infra/internal"
+	"github.com/infrahq/infra/internal/server/data/migrator"
 	"github.com/infrahq/infra/internal/server/data/querybuilder"
 	"github.com/infrahq/infra/internal/server/models"
 )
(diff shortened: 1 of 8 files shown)


internal/server/data/encryption_key.go
```diff
@@ -26,7 +31,12 @@ func (e *encryptionKeysTable) ScanFields() []any {
 	return []any{&e.Algorithm, &e.CreatedAt, &e.DeletedAt, &e.Encrypted, &e.ID, &e.KeyID, &e.Name, &e.RootKeyID, &e.UpdatedAt}
 }
 
-func CreateEncryptionKey(tx WriteTxn, key *models.EncryptionKey) error {
+// StdlibTxn is a transaction that uses only the methods from the database/sql
+// package. It excludes the metadata that exists on the ReadTxn and WriteTxn.
+// StdlibTxn is for tables that are not scoped to an org.
+type StdlibTxn migrator.DB
```
dnephin, line 37: This is necessary because these functions are called by the migrator. We could remove this if we decided to only load the key after the migrations. If we want to allow migrations access to this key we need to do something like this.

dnephin (commented):
I fixed the commit history, so this should be ready for review

BruceMacD (approved):

PR #3175, opened 2022-09-09
Reduce verbosity in GenerateRoutes

dnephin:
## Summary

Most routes in `GenerateRoutes` are added with a single. We have a few special cases that have to set fields on the `route` definition, which makes them more verbose.  

This PR changes `add` to accept the `method` and `path` separate from the `route` struct. This allows us to move the route definition to a package-level variable, and keep `GenerateRoutes` to a single line per route.


BruceMacD (approved):
looks good

PR #3177, opened 2022-09-09
Remove deprecated identity value from gin.Context

dnephin:
## Summary

We introduced `RequestContext` a little while ago to replace the untyped and mutable `gin.Context`. At the time we left a few things in `gin.Context` to keep the change small. This PR removes the last value (`identity`) from `gin.Context`.

Also updates the authorization check in `GetGroup` to match the style we use in `ListGroups` and `ListGrants`.


BruceMacD (approved):
thanks for cleaning this up, looks good

PR #3199, opened 2022-09-12
Use SQL for Create, Update, and Delete groups

dnephin:
## Summary

Updates the data query functions for `CreateGroup`, `AddUserstoGroup`, `RemoveUsersFromGroup`, and `DeleteGroup` to use SQL.

Related to #2415

### DIFF
--- a/internal/server/data/group.go
+++ b/internal/server/data/group.go
@@ -138,14 +138,10 @@ func AddUsersToGroup(tx WriteTxn, groupID uid.ID, idsToAdd []uid.ID) error {
 	return handleError(err)
 }
 
-func RemoveUsersFromGroup(db GormTxn, groupID uid.ID, idsToRemove []uid.ID) error {
-	for _, id := range idsToRemove {
-		_, err := db.Exec("DELETE FROM identities_groups WHERE identity_id = ? AND group_id = ?", id, groupID)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+func RemoveUsersFromGroup(tx WriteTxn, groupID uid.ID, idsToRemove []uid.ID) error {
+	stmt := `DELETE FROM identities_groups WHERE group_id = ? AND identity_id IN (?)`
+	_, err := tx.Exec(stmt, groupID, idsToRemove)
+	return handleError(err)
 }
 
 // TODO: do this with a join in ListGroups and GetGroup
--- a/internal/server/data/group.go
+++ b/internal/server/data/group.go
@@ -25,8 +25,8 @@ func (g *groupsTable) ScanFields() []any {
 	return []any{&g.CreatedAt, &g.CreatedBy, &g.CreatedByProvider, &g.DeletedAt, &g.ID, &g.Name, &g.OrganizationID, &g.UpdatedAt}
 }
 
-func CreateGroup(db GormTxn, group *models.Group) error {
-	return add(db, group)
+func CreateGroup(tx WriteTxn, group *models.Group) error {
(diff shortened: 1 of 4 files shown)


internal/server/models/group.go
```diff
@@ -13,9 +13,7 @@ type Group struct {
 	CreatedBy         uid.ID
 	CreatedByProvider uid.ID
 
-	Identities []Identity `gorm:"many2many:identities_groups"`
```
dnephin, line 16: This field was not used, so removing it was as easy as removing this one line. We can always add it back in the future if we need it, but we'll have to add support to `CreateGroup` to actually use the column.

Right now most code seems to use the `Groups` field on `Identity`.

internal/server/data/group.go
```diff
@@ -71,59 +93,49 @@ func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
...
-		_, err := db.Exec("DELETE FROM identities_groups WHERE identity_id = ? AND group_id = ?", id, groupID)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+// RemoveUsersFromGroup removes any user ID listed in idsToRemove from the group
+// with ID groupID.
+// Note that DeleteGroup also removes users from the group.
+func RemoveUsersFromGroup(tx WriteTxn, groupID uid.ID, idsToRemove []uid.ID) error {
```
dnephin, line 135: This changed from a for loop performing one query per userID, to a single query using an `IN` clause. There will be some limit to the size of the query I guess, but fewer round trips to the database for most cases.

`DeleteGroup` has a similar query without the `identity_id IN` clause for removing all the members of a group.

internal/server/data/group.go
```diff
@@ -71,59 +93,49 @@ func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
...
 }
 
-func AddUsersToGroup(db GormTxn, groupID uid.ID, idsToAdd []uid.ID) error {
-	for _, id := range idsToAdd {
-		// This is effectively an "INSERT OR IGNORE" or "INSERT ... ON CONFLICT ... DO NOTHING" statement which
-		// works across both sqlite and postgres
-		_, err := db.Exec("INSERT INTO identities_groups (group_id, identity_id) SELECT ?, ? WHERE NOT EXISTS (SELECT 1 FROM identities_groups WHERE group_id = ? AND identity_id = ?)", groupID, id, groupID, id)
-		if err != nil {
-			return err
+func AddUsersToGroup(tx WriteTxn, groupID uid.ID, idsToAdd []uid.ID) error {
```
dnephin, line 117: This changed from a for loop performing one query per userID, to a single query that inserts multiple entries. It also uses `ON CONFLICT DO NOTHING` now that we only need to support postgreSQL.

internal/server/data/group.go
```diff
@@ -71,59 +93,49 @@ func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
 	return result, rows.Err()
 }
 
-func DeleteGroups(db GormTxn, selectors ...SelectorFunc) error {
-	toDelete, err := ListGroups(db, nil, selectors...)
+func DeleteGroup(tx WriteTxn, id uid.ID) error {
```
dnephin, line 96: This was only ever called with a single group ID, so this function has changed to only accept that one ID, which removes the need to list groups.

Group membership also changed from using `RemoveUsersFromGroup` to a query that deletes all rows that match the group ID.

pdevine (approved):
LGTM.

PR #3200, opened 2022-09-12
feat: add prom metrics for connector->server requests

### LINKED ISSUE #2664: Add destination name label to infra_destinations metric

We are looking for a way to proactively monitor and alert (using prometheus metrics & alerts) when destinations have been disconnected from the server for more than X duration.

The recent enhancement in https://github.com/infrahq/infra/pull/2560 is very helpful, but only provides a count of connected vs disconnected destinations.

This is useful, but difficult to provide actionable information in an alert that should be able to identify the failing destination(s).

We would like to destination name to be added to the label in the `infra_destinations` metric, this will allow us to identify the specific destination(s) that is failing and make the alerts more meaningful.

## Describe alternatives you've considered

None that apply

I'm aware that this will increase cardinality of said metric, however I think the risk & potential impact is fairly low, destinations are managed entities and even an increase of say 1K time-series is generally a non-issue for Prometheus, perhaps to mitigate this can be a default opt-out and be configurable, or can be split to a separate metric so can be skipped during scrape/ingestion.

## Environment Details

```
$ infra version
```

```
 Client: 0.13.6
 Server: 0.13.6
```

```
$ kubectl version
```

```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:21:56Z", GoVersion:"go1.18.4", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.8-gke.2200", GitCommit:"81a23326454f32475752ab4ef1ebdd164a6ba96d", GitTreeState:"clean", BuildDate:"2022-04-13T09:37:04Z", GoVersion:"go1.16.15b7", Compiler:"gc", Platform:"linux/amd64"}
```

mxyng:
## Summary

Observe the requests the connector makes to the server and report them as Prometheus metrics.

## Checklist

## Related Issues

Resolves #2664

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"runtime"
 	"strconv"
+	"time"
 
 	"github.com/ssoroka/slice"
 
@@ -39,6 +40,9 @@ type Client struct {
 	// OnUnauthorized is a callback hook for the client to get notified of a 401 Unauthorized response to any query.
 	// This is useful as clients often need to discard expired access keys.
 	OnUnauthorized func()
+
+	// ObserveFunc is a callback to measure and record the status and duration of the request
+	ObserveFunc func(time.Time, *http.Request, *http.Response, error)
 }
 
 // checkError checks the resp for an error code, and returns an api.Error with
@@ -110,10 +114,17 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 		req.Header[k] = v
 	}
 
+	start := time.Now()
 	resp, err := client.HTTP.Do(req)
+
+	if client.ObserveFunc != nil {
+		client.ObserveFunc(start, req, resp, err)
+	}
+
 	if resp != nil && resp.StatusCode == 401 && client.OnUnauthorized != nil {
 		defer client.OnUnauthorized()
 	}
+
 	if err != nil {
 		if connError := HandleConnError(err); connError != nil {
 			return nil, connError
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -145,7 +145,7 @@ func Run(ctx context.Context, options Options) error {
 		Name:      "request_duration_seconds",
 		Help:      "A histogram of duration, in seconds, performing HTTP requests.",
 		Buckets:   prometheus.ExponentialBuckets(0.001, 2, 15),
-	}, []string{"host", "method", "path", "status", "error"})
+	}, []string{"host", "method", "path", "status"})
 	promRegistry.MustRegister(responseDuration)
 
 	client := &api.Client{
@@ -164,14 +164,13 @@ func Run(ctx context.Context, options Options) error {
 			cancel()
 		},
 		ObserveFunc: func(start time.Time, request *http.Request, response *http.Response, err error) {
-			errorLabel := ""
-			if err != nil {
-				errorLabel = err.Error()
-			}
-
 			statusLabel := ""
 			if response != nil {
 				statusLabel = strconv.Itoa(response.StatusCode)
+			}
+
+			if err != nil {
+				statusLabel = "-1"
 			}
 
 			responseDuration.With(prometheus.Labels{
@@ -179,7 +178,6 @@ func Run(ctx context.Context, options Options) error {
 				"method": request.Method,
 				"path":   request.URL.Path,
 				"status": statusLabel,
-				"error":  errorLabel,
 			}).Observe(time.Since(start).Seconds())
 		},
 	}
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -19,6 +20,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/goware/urlx"
 	"github.com/infrahq/secrets"
+	"github.com/prometheus/client_golang/prometheus"
 	rbacv1 "k8s.io/api/rbac/v1"
 
 	"github.com/infrahq/infra/api"
@@ -137,6 +139,15 @@ func Run(ctx context.Context, options Options) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	promRegistry := metrics.NewRegistry(internal.FullVersion())
+	responseDuration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "http_client",
+		Name:      "request_duration_seconds",
+		Help:      "A histogram of duration, in seconds, performing HTTP requests.",
+		Buckets:   prometheus.ExponentialBuckets(0.001, 2, 15),
+	}, []string{"host", "method", "path", "status", "error"})
+	promRegistry.MustRegister(responseDuration)
+
 	client := &api.Client{
 		Name:      "connector",
 		Version:   internal.Version,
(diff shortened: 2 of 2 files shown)


internal/connector/connector.go
```diff
@@ -136,11 +138,18 @@ func Run(ctx context.Context, options Options) error {
...
 
-	client := &api.Client{
-		Name:      "connector",
-		Version:   internal.Version,
-		URL:       u.String(),
-		AccessKey: accessKey,
+	promRegistry := metrics.NewRegistry(internal.FullVersion())
+	responseDuration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "serverhttp",
+		Name:      "response_duration_seconds",
```
mxyng, line 144: Not sure if this should be `request` or `response`

BruceMacD: I am on the fence too, maybe just more specific wording `response_to_connector_duration_seconds`

internal/connector/connector.go
```diff
@@ -151,7 +160,20 @@ func Run(ctx context.Context, options Options) error {
 			logging.Errorf("Unauthorized error; token invalid or expired. exiting.")
 			cancel()
 		},
-	}
+		ObserveFunc: func(start time.Time, request *http.Request, response *http.Response) {
+			statusCode := http.StatusInternalServerError
+			if response != nil {
```
BruceMacD, line 165: If the response is nil does that mean the request timed out?

BruceMacD (approved):

api/client.go
```diff
@@ -28,17 +29,35 @@ const (
...
+	ObserveFunc func(time.Time, *http.Request, *http.Response)
+}
+
+type Client struct {
+	URL       string
+	AccessKey string
+	Options   *ClientOptions
+}
+
+func NewClient(serverURL, serverAccessKey string, options ClientOptions) *Client {
```
dnephin: What's the distinction between options and these other two fields? Isn't access key just as much of an option as the rest?  You can make requests to some endpoints without an access key.

We frequently use option structs for building complex types, but in this case since the type `Client` doesn't have any state, the `Client` type itself is basically an options struct.

internal/connector/connector.go
```diff
@@ -136,11 +138,18 @@ func Run(ctx context.Context, options Options) error {
...
-	client := &api.Client{
-		Name:      "connector",
-		Version:   internal.Version,
-		URL:       u.String(),
-		AccessKey: accessKey,
+	promRegistry := metrics.NewRegistry(internal.FullVersion())
+	responseDuration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "serverhttp",
+		Name:      "response_duration_seconds",
+		Help:      "A histogram of duration, in seconds, handling server HTTP responses",
```
dnephin, line 145: I guess this needs to be changed  to something like:
```suggestion
		Help:      "A histogram of duration, in seconds, of HTTP requests to the infra server",
```

api/client.go
```diff
@@ -92,12 +111,12 @@ func request[Req, Res any](client Client, method string, path string, query Quer
 	req.URL.RawQuery = url.Values(query).Encode()
 
 	clientName, clientVersion := "client", "unknown"
-	if client.Name != "" {
-		clientName = client.Name
+	if client.Options.Name != "" {
+		clientName = client.Options.Name
```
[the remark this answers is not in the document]
mxyng: On second thought, options as a pointer is probably unnecessary

internal/connector/connector.go
dnephin: I think this needs to be named from the perspective of the person operating the connector.

Something like `infra_api_request_seconds`.

mxyng: I don't think there's any reason to be that specific. The Prometheus instance scraping this metric will add context. The final metric as seen by an end user will look something like `server_http_response_duration_seconds{job="infra-connector"}`

internal/connector/connector.go
mxyng: I think so though 500 Internal Server Error should be a good enough default to capture cases like this.

dnephin (commented):
The change seems reasonable to me. I guess someone can use this to solve #2664 by building an alert that checks for the time size the last 200 response code from this metric?

I think this metric is valuable for other purposes, but generally how I would expect to monitor something like #2664 would be with a gauge metric that emits the last time the reconciler ran successfully. If the reconcile loop is failing for any reason (not just calls to the infra API), that's a signal that the role bindings are out of date. I think monitoring the success of the reconcile operation is the most accurate signal.   I'd use the gauge metric in a monitor that checks when the value is older than some threshold, and send an alert.  Not a blocker, just something to consider.

internal/connector/connector.go
dnephin: `server_http_response_duration` is very misleading I think. It's not the server response duration, it's the clients view of the request.  Someone using this metric with a SaaS connector is going to be very confused by this name.

mxyng: The metric is scraped from `job=infra-connector` and represents the job's view of the world. This metric measures the duration of a request to the server. It's using `response` as the inverse to `http_request_duration_seconds` which measure the duration of a request to the connector. 

This is also the inverse of the server's view of the world which is measured by `http_request_duration_seconds{job='infra-server'}`.

The labels are equally as important as the metric name. We could even omit `server` from the metric name because it's captured by the `host` label which describe what the client is talking to.

api/client.go
mxyng: I was thinking URL and AccessKey were mandatory fields while `Options` contains optional fields but that's doesn't look to be the case. Will roll this back

mxyng:
@dnephin you're right that this does not capture the sync status of the cluster's roles but that was never the intent. My understanding of #2664 is that it's specifically concerned with the connectivity between the server and the connector, which tracking requests will address. The sync status of cluster roles is out of scope.

internal/connector/connector.go
dnephin: The labels definitely help for sure. I think it would be good to omit server. I think the users reading this metric won't understand that `server` implies "infra api server". 

Are you thinking that if we add a metric for timing to the kube api server we'd use the same metric name, and someone would distinguish the two from the "host" label ?

mxyng: I agree. Initially I thought `server` will help clarify things but it's causing more confusion. 

> Are you thinking that if we add a metric for timing to the kube api server we'd use the same metric name, and someone would distinguish the two from the "host" label ?

That's exactly my thought. There's no reason to be specific about the metrci name since there's many things the connector needs to communicate with, Infra server being only one of them. Others may include the k8s API server or other integrations that happen to use HTTP

dnephin: Makes sense! I wonder if `http_client_request_duration` would work for this

internal/connector/connector.go
```diff
@@ -136,6 +138,15 @@ func Run(ctx context.Context, options Options) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	promRegistry := metrics.NewRegistry(internal.FullVersion())
+	responseDuration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "http_client",
+		Name:      "request_duration_seconds",
+		Help:      "A histogram of duration, in seconds, handling HTTP requests.",
```
dnephin, line 145:
```suggestion
		Help:      "A histogram of duration, in seconds, performing HTTP requests.",
```

internal/connector/connector.go
```diff
@@ -151,6 +162,25 @@ func Run(ctx context.Context, options Options) error {
 			logging.Errorf("Unauthorized error; token invalid or expired. exiting.")
 			cancel()
 		},
+		ObserveFunc: func(start time.Time, request *http.Request, response *http.Response, err error) {
+			errorLabel := ""
+			if err != nil {
+				errorLabel = err.Error()
```
dnephin, line 168: Won't this be too high cardinality, since these are effectively infinite error strings?

I was thinking of just a `true` or `false` value for the error label.

mxyng: Technically yes. My concern with setting just `true` or `false` is it provides no information as to why it failed. With the status code, there's at least some context. Setting a `true` or `false` error is almost like setting status to `-1` on error instead of a meaningful HTTP code

dnephin: That's true. We could attempt to detect the error and reduce it to a few well known constant values, but I'm not sure it's worth the effort right now. We can always add that later as a separate `errorClass` or some other label.

Generally I would not expect metrics to tell me about the error. If I see the number of errors increasing I would go to logs to figure out more details about the errors.

mxyng: That's true. It's nice when metrics tell you what the problem is but not necessary. In this case, I'd prefer setting `status` to `-1` since `status` and `error=true` are (probably) mutually exclusive; adding a separate labels seems unnecessary

PR #3201, opened 2022-09-12
Add a scipt to sync linter deps before running lint

dnephin:
## Summary

Our custom linter is run by `golangci-lint` as a Go plugin. Go plugins must be built with the exact same version of Go, and the exact same version of all shared Go modules.

`make lint` will fail if someone updates `golangci-lint` on their local machine. This PR addresses the problem by adding a command to update the `go.mod` of the `querylinter` module based on the modules used by their local `golangci-lint`.

Usage:

From the `internal/tools/querylinter` directory, run
```
go run ./sync/sync-mod.go
```

### DIFF
--- a/internal/tools/querylinter/go.mod
+++ b/internal/tools/querylinter/go.mod
@@ -3,12 +3,12 @@ module github.com/infrahq/infra/internal/tools/querylinter
 go 1.19
 
 require (
+	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4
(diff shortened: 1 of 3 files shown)


ssoroka (approved):

dnephin:
I amended the commit to include the necessary `go.mod` changes that I missed in the previous PR.

We didn't notice it in the other PR because our tests now long run when specific files match. I'm going to fix the file filter in another PR.

Issue #3161, opened 2022-09-08
Non-existent API routes returns an HTML response


mxyng:
## Describe the bug

The UI proxy acts as a catch-all route for incoming requests. This has the side effect of routing any unknown endpoints to the UI even if the route is part of the API, e.g. `/api/nonexistent`. The request, being served by the UI, returns a HTML 404 instead of a JSON 404. For an API consumer this will be confusing.

## Expected behavior

Requests to `/api/*` should return an API response, not a UI response, e.g. a JSON 404 message.
Requests not to `/api/*` can return a UI response.

PR #3202, opened 2022-09-12
feat: transparent email verification

ssoroka:
## Summary

- adds email verification tokens to identity
- wraps common email links with a verification check and redirect
- adds supports for server handlers that redirect instead of rendering responses.

## Checklist

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -66,6 +66,7 @@ func migrations() []*migrator.Migration {
 		dropOrganizationNameIndex(),
 		sqlFunctionsMigration(),
 		setDefaultOrgID(),
+		addIdentityVerifiedFields(),
 		// next one here
 	}
 }
@@ -594,6 +595,22 @@ func setDefaultOrgID() *migrator.Migration {
 			}
 			return nil
 
+		},
+	}
+}
+
+func addIdentityVerifiedFields() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-09-01T15:00",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `
+ALTER TABLE identities
+	ADD COLUMN IF NOT EXISTS verified boolean NOT NULL DEFAULT false,
+	ADD COLUMN IF NOT EXISTS verification_token text NOT NULL DEFAULT substr(replace(translate(encode(decode(MD5(random()::text), 'hex'),'base64'),'/+','=='),'=',''), 1, 10);
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_identities_verified ON identities (organization_id, verification_token) WHERE (deleted_at IS NULL);`
+			_, err := tx.Exec(stmt)
+			return err
 		},
 	}
 }
--- a/internal/server/data/schema.sql
+++ b/internal/server/data/schema.sql
@@ -161,7 +161,9 @@ CREATE TABLE identities (
     name text,
     last_seen_at timestamp with time zone,
     created_by bigint,
-    organization_id bigint
+    organization_id bigint,
+    verified boolean DEFAULT false NOT NULL,
+    verification_token text DEFAULT substr(replace(translate(encode(decode(md5((random())::text), 'hex'::text), 'base64'::text), '/+'::text, '=='::text), '='::text, ''::text), 1, 10) NOT NULL
 );
 
 CREATE TABLE identities_groups (
@@ -287,6 +289,8 @@ CREATE UNIQUE INDEX idx_grant_srp ON grants USING btree (organization_id, subjec
 CREATE UNIQUE INDEX idx_groups_name ON groups USING btree (organization_id, name) WHERE (deleted_at IS NULL);
 
 CREATE UNIQUE INDEX idx_identities_name ON identities USING btree (organization_id, name) WHERE (deleted_at IS NULL);
+
+CREATE UNIQUE INDEX idx_identities_verified ON identities USING btree (verification_token) WHERE (deleted_at IS NULL);
 
 CREATE UNIQUE INDEX idx_organizations_domain ON organizations USING btree (domain) WHERE (deleted_at IS NULL);
 
--- a/internal/server/data/schema.sql
+++ b/internal/server/data/schema.sql
@@ -290,7 +290,7 @@ CREATE UNIQUE INDEX idx_groups_name ON groups USING btree (organization_id, name
 
 CREATE UNIQUE INDEX idx_identities_name ON identities USING btree (organization_id, name) WHERE (deleted_at IS NULL);
 
-CREATE UNIQUE INDEX idx_identities_verified ON identities USING btree (verification_token) WHERE (deleted_at IS NULL);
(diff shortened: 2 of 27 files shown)


internal/server/data/migrations.go
```diff
@@ -597,3 +598,19 @@ func setDefaultOrgID() *migrator.Migration {
...
 }
+
+func addIdentityVerifiedFields() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-09-01T15:00",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `
+ALTER TABLE identities
+	ADD COLUMN IF NOT EXISTS verified boolean NOT NULL DEFAULT false,
+	ADD COLUMN IF NOT EXISTS verification_token text NOT NULL DEFAULT substr(replace(translate(encode(decode(MD5(random()::text), 'hex'),'base64'),'/+','=='),'=',''), 1, 10);
```
dnephin, line 609: How does postgres handle duplicates if this happens to generate a random token that already existed?

internal/server/email/templates/signup.text.html
```diff
@@ -2,5 +2,9 @@ <h3>Welcome to Infra!</h3>
 
 <p>
   You can sign into your account any time from 
-  <a href="{{.Link}}">{{.Link}}</a>.
+  <a href="{{.WrappedLink}}">https://{{.Link}}</a>.
```
dnephin, line 5: I believe the link already has the scheme on it, from what I can see in `API.Signup`
```suggestion
  <a href="{{.WrappedLink}}">{{.Link}}</a>.
```

It'd be great to have some test coverage using `gotest.tools/v3/golden` to be able to preview the fully rendered email. Not a blocker for this PR though.

internal/server/models/identity.go
```diff
@@ -43,3 +47,11 @@ func (i *Identity) ToAPI() *api.User {
 func (i *Identity) PolyID() uid.PolymorphicID {
 	return uid.NewIdentityPolymorphicID(i.ID)
 }
+
+func (i *Identity) BeforeSave(_ *gorm.DB) error {
+	if len(i.VerificationToken) == 0 {
+		i.VerificationToken = generate.MathRandom(10, generate.CharsetAlphaNumeric)
```
dnephin, line 53: Are there two ways to generate a token? The default value for the database field, and this here?

Do we need both?  Doing it in the application (this code here), seems a little easier to read than the default on the database field.

internal/server/routes.go
```diff
@@ -123,6 +123,14 @@ func (s *Server) GenerateRoutes() Routes {
 	get(a, noAuthnWithOrg, "/api/providers/:id", a.GetProvider)
 	get(a, noAuthnWithOrg, "/api/providers", a.ListProviders)
 	get(a, noAuthnWithOrg, "/api/settings", a.GetSettings)
+	add(a, noAuthnWithOrg, route[api.VerifyAndRedirectRequest, *api.RedirectResponse]{
```
dnephin, line 126: I think this will conflict with the changes in #3175 (just merged). With those changes you can move the `route` out to a package-level var, so that only the group, method, and path are inline here.

internal/server/routes.go
```diff
@@ -123,6 +123,14 @@ func (s *Server) GenerateRoutes() Routes {
...
+		method:                     http.MethodGet,
+		path:                       "/link",
```
dnephin, line 128: This should probably be `/api/link`, so that we keep all paths under `/api`. Otherwise problems like #3161 become a lot harder to fix.

Thoughts on `/api/verify` instead of `/api/link` ? I guess this is trying to obscure what it's doing? Or is the idea that we might use this for other purposes later?

internal/server/routes.go
```diff
@@ -184,6 +192,8 @@ func add[Req, Res any](a *API, group *routeGroup, route route[Req, Res]) {
 	bindRoute(a, group.RouterGroup, route.method, route.path, handler)
 }
 
+var redirectResponseType = reflect.TypeOf(api.RedirectResponse{})
```
dnephin, line 195: I think the file that added `api.RedirectResponse` did not get checked in. I'm not seeing it in this branch.

internal/server/routes.go
```diff
@@ -242,11 +252,30 @@ func wrapRoute[Req, Res any](a *API, route route[Req, Res]) func(*gin.Context) e
...
+		if r, ok := responseIsRedirect(resp); ok {
+			c.Redirect(http.StatusPermanentRedirect, r.RedirectTo)
+		} else {
+			c.JSON(responseStatusCode(route.method, resp), resp)
+		}
 		return nil
 	}
 }
 
+func responseIsRedirect(resp any) (api.RedirectResponse, bool) {
```
dnephin, line 264: I wonder if we can avoid the reflection here. If `api.RedirectResponse` implemented something like:

```go
type redirectResponse interface {
    statusCoder
    RedirectTo() string
}
```

I think that would allow us to use the same pattern as `responseStatusCode`

dnephin (commented):
Looks good!

internal/server/data/migrations.go
ssoroka: 💥 
though the chances of that are about in in 61^10, which is so absurdly large we don't need to worry about it.

internal/server/email/templates/signup.text.html
ssoroka: we can already preview rendered emails in tests. see TestSendEmail

ssoroka: added an example of this

internal/server/routes.go
ssoroka: it's not strictly related to the API. If I put api in the link it'll be visible in links in emails, which I don't think is ideal.
It's not important to the user that it's for verification, and if the user has already verified, seeing verification in the link could be confusing; the link will still function properly.

internal/server/routes.go
ssoroka: a good suggestion, but I'm not sure that works. the response isn't an interface, it's a concrete type you just don't know what it is because of generics.

ssoroka: You'd have to do something like 
```
		var r interface{} = resp
		if rsp, ok := r.(isRedirect); ok {
			c.Redirect(http.StatusPermanentRedirect, rsp.RedirectURL())
		}
```

internal/server/data/schema.sql
```diff
@@ -161,7 +161,9 @@ CREATE TABLE identities (
     name text,
     last_seen_at timestamp with time zone,
     created_by bigint,
-    organization_id bigint
+    organization_id bigint,
+    verified boolean DEFAULT false NOT NULL,
+    verification_token text DEFAULT substr(replace(translate(encode(decode(md5((random())::text), 'hex'::text), 'base64'::text), '/+'::text, '=='::text), '='::text, ''::text), 1, 10) NOT NULL
```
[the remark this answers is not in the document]
ssoroka: no, it should be seeded automatically when the server starts

internal/server/routes.go
ssoroka: nm this is fine, I've got it looking good.

BruceMacD (approved):

PR #3203, opened 2022-09-13
Add more tests for well known jwks

dnephin:
## Summary

Show the endpoint working with org lookup from access key.

Also include the response status code in connector logs, if the request failed.

### DIFF
--- a/internal/connector/authn.go
+++ b/internal/connector/authn.go
@@ -104,6 +104,10 @@ func (j *authenticator) getJWK() (*jose.JSONWebKey, error) {
 	}
 	defer res.Body.Close()
 
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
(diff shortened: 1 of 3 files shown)


internal/connector/authn.go
```diff
@@ -104,6 +104,10 @@ func (j *authenticator) getJWK() (*jose.JSONWebKey, error) {
 	}
 	defer res.Body.Close()
 
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
```
dnephin, line 107: 3xx codes aren't generally errors, but the Go http client handles redirects. So if we ever get a 3xx code here it means the redirect loop was cancelled, so that is an error.

We do the same thing in our api.Client: https://github.com/infrahq/infra/blob/main/api/client.go#L50

BruceMacD (approved):
looks good 👍

PR #3204, opened 2022-09-13
ci: run core ci workflow on more paths

dnephin:
If any of the github workflow files change, or if the linter config changes we should run the core CI workflow.

### DIFF
--- a/.github/workflows/ci-core.yaml
+++ b/.github/workflows/ci-core.yaml
@@ -6,6 +6,9 @@ on:
       - go.mod
       - go.sum
       - '**/*.go'
+      - '.github/workflows/ci-core.yaml'
+      - '.golangci.yml'
+
   push:
     branches:
       - main
@@ -13,6 +16,8 @@ on:
       - go.mod
       - go.sum
       - '**/*.go'
+      - '.github/workflows/*-core.yaml'
+      - '.golangci.yml'
 
 jobs:
   lint:


.github/workflows/ci-core.yaml
```diff
@@ -6,13 +6,18 @@ on:
       - go.mod
       - go.sum
       - '**/*.go'
+      - '.github/workflows/**'
```
mxyng, line 9: Probably just need the `.github/workflows/ci-core.yaml`

dnephin: Ya, that's probably fine for `pull_request` target. For `push` to main it would be nice to run everything. I've made that change.

mxyng: I don't see a reason to run this workflow for all changes on push. This workflow triggers other workflows eventually leading to deploys so it should be as targeted as possible. Only changes related to the core code should trigger it otherwise unrelated changes will trigger image builds and deploys. At most, this should trigger on `.github/workflows/*-core.yaml`

dnephin: I generally expect all CI jobs to run on `main`, even if we don't expect the thing they are testing to have changed. I guess right now we don't really have integration tests between the different components, but once we do those will be important to run.

Running tests more frequently also helps catch flaky tests, and other problems with the build (like a bug in a new release of an unpinned docker image).

I had not considered that it would also deploy things to dev, which I guess is not desirable.  I'll change the `push` target to use `*-core.yaml`. I think that's a good option that lets us test changes to the deployment workflows as soon as they merge.

mxyng (approved):

PR #3208, opened 2022-09-14
fix: better errors on bad email

### LINKED ISSUE #3291: Stronger email validation on server side
"Dotless" domains shouldn't be allowed as they hare [prohibited](https://www.icann.org/en/announcements/details/new-gtld-dotless-domain-names-prohibited-30-8-2013-en) by ICANN

ssoroka:
should probably avoid explicitly wrong emails and give better messages about it.

closes #3291

### DIFF
--- a/internal/validate/email.go
+++ b/internal/validate/email.go
@@ -3,6 +3,7 @@ package validate
 import (
 	"fmt"
 	"net/mail"
+	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
 )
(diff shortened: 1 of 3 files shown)


internal/validate/email.go
```diff
@@ -28,6 +29,10 @@ func (e email) Validate() *Failure {
 	if addr.Name != "" {
 		return fail(e.name, fmt.Sprintf("email address must not contain display name %q", addr.Name))
 	}
+	// domain must contain at least one dot. ParseAddress already checks that it's not leading or trailing
+	if !strings.Contains(strings.Split(addr.Address, "@")[1], ".") {
```
BruceMacD, line 33: The example for the mail address shows the name being included as well:
```
// Address represents a single mail address.
// An address such as "Barry Gibbs <bg@example.com>" is represented
// as Address{Name: "Barry Gibbs", Address: "bg@example.com"}.
type Address struct {
	Name    string // Proper name; may be empty.
	Address string // user@domain
}
```

Looks like that would mean `user.name@domain` would pass this validation when we dont want it to.

BruceMacD (commented):
We should add a test for this also if possible

internal/validate/email.go
ssoroka: Not sure what you mean, bruce. This check splits on @ first, takes the domain, then looks for `.`

BruceMacD: Ah, my bad, read this wrong

BruceMacD (approved):

PR #3209, opened 2022-09-14
feat: connector service.type NodePort

mxyng:
## Summary

If connector is using service.type=NodePort, get the node list and find the first external address. If no external address is available, fallback to using the internal IP but warn the user. The internal IP is usable iff the client has a route to that network.

May be worth adding an option to prefer the internal IP which will also disable the warning.

Tested with public node port in EKS.

### DIFF
--- a/helm/charts/infra/templates/connector/clusterrole.yaml
+++ b/helm/charts/infra/templates/connector/clusterrole.yaml
@@ -17,6 +17,7 @@ rules:
       - pods
       - services
       - namespaces
+      - nodes
     verbs:
       - list
   - apiGroups:
--- a/internal/kubernetes/kubernetes.go
+++ b/internal/kubernetes/kubernetes.go
@@ -578,6 +578,57 @@ func (k *Kubernetes) Service(component string, labels ...string) (*corev1.Servic
 	return &services.Items[0], nil
 }
 
+func (k *Kubernetes) Nodes() ([]corev1.Node, error) {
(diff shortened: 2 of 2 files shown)


BruceMacD (approved):
Looks good to me

ssoroka (approved):

pdevine:
Is there an easy way to test this out locally?

mxyng:
On Linux, it should be fairly easy. The connector should report at least the internal IP of the node which should be routable by the default namespace. 

On Mac, this is significantly harder since Docker lives in a VM with no direct route. You could enter the VM and check the node port directly with `nc` and compare the results with `infra destinations list`

```
$ docker run -it --privileged --pid=host debian nsenter -t 1 -m -n nc -vz 172.20.0.2 30443
172.20.0.2 (172.20.0.2:30443) open
```

```
$ infra destinations list
  NAME          URL               STATUS     LAST SEEN
  3f541bf770ff  172.20.0.2:30443  Connected  23 seconds ago
```

Verifying the complete experience requires getting the Linux infra and kubectl binarines and sharing the mount namespace. Assuming `$PWD` has `infra` and `kubectl`:

```
$ docker run -it -v $PWD:/mnt --privileged --pid=host debian nsenter -t 1 -n --wd=/mnt bash
root@195119ec827e:/mnt# ./infra login <server>
root@195119ec827e:/mnt# ./kubectl config use-context infra:<context>
root@195119ec827e:/mnt# ./kubectl get pods
```

PR #3210, opened 2022-09-14
Use SQL for ListGroups and GetGroup

### LINKED ISSUE #2415: Isolate all references to gorm in the data package

dnephin:
## Summary

Branched from #3199 

Convert `ListGroups` and `GetGroup` to use SQL and add a few more test cases for them.

## Related Issues

Resolves #2415

### DIFF
--- a/internal/server/data/group.go
+++ b/internal/server/data/group.go
@@ -132,7 +132,7 @@ func ListGroups(tx ReadTxn, opts ListGroupsOptions) ([]models.Group, error) {
 	return result, nil
 }
 
-func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
+func ListGroupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
 	stmt := `SELECT DISTINCT group_id FROM identities_groups WHERE identity_id = ?`
 	rows, err := tx.Query(stmt, userID)
 	if err != nil {
--- a/internal/server/data/group.go
+++ b/internal/server/data/group.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"time"
 
-	"gorm.io/gorm"
-
 	"github.com/infrahq/infra/internal/server/data/querybuilder"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/uid"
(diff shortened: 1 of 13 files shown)


internal/server/data/token.go
```diff
@@ -68,7 +68,7 @@ func CreateIdentityToken(db GormTxn, identityID uid.ID) (token *models.Token, er
 		return nil, err
 	}
 
-	identityGroups, err := ListGroups(db, nil, ByGroupMember(identityID))
+	identityGroups, err := ListGroups(db, ListGroupsOptions{ByGroupMember: identityID})
```
dnephin, line 71: This was a required selector, but there's no way we can get here with a zero value for identityID because:
1. `CreateIdentityToken` is always called with a non-zero ID
2. If the ID were 0 it would the query above would return an error

internal/server/data/identity.go
```diff
@@ -157,7 +157,7 @@ func DeleteIdentities(tx GormTxn, selectors ...SelectorFunc) error {
 			return err
 		}
 
-		groups, err := ListGroups(tx, nil, ByGroupMember(i.ID))
+		groups, err := ListGroups(tx, ListGroupsOptions{ByGroupMember: i.ID})
```
dnephin, line 160: This was a required selector, but the value is always non-zero because it comes from `ListIdentities`.

internal/access/identity.go
```diff
@@ -64,7 +64,7 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return fmt.Errorf("delete identity access keys: %w", err)
 	}
 
-	groups, err := data.ListGroups(db, nil, data.ByGroupMember(id))
+	groups, err := data.ListGroups(db, data.ListGroupsOptions{ByGroupMember: id})
```
dnephin, line 67: This was a required selector, but there's no way this can be called with a zero value for `id`, because the API request validation marks it as required.

internal/server/data/group.go
```diff
@@ -33,48 +31,113 @@ func CreateGroup(tx WriteTxn, group *models.Group) error {
...
+	table := groupsTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(table))
+	if opts.Pagination != nil {
+		query.B(", count(*) OVER()")
+	}
+	query.B("FROM groups")
+	if opts.ByGroupMember != 0 {
+		query.B("JOIN identities_groups ON groups.id = identities_groups.group_id")
+		query.B("AND identities_groups.identity_id = ?", opts.ByGroupMember)
```
BruceMacD, line 91: Shouldn't the group member identity ID be a where clause rather than part of the join?

BruceMacD (approved):

internal/server/data/group.go
dnephin: Ok, I think I found an answers to this: https://dba.stackexchange.com/a/155976

It sounds like for inner joins (`JOIN` by itself is an inner join [ref](https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-JOIN)) it makes no difference to postgreSQL. It will perform the same query.

For us, the advantage to doing it here in the `JOIN` is that we don't need a second `if opts.ByGroupMember`. We can include all of the clauses in the same code block. If we were to try and use the where we'd have to split it up.

ssoroka (approved):

PR #3211, opened 2022-09-15
maintain: track provider groups separate from users

### LINKED ISSUE #2982: Keep track of where groups were syncronized from
When a group is automatically added to Infra from an identity provider it is not clear where this group came from.

Add a `providedBy` field to group entities.

### LINKED ISSUE #3138: Cannot use comma in group names
## Describe the bug
If a group synchronized from an identity provider has a comma in the name it causes an error on login.

## Expected behavior
Login succeeds and the group is added with an escaped comma.

BruceMacD:
- add provider information to trace identity group membership relations
- migrate provider user groups to provider groups

## Summary
In order to track where users had their groups assigned to them more accurately updating the database schema to track and resolve group membership. This means tracking the state of external identity provider groups with relations to their provider users, and adding provider information to identities_groups relations.

`identities_groups` relations now have the form (identity_id, group_id, provider_id, provider_group_name) to allow mapping external groups from identity providers to groups within Infra. This behavior will be a later change, this PR keeps the same behavior of automatically mapping groups from identity providers to groups within Infra that have the same name.

## Checklist

## Related Issues

Resolves #3138
Resolves #2982

### DIFF
--- a/docs/dev/identity-provider-tracking.md
+++ b/docs/dev/identity-provider-tracking.md
@@ -1,0 +1,29 @@
+# Identity Provider Tracking
+Infra tracks users and groups that exist in external identity providers and maps them access. The following database diagram shows how the relation of groups and users in identity providers are mapped to users and groups within Infra.
+
+Our identity system has two different type of relations, one of which exists in Infra (users and groups controlled by Infra), and the other relation which exists externally (information about users and groups which exist in an identity provider).
+
+To get all the groups a user belongs to as a combination of direct group membership and membership in a mapped identity provider group use the `resolved_identity_groups` view.
+
+## Users and Groups Managed by Infra
+The following tables represent users whose identity is managed by Infra and groups which are controlled by Infra.
+- **Identities**: Known information about a user. This may be a an email and password hash that can be used to login to Infra directly, or information about a user that has logged via an identity provider.
+- **Groups**: Groups which exist in Infra that membership to grants access to some resource.
+- **Identities_Groups**: Identities that have been added as a member of a group within Infra, this membership is added through either membership in an external identity provider group or through direct assignment (represented in the database as being part of an Infra provider group).
+
+## Users and Groups Managed by an Identity Provider
+- **Provider_User**: A user that exists in an identity provider.
+- **Provider_Group**: A group that exists in an identity provider.
+- **Provider_Groups_Provider_Users**: This relation shows which users of an identity provders are members of which groups in the same identity provider.
+
+```mermaid
+erDiagram
+          PROVIDER_GROUPS_MAPPINGS }o--|| PROVIDER : ""
+          GROUP ||--o{ IDENTITIES_GROUPS : "direct membership to infra group"
+          IDENTITIES_GROUPS }o--|| IDENTITY : "direct membership to infra group" 
+          IDENTITY ||--o{ PROVIDER_USER : "infra identity composed of identity provider users"
+          PROVIDER_GROUP }o--|| PROVIDER : "sync groups to infra"
+          PROVIDER_GROUP ||--o{ PROVIDER_GROUPS_PROVIDER_USERS : "sync members to infra"
+          PROVIDER_GROUPS_PROVIDER_USERS }o--|| PROVIDER_USER : "sync members to infra"
+          PROVIDER_USER }o--|| PROVIDER : "sync identities to infra"
+```
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -130,7 +130,7 @@ func UpdateIdentityInfoFromProvider(c RequestContext, oidc providers.OIDCClient)
 		return fmt.Errorf("user info provider: %w", err)
 	}
 
-	// get current identity provider groups and account status
+	// update current identity provider groups and account status
 	err = data.SyncProviderUser(ctx, db, identity, provider, oidc)
 	if err != nil {
 		if errors.Is(err, internal.ErrBadGateway) {
@@ -139,10 +139,6 @@ func UpdateIdentityInfoFromProvider(c RequestContext, oidc providers.OIDCClient)
 
 		if nestedErr := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByIssuedForID: identity.ID}); nestedErr != nil {
 			logging.Errorf("failed to revoke invalid user session: %s", nestedErr)
-		}
-
-		if nestedErr := data.DeleteProviderUsers(db, data.DeleteProviderUsersOptions{ByIdentityID: identity.ID, ByProviderID: provider.ID}); nestedErr != nil {
-			logging.Errorf("failed to delete provider user: %s", nestedErr)
 		}
 
 		return fmt.Errorf("sync user: %w", err)
--- a/internal/server/data/group.go
+++ b/internal/server/data/group.go
@@ -49,7 +49,6 @@ func GetGroup(db GormTxn, selectors ...SelectorFunc) (*models.Group, error) {
 
 func ListGroups(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.Group, error) {
 	groups, err := list[models.Group](db, p, selectors...)
-
 	if err != nil {
 		return nil, err
 	}
@@ -63,7 +62,6 @@ func ListGroups(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.
 	}
 
 	return groups, nil
-
 }
 
 func ByGroupMember(id uid.ID) SelectorFunc {
@@ -75,7 +73,10 @@ func ByGroupMember(id uid.ID) SelectorFunc {
 }
 
 func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
-	stmt := `SELECT DISTINCT group_id FROM identities_groups WHERE identity_id = ?`
+	stmt := `
+		SELECT DISTINCT group_id FROM identities_groups
+		WHERE identity_id = ?
+	`
 	rows, err := tx.Query(stmt, userID)
 	if err != nil {
 		return nil, err
@@ -101,7 +102,7 @@ func DeleteGroup(tx WriteTxn, id uid.ID) error {
 
 	_, err = tx.Exec(`DELETE from identities_groups WHERE group_id = ?`, id)
 	if err != nil {
-		return fmt.Errorf("remove useres from group: %w", err)
+		return fmt.Errorf("remove users from group: %w", err)
 	}
 
 	stmt := `
@@ -114,12 +115,27 @@ func DeleteGroup(tx WriteTxn, id uid.ID) error {
 	return handleError(err)
 }
 
-func AddUsersToGroup(tx WriteTxn, groupID uid.ID, idsToAdd []uid.ID) error {
-	query := querybuilder.New("INSERT INTO identities_groups(group_id, identity_id)")
+func AddUsersToGroup(tx WriteTxn, groupID uid.ID, providerGroupName string, providerID uid.ID, idsToAdd []uid.ID) error {
+	query := querybuilder.New("INSERT INTO identities_groups(group_id, identity_id, provider_id, provider_group_name)")
 	query.B("VALUES")
 	for i, id := range idsToAdd {
-		query.B("(?, ?)", groupID, id)
+		query.B("(?, ?, ?, ?)", groupID, id, providerID, providerGroupName)
 		if i+1 != len(idsToAdd) {
+			query.B(",")
+		}
+	}
+	query.B("ON CONFLICT DO NOTHING")
+
+	_, err := tx.Exec(query.String(), query.Args...)
+	return handleError(err)
+}
+
+func AddUserToGroups(tx WriteTxn, providerID uid.ID, identityID uid.ID, groups []models.Group) error {
+	query := querybuilder.New("INSERT INTO identities_groups(provider_id, identity_id, group_id, provider_group_name)")
+	query.B("VALUES")
+	for i, group := range groups {
+		query.B("(?, ?, ?, ?)", providerID, identityID, group.ID, group.Name)
+		if i+1 != len(groups) {
 			query.B(",")
 		}
 	}
@@ -138,11 +154,14 @@ func RemoveUsersFromGroup(tx WriteTxn, groupID uid.ID, idsToRemove []uid.ID) err
 	return handleError(err)
 }
 
-// TODO: do this with a join in ListGroups and GetGroup
 func CountUsersInGroup(tx GormTxn, groupID uid.ID) (int64, error) {
 	db := tx.GormDB()
 	var count int64
-	err := db.Table("identities_groups").Where("group_id = ?", groupID).Count(&count).Error
+	err := db.
+		Table("identities_groups").
+		Where("group_id = ?", groupID).
+		Distinct("identity_id").
+		Count(&count).Error
 	if err != nil {
 		return 0, err
 	}
--- a/internal/server/data/identity.go
+++ b/internal/server/data/identity.go
@@ -33,99 +33,75 @@ func (i *identitiesTable) ScanFields() []any {
 	return []any{&i.CreatedAt, &i.CreatedBy, &i.DeletedAt, &i.ID, &i.LastSeenAt, &i.Name, &i.OrganizationID, &i.UpdatedAt, &i.VerificationToken, &i.Verified}
 }
 
-func AssignIdentityToGroups(tx WriteTxn, user *models.Identity, provider *models.Provider, newGroups []string) error {
-	pu, err := GetProviderUser(tx, provider.ID, user.ID)
+func AssignUserToProviderGroups(tx GormTxn, providerUser *models.ProviderUser, provider *models.Provider, newGroupNames []string) error {
+	// before applying new groups, see which groups this user is currently known to be in for this provider
+	previousUserGroupsForProvider, err := ListProviderGroups(tx, ListProviderGroupOptions{ByProviderID: provider.ID, ByMemberIdentityID: providerUser.IdentityID})
 	if err != nil {
-		return err
+		return fmt.Errorf("current user idp groups: %w", err)
 	}
 
-	oldGroups := pu.Groups
-	groupsToBeRemoved := slice.Subtract(oldGroups, newGroups)
-	groupsToBeAdded := slice.Subtract(newGroups, oldGroups)
-
-	pu.Groups = newGroups
-	pu.LastUpdate = time.Now().UTC()
-	if err := UpdateProviderUser(tx, pu); err != nil {
-		return fmt.Errorf("save: %w", err)
+	// get all the groups this user was previously known to have at this provider for comparison to the new group names
+	oldGroupNames := []string{}
+	for _, g := range previousUserGroupsForProvider {
+		oldGroupNames = append(oldGroupNames, g.Name)
 	}
 
-	// remove user from groups
+	groupsToBeRemoved := slice.Subtract(oldGroupNames, newGroupNames)
+	groupsToBeAdded := slice.Subtract(newGroupNames, oldGroupNames)
+
 	if len(groupsToBeRemoved) > 0 {
-		stmt := `DELETE FROM identities_groups WHERE identity_id = ? AND group_id in (
-		   SELECT id FROM groups WHERE organization_id = ? AND name IN (?))`
-		if _, err := tx.Exec(stmt, user.ID, tx.OrganizationID(), groupsToBeRemoved); err != nil {
-			return err
+		if err := removeMemberFromProviderGroups(tx, providerUser, groupsToBeRemoved); err != nil {
+			return fmt.Errorf("remove previous idp groups: %w", err)
 		}
-		for _, name := range groupsToBeRemoved {
-			for i, g := range user.Groups {
-				if g.Name == name {
-					// remove from list
-					user.Groups = append(user.Groups[:i], user.Groups[i+1:]...)
+	}
+
+	if len(groupsToBeAdded) > 0 {
+		infraGroups := []models.Group{}
+		// make sure all the groups and provider groups exist, then add the membership relation
+		for _, name := range groupsToBeAdded {
+			// check if provider group with this name exists
+			_, err = GetProviderGroup(tx, provider.ID, name)
+			if err != nil {
+				if !errors.Is(err, internal.ErrNotFound) {
+					return err
+				}
+
+				// this is the first time this provider group has been seen, create it
+				providerGroup := &models.ProviderGroup{
+					ProviderID: provider.ID,
+					Name:       name,
+				}
+
+				if err := CreateProviderGroup(tx, providerGroup); err != nil {
+					return err
 				}
 			}
-		}
-	}
-
-	type idNamePair struct {
-		ID   uid.ID
-		Name string
-	}
-
-	stmt := `SELECT id, name FROM groups WHERE deleted_at is null AND name IN (?) AND organization_id = ?`
-	rows, err := tx.Query(stmt, groupsToBeAdded, tx.OrganizationID())
-	if err != nil {
-		return err
-	}
-	addIDs, err := scanRows(rows, func(item *idNamePair) []any {
-		return []any{&item.ID, &item.Name}
-	})
-	if err != nil {
-		return err
-	}
-
-	for _, name := range groupsToBeAdded {
-		// find or create group
-		var groupID uid.ID
-		found := false
-		for _, obj := range addIDs {
-			if obj.Name == name {
-				found = true
-				groupID = obj.ID
-				break
-			}
-		}
-		if !found {
-			group := &models.Group{
-				Name:              name,
-				CreatedByProvider: provider.ID,
-			}
-
-			if err = CreateGroup(tx, group); err != nil {
-				return fmt.Errorf("create group: %w", err)
-			}
-			groupID = group.ID
-		}
-
-		rows, err := tx.Query("SELECT identity_id FROM identities_groups WHERE identity_id = ? AND group_id = ?", user.ID, groupID)
-		if err != nil {
-			return err
-		}
-		ids, err := scanRows(rows, func(item *uid.ID) []any {
-			return []any{item}
-		})
-		if err != nil {
-			return err
-		}
-
-		if len(ids) == 0 {
-			// add user to group
-			_, err = tx.Exec("INSERT INTO identities_groups (identity_id, group_id) VALUES (?, ?)", user.ID, groupID)
+			// check if an infra group with this name exists
+			infraGroup, err := GetGroup(tx, ByName(name))
 			if err != nil {
-				return fmt.Errorf("insert: %w", handleError(err))
-			}
-		}
+				if !errors.Is(err, internal.ErrNotFound) {
+					return err
+				}
+				// this is the first time this group has been seen, create the infra version of this group
+				infraGroup = &models.Group{
+					Name:              name,
+					CreatedByProvider: provider.ID,
+				}
 
-		user.Groups = append(user.Groups, models.Group{Model: models.Model{ID: groupID}, Name: name})
+				if err := CreateGroup(tx, infraGroup); err != nil {
+					return err
+				}
+			}
+			infraGroups = append(infraGroups, *infraGroup)
+		}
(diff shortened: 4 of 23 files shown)


internal/access/identity.go
```diff
@@ -170,10 +170,6 @@ func UpdateIdentityInfoFromProvider(c RequestContext, oidc providers.OIDCClient)
 			logging.Errorf("failed to revoke invalid user session: %s", nestedErr)
 		}
 
-		if nestedErr := data.DeleteProviderUsers(db, data.ByIdentityID(identity.ID), data.ByProviderID(provider.ID)); nestedErr != nil {
```
BruceMacD, line 173: Removed this, on a failure the access is revoked, but we don't know that the user no longer exists. Its confusing when their group membership changes based on a failed login.

internal/server/authn/oidc.go
```diff
@@ -45,7 +45,7 @@ func (a *oidcAuthn) Authenticate(ctx context.Context, db data.GormTxn, requested
 		return AuthenticatedIdentity{}, fmt.Errorf("exhange code for tokens: %w", err)
 	}
 
-	identity, err := data.GetIdentity(db, data.Preload("Groups"), data.ByName(email))
```
BruceMacD, line 48: The identity groups loaded here weren't used

internal/server/data/schema.sql
```diff
@@ -292,6 +311,8 @@ CREATE UNIQUE INDEX idx_identities_name ON identities USING btree (organization_
 
 CREATE UNIQUE INDEX idx_identities_verified ON identities USING btree (organization_id, verification_token) WHERE (deleted_at IS NULL);
 
+CREATE UNIQUE INDEX idx_org_provider_group_names ON provider_groups USING btree (organization_id, provider_id, name);
```
BruceMacD, line 314: Any other index suggestions?

docs/dev/identity-provider-tracking.md
```diff
@@ -0,0 +1,11 @@
+# Identity Provider Tracking
+Infra tracks users and groups that exist in external identity providers and maps them access. The following database diagram shows how the relation of groups and users in identity providers are mapped to users and groups within Infra.
+
+```mermaid
+erDiagram
+          GROUP ||--|{ PROVIDER_GROUP : "composed of one or many"
+          IDENTITY ||--|{ PROVIDER_USER : "composed of one or many"
+          PROVIDER_GROUP||--|| PROVIDER : "exists in exactly one"
+          PROVIDER_GROUP o|--|{ PROVIDER_USER : "contains zero or more"
+          PROVIDER_USER||--|| PROVIDER : "exists in exactly one"
```
dnephin, line 10:
```suggestion
          PROVIDER_GROUP }o--|| PROVIDER : "exists in exactly one"
          PROVIDER_GROUP o|--|{ PROVIDER_USER : "contains zero or more"
          PROVIDER_USER }|--|| PROVIDER : "exists in exactly one"
```

I believe a provider has more 0-to-many groups, and has 1-to-many users.

internal/server/models/providergroup.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/infrahq/infra/uid"
+)
+
+// ProviderGroup is a local copy of the of a group from an identity provider
+// See this diagram for more details about how this model relates to a group
+// https://github.com/infrahq/infra/blob/main/docs/dev/identity-provider-tracking.md
+type ProviderGroup struct {
+	Model
+	OrganizationMember
```
dnephin, line 12: I think we don't need `OrganizationMember` here, similar to `ProviderUser`  and `identities_groups`. Since this belongs to a provider, the organization_id will always be the one associated with the provider_id.

Do we need the whole `Model` ? I see we are using the `ID` from model right now, but I suspect we don't need the other fields, similar to `ProviderUser`, and `identities_groups`.

I wonder if we replace the `ID` with something like `user_id` and `group_id` if we could remove some of the sub-selects and outer joins.

docs/dev/identity-provider-tracking.md
```diff
@@ -0,0 +1,11 @@
+# Identity Provider Tracking
+Infra tracks users and groups that exist in external identity providers and maps them access. The following database diagram shows how the relation of groups and users in identity providers are mapped to users and groups within Infra.
+
+```mermaid
+erDiagram
```
dnephin, line 5: There's also an identities_group relationship that is not represented in this diagram, is that right?

dnephin (commented):
Nice! I haven't looked over all of it yet, but left a couple suggestions/questions on the model and diagram

internal/server/data/group.go
```diff
@@ -41,19 +40,37 @@ func ListGroups(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.
...
 func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
-	stmt := `SELECT DISTINCT group_id FROM identities_groups WHERE identity_id = ?`
+	stmt := `
+		SELECT DISTINCT group_id FROM identities_groups
+		NATURAL FULL JOIN (
+			-- See identities in groups based on their provider group mappings
+			SELECT identity_id, group_id from provider_groups_identities
+			JOIN provider_groups_groups ON provider_groups_groups.provider_group_id = provider_groups_identities.provider_group_id
+		) AS normalized_identity_groups
+		WHERE identity_id = ?
```
dnephin, line 72: I'm hoping we can remove some of these full joins and sub-selects.

Would something like this work?

```sql
SELECT DISTINCT provider_groups.group_id
FROM provider_groups_identities
LEFT JOIN provider_groups ON provider_groups_identities.provider_group_id = provider_groups.id
WHERE provider_groups_identities.identity_id = ?
```

docs/dev/identity-provider-tracking.md
ssoroka: yeah sometimes the join tables are links in ERDs, but it might be useful to call it out. also I feel like you can't understand this system at all if you don't consider who owns what table. We own groups, identities, and identities_groups, and the providers own provider_users and provider_groups, and even the (implicit) providers_groups_users, which is an array on provider_users

internal/server/models/provideruser.go
```diff
@@ -12,7 +12,6 @@ type ProviderUser struct {
 	ProviderID uid.ID `gorm:"primaryKey"`
 
 	Email      string
-	Groups     CommaSeparatedStrings
```
ssoroka, line 15: So where is the provider_groups_users table now?

BruceMacD: It's this relation table:
```
CREATE TABLE provider_groups_identities (
    provider_group_id bigint NOT NULL,
    identity_id bigint NOT NULL
);
```

It doesn't tie directly to the provider user so that we don't have to do an extra join on membership look up.

internal/server/data/migrations.go
```diff
@@ -594,7 +596,124 @@ func setDefaultOrgID() *migrator.Migration {
...
+					CREATE TABLE IF NOT EXISTS provider_groups (
+						id bigint NOT NULL,
+						created_at timestamp with time zone,
+						updated_at timestamp with time zone,
+						deleted_at timestamp with time zone,
+						provider_id bigint NOT NULL,
+						name text NOT NULL,
+						organization_id bigint
+					);
+					CREATE TABLE IF NOT EXISTS provider_groups_identities (
```
ssoroka, line 619: is this named right? There should be no link between provider_groups and identities. it should be provider_groups_provider_users ?

internal/server/models/providergroup.go
BruceMacD: We will want the `Model` for SCIM groups, we need a way to track when provider group was created, updated, and disabled to keep it in sync with the external group when SCIM requests come in.

The `OrganizationMember` is on the provider group because it is possible to `ListProviderGroups` by name and the provider ID isn't required. This was a defense against that case. I'll just remove the ability to list by name for now instead.

internal/server/data/migrations.go
```diff
@@ -594,7 +596,124 @@ func setDefaultOrgID() *migrator.Migration {
...
+
+						if providerGroupID == 0 {
+							// get the corresponding infra group which already exists
+							var infraGroupID uid.ID
+							err := tx.QueryRow(`SELECT id FROM groups WHERE name = ? AND organization_id = ?`, name, groupProviderUser.OrgID).Scan(&infraGroupID)
+							if err != nil {
+								return err
+							}
+
+							// create the provider group, the infra group already exists
```
ssoroka, line 674: shouldn't the provider group be created regardless? it's a copy of what the provider says the groups are.

internal/server/data/migrations.go
```diff
@@ -594,7 +596,124 @@ func setDefaultOrgID() *migrator.Migration {
...
+								ON CONFLICT DO NOTHING
+								`,
+								providerGroupID, infraGroupID)
+							if err != nil {
+								return fmt.Errorf("map provider group to infra group: %w", err)
+							}
+						}
+						// link this provider user as a provider group member
+						_, err = tx.Exec(`
+							INSERT INTO provider_groups_identities ("provider_group_id", "identity_id")
```
ssoroka, line 698: yeah I definitely find this weird. this should be a link between provider_group and provider_user, so provider_groups_provider_users. If we want this user to be a member of our infra group we need them also in groups_identities.

internal/server/data/migrations.go
BruceMacD: I think this is a confusing comment, that is what is happening. I meant that the "parent" `Group` will already exist for each comma separated value on a `ProviderUser`. Compared to in the identity sync code where the `Group` might not exist.

internal/server/data/group.go
```diff
@@ -41,19 +40,37 @@ func ListGroups(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.
...
+				SELECT * FROM identities_groups
+				NATURAL FULL JOIN (
+					-- See identities in groups based on their provider group mappings
+					SELECT identity_id, group_id from provider_groups_identities
+					JOIN provider_groups_groups ON provider_groups_groups.provider_group_id = provider_groups_identities.provider_group_id
+				) AS normalized_identity_groups
+				) AS resolved_identity_groups
+			ON resolved_identity_groups.group_id = id
+			`).
+			Where("resolved_identity_groups.identity_id = ?", id)
```
ssoroka, line 60: I think we've taken a wrong turn here. this should never query anything other than identities_groups, as that's the canonical record of group membership.

internal/server/data/group.go
```diff
@@ -41,19 +40,37 @@ func ListGroups(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.
...
-	stmt := `SELECT DISTINCT group_id FROM identities_groups WHERE identity_id = ?`
+	stmt := `
+		SELECT DISTINCT group_id FROM identities_groups
+		NATURAL FULL JOIN (
+			-- See identities in groups based on their provider group mappings
+			SELECT identity_id, group_id from provider_groups_identities
+			JOIN provider_groups_groups ON provider_groups_groups.provider_group_id = provider_groups_identities.provider_group_id
+		) AS normalized_identity_groups
+		WHERE identity_id = ?
+	`
```
ssoroka, line 73: also this should not be querying provider_groups_identities. I see what you're going for here, that the group membership is the set of all manually added users plus all users added from the provider_group, but I think it's much easier (and much more performant) to simply update groups_identities when the provider_group says it should change. This keeps the complication from bleeding into our app.

internal/server/data/schema.sql
ssoroka: every field you use in a where clause needs an index or to be part of a composite index

ssoroka (changes requested):
I think we've really complicated things here and the complication isn't worth it. I really hate that the queries now have to query provider tables to build the picture of what our memberships are.

internal/server/models/provideruser.go
ssoroka: I'm not a fan of the membership lookup model you've selected here. Instead we should allow sync to modify our group membership table and keep the queries simple; leave the complication in the sync and don't pay for it throughout the app.

internal/server/data/group.go
ssoroka: I definitely want to simplify these, but even more I want to remove them completely by making the sync do the work of updating the memberships.

internal/server/data/migrations.go
ssoroka: ok I'm not sure that clarified it for me. :D

internal/server/data/schema.sql
```diff
@@ -220,6 +238,15 @@ CREATE TABLE providers (
     organization_id bigint
 );
 
+CREATE VIEW resolved_identity_groups AS
```
BruceMacD, line 241: Converted this complicated join to a view, this way our code for reading and writing remains simple. The complexity is in this single group joining view.

internal/server/models/provideruser.go
BruceMacD: Resolved the complexity in the new change by using a table view that makes queries simple (see the schema.sql). I tried resolving membership with application logic in this branch: https://github.com/infrahq/infra/commit/82a2adbea3c3412d45050b1ad5ccbc715ce4613a and it seems more error prone than the `VIEW` approach which makes managing group membership and resolving group membership decently understandable.

internal/server/data/migrations.go
BruceMacD: Fixed to provider_groups_provider_users to more accurately match the model.

internal/server/data/group.go
BruceMacD: Fixed with the Postgres view `resolved_identity_groups` is not our canonical record of group membership

internal/server/data/identity.go
```diff
@@ -1,123 +1,90 @@
...
-		// find or create group
-		var groupID uid.ID
-		found := false
-		for _, obj := range addIDs {
-			if obj.Name == name {
-				found = true
-				groupID = obj.ID
-				break
+		// check if an infra group with this name exists
+		infraGroup, err := GetGroup(tx, ByName(name))
```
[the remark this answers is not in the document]
BruceMacD: Moved this to only look-up the group if the provider group does not exist, which should help

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,249 @@
...
+		if err := rows.Scan(fields...); err != nil {
+			return nil, err
+		}
+
+		result = append(result, providerGroup)
+	}
+
+	// load group members
+	for i := range result {
+		members, err := listProviderGroupMembers(tx, &result[i])
```
ssoroka, line 165: this is going to be really really slow; is it really required? should be optional if so and ideally we should join it into the same query.

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,249 @@
...
+	return result, rows.Err()
+}
+
+// setProviderGroupMembers ...
+func setProviderGroupMembers(tx GormTxn, providerGroup *models.ProviderGroup) error {
+	if len(providerGroup.Members) == 0 {
+		return nil // nothing to do
+	}
+	// the org does not need to be set here since all provider users and groups are org specific
+	stmt := `INSERT INTO provider_groups_provider_users ("provider_id", "provider_user_identity_id", "provider_group_name") VALUES`
```
ssoroka, line 180: should have a trailing space

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,249 @@
...
+
+	val := 1
+	for _, user := range providerGroup.Members {
+		stmt += fmt.Sprintf("($%v, $%v, $%v),", val, val+1, val+2)
+		values = append(values, providerGroup.ProviderID, user.IdentityID, providerGroup.Name)
+		val += 3
+	}
+	// remove the trailing comma
+	stmt = strings.TrimSuffix(stmt, ",")
+	stmt += "ON CONFLICT DO NOTHING"
```
ssoroka, line 191:
```suggestion
	stmt += " ON CONFLICT DO NOTHING"
```

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,249 @@
...
+	_, err := tx.Exec(stmt, values...)
+	if err != nil {
+		return fmt.Errorf("set provider group members %w", err)
+	}
+
+	return nil
+}
+
+// addMemberToProviderGroups adds a link between a provider user and group that exists in a provider
+func addMemberToProviderGroups(tx GormTxn, user *models.ProviderUser, providerGroupNames []string) error {
```
ssoroka, line 202: stmt spacing might be nice here too. clearly not required to execute, but looks nicer in query logs.

ssoroka (changes requested):
I think we can do more to simplify this. There's no reason to keep two active join tables for membership; the provider memberships should update our memberships which simplifies querying and joining a ton; it's a lot easier if our core application knows nothing about providers and provider groups

internal/server/data/migrations.go
```diff
@@ -594,7 +596,137 @@ func setDefaultOrgID() *migrator.Migration {
...
+				`
+					CREATE TABLE IF NOT EXISTS provider_groups (
+						provider_id bigint NOT NULL,
+						name text NOT NULL,
+						created_at timestamp with time zone,
+						updated_at timestamp with time zone
+					);
+					CREATE TABLE IF NOT EXISTS provider_groups_provider_users (
+						provider_id bigint NOT NULL,
+						provider_group_name text NOT NULL,
```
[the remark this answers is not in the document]
BruceMacD: We'll need this one for tracking which provider group the users are in, the unique identifier is `group name and provider id`

internal/server/data/migrations.go
```diff
@@ -594,7 +596,137 @@ func setDefaultOrgID() *migrator.Migration {
...
+						provider_user_identity_id bigint NOT NULL
+					);
+					CREATE TABLE IF NOT EXISTS provider_groups_groups (
```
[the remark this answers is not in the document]
BruceMacD: This one is so we can map one provider group to 0-many Infra groups in the future rather than naively by name like we do at the moment

internal/server/data/migrations.go
```diff
@@ -594,7 +596,137 @@ func setDefaultOrgID() *migrator.Migration {
...
+}
+
+func addProviderGroups() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-09-07T17:05",
+		Migrate: func(tx migrator.DB) error {
+			// setup the new tables
+			_, err := tx.Exec(
+				`
+					CREATE TABLE IF NOT EXISTS provider_groups (
```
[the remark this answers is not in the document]
BruceMacD: added this

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,249 @@
...
+	table := &providerGroupTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(table))
+	query.B("FROM")
+	query.B(table.Table())
+
+	if opts.ByMemberIdentityID != 0 {
+		query.B(`
+			JOIN provider_groups_provider_users 
+			ON provider_groups.name = provider_groups_provider_users.provider_group_name
```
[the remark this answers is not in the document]
BruceMacD: Need it in this case to link the group itself rather than all groups for that provider

internal/server/data/migrations.go
```diff
@@ -594,7 +596,137 @@ func setDefaultOrgID() *migrator.Migration {
...
+					
+					-- create a resolved_identity_groups view that can be used to check if a user is a direct member of a group,
+					-- or a user is a member of an external identity provider group that is mapped to this group in Infra
+					CREATE VIEW resolved_identity_groups AS
+						SELECT DISTINCT * FROM identities_groups
+						NATURAL FULL JOIN (
+							-- Add identities to groups based on their provider group mappings
+							SELECT provider_user_identity_id as identity_id, provider_groups_groups.group_id FROM provider_groups_provider_users
+							JOIN provider_groups_groups ON provider_groups_groups.provider_id = provider_groups_provider_users.provider_id AND provider_groups_groups.provider_group_name = provider_groups_provider_users.provider_group_name
+						) AS normalized_identity_groups;
```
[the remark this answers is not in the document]
dnephin: > Views have to be dropped and recreated any time anything about the underlying tables changes

This was surprising to me. Do you know where I can learn about about that?

My expectation is that we would only need to change the view if the relationship the builds the  view changes. That's true in the case of doing the sync in-code as well, and is likely more involved than a simple `CREATE OR REPLACE VIEW`.

internal/server/data/migrations.go
```diff
@@ -655,3 +657,135 @@ func cleanCrossOrgGroupMemberships() *migrator.Migration {
...
+						group_id bigint NOT NULL
+					);
+					CREATE UNIQUE INDEX idx_provider_group_names ON provider_groups(organization_id, provider_id, name);
+					CREATE UNIQUE INDEX idx_provider_groups_provider_users ON provider_groups_provider_users(provider_id, provider_group_name, provider_user_identity_id);
+					CREATE UNIQUE INDEX idx_provider_groups_groups ON provider_groups_groups(provider_id, provider_group_name, group_id);
+					
+					-- create a resolved_identity_groups view that can be used to check if a user is a direct member of a group,
+					-- or a user is a member of an external identity provider group that is mapped to this group in Infra
+					CREATE VIEW resolved_identity_groups AS
+						SELECT DISTINCT * FROM identities_groups
```
dnephin, line 692: I did not realize we would still have an `identities_group` table. I'm assuming this is for the infra provider, is that right?

Do you know if it would be possible to manage the infra provider like we do other providers? Would that be a large change?

I suspect the full join here could be expensive, so if we have an approach to remove it (either in this PR or a follow up) that might be worth exploring.

dnephin (commented):
I think this is looking good! I was surprised that `identities_groups` was still around after these changes. I left a question about that below.

I'll continue to review tomorrow.

internal/access/identity.go
dnephin, line 173: So previously we assume that if a sync fails then the user was removed from that IDP. Now we only invalid their access keys but don't assume the user was removed. Is that right? 

Now the only way to remove this relationship is when we delete the entire provider? 

I thought we needed some way to remove them on sync because they were removed from the provider.

internal/server/models/providergroup.go
```diff
@@ -0,0 +1,22 @@
...
+type ProviderGroup struct {
+	OrganizationMember
+	CreatedAt time.Time
+	UpdatedAt time.Time
+
+	ProviderID uid.ID
+	Name       string
+
+	// for loading, not for saving
+	Members []ProviderUser
```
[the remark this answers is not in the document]
BruceMacD: Removed this in the new commit because it wasnt used anywhere

internal/server/data/migrations.go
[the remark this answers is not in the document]
BruceMacD: Removed this view in the new commit, instead adding provider_group_name and provider_id to `identities_groups` relations directly.

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return nil, handleReadError(err)
+	}
+	pg := (*models.ProviderGroup)(providerGroup)
+
+	members, err := listProviderGroupMembers(tx, pg)
+	if err != nil {
+		return nil, fmt.Errorf("retrieve provider group members: %w", err)
+	}
+	pg.Members = members
```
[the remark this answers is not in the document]
BruceMacD: Removed this loading in the new commit where it wasnt used anywhere

internal/server/data/providergroup.go
BruceMacD: Removed in the new changes

internal/server/data/providergroup.go
BruceMacD: Im not sure what you mean by this one

internal/server/data/migrations.go
BruceMacD: I took another look at this and ended up adding all group relations to the `identites_groups` table. The Infra groups get added as "from the Infra provider".

internal/access/identity.go
BruceMacD: Its more likely that their session expired. With SCIM coming it makes more sense to rely on that for a user being deleted, rather than removing them when their session expires.

Its also still possible to remove them by deleting the user (rather than the whole provider). Their session becomes invalid upon being deleted from the provider, and they can no longer login.

internal/server/data/providergroup.go
```diff
@@ -0,0 +1,252 @@
...
+// addMemberToProviderGroups adds a link between a provider user and group that exists in a provider
+func addMemberToProviderGroups(tx GormTxn, user *models.ProviderUser, providerGroupNames []string) error {
+	// the org does not need to be set here since all provider users and groups are org specific
+	stmt := `INSERT INTO provider_groups_provider_users ("provider_id", "provider_user_identity_id", "provider_group_name") VALUES`
+	values := []interface{}{}
+
+	val := 1
+	for _, grpName := range providerGroupNames {
+		stmt += fmt.Sprintf("($%v, $%v, $%v),", val, val+1, val+2)
+		values = append(values, user.ProviderID, user.IdentityID, grpName)
```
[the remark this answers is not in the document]
BruceMacD: The query builder is nicer, updated

internal/server/data/group.go
```diff
@@ -114,11 +115,11 @@ func DeleteGroup(tx WriteTxn, id uid.ID) error {
 	return handleError(err)
 }
 
-func AddUsersToGroup(tx WriteTxn, groupID uid.ID, idsToAdd []uid.ID) error {
-	query := querybuilder.New("INSERT INTO identities_groups(group_id, identity_id)")
+func AddUsersToGroup(tx WriteTxn, groupID uid.ID, providerGroupName string, providerID uid.ID, idsToAdd []uid.ID) error {
+	query := querybuilder.New("INSERT INTO identities_groups(group_id, identity_id, provider_id, provider_group_name)")
```
ssoroka, line 119: do we need something to ignore duplicates?

internal/server/data/group.go
```diff
@@ -138,11 +154,13 @@ func RemoveUsersFromGroup(tx WriteTxn, groupID uid.ID, idsToRemove []uid.ID) err
...
 }
 
-// TODO: do this with a join in ListGroups and GetGroup
 func CountUsersInGroup(tx GormTxn, groupID uid.ID) (int64, error) {
 	db := tx.GormDB()
 	var count int64
-	err := db.Table("identities_groups").Where("group_id = ?", groupID).Count(&count).Error
+	err := db.
+		Table("identities_groups").
+		Where("group_id = ?", groupID).
```
ssoroka, line 162: I think we need to count distinct users now.. so `count(distinct user_id)` should work.

internal/server/data/migrations.go
```diff
@@ -679,3 +681,245 @@ func cleanCrossOrgGroupMemberships() *migrator.Migration {
...
+							groupProviderUser.ProviderID, name, groupProviderUser.IdentityID)
+						if err != nil {
+							return fmt.Errorf("add provider group member: %w", err)
+						}
+
+						// find the relation that was previously created for this user's group relation
+						var existing uid.ID
+						err = tx.QueryRow(`
+							SELECT group_id FROM identities_groups 
+							WHERE identity_id = ? AND group_id = ? AND provider_id IS NULL`,
```
ssoroka, line 815: ignore deleted?

internal/server/data/providergroup.go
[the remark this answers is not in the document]
ssoroka: should it join on id instead? names could be problematic if they're not indexed

internal/server/data/group.go
BruceMacD: There's an `ON CONFLICT DO NOTHING` in the query builder here that should handle this

internal/server/data/group.go
```diff
@@ -75,7 +73,10 @@ func ByGroupMember(id uid.ID) SelectorFunc {
 }
 
 func groupIDsForUser(tx ReadTxn, userID uid.ID) ([]uid.ID, error) {
-	stmt := `SELECT DISTINCT group_id FROM identities_groups WHERE identity_id = ?`
+	stmt := `
+		SELECT group_id FROM identities_groups
+		WHERE identity_id = ?
+	`
```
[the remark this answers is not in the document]
BruceMacD: thanks, fixed this

internal/server/data/migrations.go
BruceMacD: This one didn't need deleted, but some of the other migration queries did when I went back to look at this, thanks

PR #3220, opened 2022-09-15
Remove secret storage from connector

dnephin:
## Summary

This PR removes `infrahq/secrets` from the connector. The connector only ever used secrets from `env`, `file`, and `plaintext`, all of which we can read directly without needing to add an `env:`, `file:`, or `plaintext:` prefix to the value. 

This PR preserves backwards compatibility by removing any `env:` or `file:` prefixes from values before looking for the file or env var.

Also:
* changes the helm chart to create a configmap without the `file:` prefix on the access key value.
* adds support for reading the access key from `INFRA_ACCESS_KEY` env var in the connector. This matches the CLI. Unlikely anyone will use this yet though, since everyone is using the helm chart, not running the connector directly.
* adds support for reading the connector proxy CA and key from literals, in addition to files. This will be useful at least for writing integration tests of the connector, but may also be useful to others deploying the connector in different envrionments.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -295,9 +295,29 @@ func newConnectorCmd() *cobra.Command {
 				return err
 			}
 
+			// backwards compat for old access key values with a prefix
+			accessKey := options.Server.AccessKey.String()
+			switch {
+			case strings.HasPrefix(accessKey, "file:"):
+				filename := strings.TrimPrefix(accessKey, "file:")
+				if err := options.Server.AccessKey.Set(filename); err != nil {
+					return err
+				}
+				logging.L.Warn().Msg("accessKey with 'file:' prefix is deprecated. Use the filename without the file: prefix instead.")
+			case strings.HasPrefix(accessKey, "env:"):
+				key := strings.TrimPrefix(accessKey, "env:")
+				options.Server.AccessKey = types.StringOrFile(os.Getenv(key))
+				logging.L.Warn().Msg("accessKey with 'env:' prefix is deprecated. Use the INFRA_ACCESS_KEY env var instead.")
+			case strings.HasPrefix(accessKey, "plaintext:"):
+				options.Server.AccessKey = types.StringOrFile(strings.TrimPrefix(accessKey, "plaintext:"))
+				logging.L.Warn().Msg("accessKey with 'plaintext:' prefix is deprecated. Use the literal value without a prefix.")
+			}
+
 			// Also accept the same env var as the CLI for setting the access key
 			if accessKey, ok := os.LookupEnv("INFRA_ACCESS_KEY"); ok {
-				options.Server.AccessKey = types.StringOrFile(accessKey)
+				if err := options.Server.AccessKey.Set(accessKey); err != nil {
+					return err
+				}
 			}
 			return runConnector(cmd.Context(), options)
 		},
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -21,6 +21,7 @@ import (
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/cmd/cliopts"
+	"github.com/infrahq/infra/internal/cmd/types"
 	"github.com/infrahq/infra/internal/connector"
 	"github.com/infrahq/infra/internal/logging"
 )
@@ -294,6 +295,10 @@ func newConnectorCmd() *cobra.Command {
 				return err
 			}
 
+			// Also accept the same env var as the CLI for setting the access key
+			if accessKey, ok := os.LookupEnv("INFRA_ACCESS_KEY"); ok {
+				options.Server.AccessKey = types.StringOrFile(accessKey)
+			}
 			return runConnector(cmd.Context(), options)
 		},
 	}
--- a/internal/cmd/types/file.go
+++ b/internal/cmd/types/file.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"errors"
 	"io"
 	"os"
 
@@ -22,10 +23,17 @@ func (s *StringOrFile) String() string {
 
 func (s *StringOrFile) Set(raw string) error {
 	fh, err := os.Open(raw)
-	if err != nil {
-		logging.L.Debug().Err(err).Msg("failed to open file")
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		// Only log a small prefix of the value, at trace level, in case the value is sensitive.
+		logging.L.Trace().
+			Str("valuePrefix", raw[:len(raw)/2]).
+			Msg("value does not appear to be a file, assuming string literal")
+
 		*s = StringOrFile(raw)
 		return nil
+	case err != nil:
+		return err
 	}
 
 	content, err := io.ReadAll(fh)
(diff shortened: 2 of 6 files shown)


helm/charts/infra/templates/connector/secret.yaml
```diff
@@ -1,7 +1,5 @@
 
 {{- if include "connector.enabled" . | eq "true" }}
-{{- $accessKey := default "" .Values.connector.config.accessKey -}}
-{{- if or (not $accessKey) (and (not (hasPrefix "file:" $accessKey)) (not (hasPrefix "env:" $accessKey))) }}
```
dnephin, line 4: I believe we want to create this secret in all cases now

mxyng: This ignores the case when a user creates the access key secret out of band and want to load an existing secret rather than creating on through Helm. One way of addressing that could be to use a different, e.g. `accessKeySecret`.

internal/cmd/cmd.go
```diff
@@ -294,6 +295,10 @@ func newConnectorCmd() *cobra.Command {
 				return err
 			}
 
+			// Also accept the same env var as the CLI for setting the access key
+			if accessKey, ok := os.LookupEnv("INFRA_ACCESS_KEY"); ok {
+				options.Server.AccessKey = types.StringOrFile(accessKey)
```
mxyng, line 300: If the access key is passed through environment variables, it should contain the actual access key and not a file reference. `StringOrFile` is unnecessary

helm/charts/infra/templates/connector/secret.yaml
mxyng: We also need to consider the migration for users who may be using `env:` or `file:` vallues. In both cases, the access key secret should not be rendered. `NOTES.txt` should also be updated to warn users about the configuration change.

mxyng (commented):
This is a breaking change. Since we're already talking about splitting connector from the chart, I suggest waiting until after the split before merging

internal/cmd/cmd.go
dnephin: Ya, in this case it is unnecessary, but it has to be used here because that's the type of the `AccessKey` field.

The field is that type because in other scenarios we do load it from a file.

helm/charts/infra/templates/connector/secret.yaml
dnephin: > This ignores the case when a user creates the access key secret out of band and want to load an existing secret rather than creating on through Helm

It would be really great to start simplifying the helm charts and being a lot more opinionated about what we support.

Using an existing secret seems like a reasonable thing to support, but do we really need to allow them to call it anything? If we require them to name the secret according to our convention, then I think this code still works, because we look up the value from the existing secret.

Is there a use case that requires us to support an arbitrarily named secret?

mxyng: The chart should not modify anything it does not own. In the case user provides an existing secret for the access key, the secret template should see it as an externally configured secret and not produce nothing.

Failing that, the best case scenario would to fail cause Helm cannot modify the resources it does not own.

The worst case is Helm takes ownership of the resource and overrides the contents with its own values. The is problematic because we don't know the origin of the resource. It could be created by something like `external-secrets` where the real value lives in some external secret store. 

Taking ownership also means the resource will subject it to its lifecycle rules. This usually means deleting the resource when the chart is uninstalled. One of the reasons why users may want to provide their own Kubernetes secret is to keep it away from Helm's lifecycle management.

dnephin: We chatted offline and I think we have a good way to solve this now:

1. If the access key is specified from the command line (`--set connector.accessKey=<acessKeyValue>`) or from a `values.yaml` 
   - create the kubernetes secret
   - set it in the deployment using `envFrom` (instead of from the configmap like we do now)
2. If the access key is not specified, we can assume the value is coming from anexisting kubernetes secret that is not managed by our chart
   - the operator will have to specify the `envFrom` using a `values.yaml`
3. If the operator is using a CSI driver to specify a secret (ex: [aws secrets manager](https://docs.aws.amazon.com/secretsmanager/latest/userguide/integrating_csi_driver.html), they will need to set a `volume`, `volumeMount`, and an env var like `INFRA_ACCESS_KEY=/path/to/secret/mount/file`

We're going to wait until we split out the helm charts for the server and connector to make this change.

dnephin (commented):
Since this change is a breaking change, I think I'm going to split it up.  I'll add some backwards compatibility to the connector to strip the `env:` and `file:` prefixes.

When we create a new helm chart for the connector we'll remove that support, and that will allow us to remove the backwards compat shim that is added here.

ssoroka:
this is probably fine, but I'm curious why this change is important?

dnephin:
Fair question! 

I think it's important for two reasons:
1. with SaaS users are still deploying the connector. They can easily be confused by the need to specify something as `env:` or `file:`. They shouldn't be required to learn that micro-format, because we have a way to accept the config from all of those sources without the prefix.
2. this setup allowed the bug fixed by #3207 to exist and not be obvious from the code. Passing around a `string` that has to be resolved is prone to errors.  I think if we want to do defered resolving of secrets we need to use a different type, so that it's obvious that you have to resolve it before using it. The fact that plaintext works without being resolved makes it even more prone to bugs, because a test will easily pass when using a plaintext value.

I think we need to fix these problems in the `server` code as well, but it's easier to do it here first.

internal/cmd/cmd.go
dnephin: Change this to use `Set` on the StringOrFile value, so that we can specify the name of a file in the env var as well. As we discussed we'll want this for case 3 in https://github.com/infrahq/infra/pull/3220#discussion_r975843443

helm/charts/infra/templates/connector/secret.yaml
dnephin: I've updated this PR to make the change backwards compatible, and removed most of the helm chart changes.

dnephin:
Now that this PR should be backwards compatible, I've marked it ready for review.

internal/cmd/types/file.go
```diff
@@ -22,10 +23,24 @@ func (s *StringOrFile) String() string {
...
-	if err != nil {
-		logging.L.Debug().Err(err).Msg("failed to open file")
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		limit := 10
+		switch {
+		case len(raw) <= 5:
+			limit = 0
+		case len(raw) <= 10:
+			limit = 5
```
ssoroka, line 33: maybe always show 50% of the characters?

ssoroka (approved):

internal/cmd/types/file.go
dnephin: Ya, that seems easier, and should be fine. I'll do that

internal/cmd/cmd.go
```diff
@@ -294,6 +295,26 @@ func newConnectorCmd() *cobra.Command {
...
+				filename := strings.TrimPrefix(options.Server.AccessKey.String(), "file:")
+				if err := options.Server.AccessKey.Set(filename); err != nil {
+					return err
+				}
+				logging.L.Warn().Msg("accessKey with 'file:' prefix is deprecated. Use the filename without the file: prefix instead.")
+			case strings.HasPrefix(options.Server.AccessKey.String(), "env:"):
+				key := strings.TrimPrefix(options.Server.AccessKey.String(), "env:")
+				options.Server.AccessKey = types.StringOrFile(os.Getenv(key))
+				logging.L.Warn().Msg("accessKey with 'env:' prefix is deprecated. Use the INFRA_ACCESS_KEY env var instead.")
+			}
```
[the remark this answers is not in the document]
dnephin: Good catch, this should trim the `plaintext:` prefix if it finds it

PR #3221, opened 2022-09-15
Fix `infra list` with destinations named with a prefix of infra

dnephin:
## Summary

This bug was causing `infra list` to hide access to destinations named with `infra...` prefix.  Also fixes a similar bug related to filtering of destination names that may have displayed some grants for a destination that did not exist , if there was another destination where the name was a prefix of the other.

Also refactor the logic a bit to remove one of the iterations. We can build the list of resource names when we find a resource that is not yet in the map.

### DIFF
--- a/internal/cmd/list.go
+++ b/internal/cmd/list.go
@@ -59,7 +59,7 @@ func list(cli *CLI) error {
 
 	keys := make([]string, 0, len(gs))
 	for k := range gs {
-		if strings.HasPrefix(k, "infra") {
+		if isInfraDestination(k) {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


internal/cmd/list.go
```diff
@@ -107,6 +92,19 @@ func list(cli *CLI) error {
...
+func destinationForResourceExists(resource string, destinations []api.Destination) bool {
+	for _, d := range destinations {
+		if isResourceForDestination(resource, d.Name) {
+			return true
+		}
+	}
+	return false
+}
+
+func isResourceForDestination(resource string, destination string) bool {
```
dnephin, line 104: If we change our `api.Grant` type to have a `Destination` field, then I think we can remove this function and compare the names directly.  

I think we need to make a similar change to `ListGrants` to allow querying for all the grants of a destination at once (instead of querying for each namespace separately).

jmorganca (approved):

pdevine:
@dnephin Didn't get a chance to comment on this before the merge, but I think there's still an issue if someone names their cluster `infra.<whatever>`.

PR #3222, opened 2022-09-15
Add a bit more validation on create and update

dnephin:
## Summary

This validation is only to catch programming errors. A user should never see these errors, they are exclusively safeguards to try and make bugs more obvious in tests.

We already have this same validation on a bunch of types (Grants and ProviderUser both have many checks). This PR adds a few more to other types.

### DIFF
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -41,20 +41,21 @@ func secretChecksum(secret string) []byte {
 	return chksm[:]
 }
 
-func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err error) {
+func validateAccessKey(accessKey *models.AccessKey) error {
 	switch {
 	case accessKey.IssuedFor == 0:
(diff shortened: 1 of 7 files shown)


internal/server/data/access_key.go
```diff
@@ -90,6 +90,10 @@ func CreateAccessKey(db GormTxn, accessKey *models.AccessKey) (body string, err
 		accessKey.Name = fmt.Sprintf("%s-%s", identityIssuedFor.Name, accessKey.ID.String())
 	}
 
+	if err := validateAccessKey(accessKey); err != nil {
```
mxyng, line 93: Why move this to later? If the key ID is not valid, it should be rejected immediately?

BruceMacD: There's a bunch of things that can optionally be generated for the key before this point, so I think this spot is right for validation if want to keep it contained in one place

BruceMacD (approved):

PR #3231, opened 2022-09-16
feat: add per-organization rate limits

mxyng:
## Summary

Add limits to how many requests can be sent to Infra server on a per-organization, per-minute basis. Only enabled if Redis is configured


internal/server/cache/cache.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/go-redis/redis/v8"
+	rate "github.com/go-redis/redis_rate/v9"
+
+	"github.com/infrahq/infra/internal/logging"
+)
+
+type Cache struct {
+	client *redis.Client
+}
+
```
ssoroka, line 16: why is something that does rate-limiting called cache?

ssoroka (commented):
Looks ok, but I might just call cache redis, considering we don't use it for a cache now, and we'll probably use it for things other than a cache in the future.

internal/server/cache/cache.go
mxyng: That's a fair point. I was thinking to generalize Redis as "cache" but you're correct that Redis is more than a cache.

internal/server/cache/cache.go
```diff
@@ -0,0 +1,67 @@
...
+	Username string
+	Password string
+	Options string
+}
+
+func NewCache(options Options) *Cache {
+	var client *redis.Client
+
+	if len(options.Host) > 0 {
+		redisOptions, err := redis.ParseURL(fmt.Sprintf("redis://%s:%d?%s", options.Host, options.Port, options.Options))
```
dnephin, line 29: Could we accept the url as a single string, instead of separate fields for each? Fewer configuration options makes the server config a little bit easier to document.

internal/server/cache/cache.go
```diff
@@ -0,0 +1,67 @@
...
+		if err != nil {
+			logging.Warnf("invalid cache options: %#v", options)
+			return nil
+		}
```
dnephin, line 33: Only logging a warning seems like the wrong behaviour here. If the operator specifies any of these options then they are expecting redis to be configured. I think we should return the error.

```suggestion
		if err != nil {
			return nil, fmt.Errorf("invalid redis configuration: %w", err)
		}
```

internal/server/cache/cache.go
```diff
@@ -0,0 +1,67 @@
...
+			panic(err)
+		}
+
+		logging.L.Debug().
+			Int("allowed", result.Allowed).
+			Int("remaining", result.Remaining).
+			Dur("retry_after", result.RetryAfter).
+			Msg("")
+		// TODO: also return result.Remaining and result.RetryAfter so the response headers can be set
+		return result.Allowed > 0
```
dnephin, line 63: Instead of a `bool` return value and using `internal.ErrTooManyRequests` from the caller, I think a structured error would allow you to return all of this data.

`internal/validate.Error`, and `data.UniqueConstraintError` are good example of this.

Ex:

```go
type RateLimitExceeded struct {
    RetryAfter time.Time
    Description string
}
```

Although I guess that does not work well for a `Remaining` header. I'm not sure if we should expose that right away. Knowing when to retry the request is pretty important, but exposing "how close you are to the limit" maybe less so?

internal/cmd/server.go
```diff
@@ -99,6 +100,10 @@ func defaultServerOptions(infraDir string) server.Options {
 			HTTPS:   ":443",
 			Metrics: ":9090",
 		},
+
+		Cache: cache.Options{
+			Port: 6379,
+		},
```
dnephin, line 106: We probably want to call this configuration `redis` as well, since redis may be used for different things, and at least initially is not used at all for caching.

dnephin (commented):
Looks good!

internal/server/cache/cache.go
mxyng: I generally dislike this kind of configuration as a single string value. It requires the operator to know the connection string format. This format is usually harder to communicate without referencing upstream documentation, in this case https://pkg.go.dev/github.com/go-redis/redis/v8#ParseURL. IMO, for a new-ish operator, it's easier to understand individual `Host`, `Port`, `Username`, `Password` fields than the full connection string.

Additionally infrastructure configurations don't usually come in a neat package: hostname comes from one source, port from another, username/password are user-defined, options are optional and user-specific. This means the user needs to gather all necessary information and compose a connection string from the results. It also means individual fields cannot be updated without changing the entire string which can lead to mistakes and misconfiguration.

I can add the single string value as an alternative but it's not something I would ever use.

internal/server/cache/cache.go
mxyng: That was the intent behind the TODO but right now I'm not sure if this improvement will be part of this PR or a separate PR

internal/server/cache/cache.go
mxyng: You're probably right. A misconfiguration should fail fast.

internal/cmd/server.go
mxyng: This has been updated locally after @ssoroka's comment

internal/server/cache/cache.go
dnephin: Ok, that makes sense. I generally prefer a single string, but I also have very little experience with terraform, so I haven't run into those problems. I don't think there's any reason to have two ways to do this right now, so let's keep it as separate fields.

internal/server/redis/redis.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(options.Host) > 0 {
+		redisOptions, err := redis.ParseURL(fmt.Sprintf("redis://%s:%d?%s", options.Host, options.Port, options.Options))
+		if err != nil {
+			return nil, fmt.Errorf("invalid redis options: %v", options)
+		}
+
+		redisOptions.Username = options.Username
+		redisOptions.Password = options.Password
+
+		client = redis.NewClient(redisOptions)
```
pdevine, line 33: Do you need to use the cluster client here? I'm assuming we would want some kind of redundancy.

internal/server/routes.go
```diff
@@ -215,12 +216,20 @@ func wrapRoute[Req, Res any](a *API, routeID routeIdentifier, route route[Req, R
...
 		if !route.noOrgRequired {
-			if org := getRequestContext(c).Authenticated.Organization; org == nil {
+			if org == nil {
 				return internal.ErrBadRequest
 			}
 		}
 
+		if org != nil {
+			// TODO: limit should be a per-organization setting
+			if err := redis.RateOK(a.server.redis, org.ID.String(), 1000); err != nil {
```
pdevine, line 228: Are we going to hit this fast? 1000 requests/min is ~ 16 requests per second, and I could see a mid/largish org hitting that fairly quickly.

mxyng: That's a good point. Raising it to 3000 for now

internal/server/redis/redis.go
mxyng: It depends how the Redis cluster is configured. Managed services like GCP's MemoryStore and AWS's ElastiCache can implement redundancy without resorting to Redis Clusters. Implementing other types of redundancy will require more complex configurations such as sentinels or cluster configurations.

For the time being, we can get away with implementing just the basic configuration

internal/server/redis/limit.go
```diff
@@ -0,0 +1,121 @@
...
+			pipe.Del(ctx, loginKeyLockout(key))
+			return nil
+		})
+	}
+}
+
+func LoginBad(r *Redis, key string, limit int) {
+	if r != nil && r.client != nil {
+		ctx := context.TODO()
+		rate := r.client.Incr(ctx, loginKey(key)).Val()
```
dnephin, line 103: We are silently ignoring errors here, and in `r.client.Pipelined` above, and in `client.SetArgs` below.  The return value is a `StatusCmd` so the linter doesn't notice that the error is being ignored.

I suspect we should at least log those errors

internal/server/redis/limit_test.go
```diff
@@ -0,0 +1,159 @@
...
+		assert.DeepEqual(t, err, OverLimitError{
+			RetryAfter: expected,
+		}, opt.DurationWithThreshold(time.Second))
+
+		LoginGood(redis, "admin@example.com")
+		err = LoginOK(redis, "admin@example.com")
+		assert.NilError(t, err)
+	})
+
+	t.Run("over limit reset after lockout period", func(t *testing.T) {
```
dnephin, line 141: Is there a test for "over limit failed again after lockout period" ?

I'm curious about what happens if they try again after the lockout period and fail the first time. Do they immediately get locked out for a very long time, or do they get another 9 attempts before the very long lockout?

internal/server/handlers.go
```diff
@@ -175,11 +176,29 @@ func wrapLinkWithVerification(link, domain, verificationToken string) string {
...
 	rCtx := getRequestContext(c)
 
+	var onSuccess, onFailure func()
+
 	var loginMethod authn.LoginMethod
 	switch {
 	case r.AccessKey != "":
 		loginMethod = authn.NewKeyExchangeAuthentication(r.AccessKey)
 	case r.PasswordCredentials != nil:
+		if err := redis.RateOK(a.server.redis, r.PasswordCredentials.Name, 10); err != nil {
```
dnephin, line 186: I wonder if these `r.PasswordCredentials.Name` should be org scoped? I guess they are email addresses so it's not super critical, but it might be easy to miss that if we ever change the name to allow them to not be org scoped.

internal/server/routes.go
```diff
@@ -215,12 +216,20 @@ func wrapRoute[Req, Res any](a *API, routeID routeIdentifier, route route[Req, R
...
 		if !route.noOrgRequired {
-			if org := getRequestContext(c).Authenticated.Organization; org == nil {
+			if org == nil {
 				return internal.ErrBadRequest
 			}
 		}
 
+		if org != nil {
+			// TODO: limit should be a per-organization setting
+			if err := redis.RateOK(a.server.redis, org.ID.String(), 3000); err != nil {
```
dnephin, line 228: Is this TODO still relevant?

dnephin (approved):
Looks great! A few questions, and a few minor suggestions, but nothing blocking.

It'd be great for us to dig into github.com/go-redis/redis_rate/v9 a bit more as well, and make sure we understanding all the implementation trade-offs made in that library, since it seems we are relying it quite a bit.

I guess our existing API metrics should give us some indication of the number of requests hitting our rate limits, right? It might be interesting to add some more fine grained metrics about exactly which of the rate limits is rejecting the request, but seems like that should be a follow up and not part of this (if we want it).

The limits we have set seem reasonable. We might consider increasing the org limit one to 5000 or 6000 to be extra safe.

internal/server/redis/redis_test.go
```diff
@@ -0,0 +1,113 @@
+package redis
+
+import (
+	"context"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/alicebob/miniredis/v2"
```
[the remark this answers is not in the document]
mxyng: The main advantage is speed and not needing to start a redis instance in order to run tests. This makes writing tests easy and fast. 

I haven't used `miniredis` previous but @pdevine has and he seems happy with it. I've also used Python's `redislite` previously for unit tests and it's a similar idea. I'm not that concerned with compatibility since the Redis API itself is relatively simple and we're not using any advanced features (yet). We can move on to an external Redis if it ever becomes a problem.

internal/server/redis/limit_test.go
mxyng: I'll add that test case. They should be locked out for a very long time.

internal/server/routes.go
mxyng: Yes it's still relevant. The rate limit may eventually be different per org so this will need to load the value from settings

internal/server/handlers.go
mxyng: Good point. I'll add that to the PR

ssoroka (approved):
lgtm. One thing I think is common to run into is the need to scale the limits by org size. Though we can tackle that later.

mxyng:
> lgtm. One thing I think is common to run into is the need to scale the limits by org size. Though we can tackle that later.

Yes definitely. That's the intent behind the TODO for per-organization limits

PR #3271, opened 2022-09-19
fix: allow users to issue access keys to themselves

ssoroka:
## Summary

fix: allow users to issue access keys to themselves

## Checklist

### DIFF
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -11,14 +11,12 @@ import (
 )
 
 func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired bool, p *data.Pagination) ([]models.AccessKey, error) {
-	var db data.GormTxn
-	var err error
-
-	if identityID == GetRequestContext(c).Authenticated.User.ID {
-		db = getDB(c)
+	rCtx := GetRequestContext(c)
+	if identityID == rCtx.Authenticated.User.ID {
+		// can list own keys
 	} else {
 		roles := []string{models.InfraAdminRole, models.InfraViewRole}
-		db, err = RequireInfraRole(c, roles...)
+		_, err := RequireInfraRole(c, roles...)
 		if err != nil {
 			return nil, HandleAuthErr(err, "access keys", "list", roles...)
 		}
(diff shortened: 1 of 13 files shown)


internal/access/access_key.go
```diff
@@ -27,9 +34,15 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired
...
-	if err != nil {
-		return "", HandleAuthErr(err, "access key", "create", models.InfraAdminRole)
+	var db data.GormTxn
+
+	if accessKey.IssuedFor == GetRequestContext(c).Authenticated.User.ID {
+		db = getDB(c) // can create access keys for yourself.
+	} else {
+		db, err = RequireInfraRole(c, models.InfraAdminRole)
+		if err != nil {
+			return "", HandleAuthErr(err, "access key", "create", models.InfraAdminRole)
```
BruceMacD, line 44:
```suggestion
			return "", HandleAuthErr(err, "access keys", "create", models.InfraAdminRole)
```
Nit: Not related to this, change.

BruceMacD (commented):
Logic all looks good, but I'd like to see a test around `access/access_key.go` that asserts a user can modify their own access key without a role

internal/access/access_key.go
dnephin: Right now I think we always use the singular form for `create` since it's creating a single key, and we use the plural for `list`, since it generally returns multiple.  I think get and delete likely use the singular as well.

I think it's fine if we want to change everything to plural, but maybe we should do that everywhere in a separate PR?

dnephin (commented):
Should users also be able to delete their own access keys? I think generally that's a thing that is allowed

internal/access/access_key.go
```diff
@@ -27,9 +34,15 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired
 }
 
 func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string, err error) {
-	db, err := RequireInfraRole(c, models.InfraAdminRole)
-	if err != nil {
-		return "", HandleAuthErr(err, "access key", "create", models.InfraAdminRole)
+	var db data.GormTxn
+
+	if accessKey.IssuedFor == GetRequestContext(c).Authenticated.User.ID {
+		db = getDB(c) // can create access keys for yourself.
```
jmorganca, line 40: Could a user use an access key to create others indefinitely (effectively nulling the point of having access keys expire)? Just a thought to explore. Perhaps we should limit this somehow to the "first" access key created by the user's user/pass or IDP login?

ssoroka: Good call, I added a check for that. There's some additional conditions to explore here, like we might want the option to not let access keys create other access keys in general. We'll have to move to a more fine-grained permission model for that (consuming our own grant system)

internal/access/access_key.go
```diff
@@ -23,20 +29,31 @@ func ListAccessKeys(c *gin.Context, identityID uid.ID, name string, showExpired
...
 
 func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string, err error) {
-	db, err := RequireInfraRole(c, models.InfraAdminRole)
-	if err != nil {
-		return "", HandleAuthErr(err, "access key", "create", models.InfraAdminRole)
+	rCtx := GetRequestContext(c)
+
+	if rCtx.Authenticated.AccessKey != nil && !rCtx.Authenticated.AccessKey.Scopes.Includes(models.ScopeAllowCreateAccessKey) {
+		// non-login access keys can not currently create other access keys.
+		return "", fmt.Errorf("%w: cannot use an access key to create other access keys", internal.ErrBadRequest)
```
BruceMacD, line 40:
```suggestion
		return "", fmt.Errorf("%w: cannot use an access key not issued from login to create other access keys", internal.ErrBadRequest)
```

This is a bit difficult to word, but here is my shot at it

BruceMacD (approved):
looks good

PR #3274, opened 2022-09-20
fix: few more bugs on new ui

hoyyeva:
## Summary
- fixed signup domain error message display
- fixed nav sidebar ui look and feel
- fixed autocomplete behaviour
- fixed provider table formatting
- removed breadcrumbs on providers add section
- hide account on the menu when the user is not login as infra local user

Screen Shot 2022-09-20 at 11 00 41 AM

### DIFF
--- a/ui/components/layouts/dashboard.js
+++ b/ui/components/layouts/dashboard.js
@@ -141,7 +141,13 @@ function Layout({ children }) {
     },
   ]
 
-  const subNavigation = [{ name: 'Account', href: '/account' }]
+  const subNavigation = [
+    {
+      name: 'Account',
+      href: '/account',
+      show: auth?.providerNames.includes('infra'),
+    },
+  ]
 
   for (const n of [...navigation]) {
     if (router.pathname.startsWith(n.href) && n.admin && !admin) {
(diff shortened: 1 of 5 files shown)


ui/components/layouts/dashboard.js
```diff
@@ -141,7 +141,13 @@ function Layout({ children }) {
     },
   ]
 
-  const subNavigation = [{ name: 'Account', href: '/account' }]
+  const subNavigation = [
+    {
+      name: 'Account',
+      href: '/account',
+      show: auth?.providerNames.includes('infra'),
```
jmorganca, line 148: While rare `providerNames` has been empty/null in the past, we may want to guard that:

```suggestion
      show: auth?.providerNames?.includes('infra'),
```

ui/pages/signup/index.js
```diff
@@ -172,14 +174,16 @@ export default function Signup() {
...
-                  errors.domain ? 'border-red-500' : 'border-gray-300'
+                  errors['org.subdomain'] ? 'border-red-500' : 'border-gray-300'
                 }`}
               />
               <span className='inline-flex select-none items-center rounded-r-lg border border-l-0 border-gray-300 bg-gray-50 px-3 text-2xs text-gray-500 shadow-sm'>
                 .{baseDomain}
               </span>
             </div>
-            {errors.domain && <ErrorMessage message={errors.domain} />}
+            {errors['org.subdomain'] && (
```
jmorganca, line 184: No need for using string keys here. Also, is it `errors?.org?.subDomain`? https://github.com/infrahq/infra/blob/4178903220e8a262b20523fb6aa529b5650f7f0f/api/signup.go#L18

```suggestion
            {errors?.org?.subdomain && (
```

ui/pages/signup/index.js
```diff
@@ -145,10 +145,12 @@ export default function Signup() {
                 }
               }}
               className={`mt-1 block w-full rounded-md shadow-sm focus:border-blue-500 focus:ring-blue-500 sm:text-sm ${
-                errors.org?.name ? 'border-red-500' : 'border-gray-300'
+                errors['org.name'] ? 'border-red-500' : 'border-gray-300'
```
jmorganca, line 148:
```suggestion
                errors?.org?.name ? 'border-red-500' : 'border-gray-300'
```

jmorganca (changes requested):
A few small comments but this is overall looking good

ui/pages/signup/index.js
hoyyeva: this will not work - i tried and it always returns `undefined` as the value

ui/pages/signup/index.js
hoyyeva: errors?.org?.name always returns `undefined`

jmorganca (approved):

PR #3283, opened 2022-09-21
fix: sorting bugs

hoyyeva:
## Summary
- fixing subject and privilege sorting 
- adding new sorting for role select

### DIFF
--- a/ui/lib/grants.js
+++ b/ui/lib/grants.js
@@ -3,9 +3,9 @@ export const descriptions = {
   admin: 'Read and write access to all resources',
   edit: 'Read and write access to most resources, but not roles',
   view: 'Read-only access to see most resources',
-  logs: 'Read and stream logs',
   exec: 'Shell to a running container',
   'port-forward': 'Use port-forwarding to access applications',
+  logs: 'Read and stream logs',
 }
 
 export function sortByPrivilege(a, b) {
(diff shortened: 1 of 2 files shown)


ui/lib/grants.js
```diff
@@ -17,25 +17,36 @@ export function sortByPrivilege(a, b) {
...
 
-  return a?.privilege?.localeCompare(b?.privilege)
+  return 0
 }
 
-export function sortByHasDescription(a, b) {
-  if (descriptions[a]) {
-    return -1
-  }
+export function sortedByPrivilegeArray(list) {
```
jmorganca, line 23: Consider a simpler approach here:

```
const KUBERNETES_ROLE_ORDER = ['cluster-admin', 'admin', 'edit', 'view', 'port-forward', 'exec', 'logs']

function sortByRole(a, b) {
  return KUBERNETES_ROLE_ORDER.indexOf(a) - KUBERNETES_ROLE_ORDER.indexOf(b)
}
```

hoyyeva: this works only if the `destination.roles` has exactly the same items as `KUBERNETES_ROLE_ORDER`. This will not work on edge cases like:
1. when `destination.roles` doesn't have all the roles that `KUBERNETES_ROLE_ORDER` has
2. when there are custom roles

jmorganca: That makes sense. In that case I think the idea here (to segment the list in half) is right but it could be a bit easier to understand (e.g. why use Set?).

Small example below:

```
const KUBERNETES_ROLE_ORDER = [
  'cluster-admin',
  'admin',
  'edit',
  'view',
  'port-forward',
  'exec',
  'logs',
]

function sortByRole(a, b) {
  return KUBERNETES_ROLE_ORDER.indexOf(a) - KUBERNETES_ROLE_ORDER.indexOf(b)
}

const input = ['admin', 'cluster-admin', 'example', 'exec']

const output = [
  ...input.filter(i => KUBERNETES_ROLE_ORDER.includes(i)).sort(sortByRole),
  ...input
    .filter(i => !KUBERNETES_ROLE_ORDER.includes(i))
    .sort((a, b) => a.localeCompare(b)),
]

console.log(output) // [ 'cluster-admin', 'admin', 'exec', 'example' ]
```

ui/lib/grants.js
```diff
@@ -17,25 +27,32 @@ export function sortByPrivilege(a, b) {
     return 1
   }
 
-  return a?.privilege?.localeCompare(b?.privilege)
+  return 0
+}
+
+export function sortByRole(list) {
```
jmorganca, line 33: nit: set a default
```suggestion
export function sortByRole(list = []) {
```

PR #3285, opened 2022-09-22
fix(website): use env for public write keys

jmorganca:


dnephin (commented):
How does this value get set? Do we need to change any of our CD jobs to set this env var?

jmorganca:
@dnephin it's set via vercel which hosts the website

dnephin (approved):

PR #3286, opened 2022-09-22
fix: update secrets with better error message

ssoroka:
To give better error messages on missing db key file.


dnephin (approved):
https://github.com/infrahq/secrets/commit/218b60c086230a3090a4b801cb1cd53c82e941ee

LGTM

It'd be great to remove our dependency on `vault/api`, which I suspect brings in a lot of unnecessary dependencies.

ssoroka:
Should just be the vault client. Definitely open to alternatives.

PR #3289, opened 2022-09-22
fix: typo in web signup error handling

jmorganca:


dnephin (approved):
I don't know this code, but the fix seems right to me!

PR #3290, opened 2022-09-22
fix: scope group names

BruceMacD:
## Summary
When looking up group names on IDP login we should scope the look-up by org.

## Checklist

### DIFF
--- a/internal/server/data/identity.go
+++ b/internal/server/data/identity.go
@@ -31,8 +31,8 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
 	// remove user from groups
 	if len(groupsToBeRemoved) > 0 {
 		stmt := `DELETE FROM identities_groups WHERE identity_id = ? AND group_id in (
-		   SELECT id FROM groups WHERE name IN (?))`
-		if _, err := tx.Exec(stmt, user.ID, groupsToBeRemoved); err != nil {
+		   SELECT id FROM groups WHERE organization_id = ? AND name IN (?))`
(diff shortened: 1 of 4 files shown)


ssoroka (approved):

internal/server/data/identity.go
```diff
@@ -51,8 +51,8 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
 	}
 	var addIDs []idNamePair
 
-	stmt := `SELECT id, name FROM groups WHERE name IN (?)`
-	rows, err := tx.Query(stmt, groupsToBeAdded)
+	stmt := `SELECT id, name FROM groups WHERE name IN (?) AND organization_id = ?`
+	rows, err := tx.Query(stmt, groupsToBeAdded, user.OrganizationID)
```
dnephin, line 55: minor: should we use `tx.OrganizationID()` here, for consistent with other queries?

`user.OrganizationID` should be safe, even if it's not set, since the 0 value won't match anything.

internal/server/data/migrations.go
```diff
@@ -614,3 +614,44 @@ CREATE UNIQUE INDEX IF NOT EXISTS idx_identities_verified ON identities (organiz
 		},
 	}
 }
+
+func cleanCrossOrgGroupMemberships() *migrator.Migration {
```
dnephin, line 618: Might be worth checking prod first to see if this migration finds any groups, so that we know if anyone was impacted or not.

dnephin (approved):
Looks good, couple minor suggestions

PR #3294, opened 2022-09-23
maintain: set dev version

mxyng:
## Checklist

## Related Issues

Resolves #


dnephin (approved):
Nice! I like how this simplifies things.

Anything blocking merging this?

mxyng:
Nothing in particular. It really depends if this addresses the problems @pdevine is trying to solve.

pdevine:
I like the simplification in `version.go`. Can you modify the `Makefile` to remove the excess version lines at the top of it? Those are effectively useless right now.

I am slightly nervous about development getting weird versions, however I think it's pretty obvious that you're getting a bogus value with this change. The one thing it doesn't address is whether your gh repo is stale, although that's not entirely addressed by the other change either. On other projects I have worked on we have included something like: `18a3cd029d79f17 (stale)` after the version which might be a nice addition to this.

pdevine (approved):

PR #3296, opened 2022-09-23
feat: make name optional for provider

kimskimchi:
## Things changed:

- UI form 
- Provider name will default to the kind (eg. 'okta'), and append a randomly generated string if it is taken
- Reword grey box text from "User or group" to "Enter user or group"
- Introduces a alphanumeric without vowels, for generated strings that will be user-facing to prevent accidental profanity

### API will now generate name of provider if not provided. 

https://user-images.githubusercontent.com/20054646/192228163-3d62b23f-8103-4ce4-bc32-c5ece0a3cbd7.mov

### The UI is slightly rearranged. 
  - _**Previous:**_ 

  image

  - _**After:**_
  image

### DIFF
--- a/internal/server/providers.go
+++ b/internal/server/providers.go
@@ -87,7 +87,7 @@ func (a *API) CreateProvider(c *gin.Context, r *api.CreateProviderRequest) (*api
 			return nil, fmt.Errorf("Error while generating name for provider: %w", err)
 		}
 		if len(providers) > 0 {
-			randomString, err := generate.CryptoRandom(6, generate.CharsetAlphaNumeric)
+			randomString, err := generate.CryptoRandom(6, generate.CharsetAlphaNumericNoVowels)
 			if err != nil {
 				return nil, fmt.Errorf("Error while generating name for provider: %w", err)
 			}
--- a/internal/server/providers.go
+++ b/internal/server/providers.go
@@ -13,6 +13,7 @@ import (
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/access"
 	"github.com/infrahq/infra/internal/generate"
+	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 )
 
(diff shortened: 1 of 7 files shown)


ssoroka:
This feels like it's worse than showing the default populated name. why do we want this?

internal/server/providers.go
```diff
@@ -75,6 +76,15 @@ func (a *API) CreateProvider(c *gin.Context, r *api.CreateProviderRequest) (*api
 	}
 	provider.Kind = kind
 
+	// If name is not provided, generate based on provider kind
+	if provider.Name == "" {
+		randomString, err := generate.CryptoRandom(6, generate.CharsetAlphaNumeric)
+		if err != nil {
+			return nil, errors.New("Error while generating name for provider")
+		}
+		provider.Name = r.Kind + "-" + randomString
```
dnephin, line 85: I think there was some discussion about not adding the random suffix, and instead using the kind as the name without any suffix.
Do you think that could work?

If someone adds multiple providers of the same kind they would be forced to pick a name, but I think we expect almost everyone to use only a single provider of any kind, so that should be very rare.

dnephin (commented):
Looks good! One question about the implementation below.

I would also be good to add a case to `TestAPI_CreateProvider` to show the provider created with the default name. I'm happy to help with that if you have any questions

dnephin:
> This feels like it's worse than showing the default populated name. why do we want this?

I believe it's because we want to minimize the work required to setup a provider, and we expect that almost everyone will only ever use a single provider (and at most one provider of each kind), so requiring them to name `okta` as `okta` can be avoided.

internal/server/providers.go
kimskimchi: Yep! Just uploaded a video of the new behavior. It will check how of that kind there is and append a number

ui/components/grant-form.js
```diff
@@ -65,7 +65,7 @@ export default function GrantForm({ grants, roles, onSubmit = () => {} }) {
         >
           <Combobox.Input
             className={`block w-full rounded-md border-gray-300 text-xs shadow-sm focus:border-blue-500 focus:ring-blue-500`}
-            placeholder='User or group'
+            placeholder='Enter user or group'
```
kimskimchi, line 68: So that it's clearer for user that this is a text box that they need to enter

dnephin (commented):
Implementation looks good to me! I think we're just missing a test case for the new behaviour

kimskimchi:
> Implementation looks good to me! I think we're just missing a test case for the new behaviour

Yay! okay, will implement test and finalize the pr :)

ui/pages/providers/add.js
```diff
@@ -187,37 +186,39 @@ export default function ProvidersAddDetails() {
...
-            {errors.name && <ErrorMessage message={errors.name} />}
           </div>
           <div className='w-full'>
             <div className='mb-1 flex'>
               <h3 className='text-base font-medium leading-6 text-gray-900'>
-                Additional Information
+                Information
               </h3>
-              &nbsp;
+              &nbsp; &nbsp;
```
hoyyeva, line 195: i think it looks like there is an extra space on UI. what we can do is to put the `space-x-2` on the <div /> on line 191 to the spacing looks nice

dnephin (approved):
LGTM!

hoyyeva (approved):

PR #3297, opened 2022-09-23
fix: allow users to delete their own access keys from cli

ssoroka:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -148,10 +148,19 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
 				return err
 			}
 
+			config, err := currentHostConfig()
+			if err != nil {
+				return err
+			}
+
 			logging.Debugf("call server: list access keys named %q", args[0])
 			keys, err := client.ListAccessKeys(api.ListAccessKeysRequest{Name: args[0]})
 			if err != nil {
-				return handleListKeysMissingPrivilege(err)
+				var err2 error
+				keys, err2 = client.ListAccessKeys(api.ListAccessKeysRequest{UserID: config.UserID, Name: args[0]})
+				if err2 != nil {
+					return handleListKeysMissingPrivilege(err)
+				}
 			}
 
 			if keys.Count == 0 && !force {
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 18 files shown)


internal/cmd/keys.go
```diff
@@ -148,10 +148,19 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
...
 			logging.Debugf("call server: list access keys named %q", args[0])
 			keys, err := client.ListAccessKeys(api.ListAccessKeysRequest{Name: args[0]})
 			if err != nil {
-				return handleListKeysMissingPrivilege(err)
+				var err2 error
+				keys, err2 = client.ListAccessKeys(api.ListAccessKeysRequest{UserID: config.UserID, Name: args[0]})
+				if err2 != nil {
+					return handleListKeysMissingPrivilege(err)
+				}
 			}
```
dnephin, line 164: Is it always correct to assume that the error is related to authorization based on identity?

I think an admin could be removing a key for a different user, and get an error for some other reason.

I wonder if we could improve our API to remove the need to list the keys first. We have a unique identifier for the key, it should be possible to delete it based on that identifier.

internal/server/data/access_key.go
```diff
@@ -194,6 +196,9 @@ func GetAccessKey(tx ReadTxn, opts GetAccessKeysOptions) (*models.AccessKey, err
 	if opts.ByID > 0 {
 		query.B("and id = ?", opts.ByID)
 	}
+	if opts.ByName != "" {
+		query.B("and name = ?", opts.ByName)
```
dnephin, line 200: Ah, I missed this in the previous PR, but we are missing a query clause when we use `ByID` and `ByName`.

If `ByKeyID` is set we need to omit `organization_id = ?` because we need to query for all keys, we don't know the org yet.

Now that we've added `ByID` and `byName` though, we have to add `organization_id = ?` if either of those are set.

That might be a good reason to have two separate functions (one for `byKeyID`, and the other that takes options but always set an `organization_id`).  I'd also be fine keeping one function and adding 

```go
if opts.ByID != 0 || opts.ByName != "" {
    query.B("organization_id = ? , tx.OrganizationID())
}
```

dnephin (commented):
API changes looks good, but I noticed a problem with the query that I had missed in the previous PR

internal/server/models/access_key.go
```diff
@@ -50,3 +50,7 @@ func (ak *AccessKey) ToAPI() *api.AccessKey {
 		ExtensionDeadline: api.Time(ak.ExtensionDeadline),
 	}
 }
+
+func (ak *AccessKey) Token() string {
+	return ak.KeyID + "." + ak.Secret
+}
```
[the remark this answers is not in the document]
dnephin: I think when reading out of  the database the `KeyID` is not blank, so we'd return `<keyID>.`. Returning `""` is probably better in that case, but no substitute for the godoc.

internal/server/data/access_key.go
```diff
@@ -174,25 +174,52 @@ func ListAccessKeys(tx ReadTxn, opts ListAccessKeyOptions) ([]models.AccessKey,
...
+	err := tx.QueryRow(query.String(), query.Args...).Scan(accessKey.ScanFields()...)
+	if err != nil {
+		return nil, handleReadError(err)
+	}
+	return (*models.AccessKey)(accessKey), nil
+}
+
+// GetAccessKey by any unique field. This must be scoped by organizationID as it's callable by users.
+func GetAccessKey(tx ReadTxn, opts GetAccessKeysOptions) (*models.AccessKey, error) {
+	if opts.ByID == 0 && len(opts.ByKeyID) == 0 && len(opts.ByName) == 0 {
```
dnephin, line 205: I think we should probably remove `ByKeyID` from `GetAccessKeysOptions` (and the two places we reference that below here), so that there's not two ways to get the access key by KeyID. Otherwise it's possible someone uses the wrong function.

PR #3300, opened 2022-09-23
improve: move database dialector to data package

mxyng:
## Summary

Move the database connection and configuration parsing to the data package. This allows the data package to more fully abstract the database. In particular, the database password value is resolved before passing it to `NewDB()` so it doesn't need the server secret storage nor import secrets 

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgerrcode"
+	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
 
 	"github.com/infrahq/infra/internal"
@@ -24,6 +25,8 @@ import (
 )
 
 type NewDBOptions struct {
+	DSN string
+
 	EncryptionKeyProvider EncryptionKeyProvider
 	RootKeyID             string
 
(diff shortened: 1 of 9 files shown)


internal/server/data/data.go
```diff
@@ -24,6 +25,15 @@ import (
 )
 
 type Options struct {
+	Host       string
+	Port       int
+	Name       string
+	Username   string
+	Password   string
+	Parameters string
```
dnephin, line 33: This is the only part that seems not ideal.  The `data` package shouldn't need to be aware of all of these extra settings (which only exist because of a UX decision).  I think by moving all of this to data we end making the internal package too aware of a specific interface (how we configure the DB from the CLI).

It'd be great if we could keep all of this in the `cmd` package (or maybe leave it in `server`) , and only accept the `DSN` string below as part of the options.

dnephin (commented):
Nice! In general I think this is a great change, just one concern.

internal/server/data/data.go
mxyng: I think this is a matter of perspective. IMO, how to connect to the database between the configuration layer and the data layer while the server is merely a middle man. With that in mind, it makes sense to me to have the server blindly pass along those configuration values and get a `db` object in return. 

Now, there are some compromises such as the `DB*` fields in `server.Options` but those should (eventually) exist only in `data.Options`. The secret and encryption key fields are also problematic since they need to be resolved by the server. Other than those, the server doesn't really need to know anything about how to connect to the database.

dnephin: Definitely a matter of perspective! I always try to push as much of the complexity of reading user input are read from a user. That way if we change something based on user feedback, it doesn't require changing things in multiple places.

I think it's also not obvious to a reader that this struct impacts the user directly. It's not safe to remove or rename one of these fields, because that would break a user config. And the tests for that need to be in `cmd` because that's where user interaction happens.  

We could document that on this struct, but moving it to `cmd` makes it even more obvious.

dnephin (approved):
LGTM! I think this is going to make it easier to add row level security and other db options in the future.

PR #3301, opened 2022-09-23
improve: add configurable db connection options

mxyng:
## Summary

Move default server options from `cmd/server.go` to `server/server.go`. This seems like a more reasonable place to set defaults. Specifically, it can set defaults for nested options without exposing internal structures like `data.Options`

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -26,13 +26,17 @@ import (
 type NewDBOptions struct {
 	EncryptionKeyProvider EncryptionKeyProvider
 	RootKeyID             string
+
+	MaxOpenConnections int
+	MaxIdleConnections int
+	MaxIdleTimeout     time.Duration
 }
 
 // NewDB creates a new database connection and runs any required database migrations
 // before returning the connection. The loadDBKey function is called after
 // initializing the schema, but before any migrations.
 func NewDB(connection gorm.Dialector, dbOpts NewDBOptions) (*DB, error) {
-	db, err := newRawDB(connection)
+	db, err := newRawDB(connection, dbOpts)
 	if err != nil {
 		return nil, fmt.Errorf("db conn: %w", err)
 	}
@@ -212,7 +216,7 @@ func (t *Transaction) WithOrgID(orgID uid.ID) *Transaction {
 }
 
 // newRawDB creates a new database connection without running migrations.
-func newRawDB(connection gorm.Dialector) (*gorm.DB, error) {
+func newRawDB(connection gorm.Dialector, options NewDBOptions) (*gorm.DB, error) {
 	db, err := gorm.Open(connection, &gorm.Config{
 		Logger: logging.NewDatabaseLogger(time.Second),
 	})
@@ -225,10 +229,9 @@ func newRawDB(connection gorm.Dialector) (*gorm.DB, error) {
 		return nil, fmt.Errorf("getting db driver: %w", err)
 	}
 
-	// TODO: make these configurable from server config
-	sqlDB.SetMaxIdleConns(900)
-	sqlDB.SetMaxOpenConns(1000)
-	sqlDB.SetConnMaxIdleTime(5 * time.Minute)
+	sqlDB.SetMaxOpenConns(options.MaxOpenConnections)
+	sqlDB.SetMaxIdleConns(options.MaxIdleConnections)
+	sqlDB.SetConnMaxIdleTime(options.MaxIdleTimeout)
 
 	return db, nil
 }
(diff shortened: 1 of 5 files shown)


internal/server/server.go
```diff
@@ -392,3 +396,30 @@ func configureEmail(options Options) {
...
+
+		Addr: ListenerOptions{
+			HTTP:    ":80",
+			HTTPS:   ":443",
+			Metrics: ":9090",
+		},
+
+		DB: data.Options{
+			MaxOpenConnections: 1000,
+			MaxIdleConnections: 1000,
```
ssoroka, line 421: looks good. want to make the defaults something smaller?

ssoroka (approved):

internal/server/server.go
```diff
@@ -392,3 +396,30 @@ func configureEmail(options Options) {
 		email.SMTPServer = options.SMTPServer
 	}
 }
+
+func DefaultOptions(infraDir string) Options {
```
dnephin, line 400: What's the motivation for moving this to the server package?

Generally I would expect defaults to be set from the CLI, because defaults depend on the environment, and the CLI is what integrates the server package with a specific environment.

internal/server/data/data.go
```diff
@@ -23,16 +23,20 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-type NewDBOptions struct {
+type Options struct {
```
dnephin, line 26: `NewDBOptions` is somewhat more explicit. There are a lot of functions in the `data` package, and many structs named `<name of operation>Options`. Doesn't calling this `Options` make it less obvious how to use it?

dnephin (commented):
Change looks good!  Mostly I'm curious about the motivation for renaming the options struct, and for moving the defaults.

internal/server/server.go
dnephin: Ah,I see
> it can set defaults for nested options without exposing internal structures like data.Options

Either way we have to export that struct, right? Why does it matter if it's important from `server` or from `cmd` (`cmd` has an indirect dependency on `data` either way).

internal/server/server.go
dnephin: Oh ya, I think we had talked about defaulting MaxOpen to 100 if that's what Postgres uses as a default. Is that right? MaxIdle could probably be even smaller.

internal/server/server.go
dnephin: I think it's a common problem to have CLI configuration leak all over the app.

In a past role I had a colleague who has very insistent on handling all the configuration in the `cmd` or `main` package.  Everything would get constructed in one place, and the rest of the application didn't need to concern itself with configuration.

At first I was skeptical, but it worked out really well. 

I think our setup today is pretty close to that. If we were to move `server.New` into `cmd`, we could keep all of these `cmd` configuration options in one place. 

I think this might be moving us in the wrong direction, where configuration is now spread out everywhere instead of centralized in one place.

mxyng: Configuration is already spread out everywhere. When adding a new server option, you must add the field to `server.Options` then set defaults in `cmd.defaultServerOptions`. Not doing the latter can produce unexpected results.

> defaults depend on the environment, and the CLI is what integrates the server package with a specific environment

Maybe this is the main difference between our perspectives. I see the CLI as the entrypoint into the server package but ultimately the defaults are the server's defaults, not for the entrypoint. Adding another entrypoint doesn't necessarily change the defaults.

dnephin: The `server` package is a big risk. I've seen this pattern a few times now. It can very quickly become a large ball of integrating everything. It's the foundation of a difficult to decompose monolith.

Today our server is already:
1. the backend for API handlers
2. the package that contains API handlers, API routing, and API middlware
3. our IaaC implementation from config file
4. the management of goroutines for the `infra server` process

The changes here move even more of the `infra server` CLI command responsibilities into the `server` package, which I think is not really the right direction. We need to be very careful about expanding the responsibilities of `server`, and find opportunities to shrink it.

I think we should be looking to move the `server.New` function into `cmd` (or `cmd/server`) which would allow us to move `server.Options` into that package as well. The `server` package would receive a fully constructed `Server`. Accepting dependencies (instead of parameters used to construct dependencies) is often called dependency injection, and I have found it makes code much easier to test, and much easier to change.

By moving the defaults into `server`, I think we're moving away from a dependency injection approach, and doubling down on "passing in settings" instead of looking to pass in dependencies.

I think we could accomplish the goals of this PR without moving the defaults. In the short term I think the risk is small, so I don't want to block this PR, I'm ok if you merge it this way. But I also think it's likely we'll eventually be moving these defaults back into the `cmd` package.

mxyng: I want to avoid more tech debt so I'll update this PR without the default changes

dnephin (approved):

PR #3303, opened 2022-09-23
fix: faster global counts

ssoroka:
## Summary

Adds a QuickGlobalCount function for super fast global counts at the cost of a tiny bit of accuracy.

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 	"unicode"
@@ -15,6 +16,7 @@ import (
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgerrcode"
 	"gorm.io/gorm"
+	"gorm.io/gorm/schema"
 
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/logging"
@@ -505,6 +507,24 @@ func GlobalCount[T models.Modelable](tx GormTxn, selectors ...SelectorFunc) (int
 
 	var count int64
 	if err := db.Model((*T)(nil)).Count(&count).Error; err != nil {
+		return -1, err
+	}
+
+	return count, nil
+}
+
+// QuickGlobalCount gives the count of all records, not scoped by org, and not filterable, but very fast.
+func QuickGlobalCount[T models.Modelable](tx GormTxn) (int64, error) {
+	db := tx.GormDB()
+
+	s, err := schema.Parse((*T)(nil), &sync.Map{}, schema.NamingStrategy{})
+	if err != nil {
+		return 0, fmt.Errorf("trying to get table name: %w", err)
+	}
+	tableName := s.Table
+
+	var count int64
+	if err := db.Raw("SELECT reltuples AS count FROM pg_class WHERE relname = ?", tableName).Scan(&count).Error; err != nil {
 		return -1, err
 	}
 
--- a/internal/server/metrics.go
+++ b/internal/server/metrics.go
@@ -35,7 +35,7 @@ func setupMetrics(db *data.DB) *prometheus.Registry {
 		Name:      "groups",
 		Help:      "The total number of groups",
 	}, []string{}, func() []metrics.Metric {
-		count, err := data.GlobalCount[models.Group](db)
+		count, err := data.QuickGlobalCount[models.Group](db)
 		if err != nil {
 			logging.L.Warn().Err(err).Msg("groups")
 			return []metrics.Metric{}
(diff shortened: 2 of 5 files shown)


internal/server/data/data.go
```diff
@@ -511,6 +513,24 @@ func GlobalCount[T models.Modelable](tx GormTxn, selectors ...SelectorFunc) (int
...
 
+// QuickGlobalCount gives the count of all records, not scoped by org, and not filterable, but very fast.
+func QuickGlobalCount[T models.Modelable](tx GormTxn) (int64, error) {
+	db := tx.GormDB()
+
+	s, err := schema.Parse((*T)(nil), &sync.Map{}, schema.NamingStrategy{})
+	if err != nil {
+		return 0, fmt.Errorf("trying to get table name: %w", err)
+	}
+	tableName := s.Table
```
dnephin, line 524: Could we do this without involving gorm, since we are actively working on removing it?

We should be able to pass the table name as an argument, right? 

We could move the queries for `EnqueueHeartbeat` to something like `data.TableCounts` to not expose table names in that case, and expose type specific functions for the two calls in `setupMetrics`.

internal/server/metrics.go
```diff
@@ -35,7 +35,7 @@ func setupMetrics(db *data.DB) *prometheus.Registry {
 		Name:      "groups",
 		Help:      "The total number of groups",
 	}, []string{}, func() []metrics.Metric {
-		count, err := data.GlobalCount[models.Group](db)
+		count, err := data.QuickGlobalCount[models.Group](db)
```
dnephin, line 38: Why do some of these use quick and others the old GlobalCount ?

dnephin (commented):
Seems like a good thing to change. Couple questions about the implementation

internal/server/data/data.go
ssoroka: I'd like to be able to get the table name directly from the models. Doesn't appear we're set up for that yet.

internal/server/metrics.go
ssoroka: You can't do a quick count on a global count that has where conditions

internal/server/data/data.go
dnephin: We do have the table name on some of the unexported aliases `table` types in `data`, but ya, not yet exported anywhere, and most models don't have a `table` equivalent yet.

internal/server/metrics.go
dnephin: Cool, makes sense. I'll be looking to remove `GlobalCount` when we replace gorm, and change it to something like:

```go
type CountOptions struct {
    // AllOrganizations instructs count to return a global count, not only the
    // count for the request organization.
    AllOrganizations true
}

func CountGrants(tx ReadTxn, opts CountOptions) (int64, error) { ... }

```

ssoroka: that's fine. Global counts are generally a pain. A bunch of the counts we're doing aren't going to scale, and that includes the ones currently in pagination.

internal/server/data/data.go
ssoroka: 👍🏼 seems like this is fine for now as it will be replaced soon anyway.

dnephin: That's one way to look at it! It'd be nice if new code started using the new style of queries, instead of adding to the number of queries that need to be changed.

For now we could add a `Table()` method to the few models that are used with `QuickGlobalCount`, or create a `map[reflect.Type]string` here to look up the name from the mode type. Either of those should allow this to use regular sql without gorm, right?

internal/server/data/data.go
```diff
@@ -511,6 +513,24 @@ func GlobalCount[T models.Modelable](tx GormTxn, selectors ...SelectorFunc) (int
...
+
+	var count int64
+	if err := db.Raw("SELECT reltuples AS count FROM pg_class WHERE relname = ?", tableName).Scan(&count).Error; err != nil {
```
mxyng, line 527: This returns the wrong count as it includes the deleted groups

pdevine: well, deleted anything, really. I'm not sure there's an easy way around this given we're doing soft deletes.

mxyng: Right, any resource that has soft delete. It's not worth making this change if it counts the deleted resources since the count will have no meaning

ssoroka: I mean it's kind of nice for growth analytics. you don't really want deleted records to reduce your count of how much the product is getting used. :D 
These counts with conditions all have to go long term, as they'll eventually start to take minutes and 10s of minutes to run. There's some more complicated alternatives like moving deleted records to separate tables, or maintaining the count yourself in redis.

dnephin: Definitely, I think global counts become not that interesting pretty quickly. We will hopefully start to be a lot more interested in things like "number of users and groups per org", and "number of active orgs". I'd expect us to get that from our analytics platform, or exporting the data to an analytics database which can support this type of query.

ssoroka:
Realized this was counting DeletedAt records. After some discussion with the team, it's best to continue doing this the expensive way and replace it with something else later, eg nightly redis-based counts with incr/decr on add/remove of records.

PR #3327, opened 2022-09-26
Add TxOptions to txn.Begin

dnephin:
This will allow us to start read only transactions, or transactions with different isolation levels.

We're not using that yet, but after #3302 merges we'll be able to use it for handlers. An example of that can be seen here #3328.

We will also need this for long-polling. We need to use an isolation level that ensures that multiple queries in the same transaction all read the same snapshot (repeatable read).

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -41,7 +41,7 @@ func NewDB(connection gorm.Dialector, dbOpts NewDBOptions) (*DB, error) {
 		return nil, fmt.Errorf("db conn: %w", err)
 	}
 	dataDB := &DB{DB: db}
-	tx, err := dataDB.Begin(context.TODO())
+	tx, err := dataDB.Begin(context.TODO(), nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


internal/server/data/data.go
```diff
@@ -122,9 +122,8 @@ func (d *DB) GormDB() *gorm.DB {
 	return d.DB
 }
 
-// TODO: accept sql.TxOptions when we remove gorm
-func (d *DB) Begin(ctx context.Context) (*Transaction, error) {
-	tx := d.DB.WithContext(ctx).Begin()
+func (d *DB) Begin(ctx context.Context, opts *sql.TxOptions) (*Transaction, error) {
```
ssoroka, line 125: could consider using variadic arguments to allow the opts to be optional.
```suggestion
func (d *DB) Begin(ctx context.Context, opts ...sql.TxOptions) (*Transaction, error) {
```

dnephin: I see `gorm` uses this pattern, but I consider this dangerous and I don't think it should ever be used.

varargs are great if you are going to use all the args, but using varargs for a single optional argument and ignoring the rest is going to cause confusion for anyone reading the code. Adding a `nil` for the parameter isn't really a big inconvenience. We could also follow the stdlib and have two functions for `Begin`, but at this time it didn't seem necessary since it was easy enough to add the nils.

ssoroka: Well, it's just a suggestion. The nil argument is a bit ugly, especially when that's the default case. From the project I've seen, this is a really common pattern and familiar to many Go devs.

PR #3328, opened 2022-09-26
Allow routes to specify transaction options

dnephin:
## Summary

Branched from #3302, including the changes from #3327

This PR adds TxOptions to `routeSettings`, and uses it to create read-only transactions for some routes. This would also allow us to use different isolation levels if necessary.

### DIFF
--- a/internal/server/migrations.go
+++ b/internal/server/migrations.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"database/sql"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
@@ -30,9 +31,15 @@ func (a *API) deprecatedRoutes(noAuthnNoOrg *routeGroup) {
 	type SignupEnabledResponse struct {
 		Enabled bool `json:"enabled"`
 	}
-	addDeprecated(a, noAuthnNoOrg, http.MethodGet, "/api/signup",
-		func(c *gin.Context, _ *api.EmptyRequest) (*SignupEnabledResponse, error) {
+
+	add(a, noAuthnNoOrg, http.MethodGet, "/api/signup", route[api.EmptyRequest, *SignupEnabledResponse]{
+		handler: func(c *gin.Context, _ *api.EmptyRequest) (*SignupEnabledResponse, error) {
 			return &SignupEnabledResponse{Enabled: false}, nil
 		},
-	)
+		routeSettings: routeSettings{
+			omitFromTelemetry: true,
+			omitFromDocs:      true,
+			txnOptions:        &sql.TxOptions{ReadOnly: true},
+		},
+	})
 }
--- a/internal/server/routes.go
+++ b/internal/server/routes.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"database/sql"
 	"fmt"
 	"net/http"
 	"path"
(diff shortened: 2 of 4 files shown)


internal/server/migrations.go
```diff
@@ -30,9 +31,15 @@ func (a *API) deprecatedRoutes(noAuthnNoOrg *routeGroup) {
...
-	addDeprecated(a, noAuthnNoOrg, http.MethodGet, "/api/signup",
-		func(c *gin.Context, _ *api.EmptyRequest) (*SignupEnabledResponse, error) {
+
+	add(a, noAuthnNoOrg, http.MethodGet, "/api/signup", route[api.EmptyRequest, *SignupEnabledResponse]{
+		handler: func(c *gin.Context, _ *api.EmptyRequest) (*SignupEnabledResponse, error) {
 			return &SignupEnabledResponse{Enabled: false}, nil
 		},
-	)
+		routeSettings: routeSettings{
+			omitFromTelemetry: true,
```
mxyng, line 40: Curious why this endpoint is omitted from telemetry and docs

internal/server/migrations.go
```diff
@@ -30,9 +31,15 @@ func (a *API) deprecatedRoutes(noAuthnNoOrg *routeGroup) {
 	type SignupEnabledResponse struct {
 		Enabled bool `json:"enabled"`
 	}
-	addDeprecated(a, noAuthnNoOrg, http.MethodGet, "/api/signup",
-		func(c *gin.Context, _ *api.EmptyRequest) (*SignupEnabledResponse, error) {
+
+	add(a, noAuthnNoOrg, http.MethodGet, "/api/signup", route[api.EmptyRequest, *SignupEnabledResponse]{
```
mxyng, line 35: Is this endpoint no longer deprecated? Seems like a weird change

internal/server/routes.go
```diff
@@ -205,7 +207,7 @@ func wrapRoute[Req, Res any](a *API, routeID routeIdentifier, route route[Req, R
 			return err
 		}
 
-		tx, err := a.server.db.Begin(c.Request.Context(), nil)
+		tx, err := a.server.db.Begin(c.Request.Context(), route.txnOptions)
```
BruceMacD, line 210: Applying the transaction options here after the authentication is checked means that a read only route can still update the user's "last seen at" time, correct?

internal/server/migrations.go
dnephin: I removed the `addDeprecated` function because all it did was define the `routeSettings`, and I don't think we can safely assume that all deprecated routes will have the same `routeSettings`. They should all be `omitFromDocs=true`, but for other settings like `txnOptions` and `omitFromTelemetry` I think it may depend on the route. We may want telemetry for the route to see how many people are still using a deprecated one.

The diff here is moving the logic from `addDeprecated` inline so that we can give each deprecated route its own settings.

This diff is in the `deprecatedRoutes` method, so still documented as deprecated.

internal/server/routes.go
dnephin: Exactly, any of the writes that are common to most requests (update users last seen at, update destination last seen at, and extending the access key expiry time) happen in the `authenticateRequest` function which is two calls above this. It always gets its own write transaction that it commits before returning.

BruceMacD (approved):

Issue #3009, opened 2022-08-23
Use a transaction for migrations


dnephin:
Currently we don't use a transaction, but postgres supports running most operations in a transaction. We should look into adding a transaction for running migrations so that if one fails the whole thing is rolled back.

dnephin:
Fixed by #3157 and #https://github.com/infrahq/infra/pull/3329

PR #3329, opened 2022-09-26
Use a transaction in the migration test

dnephin:
## Summary

#3009 added a transaction for migrations, but I missed adding it to the test function.

This PR adds a transaction for each test case. I used a separate transaction for `migrate` and `expected` because using the same transaction seem to produce a different error than the one we saw in production.  This transaction setup should match production, because we commit the transaction used for migrations before performing any other query.

Includes some fixes that will likely be merged before this, so I'll rebase once those are in.

### DIFF
--- a/internal/server/data/migrations_test.go
+++ b/internal/server/data/migrations_test.go
@@ -2,6 +2,7 @@ package data
 
 import (
 	"bytes"
+	"context"
 	"database/sql"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


mxyng (approved):

internal/server/data/migrations_test.go
```diff
@@ -158,6 +173,7 @@ func TestMigrations(t *testing.T) {
 		{
 			label: testCaseLine("202206281027"),
 			setup: func(t *testing.T, db WriteTxn) {
+				t.Skip("this migration no longer works with transactions")
```
mxyng, line 176: Should add a TODO here to fix later

BruceMacD (approved):
tested and validated locally

internal/server/data/migrations_test.go
dnephin: Ya, I wonder if we should fix this, or remove it since it's an old migration.

PR #3331, opened 2022-09-26
Fix gofmt

dnephin:
## Summary

Looks like a recent PR merged without `gofmt` being applied. 

I think when I added the query linter I introduced a bug into our CI config.  `make lint` by default runs fixes, but we don't detect if files have changed in that job, so we were not getting lint failures for problems that were auto-fixed.

This PR fixes the `gofmt`, and fixes the CI config to run without `--fix` so that formatting problems are reported as errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,9 +43,11 @@ postgres:
 	@echo 'export POSTGRESQL_CONNECTION="host=localhost port=5432 user=postgres dbname=postgres password=password123"'
 
 
+LINT_ARGS ?= --fix
+
 lint:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -43,9 +43,11 @@ postgres:
 	@echo 'export POSTGRESQL_CONNECTION="host=localhost port=5432 user=postgres dbname=postgres password=password123"'
 
 
+LINT_FIX ?= --fix
```
dnephin, line 46: Maybe `LINT_ARGS` ? Since you can set any arguments with this. I'll make that change.

pdevine (approved):

mxyng (approved):

Issue #3333, opened 2022-09-27
Secret value with `:` is misinterpreted as a secret from a provider


mxyng:
## Describe the bug

If the secret value contains `:`, and the secret is loaded through environment variables, the value will be interpreted as a secret from a secret provider. Since this will probably not resolve to a valid provider, it will fail and the server will refuse to start up. 

e.g. configuring a secret from env using Helm

```
server:
  env:
    - name: INFRA_SERVER_DB_PASSWORD
      value: !@#$%&*()-_=+[]{}<>:?
```

The server will try to load secret `?` from provider `!@#$%&*()-_=+[]{}<>`

## Expected behavior

Two solutions come to mind.

1. Disable secret look when value comes from the environment
2. Use the value verbatim if the provider does not exist.

dnephin:
I think we should do both of those. #2186 is a related proposal that uses the issue described here as one of the motivations.

I think any literal string should be used as a literal string. Only the config file should support these indirect lookup of secrets by using a nested structure.

PR #3334, opened 2022-09-27
fix: do not load client secret from infra secrets

BruceMacD:
## Summary
Loading from Infra secrets should only be possible internally and through config loading, it should not be possible to send the secret outside of Infra.

## Checklist

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/gin-gonic/gin"
-	"github.com/infrahq/secrets"
 	"gopkg.in/square/go-jose.v2"
 
 	"github.com/infrahq/infra/api"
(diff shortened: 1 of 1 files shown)


ssoroka (approved):
I think this is right. the GetSecret stuff needs to be config only, and not work from the UI or API

internal/server/config.go
```diff
@@ -686,13 +686,18 @@ func (s Server) loadProviders(db data.GormTxn, providers []Provider) error {
...
 
-func (Server) loadProvider(db data.GormTxn, input Provider) (*models.Provider, error) {
+func (s Server) loadProvider(db data.GormTxn, input Provider) (*models.Provider, error) {
 	// provider kind is an optional field
 	kind, err := models.ParseProviderKind(input.Kind)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse provider in config load: %w", err)
 	}
 
+	clientSecret, err := secrets.GetSecret(input.ClientSecret, s.secrets)
```
dnephin, line 696: Where did we do this lookup previously? I think this could be a breaking change for anyone on-prem that is using a  `:` in their client secret #3333

BruceMacD: This wouldn't have worked before either so it wont be a breaking change in this case

PR #3338, opened 2022-09-27
Add a check to the data migration test

dnephin:
This will ensure that we don't leave stale test data in the database, which could cause subsequent test cases to fail.

Also adds a missing `DELETE` to cleanup some stale entries from an existing test case.

### DIFF
--- a/internal/server/data/migrations_test.go
+++ b/internal/server/data/migrations_test.go
@@ -677,6 +677,7 @@ DELETE FROM settings WHERE id=24567;
 				stmt := `
 					DELETE FROM groups;
 					DELETE FROM identities;
+					DELETE from identities_groups;
 				`
 				_, err := tx.Exec(stmt)
 				assert.NilError(t, err)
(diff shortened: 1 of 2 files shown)


internal/server/data/migrations_test.go
```diff
@@ -773,6 +775,27 @@ in ./migrations.go, add a test case to this test, and run the tests again.
...
+
+		data := dumpSchema(t, os.Getenv("POSTGRESQL_CONNECTION"),
+			"--section=data",
+			"--exclude-table=testing.migrations",
+			"--inserts",
+			"--no-comments")
+		stmts, err := schema.TrimComments(data)
+		assert.NilError(t, err)
+
+		if !assert.Check(t, is.Equal(stmts, "")) {
```
dnephin, line 792: This line here should print something like this when it finds stale data:

```
        --- stmts
        +++ →
        @@ -1,3 +1 @@
        -INSERT INTO testing.identities_groups VALUES (1003, 1001);
        -INSERT INTO testing.identities_groups VALUES (1004, 1002);
```

BruceMacD (approved):
This fixes what I ran into, thanks

internal/server/data/migrations_test.go
```diff
@@ -838,7 +861,8 @@ func dumpSchema(t *testing.T, conn string) string {
 
 	out := new(bytes.Buffer)
 	// https://www.postgresql.org/docs/current/app-pgdump.html
-	cmd := exec.Command("pg_dump", "--no-owner", "--no-tablespaces", "--schema-only", "--schema=testing")
+	args = append([]string{"--no-owner", "--no-tablespaces", "--schema=testing"}, args...)
```
mxyng, line 864: Wouldn't this be neater if the static options come after?

```suggestion
	args = append(args, "--no-owner", "--no-tablespaces", "--schema=testing")
```

mxyng (approved):

internal/server/data/migrations_test.go
dnephin: Ya, that should work as well. I was concerned about modifying the backing array of `args`, but since this variable comes from a varargs, I don't know if that's really a problem.

The one difference is that I think by putting `args` at the end the caller can decide to override some of the pre-defined args, but maybe we won't need that. 

I think it should be safe to move `args` to the first argument.

PR #3339, opened 2022-09-27
improve: add tests for timeouts

ssoroka:


internal/server/data/data_test.go
```diff
@@ -213,3 +214,37 @@ func TestDB_Begin(t *testing.T) {
...
+			started := time.Now()
+
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+			defer cancel()
+			tx := db.WithContext(ctx)
+			err := tx.Exec("select pg_sleep(10);").Error
+			assert.Error(t, err, "timeout: context deadline exceeded")
+
+			elapsed := time.Since(started)
+			assert.Assert(t, elapsed < 3*time.Second, "query should have timed out and been cancelled")
```
mxyng, line 231: This assertion can be stricter if you use `assert.DeepEqual` with `DurationWithThreshold`. Not blocking since this is likely sufficient

mxyng (approved):
LGTM

internal/server/data/data_test.go
```diff
@@ -213,3 +214,37 @@ func TestDB_Begin(t *testing.T) {
 		})
 	})
 }
+
+func TestLongRunningQueriesAreCancelled(t *testing.T) {
+	runDBTests(t, func(t *testing.T, db *DB) {
```
dnephin, line 219: Since we're sleeping in this test, I think we need to add the following to the top of the test to allow it to be skipped

```go
if testing.Short() {
    t.Skip("too slow for short run")
}
```

internal/server/data/data_test.go
```diff
@@ -213,3 +214,37 @@ func TestDB_Begin(t *testing.T) {
...
+
+		t.Run("Gorm", func(t *testing.T) {
+			started := time.Now()
+
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
dnephin, line 224: Could use a timeout of maybe `200 * time.Millisecond` to spend less time waiting on the test?

internal/server/routes_test.go
```diff
@@ -280,3 +281,40 @@ func TestInfraVersionHeader(t *testing.T) {
 }
 
 var apiVersionLatest = internal.FullVersion()
+
+func TestRequestTimeout(t *testing.T) {
+	srv := setupServer(t)
```
dnephin, line 286: Would be good to have this here too:

```go
if testing.Short() {
    t.Skip("too slow for short run")
}
```

internal/server/routes_test.go
```diff
@@ -280,3 +281,40 @@ func TestInfraVersionHeader(t *testing.T) {
...
+	routes := srv.GenerateRoutes()
+	router, ok := routes.Handler.(*gin.Engine)
+	assert.Assert(t, ok)
+	a := &API{server: srv}
+
+	group := &routeGroup{RouterGroup: router.Group("/", TimeoutMiddleware(2*time.Second)), noAuthentication: true, noOrgRequired: true}
```
dnephin, line 292: Could do 200ms here as well

dnephin (commented):
Looks good! 

We had a test case called `TestDBTimeout` in `internal/server`. Can we remove that one now that we have this new one that looks to be closer to a production setup?

We may also be able to remove `TestRequestTimeoutError` since I think that is also covered by this new test.

I'd be great to add the `t.Skip` to these two tests when `testing.Short()`, and I think we could sleep for less time to reduce overall runtime when not using `-short`.

internal/server/data/data_test.go
ssoroka: yeah, was a bit worried about slow CI servers

dnephin (approved):
LGTM, waiting 1 second seems fine for now. 

We could do something where we pick the deadline based on the `CI` env var being set or not (longer deadline for CI shorter for local), but not critical right now.

PR #3340, opened 2022-09-27
fix: cli redirect

jmorganca:


mchiang0610 (approved):
Thanks for catching this.

PR #3345, opened 2022-09-29
maintain: migrate provider user data to direct sql

BruceMacD:
## Summary
No functional changes here. As a part of SCIM user management I need to be able to apply custom SCIM queries to the provider user look-up. This is much easier using direct SQL so updating the provider user to use the direct SQL now.

## Checklist

### DIFF
--- a/internal/server/data/provideruser.go
+++ b/internal/server/data/provideruser.go
@@ -8,10 +8,34 @@ import (
 
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/internal/server/data/querybuilder"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/internal/server/providers"
 	"github.com/infrahq/infra/uid"
 )
+
+type providerUserTable models.ProviderUser
+
+func (p providerUserTable) Table() string {
+	return "provider_users"
+}
+
+func (p providerUserTable) Columns() []string {
+	return []string{"identity_id", "provider_id", "email", "groups", "last_update", "redirect_url", "access_token", "refresh_token", "expires_at"}
+}
+
+func (p providerUserTable) Values() []any {
+	return []any{p.IdentityID, p.ProviderID, p.Email, p.Groups, p.LastUpdate, p.RedirectURL, p.AccessToken, p.RefreshToken, p.ExpiresAt}
+}
+
+func (p *providerUserTable) ScanFields() []any {
+	return []any{&p.IdentityID, &p.ProviderID, &p.Email, &p.Groups, &p.LastUpdate, &p.RedirectURL, &p.AccessToken, &p.RefreshToken, &p.ExpiresAt}
+}
+
+func (p *providerUserTable) OnInsert() error {
+	p.LastUpdate = time.Now().UTC()
+	return nil
+}
 
 func validateProviderUser(u *models.ProviderUser) error {
 	switch {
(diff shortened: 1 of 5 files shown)


internal/server/data/provideruser_test.go
```diff
@@ -73,17 +73,13 @@ func TestSyncProviderUser(t *testing.T) {
 					err = CreateIdentity(db, user)
 					assert.NilError(t, err)
 
-					pu := &models.ProviderUser{
-						ProviderID: provider.ID,
-						IdentityID: user.ID,
+					pu, err := CreateProviderUser(db, provider, user)
```
mxyng, line 76: I'm a little confused by this change. Was it not necessary to create the provider user explicitly before? Why is it needed now?

dnephin: I suspect because we were previously using `gorm.DB.Save` that the update below was actually using sql `INSERT` once it found there was not existing record.

Now that we explicitly only use sql `UPDATE` in update functions we have to insert it first.

We might be able to insert all the values in `CreateProviderUser`, instead of calling update immediately after.

internal/server/data/provideruser.go
```diff
@@ -47,26 +70,88 @@ func CreateProviderUser(db GormTxn, provider *models.Provider, ident *models.Ide
...
+	query.B("SET")
+	query.B(columnsForUpdate(pu), pu.Values()...)
+	query.B("WHERE provider_id = ? AND identity_id = ?;", providerUser.ProviderID, providerUser.IdentityID)
+	_, err := tx.Exec(query.String(), query.Args...)
+	return handleError(err)
 }
 
-func ListProviderUsers(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.ProviderUser, error) {
-	return list[models.ProviderUser](db, p, selectors...)
+func ListProviderUsers(tx ReadTxn, providerID uid.ID) ([]models.ProviderUser, error) {
```
dnephin, line 96: Looks like this is only used by `DeleteProviders` in the same package. We could un-export this function and rely on the test coverage for `DeleteProviders` to cover this query.

internal/server/data/provideruser.go
```diff
@@ -47,26 +70,88 @@ func CreateProviderUser(db GormTxn, provider *models.Provider, ident *models.Ide
...
+	table := &providerUserTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(table))
+	query.B("FROM")
+	query.B(table.Table())
+	query.B("WHERE provider_id = ?", providerID)
+	query.B("ORDER BY email ASC")
```
dnephin, line 103: I don't think we have an index on `email`, so we might want to remove this `ORDER BY`. It seems like the order doesn't matter since this is only used internally to delete a provider. If we mention that there is no sort order in the `godoc` for this function hopefully the lack of sort won't be surprising to anyone.

dnephin (approved):
LGTM!

I think our test coverage of these functions is a little light, I see a few calls in `TestDeleteProviderUser`, but no assertions about the returned value from `Create` or `Get`.

As I've been converting these I have been trying to add more complete test coverage of the returning values (and error conditions like not found). Would also be fine as a follow up

internal/server/data/provideruser.go
```diff
@@ -47,26 +70,88 @@ func CreateProviderUser(db GormTxn, provider *models.Provider, ident *models.Ide
...
 
-func ListProviderUsers(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]models.ProviderUser, error) {
-	return list[models.ProviderUser](db, p, selectors...)
+func ListProviderUsers(tx ReadTxn, providerID uid.ID) ([]models.ProviderUser, error) {
+	table := &providerUserTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(table))
+	query.B("FROM")
+	query.B(table.Table())
+	query.B("WHERE provider_id = ?", providerID)
```
dnephin, line 102: While reading over these queries it is now a lot more obvious that we always use `provider_id` but not always `identity_id`. I opened #3347 to fix the primary key index so that we can make use of it for all of these queries.

internal/server/data/provideruser_test.go
BruceMacD: Yeah, the update was used as a a "create or update" function before. Which our new insert doesn't do. Creating a provider user doesn't allow specifying these fields, so we have to use the update to add them. We might need a logic change here in a different PR

PR #3347, opened 2022-09-29
Fix order of provider_user index

dnephin:
`gorm` created the index as `PRIMARY KEY (identity_id, provider_id)`

Looking at the queries in #3345 we can see that while some queries use a `WHERE` clause with both of these, two of the queries can use only the `provider_id`. The order of the columns in the index matters. 

https://www.postgresql.org/docs/current/indexes-multicolumn.html has this to say about it (the default index type should be b-tree, even though it's not in the statement, [ref](https://www.postgresql.org/docs/current/indexes-unique.html)).

>A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns. The exact rule is that equality constraints on leading columns, plus any inequality constraints on the first column that does not have an equality constraint, will be used to limit the portion of the index that is scanned. Constraints on columns to the right of these columns are checked in the index, so they save visits to the table proper, but they do not reduce the portion of the index that has to be scanned.

Since this index includes all the rows in the table, we would have to scan the full index if we did not include an `identity_id`.

We always query with a `provider_id`, but not always with an `identity_id`, so re-ordering this index ensures we can always use it optimally.

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -68,6 +68,7 @@ func migrations() []*migrator.Migration {
 		setDefaultOrgID(),
 		addIdentityVerifiedFields(),
 		cleanCrossOrgGroupMemberships(),
+		fixProviderUserIndex(),
 		// next one here
 	}
 }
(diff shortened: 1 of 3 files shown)


BruceMacD (approved):
good catch

PR #3350, opened 2022-09-29
Allow the connector to run outside of a kube pod (part 2)

### LINKED ISSUE #2500: connector: support running the connector outside of a kube pod

dnephin:
## Summary

Branched from #2613

This PR completes the work to allow the connector to run outside of a kube pod (primarily for testing).

Two options are added to the connector config (both can be omitted when run from inside the cluster, and are documented as being optional).
* `endpointAddr` - is the address of the connector proxy that is sent to the Infra API, to tell clients where to connect. Sometimes this will be the same value as `addrs.https`, but it may be different if there are load balancers or other proxies in-between. When set this allows us to skip the "lookup service host and port" steps, which require the service to exist in kuberenetes.
* `kubernetes.authToken` - gives us a way to specify a token (likely for a service account) that will be used to communicate with the kube API. I set this up locally using the following:

```
kubectl create serviceaccount infra-connector
kubectl create token infra-connector > kube.token
kubectl create clusterrolebinding infra-connector --clusterrole=cluster-admin --serviceaccount=default:infra-connector
```

For now the kube API hostname/port, and kube API CA are both still being looked up from either in-cluster, or the kubeconfig, but it might make more sense to specify those as options directly (instead of falling back to kubeconfig).

Resolves #2500

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -44,6 +44,8 @@ type Options struct {
 	// This value is sent to the infra API server to update the
 	// Destination.Connection.URL.
 	EndpointAddr types.HostPort
+
+	Kubernetes KubernetesOptions
 }
 
 type ServerOptions struct {
@@ -58,6 +60,19 @@ type ListenerOptions struct {
 	Metrics string
 }
 
+type KubernetesOptions struct {
+	// AuthToken may be used to override the token used to authenticate with the
+	// kubernetes API server. When the connector is run in-cluster, the
+	// service account associated with the pod will be used by default.
+	// When run outside of cluster there is no default, and this value must
+	// be set to a token that has permission to impersonate users in the cluster.
+	AuthToken types.StringOrFile
+
+	// Host is looked up from either the in-cluster config, or the kubectl
+	// config.
+	// CA is looked up from either the in-cluster config or the kubectl config.
+}
+
 // connector stores all the dependencies for the connector operations.
 type connector struct {
 	k8s         *kubernetes.Kubernetes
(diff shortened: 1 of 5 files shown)


internal/connector/connector.go
```diff
@@ -196,8 +219,9 @@ func Run(ctx context.Context, options Options) error {
 		MinVersion: tls.VersionTLS12,
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(proxyHost)
+	proxy := httputil.NewSingleHostReverseProxy(kubeAPIAddr)
 	proxy.Transport = proxyTransport
+	proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
```
BruceMacD, line 224: Where does this log to, the standard logger?

BruceMacD (approved):

internal/connector/connector.go
```diff
@@ -93,6 +115,7 @@ func Run(ctx context.Context, options Options) error {
 	}
 
 	// server is localhost which should never be the case. try to infer the actual host
+	// TODO: do this lookup when Host is "" instead of localhost
```
dnephin, line 118: I created #3380 for this issue, I'll add the link to this comment in a follow up.

PR #3351, opened 2022-09-29
draft: add long-polling support for list grant

dnephin:
## Summary

This PR adds API support for performing a `ListGrants` query that blocks until there is an update to the query. Updates are identified by the `UpdateIndex`, which is incremented on every insert and delete. 

The client sends in the last `UpdateIndex` they saw (or `1` if they don't have any previous state), and the API blocks until it gets a notification about an update. The API then returns the result of the query and the new `UpdateIndex` to use in the next query.

This is a draft until some of these TODOs are resolved. I may look to extract parts of this into smaller PRs.


dnephin:
Related by #3400 and #3436

PR #3352, opened 2022-09-30
feat(agent): add retries to tolerate minor interruptions

mxyng:
## Summary

Add a counter to the infra agent. When a request fails, it increments the counter and halt the loop. If the counter exceeds the threshold, the agent will exit. This allows the agent to tolerate some interruption from the upstream service. 

The previous agent will exit immediately after getting an error. This can be frustrating if there's any minor interruptions, maybe due to network instability or long request times.

### DIFF
--- a/internal/cmd/agent.go
+++ b/internal/cmd/agent.go
@@ -19,6 +19,13 @@ import (
 	"github.com/infrahq/infra/internal/repeat"
 )
 
+var countCtxKey = key{}
+
+type counter struct {
+	count     int
+	threshold int
+}
+
 func newAgentCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:    "agent",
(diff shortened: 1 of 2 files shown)


internal/logging/logger.go
```diff
@@ -57,6 +58,7 @@ func isTerminal() bool {
 // UseFileLogger changes L to a logger that writes log output to a file that is
 // rotated.
 func UseFileLogger(filepath string) {
+	zerolog.TimeFieldFormat = time.RFC3339
```
mxyng, line 61: File logger should contain timestamps in a human readable format

internal/cmd/agent.go
```diff
@@ -143,23 +152,41 @@ func writeAgentConfig(pid int) error {
...
-		logging.Errorf("agent failed to get user destination grants: %v\n", err)
-		cancel()
+		maybeCancel("failed to get grants", err, count, cancel)
+		return
 	}
+
 	if err := writeKubeconfig(user, destinations, grants); err != nil {
-		logging.Errorf("agent failed to update kube config: %v\n", err)
-		cancel()
+		maybeCancel("failed to update kubeconfig", err, count, cancel)
```
BruceMacD, line 186: I wonder if this could leave the kube config in an unexpected state

BruceMacD (approved):
looks good to me

internal/cmd/agent.go
mxyng: You mean if it was interrupted while modifying the kubeconfig? That shouldn't happen. `writeKubeconfig` writes to a temporary file then does an atomic move.

On closer inspection, `writeKubeconfig` may not need retry since any error is unlikely to be temporary

PR #3353, opened 2022-09-30
fix: domain error msg did not show up

### LINKED ISSUE #3332: ui: no error msg when org name is not available
## Describe the bug

No message pops up when org name is not available

## Expected behavior

https://user-images.githubusercontent.com/20054646/192399189-a1b620ac-0e63-4f7f-bac9-67104864a71e.mov

hoyyeva:
## Summary
The problem that the domain error msg did not show up was because the fieldname is not consistent for different type of error message 

for domain field, there is two types of error messages 
1. validation error 
2. conflict (when the name is being taken)

and between these two error messages, the fieldnames are different:

{"code":400,"message":"validation failed: **org.subDomain**: infra is reserved and can not be used","fieldErrors":[{"fieldName":"org.subDomain","errors":["infra is reserved and can not be used"]}]}

v.s.

{"code":409,"message":"create org on sign-up: creating org: an organization with that domain already exists","fieldErrors":[{"fieldName":"**domain**","errors":[" an organization with that domain already exists"]}]}

If it is possible we should standardize it for different type of error messages so that the client side doesn't need to check what type of error it is. 

## 📷 

Screen Shot 2022-09-30 at 10 00 37 AM
Screen Shot 2022-09-30 at 9 40 12 AM
Screen Shot 2022-09-30 at 10 23 45 AM

## Related Issues

Resolves #3332

### DIFF
--- a/ui/pages/signup/index.js
+++ b/ui/pages/signup/index.js
@@ -52,6 +52,10 @@ export default function Signup() {
         for (const error of e.fieldErrors) {
           errors[error.fieldName.toLowerCase()] =
             error.errors[0] || 'invalid value'
+
+          if (error.fieldName.toLowerCase() === 'org.subdomain') {
+            errors.domain = errors[error.fieldName.toLowerCase()]
+          }
         }
         setErrors(errors)
       } else {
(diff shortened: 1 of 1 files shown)


BruceMacD:
Should we close this in favor of #3355?

PR #3354, opened 2022-09-30
fix: improve error message on login page

hoyyeva:
## Summary
improve how we handle error message on login page


ui/pages/login/index.js
```diff
@@ -201,12 +218,14 @@ export default function Login() {
             data-testid='form-field-password'
             onChange={e => {
               setPassword(e.target.value)
+              setErrors({})
               setError('')
             }}
             className={`mt-1 block w-full rounded-md  shadow-sm focus:border-blue-500 focus:ring-blue-500 sm:text-sm ${
-              error ? 'border-red-500' : 'border-gray-300'
+              errors.password ? 'border-red-500' : 'border-gray-300'
```
BruceMacD, line 225: Is there a case where we return an error specific to the password on login? The server should be vague in its error in this case so we don't leak info.

ui/pages/login/index.js
```diff
@@ -138,7 +140,20 @@ export default function Login() {
...
+      if (e.fieldErrors) {
+        const errors = {}
+        for (const error of e.fieldErrors) {
+          errors[error.fieldName.toLowerCase()] =
+            error.errors[0] || 'invalid value'
+        }
+        setErrors(errors)
+      } else {
+        if (e.code === 401 && e.message === 'unauthorized') {
+          setError('Invalid Credentials')
```
jmorganca, line 152: We should keep consistent casing here with other errors. I think only the first letter should be capitalized?

ui/pages/login/index.js
```diff
@@ -220,11 +239,7 @@ export default function Login() {
         <button className='mt-4 mb-2 flex w-full cursor-pointer justify-center rounded-md border border-transparent bg-blue-500 py-2 px-4 text-sm font-medium text-white shadow-sm hover:bg-blue-700 focus:outline-none focus:ring-2 focus:ring-blue-500 focus:ring-offset-2'>
           Log in
         </button>
-        {error && (
-          <p className='absolute -bottom-3.5 mx-auto w-full text-center text-2xs text-red-500'>
-            {error}
-          </p>
-        )}
+        {error && <ErrorMessage message={error} />}
```
jmorganca, line 242: Would recommend using a <p> tag here as we need to align the error message nicely with the font size and spacing, and to keep it obvious as to what's being rendered.

jmorganca (commented):
Nice work. Some minor comments!

jmorganca (approved):

PR #3355, opened 2022-09-30
fix: use the correct field name in org conflict error

dnephin:
Use the field name `org.subDomain` so that the UI, and other clients, can correctly associate the error with an input field.

Changing the `UniqueConstraintError.Column` to `org.subDomain` works to fix the field, but then the error becomes "an organization with that org.subDomain already exists", which is unfortunate. We could have added another field to `data.UniqueConstraintError` that would only be used by the API, but that seems like it's fixing the problem in the wrong place.

This PR takes the approach of changing `sendAPIError` to handle an already constructed `api.Error`. This allows API handlers to deal with special cases themselves. This approach also avoids a problem with unwrapping the error to make that modification directly, which would cause us to lose any fmt.Errorf context we added to the error.

Related to #3353
Resolves https://github.com/infrahq/infra/issues/3332

### DIFF
--- a/internal/server/data/passwordresettoken.go
+++ b/internal/server/data/passwordresettoken.go
@@ -30,6 +30,7 @@ retry:
 
 	tries++
 	if err = save(db, prt); err != nil {
+		// TODO: must use errors.As for error types
 		if tries <= 3 && errors.Is(err, UniqueConstraintError{}) {
 			logging.Warnf("generated random token %q already exists in the database", token)
 			goto retry // on the off chance the token exists.
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"sort"
 	"strconv"
-	"strings"
 
 	"github.com/gin-gonic/gin"
 
@@ -33,10 +32,15 @@ func sendAPIError(c *gin.Context, err error) {
 	var uniqueConstraintError data.UniqueConstraintError
 	var authzError access.AuthorizationError
 	var overLimitError redis.OverLimitError
+	var apiError api.Error
 
 	log := logging.L.Debug()
 
 	switch {
+	case errors.As(err, &apiError):
+		// the handler has already created an appropriate error to return
+		resp = &apiError
+
 	case errors.Is(err, internal.ErrUnauthorized):
 		resp.Code = http.StatusUnauthorized
 		// hide the error text, it may contain sensitive information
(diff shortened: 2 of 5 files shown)


internal/server/errors.go
```diff
@@ -54,18 +58,7 @@ func sendAPIError(c *gin.Context, err error) {
...
 
 	case errors.As(err, &uniqueConstraintError):
-		resp.Code = http.StatusConflict
-		resp.Message = err.Error()
-		// remove the error trace from field error message
-		errMsg := err.Error()
-		errTrace := strings.Split(err.Error(), ":")
-		if len(errTrace) > 0 {
-			errMsg = errTrace[len(errTrace)-1]
-		}
```
dnephin, line 64: It's not obvious from the diff (since this code moved), but I changed how we accomplished this. I noticed there was a leading space in the error message, and I believe it was coming from here.

Instead of trimming the error, I use the message from `uniqueConstraintError.Error()` instead of `err.Error()`. The `uniqueConstraintError.Error()` doesn't have the "trace" part of the error because those parts are removed by `errors.As`, which calls `Unwrap` to remove them.

I updated the test case to show it still works the same way (and fixes the leading space in the error message).

internal/server/data/passwordresettoken.go
```diff
@@ -30,6 +30,7 @@ retry:
 
 	tries++
 	if err = save(db, prt); err != nil {
+		// TODO: must use errors.As for error types
 		if tries <= 3 && errors.Is(err, UniqueConstraintError{}) {
```
dnephin, line 34: I believe `errors.Is` only works on typed errors if they implement `Is`. We could add an `Is` method to `UniqueConstraintError`, or we should use `errors.As` here.

internal/server/errors_test.go
```diff
@@ -80,15 +80,32 @@ func TestSendAPIError(t *testing.T) {
 			result: api.Error{Code: http.StatusNotImplemented, Message: "not implemented"},
 		},
 		{
-			err: data.UniqueConstraintError{Table: "user", Column: "name"},
+			err: fmt.Errorf("with context: %w",
+				data.UniqueConstraintError{Table: "user", Column: "name"}),
 			result: api.Error{
 				Code:    http.StatusConflict,
-				Message: "a user with that name already exists",
+				Message: "with context: a user with that name already exists",
```
ssoroka, line 87: do we want the context for user-facing messages?

ssoroka (approved):

internal/server/errors_test.go
dnephin: I guess we probably do, since it's unlikely to be sensitive. This was the previous behaviour, but we didn't have any test that demonstrated it. I kept that behaviour for now.

It might be fine to remove it for any error that has structured field errors, since maybe those will provide enough context.

PR #3356, opened 2022-09-30
Use SQL for Destination{Create,Update,Delete}

dnephin:
## Summary

This PR converts the three destination write queries to use SQL.

`DestinationUpdate` is a bit of a special case. We don't receive the created_at time, so we have to exclude it from the query. I did that by creating a second set of methods for `Columns` and `Values`. We could also select the row first, but it seems like we should be able to do an update and omit the rows that we don't care to update. If this becomes a more common pattern then maybe we can support that difference in the Table interface somehow.

Also updated `CountDestinationsByConnectedVersion` to remove a sub-select.

## Related Issues

Related to #2415

### DIFF
--- a/internal/server/data/destination.go
+++ b/internal/server/data/destination.go
@@ -74,23 +74,23 @@ func DeleteDestinations(db GormTxn, selector SelectorFunc) error {
 	return internal.ErrNotFound
 }
 
-type destinationsCount struct {
+type DestinationsCount struct {
 	Connected bool
 	Version   string
 	Count     float64
 }
 
-func CountDestinationsByConnectedVersion(tx ReadTxn) ([]destinationsCount, error) {
+func CountDestinationsByConnectedVersion(tx ReadTxn) ([]DestinationsCount, error) {
 	timeout := time.Now().Add(-5 * time.Minute)
 
 	stmt := `
-		SELECT *, COUNT(*) AS count
-		FROM (
-			SELECT COALESCE(version, '') AS version, last_seen_at >= ? AS connected
-			FROM destinations
-			WHERE deleted_at IS NULL
-		) AS d
-		GROUP BY version, connected`
+		SELECT COALESCE(version, '') as version,
+			   last_seen_at >= ? as connected,
+			   count(*)
+		FROM destinations
+		WHERE deleted_at IS NULL
+		GROUP BY connected, version
+	`
 	rows, err := tx.Query(stmt, timeout)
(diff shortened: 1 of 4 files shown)


internal/server/data/destination.go
```diff
@@ -38,51 +75,42 @@ func ListDestinations(db GormTxn, p *Pagination, selectors ...SelectorFunc) ([]m
...
-	if len(toDelete) > 0 {
-		ids := make([]uid.ID, 0)
-		for _, g := range toDelete {
-			ids = append(ids, g.ID)
-		}
-
-		return deleteAll[models.Destination](db, ByIDs(ids))
-	}
-
-	return internal.ErrNotFound
```
dnephin, line 56: Previously we were returning `ErrNotFound` if the ID was not found, but we don't do that for any other types. I suspect we should be consistent with the behaviour. 

If we want to return a not found on a missing delete then we can check `result.RowsAffected() > 0` from the result returned by `tx.Exec`.

internal/server/data/destination.go
```diff
@@ -4,30 +4,67 @@ import (
...
+}
+
+func (d *destinationsTable) ScanFields() []any {
+	return []any{&d.ConnectionCA, &d.ConnectionURL, &d.CreatedAt, &d.DeletedAt, &d.ID, &d.LastSeenAt, &d.Name, &d.OrganizationID, &d.Resources, &d.Roles, &d.UniqueID, &d.UpdatedAt, &d.Version}
+}
+
+// destinationsUpdateTable is used to update the destination. It excludes
+// the CreatedAt field, because that field is not part of the input to
+// UpdateDestination.
+type destinationsUpdateTable models.Destination
```
BruceMacD, line 32: This feels a bit off, but I can't come up with a better way to do updates either in our current model

BruceMacD (approved):

internal/server/data/destination.go
dnephin: It is unfortunate, there are some option options:

1. We could `select created_at` and set the value on the struct, before the `UPDATE`
2. We could not use `insert()` and copy out the implementation into `CreateDestination`. 

Maybe option 2 is better, I'll see what it looks like

dnephin: I guess the problem with 2 is that we have to add `CreatedAt` back in for both `Get` and `List`, which means we end up having to deal with the special case in 3 places, instead of only `Update` being a special case.

PR #3357, opened 2022-09-30
Use SQL for GetDestination and ListDestination

dnephin:
## Summary

Update the queries for GetDestination and ListDestination to use SQL, and add a few more test cases.

Related to #2415

### DIFF
--- a/internal/access/destination.go
+++ b/internal/access/destination.go
@@ -29,7 +29,7 @@ func SaveDestination(rCtx RequestContext, destination *models.Destination) error
 
 func GetDestination(c *gin.Context, id uid.ID) (*models.Destination, error) {
 	db := getDB(c)
-	return data.GetDestination(db, data.ByID(id))
+	return data.GetDestination(db, data.GetDestinationOptions{ByID: id})
 }
 
 func ListDestinations(c *gin.Context, uniqueID, name string, p *data.Pagination) ([]models.Destination, error) {
--- a/internal/access/destination.go
+++ b/internal/access/destination.go
@@ -28,14 +28,18 @@ func SaveDestination(rCtx RequestContext, destination *models.Destination) error
 }
 
 func GetDestination(c *gin.Context, id uid.ID) (*models.Destination, error) {
-	db := getDB(c)
-	return data.GetDestination(db, data.GetDestinationOptions{ByID: id})
+	rCtx := GetRequestContext(c)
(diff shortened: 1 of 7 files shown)


internal/server/data/destination.go
```diff
@@ -67,12 +68,89 @@ func UpdateDestination(tx WriteTxn, destination *models.Destination) error {
...
+	if err != nil {
+		return nil, handleReadError(err)
+	}
+	return (*models.Destination)(&destination), nil
+}
+
+type ListDestinationsOptions struct {
+	ByUniqueID string
+	// TODO: name is not indexed
+	ByName string
```
dnephin, line 106: I think we'll likely need to restore the unique constraint on destination.name, which will act as the index for these queries.

BruceMacD (approved):
looks good 👍

internal/server/data/destination.go
dnephin: #3392

internal/access/destination.go
```diff
@@ -29,13 +29,17 @@ func SaveDestination(rCtx RequestContext, destination *models.Destination) error
...
-	db := getDB(c)
-	return data.ListDestinations(db, p, data.ByOptionalUniqueID(uniqueID),
-		data.ByOptionalName(name))
+	rCtx := GetRequestContext(c)
+	opts := data.ListDestinationsOptions{
+		ByName:     name,
+		ByUniqueID: uniqueID,
+		Pagination: p,
+	}
+	return data.ListDestinations(rCtx.DBTxn, opts)
```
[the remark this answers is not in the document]
dnephin: made that change here, since I was rebasing some conflicts anyway

internal/server/data/destination.go
```diff
@@ -67,12 +68,92 @@ func UpdateDestination(tx WriteTxn, destination *models.Destination) error {
...
-	return get[models.Destination](db, selectors...)
+type GetDestinationOptions struct {
+	// ByID instructs GetDestination to return the row matching this ID. When
+	// this value is set, all other fields on this strut will be ignored
+	ByID uid.ID
+	// ByUniqueID instructs GetDestination to return the row matching this
+	// uniqueID.
+	ByUniqueID string
+	// ByName instructs GetDestination to return the row matching this name.
+	ByName string
```
dnephin, line 79: I added this now that we have a unique constraint on the name field. There's a test case for it as well

PR #3375, opened 2022-10-03
fix: dont require server config base url on login

### LINKED ISSUE #3374: UI requires base domain to be set or login gets stuck at spinner
## Describe the bug
While logging in the UI waits for all the necessary information before finishing the login process. One component it waits for is the base URL, if this optional server configuration field is not set the login will get stuck on the spinner screen.

## Expected behavior
The login proceeds if the base URL is empty in server config.

## Environment
Infra server version 0.15.2

BruceMacD:
## Summary
When no server URL is set in server configuration set the default server URL to the current host on response. This allows the login response to continue when the optional `baseURL` server configuration is not set, but it also ensures that we wait until we have checked the server to see if there is one set.

## Checklist

## Related Issues

Resolves #3374

### DIFF
--- a/ui/lib/serverconfig.js
+++ b/ui/lib/serverconfig.js
@@ -3,7 +3,7 @@ import useSWR from 'swr'
 export function useServerConfig() {
   const { data: { isEmailConfigured, isSignupEnabled, baseDomain } = {} } =
     useSWR(`/api/server-configuration`, {
-      revalidateIfStale: false
+      revalidateIfStale: false,
     })
 
   return {
--- a/ui/lib/serverconfig.js
+++ b/ui/lib/serverconfig.js
@@ -2,7 +2,9 @@ import useSWR from 'swr'
 
 export function useServerConfig() {
   const { data: { isEmailConfigured, isSignupEnabled, baseDomain } = {} } =
-    useSWR(`/api/server-configuration`)
+    useSWR(`/api/server-configuration`, {
+      revalidateIfStale: false
+    })
 
   return {
     isEmailConfigured,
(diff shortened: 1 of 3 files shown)


ui/lib/serverconfig.js
```diff
@@ -4,6 +4,11 @@ export function useServerConfig() {
   const { data: { isEmailConfigured, isSignupEnabled, baseDomain } = {} } =
     useSWR(`/api/server-configuration`, {
       revalidateIfStale: false,
+      onSuccess: (data) => {
+        if (data.baseDomain === "") {
+          data.baseDomain = window.location.host
```
jmorganca, line 9: Great catch! This should live closer to where it's being rendered or used (vs mutating the data sent by the server).

BruceMacD: Updated the logic to just use the window location

ui/pages/login/index.js
```diff
@@ -39,9 +39,16 @@ export function saveToVisitedOrgs(domain, baseDomain, orgName) {
       name: orgName,
     })
 
+    // set the cookie domain to a general base domain
+    let cookieDomain = window.location.host
+    let parts = cookieDomain.split('.')
+    if (parts.length > 2) {
+      cookieDomain = parts.slice(-2).join('.') // join the last two parts of the domain
```
jmorganca, line 46: In this case, would it be better to set it to all but the first part? e.g. `myorg.dev.test.com` may not work here

BruceMacD: The "last 2 parts" here would be `test.com` which in my tests gets the behavior we want of sending the cookie on any domain at `test.com/login/organizations`.

Screen Shot 2022-10-04 at 10 40 14 AM

BruceMacD: Fixed this to just remove the first part that is the org subdomain, this will work in dev and prod as you pointed out

PR #3376, opened 2022-10-03
fix: minor ui bugs

hoyyeva:
## Summary
- fix the default select for role select component
- fix the login with button's text overflow
- hide revoke when there is only one admin on the list on admin setting page
- sort user on the users table to make sure the self is always on the top of the list

### DIFF
--- a/ui/pages/settings/index.js
+++ b/ui/pages/settings/index.js
@@ -53,7 +53,7 @@ function AdminList({ grants, users, groups, onRemove, auth, selfGroups }) {
             const [deleteId, setDeleteId] = useState(null)
 
             return (
-              grants.length > 1 && (
+              grants?.length > 1 && (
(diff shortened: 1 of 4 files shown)


ui/pages/settings/index.js
```diff
@@ -51,46 +51,50 @@ function AdminList({ grants, users, groups, onRemove, auth, selfGroups }) {
...
-                        </span>
-                        ?
-                      </>
-                    ) : (
-                      grantsList?.find(grant => grant.id === deleteId)?.message
-                    )
-                  }
-                />
-              </div>
+              grants.length > 1 && (
```
jmorganca, line 56:
```suggestion
              grants?.length > 1 && (
```

jmorganca (approved):

PR #3377, opened 2022-10-03
Add a test for running the connector

dnephin:
## Summary

Branched from #3350

This PR adds a test case that runs the connector against an infra API server and a fake kubernetes API and checks that the connector behaves as expected. I think we'll want other test cases for this logic to cover different conditions, but this gives us some basic high level coverage of one of the happy paths.

This PR includes a few bug fixes and other changes required to get the test passing:
* use the right file and line number in logs that use the helper functions in `logging`. Without this change the logs were always reporting the line number in `logging.go`, which is not helpful.
* fix the error handling in connector proxy shutdown. There was a data race with the `err` variable that we were not previously aware of because no tests ran that code
* added support for specifying the kube API address, and trusted CA from connector options (instead of requiring an in-cluster config).
* changed the `runAndWait` test helper to accept a context instead of return one, so that it can be used multiple times in the same test (it's used here to run both the server and the connector in the new test case)
* improve some error message and log lines to be more clear


internal/cmd/connector_test.go
```diff
@@ -0,0 +1,269 @@
...
+			Method: "PUT",
+			URL:    &url.URL{Path: "/apis/rbac.authorization.k8s.io/v1/namespaces/ns1/rolebindings/infra:logs"},
+		},
+	}
+	sort.Slice(fakeKube.writes, func(i, j int) bool {
+		return fakeKube.writes[i].URL.Path < fakeKube.writes[j].URL.Path
+	})
+	assert.DeepEqual(t, fakeKube.writes, expectedWrites, cmpHTTPRequestShallow)
+
+	// TODO: check proxy is listening
```
dnephin, line 155: Checking the proxy requires a way to get the random assigned port, which requires a larger change to the connector. I'm planning on making that change along with splitting up the goroutines. It should be similar to what I did for the server in #1430

internal/connector/connector.go
```diff
@@ -233,6 +238,7 @@ func Run(ctx context.Context, options Options) error {
 	}
 
 	go func() {
+		// TODO: failing to bind should cause process to exit
```
dnephin, line 241: This TODO will be addressed along with the other one I commented on. It's all related to the goroutine lifetime.

internal/cmd/connector_test.go
```diff
@@ -0,0 +1,269 @@
...
+	t.Cleanup(cancel)
+
+	runAndWait(ctx, t, func(ctx context.Context) error {
+		return connector.Run(ctx, opts)
+	})
+
+	// check destination has been registered
+	var destination *models.Destination
+	poll.WaitOn(t, func(t poll.LogT) poll.Result {
+		destination, err = data.GetDestination(srv.DB(), data.ByName("testing"))
```
dnephin, line 97: Reminder to myself that this will conflict with #3357, and to not try and merge these two PRs at the same time. If that one has merged I need to update this code to use the options struct.

internal/cmd/connector_test.go
```diff
@@ -0,0 +1,269 @@
...
+		fakeKube.writesLock.Lock()
+		defer fakeKube.writesLock.Unlock()
+		if len(fakeKube.writes) >= 3 {
+			return poll.Success()
+		}
+		return poll.Continue("request count %d waiting for 3", len(fakeKube.writes))
+	})
+
+	// check kube bindings were updated
+	expectedWrites := []*http.Request{
```
mxyng, line 136: No assertions on the subject?

internal/logging/logger.go
```diff
@@ -70,19 +70,19 @@ func UseFileLogger(filepath string) {
 }
 
 func Debugf(format string, v ...interface{}) {
-	L.Debug().Msgf(format, v...)
+	L.Debug().CallerSkipFrame(1).Msgf(format, v...)
```
mxyng, line 73: Thanks for fixing this!

mxyng (approved):
Nice test!

internal/cmd/connector_test.go
dnephin: Makes sense, I guess that's in the body of the request?

Issue #3378, opened 2022-10-03
SCIM Inbound User Provisioning


BruceMacD:
## Is your feature request related to a problem? Please describe.
SCIM User endpoints should expose controls for an identity provider to keep identities in Infra in sync/

## Describe the solution you'd like
Inbound provisioning SCIM user support.

PR #3379, opened 2022-10-03
feat: SCIM list provider users

BruceMacD:
- add SCIM list users endpoint
- add SCIM role

## Summary
This change adds the necessary logic for a SCIM identity provider to list the users it has associated with it. 

It does not include the changes necessary to create an access key for an identity provider (which this functionality will rely on) so the code path can't be accessed yet.

## Checklist

## Related Issues

Part of #3378

### DIFF
--- a/internal/server/data/identity.go
+++ b/internal/server/data/identity.go
@@ -64,7 +64,7 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
 		}
 		addIDs = append(addIDs, item)
 	}
-	if rows.Err() != nil {
+	if err := rows.Err(); err != nil {
 		return err
 	}
 
@@ -105,7 +105,7 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
 			}
 			ids = append(ids, item)
 		}
-		if rows.Err() != nil {
+		if err := rows.Err(); err != nil {
 			return err
 		}
 
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -69,6 +69,7 @@ func migrations() []*migrator.Migration {
 		addIdentityVerifiedFields(),
 		cleanCrossOrgGroupMemberships(),
 		fixProviderUserIndex(),
+		addProviderUserSCIMFields(),
 		// next one here
 	}
 }
@@ -690,6 +691,24 @@ ALTER TABLE provider_users DROP CONSTRAINT IF EXISTS provider_users_pkey;
 ALTER TABLE provider_users ADD CONSTRAINT
     provider_users_pkey PRIMARY KEY (provider_id, identity_id);
 `
+			_, err := tx.Exec(stmt)
+			return err
+		},
+	}
+}
+
+func addProviderUserSCIMFields() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-09-28T13:00",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `
+				ALTER TABLE provider_users
+				ADD COLUMN IF NOT EXISTS given_name text,
+				ADD COLUMN IF NOT EXISTS family_name text,
+				ADD COLUMN IF NOT EXISTS active boolean DEFAULT true;
+
+				CREATE UNIQUE INDEX IF NOT EXISTS idx_emails_providers ON provider_users (email, provider_id);
+			`
 			_, err := tx.Exec(stmt)
 			return err
 		},
--- a/internal/server/data/provideruser.go
+++ b/internal/server/data/provideruser.go
@@ -93,29 +93,60 @@ func UpdateProviderUser(tx WriteTxn, providerUser *models.ProviderUser) error {
 	return handleError(err)
 }
 
-func listProviderUsers(tx ReadTxn, providerID uid.ID) ([]models.ProviderUser, error) {
+func ListProviderUsers(tx ReadTxn, providerID uid.ID, p *SCIMParameters) ([]models.ProviderUser, error) {
 	table := &providerUserTable{}
 	query := querybuilder.New("SELECT")
 	query.B(columnsForSelect(table))
 	query.B("FROM")
 	query.B(table.Table())
 	query.B("WHERE provider_id = ?", providerID)
+
+	query.B("ORDER BY email ASC")
+
+	if p != nil {
+		// apply scim parameters
+		if p.Count != 0 {
+			query.B("LIMIT ?", p.Count)
+		}
+		if p.StartIndex != 0 {
+			query.B("OFFSET ?", p.StartIndex)
+		}
+	}
+
 	rows, err := tx.Query(query.String(), query.Args...)
 	if err != nil {
 		return nil, err
 	}
-	defer rows.Close()
 
 	var result []models.ProviderUser
 	for rows.Next() {
 		var pu models.ProviderUser
 
 		if err := rows.Scan((*providerUserTable)(&pu).ScanFields()...); err != nil {
+			rows.Close()
 			return nil, err
 		}
 		result = append(result, pu)
 	}
-	return result, rows.Err()
+
+	if err := rows.Close(); err != nil {
+		return nil, err
+	}
+
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	if p != nil {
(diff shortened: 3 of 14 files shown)


internal/server/data/identity.go
```diff
@@ -64,7 +64,7 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
 		}
 		addIDs = append(addIDs, item)
 	}
-	if rows.Err() != nil {
+	if err := rows.Err(); err != nil {
```
BruceMacD, line 67: Looks like this would have returned nil previously, which doesnt seem intentional.

internal/server/data/provideruser.go
```diff
@@ -93,29 +93,60 @@ func UpdateProviderUser(tx WriteTxn, providerUser *models.ProviderUser) error {
...
 	query.B(columnsForSelect(table))
 	query.B("FROM")
 	query.B(table.Table())
 	query.B("WHERE provider_id = ?", providerID)
+
+	query.B("ORDER BY email ASC")
+
+	if p != nil {
+		// apply scim parameters
+		if p.Count != 0 {
```
BruceMacD, line 108: These are similar to our pagination, but slightly more complicated. Our pagination only allows set page sizes, but SCIM requires a "start index" which offsets the results.

internal/server/data/provideruser.go
```diff
@@ -93,29 +93,60 @@ func UpdateProviderUser(tx WriteTxn, providerUser *models.ProviderUser) error {
...
 
-func listProviderUsers(tx ReadTxn, providerID uid.ID) ([]models.ProviderUser, error) {
+func ListProviderUsers(tx ReadTxn, providerID uid.ID, p *SCIMParameters) ([]models.ProviderUser, error) {
 	table := &providerUserTable{}
 	query := querybuilder.New("SELECT")
 	query.B(columnsForSelect(table))
 	query.B("FROM")
 	query.B(table.Table())
 	query.B("WHERE provider_id = ?", providerID)
+
```
dnephin, line 103: To prevent someone guessing an ID and querying for other orgs, I think we would also need to do something like this:

```
INNER JOIN providers ON provider_users.provider_id = providers.id AND providers.organization_id = ?
```

Right now that's not a risk because we get the providerID from the access key, so not from user input.

I was wrong about not needing organization_id in this case previously, so I thought I'd mention it.

internal/server/data/provideruser_test.go
```diff
@@ -254,3 +255,108 @@ func TestDeleteProviderUser(t *testing.T) {
...
+			setup: func(t *testing.T, tx *Transaction) (providerID uid.ID, p *SCIMParameters, expected []models.ProviderUser) {
+				provider := &models.Provider{
+					Name: "mockta",
+					Kind: models.ProviderKindOkta,
+				}
+
+				err := CreateProvider(tx, provider)
+				assert.NilError(t, err)
+
+				pu := createTestProviderUser(t, tx, provider, "david@example.com")
```
dnephin, line 277: Not a blocker, but I've been trying to include a few records that I expect not to match the query in most of the tests for `List` operations. Something like same email different provider.

internal/server/data/provideruser_test.go
```diff
@@ -254,3 +255,108 @@ func TestDeleteProviderUser(t *testing.T) {
...
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				tx := txnForTestCase(t, db, org.ID)
+
+				providerID, p, expected := tc.setup(t, tx)
+
+				result, err := ListProviderUsers(tx, providerID, p)
+
+				assert.NilError(t, err)
+				assert.DeepEqual(t, result, expected, cmpTimeWithDBPrecision)
```
dnephin, line 343: It'd be great to include an assert for the `TotalCount` as well, which gets set on `p` I think.

internal/server/models/provideruser.go
```diff
@@ -20,6 +23,26 @@ type ProviderUser struct {
...
+			{
+				Primary: true,
+				Value:   pu.Email,
+			},
+		},
+		Active: pu.Active,
+		Meta: api.SCIMMetadata{
+			ResourceType: "User",
+		},
+	}
```
dnephin, line 47: GivenName and FamilyName are not in here. Is that expected? Are those fields only for us to consume, and will be used in a future PR?

dnephin (approved):
LGTM! Mostly questions, some of which I suspect will be answered when the create/update endpoints are added.

internal/server/data/migrations.go
```diff
@@ -695,3 +696,21 @@ ALTER TABLE provider_users ADD CONSTRAINT
...
+
+func addProviderUserSCIMFields() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-09-28T13:00",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `
+				ALTER TABLE provider_users
+				ADD COLUMN IF NOT EXISTS given_name text,
+				ADD COLUMN IF NOT EXISTS family_name text,
+				ADD COLUMN IF NOT EXISTS active boolean DEFAULT true;
```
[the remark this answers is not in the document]
BruceMacD: Yeah, control of these will be set through the SCIM create user endpoint, I'll make some updates to the just-in-time logic too. These are required fields for the SCIM spec.

internal/server/data/provideruser.go
```diff
@@ -93,29 +93,60 @@ func UpdateProviderUser(tx WriteTxn, providerUser *models.ProviderUser) error {
...
+		if p.StartIndex != 0 {
+			query.B("OFFSET ?", p.StartIndex)
+		}
+	}
+
 	rows, err := tx.Query(query.String(), query.Args...)
 	if err != nil {
 		return nil, err
 	}
-	defer rows.Close()
```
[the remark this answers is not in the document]
BruceMacD: I was closing it to use same transaction to read the count immediately afterwards, I switched it back now though when I updated how the count was read to match the other queries

internal/server/data/provideruser.go
```diff
@@ -93,29 +93,60 @@ func UpdateProviderUser(tx WriteTxn, providerUser *models.ProviderUser) error {
...
+	if err := rows.Close(); err != nil {
+		return nil, err
+	}
+
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	if p != nil {
+		row := tx.QueryRow(`SELECT count(identity_id) FROM provider_users WHERE provider_id = ?`, providerID)
```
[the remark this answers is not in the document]
BruceMacD: I just didn't notice it when writing the query, updated to use the same pattern now

internal/server/routes.go
```diff
@@ -98,6 +99,16 @@ func (s *Server) GenerateRoutes() Routes {
...
 
+	// SCIM inbound provisioning
+	add(a, authn, http.MethodGet, "/api/scim/v2/Users", route[api.SCIMParametersRequest, *api.ListProviderUsersResponse]{
+		handler: a.ListProviderUsers,
+		routeSettings: routeSettings{
+			omitFromTelemetry:          true,
+			omitFromDocs:               true,
+			infraVersionHeaderOptional: true,
+		},
+	})
```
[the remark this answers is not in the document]
BruceMacD: Fixed 👍

internal/server/scim.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for _, user := range users {
+		result.Resources = append(result.Resources, *user.ToAPI())
+	}
+
+	return result, nil
+}
+
+func SCIMParametersFromRequest(f *api.SCIMParametersRequest) data.SCIMParameters {
+	startIndex, count := 0, 100
```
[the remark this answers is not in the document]
BruceMacD: Good point, removed the default

internal/server/data/provideruser.go
BruceMacD: Good point, I've added this protection

internal/server/data/provideruser_test.go
BruceMacD: Added some assertions here

internal/server/models/provideruser.go
BruceMacD: I actually missed them because I wasn't setting them anywhere yet, good catch

PR #3381, opened 2022-10-03
fix: group trait telemetry

jmorganca:


internal/server/telemetry.go
```diff
@@ -100,17 +100,22 @@ func (t *Telemetry) EnqueueHeartbeat() {
...
+			oid = org.ID.String()
 		}
 	}
 
-	t.Event(event, uid, properties...)
+	t.Event(event, uid, oid, properties...)
 }
 
-func (t *Telemetry) Event(event string, userId string, properties ...map[string]interface{}) {
+func (t *Telemetry) Event(event string, userId string, orgId string, properties ...map[string]interface{}) {
```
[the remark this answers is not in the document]
jmorganca: It's not an optional property and should be included in every event.

internal/server/telemetry.go
```diff
@@ -147,13 +154,19 @@ func (t *Telemetry) User(id string, name string) {
...
 		return
 	}
 	err := t.Enqueue(analytics.Group{
-		GroupId:   id,
-		Traits:    analytics.NewTraits().SetName(name),
+		GroupId: id,
+		UserId:  userID,
+		Traits: map[string]interface{}{
+			"name":   name,
+			"$name":  name,
```
BruceMacD, line 166: What is the difference between `name` and `$name` here?

jmorganca: One is a "super property". Some tools require both

BruceMacD (approved):
looks good to me

mxyng (approved):

PR #3383, opened 2022-10-04
fix: remove server lookup

### LINKED ISSUE #3380: connector should lookup server addr only when unset, not when the value is localhost
If someone runs the connector with `--server-url=localhost` the connector will assume that url is wrong and attempt to lookup the server URL from the kubernetes API using the `server` service name.

https://github.com/infrahq/infra/blob/f8689e0c39893fc2a85e89392a2b610ae18f19c8/internal/connector/connector.go#L95-L96

I believe this was from a time when we expected the server and connector to often be run in the same cluster. 

It seems fine to lookup the server url when the value is empty, but if someone has specified a url for the server, let's not assume it was the wrong value and attempt to override it.

We could choose to remove this lookup entirely, or we could do the lookup only when the value is the empty string.

I think we may want to always default to a server url of `api.infrahq.com`, which means the server url should never be empty, and we could remove this lookup.

mxyng:
## Summary

Remove the server lookup when server config value is `localhost`. The original use case for this was to provide a simple way to configure a in-cluster connector. This is no longer useful. Users should provide the server URL explicitly

Resolves #3380

### DIFF
--- a/internal/kubernetes/kubernetes.go
+++ b/internal/kubernetes/kubernetes.go
@@ -542,8 +542,8 @@ func readNamespaceFromInClusterFile() (string, error) {
 	return string(contents), nil
 }
 
-// Find the first suitable Service, filtering on infrahq.com/component
-func (k *Kubernetes) Service(component string, labels ...string) (*corev1.Service, error) {
+// Find the first suitable Service, filtering on app.infrahq.com/component
+func (k *Kubernetes) Service(labels ...string) (*corev1.Service, error) {
 	clientset, err := kubernetes.NewForConfig(k.Config)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


internal/kubernetes/kubernetes.go
```diff
@@ -536,21 +536,15 @@ func (k *Kubernetes) Service(component string, labels ...string) (*corev1.Servic
 		return nil, err
 	}
 
-	selector := []string{
-		fmt.Sprintf("app.infrahq.com/component=%s", component),
-	}
```
dnephin, line 541: Doesn't this cause a problem for on-prem when someone does try to deploy them both in the same cluster, because it could match any component (server, ui, etc). Shouldn't this  change to hardcode `connector` as the component here, instead of removing this selector?

internal/kubernetes/kubernetes.go
```diff
@@ -630,7 +624,7 @@ func (k *Kubernetes) Endpoint() (string, int, error) {
 		return k.NodePort(service)
 	case corev1.ServiceTypeLoadBalancer:
 		if len(service.Status.LoadBalancer.Ingress) == 0 {
-			return "", -1, fmt.Errorf("load balancer has no ingress objects")
+			return "", -1, fmt.Errorf("no address available for load balancer")
```
dnephin, line 627: Maybe something to indicate it may still resolve itself? Ex:

> no address available for load balancer yet, it may still be provisioning

Or something along those lines

dnephin (commented):
Nice! Thanks for making this change. One question and one suggestion on the improved error message

internal/kubernetes/kubernetes.go
```diff
@@ -525,7 +525,7 @@ func readNamespaceFromInClusterFile() (string, error) {
 }
 
 // Find the first suitable Service, filtering on infrahq.com/component
-func (k *Kubernetes) Service(component string, labels ...string) (*corev1.Service, error) {
+func (k *Kubernetes) Service(labels ...string) (*corev1.Service, error) {
```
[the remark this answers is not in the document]
dnephin: I suspect we still want it, for the case where someone does happen to be running both the server and connector in the same cluster.

internal/kubernetes/kubernetes.go
mxyng: I actually want to remove this label from the chart. The scenario you're describing is not possible if the server and connector are two separate Helm deployments. I can keep this until that separation is enforced.

PR #3384, opened 2022-10-04
feat(connector): create destination as soon as possible

mxyng:
## Summary

Update the connector to create itself as quickly as possible instead of waiting for a valid connection. This is mainly due to LoadBalancer potentially taking a significant amount of time to initialize. Creating the destination quickly allows the user to continue without waiting.

Update UI and CLI to output status == "Pending" when the destination is disconnected and does not have a connection URL. Omit pending and disconnected destinations when syncing kubeconfig

### DIFF
--- a/api/destination.go
+++ b/api/destination.go
@@ -30,7 +30,7 @@ type DestinationConnection struct {
 
 func (r DestinationConnection) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
-		validate.Required("url", r.URL),
+		validate.Required("ca", r.CA),
 	}
 }
 
--- a/internal/cmd/destinations.go
+++ b/internal/cmd/destinations.go
@@ -14,6 +14,7 @@ import (
 
 const (
 	DestinationStatusConnected    = "Connected"
+	DestinationStatusPending      = "Pending"
 	DestinationStatusDisconnected = "Disconnected"
 )
 
@@ -79,8 +80,13 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
 
 				var rows []row
 				for _, d := range destinations {
-					status := DestinationStatusDisconnected
-					if d.Connected {
+					var status string
+					switch {
+					case !d.Connected:
+						status = DestinationStatusDisconnected
+					case d.Connection.URL == "":
+						status = DestinationStatusPending
+					default:
 						status = DestinationStatusConnected
 					}
 
--- a/internal/cmd/list_test.go
+++ b/internal/cmd/list_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"net/http"
 	"path/filepath"
 	"testing"
 	"time"
@@ -74,6 +75,19 @@ func TestListCmd(t *testing.T) {
 		},
 	})
 	assert.NilError(t, err)
+
+	for _, uniqueID := range []string{"space", "moon", "maintain"} {
+		// set client.Headers so each destination becomes connected
+		c.Headers = http.Header{
+			"Infra-Destination": {uniqueID},
+		}
+
+		_, err := c.ListGrants(api.ListGrantsRequest{})
+		assert.NilError(t, err)
+	}
+
+	// reset client.Headers
+	c.Headers = http.Header{}
 
 	users, err := c.ListUsers(api.ListUsersRequest{})
 	assert.NilError(t, err)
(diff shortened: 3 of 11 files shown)


ui/pages/destinations/index.js
```diff
@@ -112,12 +112,18 @@ export default function Destinations() {
                 <div
                   className={`h-2 w-2 flex-none rounded-full border ${
                     info.getValue()
-                      ? ' border-teal-500/50 bg-teal-400'
+                      ? 'border-teal-500/50 bg-teal-400'
+                      : info.row.original.connection.url === ''
+                      ? 'border-yellow-500 bg-yellow-500'
```
jmorganca, line 117:
```suggestion
                      ? 'border-yellow-600 bg-yellow-500'
```

internal/cmd/destinations.go
```diff
@@ -14,6 +14,7 @@ import (
```
jmorganca, line 14: This is looking pretty good. Should we filter out entries from `infra list` as they will be pending and inaccessible?

mxyng: Possibly though I'm wary of hiding pending destinations as it has a similar problem as not creating the destination until it can be connected.

BruceMacD: I feel list they should still have a pending label in `infra list`, otherwise they will get added to the kube config before they can be accessed (and that could take a while) which is confusing to me

internal/connector/connector.go
```diff
@@ -297,23 +297,26 @@ func httpTransportFromOptions(opts ServerOptions) *http.Transport {
 func syncDestination(con connector) error {
 	endpoint, err := getEndpointHostPort(con.k8s, con.options)
 	if err != nil {
-		return err
+		logging.L.Warn().Err(err).Msg("could not get host")
```
BruceMacD, line 300: maybe some more info here on why not getting the host could be happening would be useful, looks like something that is unexpected otherwise

api/destination.go
```diff
@@ -30,7 +30,7 @@ type DestinationConnection struct {
 
 func (r DestinationConnection) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
-		validate.Required("url", r.URL),
+		validate.Required("ca", r.CA),
```
BruceMacD, line 33: Could this requirement break old connectors?

internal/server/models/destination.go
```diff
@@ -26,7 +26,7 @@ func (d *Destination) ToAPI() *api.Destination {
 	connected := false
 	// TODO: this should be configurable
 	// https://github.com/infrahq/infra/issues/2505
-	if time.Since(d.LastSeenAt) < 5*time.Minute {
+	if time.Since(d.LastSeenAt) < 5*time.Minute && d.ConnectionURL != "" {
```
BruceMacD, line 29: Does this mean pending destinations never timeout and disconnect?

api/destination.go
mxyng: No, current connection already consistently pass Connection.CA so this doesn't really do anything. Alternatively, I can remove the validation rules entirely.

internal/server/models/destination.go
mxyng: That's right. Destinations will be in the Pending state until they report a connection URL. This might change as I can see it leading to some confusion. Instead, the status should be 1) Disconnected if not seen within last X minutes, 2) Pending if seen but there's no connection string, and 3) Connected if seen and has a connection string. Will update the UI as well.

internal/cmd/destinations.go
jmorganca: For now let's avoid putting them in `infra list` and generating a kubeconfig for pending clusters. Users can always run `infra destinations list`

BruceMacD (approved):

internal/cmd/list_test.go
```diff
@@ -75,6 +76,19 @@ func TestListCmd(t *testing.T) {
 	})
 	assert.NilError(t, err)
 
+	for _, uniqueID := range []string{"space", "moon", "maintain"} {
+		// set client.Headers so each destination becomes connected
```
dnephin, line 80: I had this problem as well in #3391, I thought it might make sense to have `CreateDestination` mark it as "last seen at", like this: https://github.com/infrahq/infra/pull/3391/files#diff-487038165470a56e27ec73856c4cffb0b0808e9d00fd6a3481dd59d24f4e150c

PR #3385, opened 2022-10-04
improve: table revisit

hoyyeva:
## Summary
couple improvement on the tables
- made the interaction more consistent (other than the users page - since we need more time to design and implement the user detail page)
- modified the provider edit page
- modified the width of the columns so that the first column of the table has same length throughout the app
- removed `added` columns from all the tables because it is not a very useful information for users

## 📹 
https://user-images.githubusercontent.com/63033505/193854442-ef0acc38-a05b-4a22-a140-461350093154.mov


jmorganca (approved):
LGTM

PR #3391, opened 2022-10-04
connector: Use a separate goroutine for update destination

dnephin:
This should allow us to have different life-cycles for each of these operations.

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -12,12 +12,12 @@ import (
 	"net/http/httputil"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/goware/urlx"
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
 	rbacv1 "k8s.io/api/rbac/v1"
 
 	"github.com/infrahq/infra/api"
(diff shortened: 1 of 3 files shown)


internal/connector/connector.go
```diff
@@ -193,15 +192,38 @@ func Run(ctx context.Context, options Options) error {
...
+			}
+		}
+	})
+	group.Go(func() error {
+		for {
+			if err := syncDestination(con); err != nil {
+				logging.Errorf("failed to update destination in infra: %v", err)
+			}
+			// TODO: should this exit after the first successful update?
+			wait(ctx, 30*time.Second)
```
dnephin, line 221: I'm wondering what the right behaviour here should be. I think there's a few different cases:

1. the sync failed - we should probably retry pretty quickly, but maybe with some exponential backoff (start at 2s retry, with a max backoff of 1 minute?)
2. the sync was successful, but there's no `Connection.URL` yet (relevant once we merge #3384) - probably the same behaviour as above. Short delay with some backoff
3. the sync was successful, and there is a `Connection.URL` - do we actually want to continue to check for updates in this scenario? I guess it needs to detect new namespaces and roles being added to the cluster? So maybe a 5 minute sleep?

mxyng: The URL could also change. If the service is using NodePort, the original node may have scaled down and replaced with another node. If the service is LoadBalancer, the load balancer may have been updated.

mxyng (approved):

PR #3398, opened 2022-10-05
fix: delete users from specific identity provider

### LINKED ISSUE #2838: Consolidate duplicate code paths for deleting users

### LINKED ISSUE #3397: When an identity is soft-deleted provider user records should be removed
## Describe the bug
When an identity is deleted we don't clean up the "provider_user" records that correspond with this identity. These records are used to track where an identity exists, but do not provide an access.

## Expected behavior
When the identity record that corresponds to a "provider_user" record is deleted we should also remove the "provider_user" tracking.

## Environment

```
$ infra version
Client: 0.15.2
```

BruceMacD:
- infra users command adds/removes from the infra identity provider specifically
- consolidate identity deletion logic
- delete identity when it no longer exists in any identity provider

## Summary
When a user is deleted using `infra users add` or `infra users rm` only delete the resources associated with the identity through the Infra provider. Users from other identity providers can be removed by removing the associated provider, or through SCIM (soon).

## Checklist

## Related Issues

Resolves #3397
Resolves #2838

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -39,7 +39,6 @@ func CreateIdentity(c *gin.Context, identity *models.Identity) error {
 	return data.CreateIdentity(db, identity)
 }
 
-// TODO (https://github.com/infrahq/infra/issues/2318) remove provider user, not user.
 func DeleteIdentity(c *gin.Context, id uid.ID) error {
 	rCtx := GetRequestContext(c)
 	self, err := isIdentitySelf(c, id)
@@ -60,39 +59,7 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
 		return HandleAuthErr(err, "user", "delete", models.InfraAdminRole)
 	}
 
-	if err := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByIssuedForID: id}); err != nil {
-		return fmt.Errorf("delete identity access keys: %w", err)
-	}
-
-	groups, err := data.ListGroups(db, nil, data.ByGroupMember(id))
-	if err != nil {
-		return fmt.Errorf("list groups for identity: %w", err)
-	}
-	for _, group := range groups {
-		err = data.RemoveUsersFromGroup(db, group.ID, []uid.ID{id})
-		if err != nil {
-			return fmt.Errorf("delete group membership for identity: %w", err)
-		}
-	}
-	// if an identity does not have credentials in the Infra provider this won't be found, but we can proceed
-	credential, err := data.GetCredential(db, data.ByIdentityID(id))
-	if err != nil && !errors.Is(err, internal.ErrNotFound) {
-		return fmt.Errorf("get delete identity creds: %w", err)
-	}
-
-	if credential != nil {
-		err := data.DeleteCredential(db, credential.ID)
-		if err != nil {
-			return fmt.Errorf("delete identity creds: %w", err)
-		}
-	}
-
-	err = data.DeleteGrants(db, data.DeleteGrantsOptions{BySubject: uid.NewIdentityPolymorphicID(id)})
-	if err != nil {
-		return fmt.Errorf("delete identity creds: %w", err)
-	}
-
-	return data.DeleteIdentity(db, id)
+	return data.DeleteIdentities(db, data.InfraProvider(db).ID, data.ByID(id))
 }
 
 func ListIdentities(c *gin.Context, name string, groupID uid.ID, ids []uid.ID, showSystem bool, p *data.Pagination) ([]models.Identity, error) {
--- a/internal/access/identity_test.go
+++ b/internal/access/identity_test.go
@@ -98,6 +98,9 @@ func TestDeleteIdentityCleansUpResources(t *testing.T) {
 	_, err = data.GetIdentity(db, data.ByID(identity.ID))
 	assert.ErrorIs(t, err, internal.ErrNotFound)
 
+	_, err = data.GetProviderUser(db, infraProvider.ID, identity.ID)
+	assert.ErrorIs(t, err, internal.ErrNotFound)
+
 	_, err = data.GetAccessKeyByKeyID(db, keyID)
 	assert.ErrorIs(t, err, internal.ErrNotFound)
 
--- a/internal/access/identity_test.go
+++ b/internal/access/identity_test.go
@@ -53,6 +53,9 @@ func TestDeleteIdentityCleansUpResources(t *testing.T) {
 	err := data.CreateIdentity(db, identity)
 	assert.NilError(t, err)
 
+	_, err = data.CreateProviderUser(db, infraProvider, identity)
+	assert.NilError(t, err)
+
 	// create some resources for this identity
 
 	keyID := generate.MathRandom(models.AccessKeyKeyLength, generate.CharsetAlphaNumeric)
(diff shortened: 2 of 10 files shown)


internal/access/identity.go
```diff
@@ -60,39 +59,7 @@ func DeleteIdentity(c *gin.Context, id uid.ID) error {
...
-		}
-	}
-
-	err = data.DeleteGrants(db, data.DeleteGrantsOptions{BySubject: uid.NewIdentityPolymorphicID(id)})
-	if err != nil {
-		return fmt.Errorf("delete identity creds: %w", err)
-	}
-
-	return data.DeleteIdentity(db, id)
+	return data.DeleteIdentities(db, data.InfraProvider(db).ID, data.ByID(id))
```
dnephin, line 62: I think this change also fixes #2838

internal/server/data/identity.go
```diff
@@ -161,31 +163,61 @@ func DeleteIdentities(tx GormTxn, selectors ...SelectorFunc) error {
...
+	return nil
+}
+
+func deleteReferencesToIdentities(tx GormTxn, providerID uid.ID, toDelete []models.Identity) (unreferencedIdentityIDs []uid.ID, err error) {
+	for _, i := range toDelete {
+		if err := DeleteAccessKeys(tx, DeleteAccessKeysOptions{ByIssuedForID: i.ID, ByProviderID: providerID}); err != nil {
+			return []uid.ID{}, fmt.Errorf("delete identity access keys: %w", err)
+		}
+		// if an identity does not have credentials in the Infra provider this won't be found, but we can proceed
+		credential, err := GetCredential(tx, ByIdentityID(i.ID))
```
dnephin, line 184: I'm not sure if this is correct. I believe we can get here from `data.DeleteProvider`, and if someone deletes an `okta` provider, won't this now delete the users infra credentials as well (because `data.DeleteIdentities` removes credentials by identity ID) ?

dnephin (commented):
Looks great! I think this will resolves a couple problems.

There are a lot of calls to `DeleteIdentity` so I think I'll need another pass to think through all those scenarios and see how it matches up to our test coverage.  Left a few questions from my first pass.

internal/access/identity.go
BruceMacD: Didn't know this was out there, this PR will close this issue also

internal/server/config.go
```diff
@@ -902,6 +902,12 @@ func (s Server) loadUser(db data.GormTxn, input User) (*models.Identity, error)
 		if err := data.CreateIdentity(db, identity); err != nil {
 			return nil, err
 		}
+
+		_, err = data.CreateProviderUser(db, data.InfraProvider(db), identity)
```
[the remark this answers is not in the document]
BruceMacD: `CreateCredential` creating the provider user would make a lot of sense, but the limitation there at the moment is that user can be declared in grants without having a credential linked to them.

Ex:
```
infra grants add dne@example.com infra --role admin --force
Created user "dne@example.com"
Created grant to "infra" for "dne@example.com"
```

internal/access/identity_test.go
```diff
@@ -53,6 +53,9 @@ func TestDeleteIdentityCleansUpResources(t *testing.T) {
 	err := data.CreateIdentity(db, identity)
 	assert.NilError(t, err)
 
+	_, err = data.CreateProviderUser(db, infraProvider, identity)
+	assert.NilError(t, err)
```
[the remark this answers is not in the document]
BruceMacD: The assertion that it is missing is down in line 98:
```
_, err = data.GetAccessKeyByKeyID(db, keyID)
assert.ErrorIs(t, err, internal.ErrNotFound)
```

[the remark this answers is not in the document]
dnephin: How does the access key not found relate to the provider user?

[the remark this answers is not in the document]
BruceMacD: Apologies, I was on the wrong branch locally, so I looked at the wrong line number here. I added the assertion that the provider user is deleted to this test.

internal/server/data/identity.go
BruceMacD: Good catch, added a check to not delete credentials when not removing an identity from the infra provider and test for this case.

internal/server/data/identity_test.go
```diff
@@ -143,31 +145,91 @@ var cmpModelsIdentityShallow = cmp.Comparer(func(x, y models.Identity) bool {
...
+					Kind: models.ProviderKindOIDC,
+				}
+				err = CreateProvider(tx, provider)
+				assert.NilError(t, err)
+
+				_, err = CreateProviderUser(tx, provider, id)
+				assert.NilError(t, err)
+
+				return provider.ID, *id
+			},
```
dnephin, line 206: Do we have any test coverage to show that groups and grants are removed when it's the last providerUser, and not removed when it's not the last user?

internal/server/data/identity.go
```diff
@@ -161,31 +163,63 @@ func DeleteIdentities(tx GormTxn, selectors ...SelectorFunc) error {
...
+		// if this identity only exists from an infra identity provider, remove all their groups and grants, then delete the user
+		user, err := GetIdentity(tx, Preload("Providers"), ByID(i.ID))
 		if err != nil {
-			return err
+			return []uid.ID{}, fmt.Errorf("check user providers: %w", err)
 		}
 
-		for _, group := range groups {
-			err = RemoveUsersFromGroup(tx, group.ID, []uid.ID{i.ID})
+		if len(user.Providers) == 0 {
```
dnephin, line 206: I don't think I follow this comment. This happens when they are removed from the last provider (regardless of which provide it was), doesn't it? Or because we don't call this from `UpdateIdentityInfoFromProvider` then presumably the last provider was infra?

dnephin (approved):
`DeleteIdentities` does a lot of things, and I think we might be missing assertions for the conditional removal of grants and groups. With that test coverage I think this looks great!

Should `access.UpdateIdentityInfoFromProvider` be calling `data.DeleteIdentities` now to clean up? It's currently only deleting the access key and provider_user, but I would expect it to need to do more if that was the last provider.

internal/server/data/identity.go
BruceMacD: This was a bad comment, fixed

PR #3399, opened 2022-10-05
fix: remove provider users for deleted identities

### LINKED ISSUE #3397: When an identity is soft-deleted provider user records should be removed

BruceMacD:
## Summary
This should be merged after #3398
When an identity was deleted through the CLI the associated provider user was not deleted. This migration cleans up provider_user records that reference a deleted identity.

## Checklist

## Related Issues

Resolves #3397

### DIFF
--- a/internal/server/data/migrations_test.go
+++ b/internal/server/data/migrations_test.go
@@ -775,10 +775,10 @@ DELETE FROM settings WHERE id=24567;
 				assert.NilError(t, err)
 			},
 			expected: func(t *testing.T, db WriteTxn) {
-				var remainingProviderUser uid.ID
-				err := db.QueryRow(`SELECT identity_id FROM provider_users`).Scan(&remainingProviderUser)
+				var count int
+				err := db.QueryRow(`SELECT count(*) from provider_users where identity_id = 1111`).Scan(&count)
 				assert.NilError(t, err)
-				assert.Equal(t, remainingProviderUser, uid.ID(2222))
+				assert.Equal(t, count, 0)
 			},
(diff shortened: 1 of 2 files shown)


internal/server/data/migrations_test.go
```diff
@@ -728,6 +728,29 @@ DELETE FROM settings WHERE id=24567;
...
+				_, err := db.Exec("DELETE FROM identities")
+				assert.NilError(t, err)
+				_, err = db.Exec("DELETE FROM provider_users")
+				assert.NilError(t, err)
+			},
+			expected: func(t *testing.T, db WriteTxn) {
+				var remainingProviderUser uid.ID
+				err := db.QueryRow(`SELECT identity_id FROM provider_users`).Scan(&remainingProviderUser)
+				assert.NilError(t, err)
+				assert.Equal(t, remainingProviderUser, uid.ID(2222))
```
dnephin, line 751: minor: I think this test could pass even when the first record is not deleted, if the DB decided to return the records in a different order than we expect. I'm not sure if that's possible. 
Another option might be to `SELECT count(*) from provider_users where identity_id = 1111` which should give us a count of 0.

dnephin (approved):
LGTM

PR #3402, opened 2022-10-06
feat: add endpoint to get a user's own organization

pdevine:
## Summary

This change modifies the `GET /api/organizations/:id` endpoint to now support `GET /api/organizations/self` similar to how `GET /api/identities/self` works. The user does not need to be an infra support admin to retrieve their own organization details.

Additionally, the `created` and `updated` fields inside of the JSON response are now being properly updated (instead of returning `null` values), and a unit test for the endpoint has been added.

## Checklist

### DIFF
--- a/api/organization.go
+++ b/api/organization.go
@@ -13,6 +13,10 @@ type Organization struct {
 	Domain  string `json:"domain"`
 }
 
+type GetOrganizationRequest struct {
+	ID IDOrSelf `uri:"id"`
+}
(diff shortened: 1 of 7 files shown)


jmorganca (approved):

BruceMacD (approved):

ssoroka (commented):
might be simpler to add organization id to the identity/self endpoint?

PR #3403, opened 2022-10-06
fix postgres-dev to only listen on localhost

dnephin:
## Summary

The default for docker port mapping is `0.0.0.0` (all interfaces), which can be a security risk. If a developer laptop is ever connected to an untrusted network, someone on that network could connect to the database and use `COPY` and other postgres operations to read files or even run processes on the laptop.

By mapping the host port to 127.0.0.1 we ensure that no one can connect to the dev database except for the processes on the local host.

If you are using Docker for Mac I could use your help to confirm this works with that setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,9 @@ postgres:
 	docker run -d --name=postgres-dev --rm \
 		-e POSTGRES_PASSWORD=password123 \
 		--tmpfs=/var/lib/postgresql/data \
-		-p 15432:5432 \
-		postgres:14-alpine -c fsync=off -c full_page_writes=off
+		-p 127.0.0.1:15432:5432 \
+		postgres:14-alpine -c fsync=off -c full_page_writes=off \
+			-c listen_addresses=127.0.0.1 -c max_connections=100
 	@echo
 	@echo Copy the line below into the shell used to run tests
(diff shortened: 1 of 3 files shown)


dnephin:
I missed the `docker-compose.dev.yml` which I think needs the same fix, I'll do that now

Makefile
```diff
@@ -36,8 +36,9 @@ postgres:
 	docker run -d --name=postgres-dev --rm \
 		-e POSTGRES_PASSWORD=password123 \
 		--tmpfs=/var/lib/postgresql/data \
-		-p 15432:5432 \
-		postgres:14-alpine -c fsync=off -c full_page_writes=off
+		-p 127.0.0.1:15432:5432 \
+		postgres:14-alpine -c fsync=off -c full_page_writes=off \
+			-c listen_addresses=127.0.0.1 -c max_connections=100
```
dnephin, line 41: These are the default settings, but I'd like to include them here to make it easier to change them.

On linux setting `listen_addresses` to `0.0.0.0` allows tests to bypass the port mapping proxy and connect directly to the container.  Setting `max_connections` can be useful when testing concurrency.

BruceMacD (approved):
Confirmed this works on Docker Desktop for Mac

dnephin:
Thanks!

PR #3404, opened 2022-10-06
improve: org name and notification

hoyyeva:
## Summary
- improve notification component to clear the timeout when unmount and cancel the timer when click away
- add org name to the profile dropdown menu

## 📷 
Screen Shot 2022-10-06 at 3 18 41 PM

### DIFF
--- a/ui/pages/destinations/[id]/index.js
+++ b/ui/pages/destinations/[id]/index.js
@@ -120,7 +120,7 @@ export default function DestinationDetail() {
       </Head>
       <header className='mt-6 mb-12 space-y-4'>
         <div className=' flex flex-col justify-between md:flex-row md:items-center'>
-          <h1 className='flex truncate py-1 font-display text-xl font-medium'>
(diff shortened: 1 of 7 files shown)


ui/pages/destinations/[id]/index.js
```diff
@@ -120,7 +120,7 @@ export default function DestinationDetail() {
       </Head>
       <header className='mt-6 mb-12 space-y-4'>
         <div className=' flex flex-col justify-between md:flex-row md:items-center'>
-          <h1 className='flex truncate py-1 font-display text-xl font-medium'>
```
Comment (jmorganca) [inline ui/pages/destinations/[id]/index.js:123]: Why 75%? Is the idea to add a buffer on the right to truncate?

jmorganca (approved):

ui/pages/destinations/[id]/index.js
Comment (hoyyeva) [reply ui/pages/destinations/[id]/index.js:123]: yup that is the reason why we add 75% there

PR #3406, opened 2022-10-06
fix: edit provider only update changed fields

mxyng:
## Summary

The client secret field has an initial value which gets passed to the API request even if it's unchanged. This can lead to unexpected changes to the client secret if only the name is updated.

### DIFF
--- a/ui/pages/providers/[id].js
+++ b/ui/pages/providers/[id].js
@@ -9,8 +9,6 @@ import Dashboard from '../../components/layouts/dashboard'
 import RemoveButton from '../../components/remove-button'
 import Notification from '../../components/notification'
 
-const CLIENT_SECRET_INIT = '***********'
-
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):

ui/pages/providers/[id].js
```diff
@@ -210,7 +208,8 @@ export default function ProvidersEditDetails() {
                 Client Secret
               </label>
               <input
-                type='password'
+                type='search'
```
Comment (mxyng) [inline ui/pages/providers/[id].js:211]: This change will disable password manager integration such as autofill and saving to the password manager but unhides the client secret while it's being edited

hoyyeva (approved):

ui/pages/providers/[id].js
Comment (mxyng) [reply ui/pages/providers/[id].js]: Keeping this as `type=password` for now

PR #3407, opened 2022-10-06
feat: create access keys for idp scim

BruceMacD:
## Summary
Allow creating access keys for identity providers to make SCIM requests. This change only adds CLI support, waiting for full user SCIM support to add to the UI.

Example CLI interaction.
```
$ infra providers add okta --url dev-123.okta.com --client-id xyz --client-secret xxx --kind okta --scim
Connected provider "okta" (dev-123.okta.com) to infra
Access key for SCIM provisioning: aaaaaaaa.bbbbbbbbb

$ infra providers edit okta --scim
New access key for SCIM provisioning: aaaaaa.bbbbbbbb
```

## Checklist

## Related Issues

Part of #3378

### DIFF
--- a/internal/cmd/providers.go
+++ b/internal/cmd/providers.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/yaml"
@@ -65,16 +66,17 @@ func (o providerAPIOptions) Validate(providerKind string) error {
 
 type providerEditOptions struct {
 	ClientSecret       string
+	SCIM               bool
 	ProviderAPIOptions providerAPIOptions
 }
 
 func (o providerEditOptions) Validate(providerKind string) error {
-	if o.ClientSecret == "" && o.ProviderAPIOptions.PrivateKey == "" && o.ProviderAPIOptions.ClientEmail == "" && o.ProviderAPIOptions.WorkspaceDomainAdminEmail == "" {
-		return fmt.Errorf("Please specify a field to update.'\n\n%s", newProvidersEditCmd(nil).UsageString())
+	if !o.SCIM && o.ClientSecret == "" {
+		return fmt.Errorf("scim or client-secret flag are required for updates")
 	}
 
-	if providerKind != "google" && o.ClientSecret == "" {
-		return fmt.Errorf("Client secret flag must be specified when updating an identity provider that isn't of kind Google")
+	if o.ClientSecret == "" && o.ProviderAPIOptions.PrivateKey == "" && o.ProviderAPIOptions.ClientEmail == "" && o.ProviderAPIOptions.WorkspaceDomainAdminEmail == "" {
+		return fmt.Errorf("client secret, private key, client email, and workspace domain admin email are required.'\n\n%s", newProvidersEditCmd(nil).UsageString())
 	}
 
 	return o.ProviderAPIOptions.Validate(providerKind)
@@ -92,13 +94,14 @@ $ infra providers edit okta --client-secret VT_oXtkEDaT7UFY-C3DSRWYb00qyKZ1K1VCq
 # Connect Google to Infra with group sync
 $ infra providers edit google --client-secret VT_oXtkEDaT7UFY-C3DSRWYb00qyKZ1K1VCq7YzN --service-account-key ~/client-123.json --service-account-email hello@example.com --workspace-domain-admin admin@example.com
 `,
-		Args: ExactArgs(1),
+		Args: MaxArgs(5),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return updateProvider(cli, args[0], opts)
 		},
 	}
 
 	cmd.Flags().StringVar(&opts.ClientSecret, "client-secret", "", "Set a new client secret")
+	cmd.Flags().BoolVar(&opts.SCIM, "scim", false, "Create a new access key for SCIM provisioning")
 	cmd.Flags().Var((*types.StringOrFile)(&opts.ProviderAPIOptions.PrivateKey), "service-account-key", "The private key used to make authenticated requests to Google's API")
 	cmd.Flags().StringVar(&opts.ProviderAPIOptions.ClientEmail, "service-account-email", "", "The email assigned to the Infra service client in Google")
 	cmd.Flags().StringVar(&opts.ProviderAPIOptions.WorkspaceDomainAdminEmail, "workspace-domain-admin", "", "The email of your Google workspace domain admin")
@@ -168,6 +171,7 @@ type providerAddOptions struct {
 	ClientID           string
 	ClientSecret       string
 	Kind               string
+	SCIM               bool
 	ProviderAPIOptions providerAPIOptions
 }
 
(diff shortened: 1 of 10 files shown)


internal/cmd/providers.go
```diff
@@ -65,16 +66,17 @@ func (o providerAPIOptions) Validate(providerKind string) error {
...
-	if o.ClientSecret == "" && o.ProviderAPIOptions.PrivateKey == "" && o.ProviderAPIOptions.ClientEmail == "" && o.ProviderAPIOptions.WorkspaceDomainAdminEmail == "" {
-		return fmt.Errorf("Please specify a field to update.'\n\n%s", newProvidersEditCmd(nil).UsageString())
+	if !o.SCIM && o.ClientSecret == "" {
+		return fmt.Errorf("scim or client-secret flag are required for updates")
 	}
 
-	if providerKind != "google" && o.ClientSecret == "" {
-		return fmt.Errorf("Client secret flag must be specified when updating an identity provider that isn't of kind Google")
+	if o.ClientSecret == "" && o.ProviderAPIOptions.PrivateKey == "" && o.ProviderAPIOptions.ClientEmail == "" && o.ProviderAPIOptions.WorkspaceDomainAdminEmail == "" {
+		return fmt.Errorf("client secret, private key, client email, and workspace domain admin email are required.'\n\n%s", newProvidersEditCmd(nil).UsageString())
```
dnephin, line 79: Should the condition be `||` instead of `&&` ? The error message reads like all of the values are required, but the condition is "one of" them is required.

I wonder if we should do the `newProvidersEditCmd(nil).UsageString()` printing from the caller. I guess any validation of options would benefit from the usage text.

internal/cmd/providers.go
```diff
@@ -92,13 +94,14 @@ $ infra providers edit okta --client-secret VT_oXtkEDaT7UFY-C3DSRWYb00qyKZ1K1VCq
 # Connect Google to Infra with group sync
 $ infra providers edit google --client-secret VT_oXtkEDaT7UFY-C3DSRWYb00qyKZ1K1VCq7YzN --service-account-key ~/client-123.json --service-account-email hello@example.com --workspace-domain-admin admin@example.com
 `,
-		Args: ExactArgs(1),
+		Args: MaxArgs(5),
```
dnephin, line 97: Where do these extra 4 arguments get read? In `RunE` below it seems like we only grab `args[0]` still.

internal/cmd/providers.go
```diff
@@ -245,6 +249,26 @@ $ infra providers add google --url accounts.google.com --client-id 0oa3sz06o6do0
...
+					TTL:               api.Duration(time.Hour * 87600), // 10 years
+					ExtensionDeadline: api.Duration(time.Hour * 87600), // 10 years
+				})
+				if err != nil {
+					if api.ErrorStatusCode(err) == 403 {
+						logging.Debugf("%s", err.Error())
+						return Error{
+							Message: "Cannot create SCIM key: missing privileges for CreateKey",
+						}
+					}
```
dnephin, line 266: I think we don't need this anymore because we return pretty reasonable errors from the api, but I'm ok if we want to keep it.

internal/server/data/provider.go
```diff
@@ -86,6 +86,10 @@ func DeleteProviders(db GormTxn, selectors ...SelectorFunc) error {
 		if err := DeleteAccessKeys(db, DeleteAccessKeysOptions{ByProviderID: p.ID}); err != nil {
 			return fmt.Errorf("delete access keys: %w", err)
 		}
+
+		if err := DeleteAccessKeys(db, DeleteAccessKeysOptions{ByIssuedForID: p.ID}); err != nil {
```
dnephin, line 90:
```suggestion
		// delete any access keys used for SCIM
		if err := DeleteAccessKeys(db, DeleteAccessKeysOptions{ByIssuedForID: p.ID}); err != nil {
```

internal/cmd/providers.go
```diff
@@ -245,6 +249,26 @@ $ infra providers add google --url accounts.google.com --client-id 0oa3sz06o6do0
 			}
 
 			cli.Output("Connected provider %q (%s) to infra", args[0], opts.URL)
+
+			if opts.SCIM {
+				key, err := client.CreateAccessKey(&api.CreateAccessKeyRequest{
+					UserID:            provider.ID,
```
dnephin, line 255: This call to create access key , and the one below in `updateProvider` both set the `ProviderID` to the infra provider, right?  

Could we set the `ProviderID` on this key to the scim provider as well? That would give us an easy way of finding these access keys that are for provider scim without having to first do the user lookup and fallback to the provider lookup.

dnephin (commented):
Looks good! I'm always a little weary of overloading a field, but I think this is a reasonable way of avoiding the `scim` user.

It could be surprising that we store either a providerId or UserId as the accessKey.IssuedFor. I think `models.AccessKey.IssusedFor` could use some more godoc to explain that the ID could be one or the other, and a bit about why.

Left a couple other suggestions for how to make this a bit more obvious below.

internal/cmd/providers.go
BruceMacD: The ProviderID here indicates who granted the access key so we can' use it to prevent the fallback lookup. For example if a user signed in using Okta the provider ID would the Okta provider's ID, but we don't want them to be able to provision users within the Okta provider.

dnephin: We can't use that field on its own, but I was wondering if we could do something like:

```go
if accessKey.ProviderID == accessKey.IssusedFor { ... }
```

That gives us a way to identify the skim keys without having to compare the row to the user table. 
* any key where the providerId and issusedFor match is a scim key
* any key where they do not match is a regular user key

Otherwise we have to compare the field to another table to determine what kind of key it is.

internal/cmd/providers.go
BruceMacD: This edit command validation was wrong, fixed it

internal/cmd/providers.go
BruceMacD: This was a mistake, I was thinking of the flags as args, reverted back to 1

internal/cmd/providers.go
BruceMacD: No reason to keep it around if its not useful, removed

internal/cmd/providers.go
BruceMacD: Clever, I'll add this

BruceMacD: Made this change:
https://github.com/infrahq/infra/pull/3407/commits/1e5d87d69eb362b6fbb4a3ef2b8126b609874ef2

internal/cmd/providers.go
```diff
@@ -92,13 +104,14 @@ $ infra providers edit okta --client-secret VT_oXtkEDaT7UFY-C3DSRWYb00qyKZ1K1VCq
 # Connect Google to Infra with group sync
 $ infra providers edit google --client-secret VT_oXtkEDaT7UFY-C3DSRWYb00qyKZ1K1VCq7YzN --service-account-key ~/client-123.json --service-account-email hello@example.com --workspace-domain-admin admin@example.com
 `,
-		Args: ExactArgs(1),
+		Args: MaxArgs(1),
```
dnephin, line 107: This needs to be `ExactArgs`, right? Because we use `args[0]` unconditionally, so no args will cause a panic.

dnephin (approved):
LGTM! Just one small thing on the arg validation

PR #3408, opened 2022-10-06
feat: parse scim query filters

BruceMacD:
## Summary
Parse SCIM query filters into SQL. 

Branched from #3405

## Checklist

## Related Issues

Part of #3378

### DIFF
--- a/internal/server/data/provideruser.go
+++ b/internal/server/data/provideruser.go
@@ -132,7 +132,7 @@ func ListProviderUsers(tx ReadTxn, opts ListProviderUsersOptions) ([]models.Prov
 	}
 	if opts.SCIMParameters != nil && opts.SCIMParameters.Filter != nil {
 		query.B("AND (")
-		err := filterSQL(*opts.SCIMParameters.Filter, query)
+		err := filterSQL(opts.SCIMParameters.Filter, query)
 		if err != nil {
 			return nil, fmt.Errorf("apply filter: %w", err)
 		}
@@ -246,8 +246,8 @@ func SyncProviderUser(ctx context.Context, tx WriteTxn, user *models.Identity, p
 }
 
 type SCIMParameters struct {
-	Count      int                // the number of items to return
-	StartIndex int                // the offset to start counting from
-	TotalCount int                // the total number of items that match the query
-	Filter     *filter.Expression // a filter to apply to the results
+	Count      int               // the number of items to return
+	StartIndex int               // the offset to start counting from
+	TotalCount int               // the total number of items that match the query
+	Filter     filter.Expression // a filter to apply to the results
 }
--- a/internal/server/data/provideruser.go
+++ b/internal/server/data/provideruser.go
@@ -6,13 +6,14 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/scim2/filter-parser/v2"
+
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data/querybuilder"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/internal/server/providers"
 	"github.com/infrahq/infra/uid"
-	"github.com/scim2/filter-parser/v2"
 )
 
 type providerUserTable models.ProviderUser
--- a/internal/server/data/provideruser.go
+++ b/internal/server/data/provideruser.go
@@ -12,6 +12,7 @@ import (
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/internal/server/providers"
 	"github.com/infrahq/infra/uid"
+	"github.com/scim2/filter-parser/v2"
 )
 
 type providerUserTable models.ProviderUser
@@ -128,6 +129,14 @@ func ListProviderUsers(tx ReadTxn, opts ListProviderUsersOptions) ([]models.Prov
 	if opts.HideInactive {
 		query.B("AND active = ?", opts.HideInactive)
 	}
+	if opts.SCIMParameters != nil && opts.SCIMParameters.Filter != nil {
+		query.B("AND (")
+		err := filterSQL(*opts.SCIMParameters.Filter, query)
+		if err != nil {
+			return nil, fmt.Errorf("apply filter: %w", err)
+		}
+		query.B(")")
+	}
 
 	query.B("ORDER BY email ASC")
 
@@ -236,8 +245,8 @@ func SyncProviderUser(ctx context.Context, tx WriteTxn, user *models.Identity, p
 }
 
 type SCIMParameters struct {
-	Count      int // the number of items to return
-	StartIndex int // the offset to start counting from
-	TotalCount int // the total number of items that match the query
-	// TODO: filter query param
+	Count      int                // the number of items to return
+	StartIndex int                // the offset to start counting from
+	TotalCount int                // the total number of items that match the query
+	Filter     *filter.Expression // a filter to apply to the results
 }
--- a/internal/server/data/provideruser_test.go
+++ b/internal/server/data/provideruser_test.go
@@ -2,10 +2,12 @@ package data
 
 import (
 	"context"
+	"fmt"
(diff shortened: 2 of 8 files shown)


internal/server/data/provideruser.go
```diff
@@ -104,9 +106,16 @@ func ListProviderUsers(tx ReadTxn, providerID uid.ID, p *SCIMParameters) ([]mode
...
 	query.B("INNER JOIN providers ON provider_users.provider_id = providers.id AND providers.organization_id = ?", tx.OrganizationID())
 	query.B("WHERE provider_id = ?", providerID)
-
+	if p != nil && p.Filter != nil {
+		filter, err := filterSQL(*p.Filter)
+		if err != nil {
+			return nil, fmt.Errorf("apply filter: %w", err)
+		}
+		query.B("AND (")
+		query.B(filter)
```
dnephin, line 115: To fix the warning from the querylinter, and ensure that we're escaping all the parameters, I think `filterSQL` and the functions it calls can accept a `query *querybuilder.Query`. 

That way the functions themselves can build up the query and use the second argument to `query.B` to add the untrusted input.

internal/server/data/scim.go
```diff
@@ -0,0 +1,69 @@
...
+	case c == filter.EQ:
+		return fmt.Sprintf("= '%s'", compare), nil
+	case c == filter.NE:
+		return fmt.Sprintf("!= '%s'", compare), nil
+	case c == filter.SW:
+		return fmt.Sprintf("LIKE '%s%%'", compare), nil
+	case c == filter.CO:
+		return fmt.Sprintf("LIKE '%%%s%%'", compare), nil
+	case c == filter.EW:
+		return fmt.Sprintf("LIKE '%%%s'", compare), nil
```
dnephin, line 66: I was confused about the extra `%%`, but I realize now it's to escape the `%` in the `fmt.Sprintf`. I think by passing in the query builder this can be:

```go
query.B("LIKE ?", "%"+compare)
```

This is using `%s`, so I'm guessing despite the parameter being `any`, it's always a string?

internal/server/data/scim.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	return supportedColumns[a.String()], nil
+}
+
+func sqlComparator(c filter.CompareOperator, compare any) (string, error) {
+	switch {
+	case c == filter.PR:
+		return "IS NOT NULL", nil
+	case c == filter.EQ:
+		return fmt.Sprintf("= '%s'", compare), nil
```
dnephin, line 58: If `sqlComparator` accepts a `query *querybuilder.Query` this can use parameter substitution like:

```go
    query.B("= ?", compare)
```

internal/server/data/scim.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			return "", fmt.Errorf("attribute path: %w", err)
+		}
+		return fmt.Sprintf("%s %s", column, comparison), nil
+	}
+	return "", fmt.Errorf("unable to parse filter, unrecognized format")
+}
+
+func sqlColumn(a filter.AttributePath) (string, error) {
+	if supportedColumns[a.String()] == "" {
```
dnephin, line 47: I think converting the map to a switch/case would allow this to pass string literals to the query builder:

```go
switch a.String() {
case "id":
    query.B("identity_id")

```

dnephin (commented):
Nice! This definitely looks like the most complicated part of the SCIM API so far. 

I took a quick look at the `scim2/filter-parser`. The test coverage looks a little light, so we should make sure we have good coverage for the main cases we care about.

Do the identity providers document which filters they use, or are we kind of forced to support the entire spec?

I think the approach you've taken here to translate the expression into SQL should work well. I left some comments for making it more resilient to injection attacks (and also appease the querylinter at the same time).

BruceMacD:
@dnephin 
> Do the identity providers document which filters they use, or are we kind of forced to support the entire spec?
They don't document the filters super clearly, the most common one seems to be the "email eq x" filter to see if a user exists yet. Many other inbound scim implementations dont implement all filters though, so this seems like a good base.

internal/server/scim.go
```diff
@@ -22,6 +25,13 @@ func ListProviderUsers(c *gin.Context, r *api.SCIMParametersRequest) (*api.ListP
 		StartIndex: r.StartIndex,
 		Count:      r.Count,
 	}
+	if r.Filter != "" {
+		exp, err := filter.ParseFilter([]byte(r.Filter))
```
dnephin, line 29: I couldn't find any fuzz tests in the scim2/filter-parser repo, so we should write some that exercise `filter.ParseFilter` There are a few examples of that for `uid.ID`.  I'd be happy to pair on that with you, if you're interested. 

I guess we mostly want to test that it doesn't panic, and that the parse finishes in a relatively short duration (effectively testing that it doesn't get stuck in some infinite loop on bad input).

internal/server/data/provideruser.go
```diff
@@ -236,8 +246,8 @@ func SyncProviderUser(ctx context.Context, tx WriteTxn, user *models.Identity, p
...
 
 type SCIMParameters struct {
-	Count      int // the number of items to return
-	StartIndex int // the offset to start counting from
-	TotalCount int // the total number of items that match the query
-	// TODO: filter query param
+	Count      int                // the number of items to return
+	StartIndex int                // the offset to start counting from
+	TotalCount int                // the total number of items that match the query
+	Filter     *filter.Expression // a filter to apply to the results
```
dnephin, line 252: `filter.Expression` is an interface, so I think this should not be a pointer. An interface allows us to do `!= nil` without it being a pointer.

```suggestion
	Filter     filter.Expression // a filter to apply to the results
```

internal/server/data/provideruser.go
```diff
@@ -128,6 +130,14 @@ func ListProviderUsers(tx ReadTxn, opts ListProviderUsersOptions) ([]models.Prov
 	if opts.HideInactive {
 		query.B("AND active = ?", opts.HideInactive)
 	}
+	if opts.SCIMParameters != nil && opts.SCIMParameters.Filter != nil {
+		query.B("AND (")
+		err := filterSQL(*opts.SCIMParameters.Filter, query)
```
dnephin, line 135:
```suggestion
		err := filterSQL(opts.SCIMParameters.Filter, query)
```

If we make this not a pointer we can remove the dereference here as well.

internal/server/scim.go
```diff
@@ -22,6 +25,13 @@ func ListProviderUsers(c *gin.Context, r *api.SCIMParametersRequest) (*api.ListP
...
+		if err != nil {
+			return nil, fmt.Errorf("parse SCIM filter expression: %w", err)
+		}
+		p.Filter = &exp
```
dnephin, line 33:
```suggestion
		p.Filter = exp
```

If we make this not a pointer we no longer need to take the address

internal/server/data/provideruser_test.go
```diff
@@ -327,6 +329,158 @@ func TestListProviderUsers(t *testing.T) {
...
+					Name: "mockta",
+					Kind: models.ProviderKindOkta,
+				}
+
+				err := CreateProvider(tx, provider)
+				assert.NilError(t, err)
+
+				_ = createTestProviderUser(t, tx, provider, "david@example.com")
+				pu := createTestProviderUser(t, tx, provider, "lucy@example.com")
+				filter, err := filter.ParseFilter([]byte(fmt.Sprintf("id eq \"%d\"", pu.IdentityID)))
```
dnephin, line 345: Not a blocker, just an fyi, I always try to use tilde "raw string literals" when the string needs to include double quotes, to avoid having to escape them:

```go
fmt.Sprintf(`id eq "%d"`, pu.IdentityID)
```

Raw strings don't allow `\n` and those kinds of escapes, but otherwise work the same way.

dnephin (approved):
Looks good! One suggestion for removing the pointer to interface

internal/server/scim.go
BruceMacD: Added a fuzz test that makes sure this completes in 5 seconds and does not panic, good suggestion

PR #3413, opened 2022-10-07
maintain: clean up architecture...and the rest

technovangelist:

## Summary

The architecture page had issues, so fixed them up. And then looked at active vs passive voice, better grammar throughout, and so much more.

## Checklist

## Related Issues

Resolves #


docs/manage/connectors/kubernetes.md
```diff
@@ -42,17 +42,48 @@ helm upgrade --install infra-connector infrahq/infra \
 
 {% callout type="info" %}
 
-Note: it may take a few minutes for the LoadBalancer to be provisioned.
+Note: it may take a few minutes for the Load Balancer to be provisioned.
```
mxyng, line 45: This shouldn't be changed since it's the Kubernetes resource name

docs/reference/cli.md
```diff
@@ -2,812 +2,835 @@
 title: CLI Reference
```
mxyng, line 2: This is generated so it shouldn't be manually changed

technovangelist: Oh, how is it generated? I can update that instead

mxyng: It's generated by `internal/docgen`

mchiang0610 (approved):
saw @mxyng's comment above per CLI docs need to be generated. Let's get that in. Otherwise this is a step forward, and we can make more iterations as we go.

PR #3414, opened 2022-10-07
Add test coverage of email templates 

dnephin:
## Summary

I was looking to make a change to some date formatting in one of the emails (to move it out of the `data` package), but I noticed that the change I was going to make would not be covered by tests.

This PR improves our test coverage for sending emails by:
* removing the use of package-level variables, and capturing the email into a strut
* making assertions about the text and html in the email

The expected values here will change, and can automatically be updated by running the tests with `-update` . The value here is that we get to preview the source of the email, and we'll catch unintended changes to the template.


BruceMacD (approved):
looks good to me

PR #3429, opened 2022-10-11
fix: only remove password reset scope on cred update

### LINKED ISSUE #3348: UpdateCredentials clears all access key scopes
## Describe the bug

When resetting a password (either from the CLI after login, or from the verify password reset flow) the `updateCredentials` call removes all scopes from the access key, instead of only removing the password reset scope:

https://github.com/infrahq/infra/blob/3eb34bbe945280549c6e289a461cb6e55565f1a5/internal/access/credential.go#L115

This prevents anyone who has reset their password from creating an access key during that session. They will need to logout and login again.

BruceMacD:
## Summary
When a credential is update the access key needs to have the "password reset required" scope removed, but keep the ability to create access keys. This change ensures that the key has the only the "password reset required scope" removed so any scopes we add in the future will also continue to work.

## Checklist

## Related Issues

Resolves #3348


dnephin (approved):
LGTM!

PR #3431, opened 2022-10-11
maintain: cli docgen fixes to clean up the resulting file

technovangelist:

## Summary

The docgen process was generating cli-reference.md, but all the docs point to cli.md after renaming a month or so back. So this update generates the right doc, and cleans up the generated doc a bit.

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/docgen/main.go
+++ b/internal/docgen/main.go
@@ -8,7 +8,7 @@ import (
 )
 
 func main() {
-	f, err := os.Create("./docs/reference/cli-reference.md")
+	f, err := os.Create("./docs/reference/cli.md")
 	if err != nil {
 		log.Println(err.Error())
 		return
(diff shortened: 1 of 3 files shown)


internal/docgen/main.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func main() {
-	f, err := os.Create("./docs/reference/cli-reference.md")
```
mchiang0610, line 11: we should add a redirect for anything pointing to cli-reference to cli

technovangelist: #3432

BruceMacD (approved):
Thanks for tidying this up, I ran the command to re-generate the associated docs here, we keep these files in the repo so I added the generated file to the PR.

PR #3432, opened 2022-10-11
maintain: add redirect for cli-reference to docs

technovangelist:

## Summary

A month or so back, we simplified a lot of the docs. One of the changes was moving reference/cli-reference to reference/cli. I added most of the redirects, but managed to miss this one. 

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/website/next.config.js
+++ b/website/next.config.js
@@ -111,15 +111,11 @@ module.exports = phase => ({
           permanent: true,
         }
       }),
-      ...[
-        '/docs/reference/cli-reference',
-      ].map(source => {
-        return {
-          source,
-          destination: '/docs/reference/cli',
-          permanent: true,
-        }
-      }),
+      {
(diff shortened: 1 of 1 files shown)


website/next.config.js
```diff
@@ -111,6 +111,15 @@ module.exports = phase => ({
           permanent: true,
         }
       }),
+      ...[
```
jmorganca, line 114: This can be

```
{
  source: '/docs/reference/cli-reference',
  destination: '/docs/reference/cli',
  permanent: true,
}
```

technovangelist: was trying to be consistent with the other redirects, but that works too. changed

BruceMacD (approved):
just need to amend the commit without the capital for this to pass the checks:
`maintain: add redirect for cli-reference to docs`

PR #3434, opened 2022-10-12
Require oldpassword

pdevine:
## Summary

When a user changes their password, they need to specify their old password in addition to their new password. This doesn't modify the forgot password screen which validates a user's identity by sending them a password reset link email.

## Checklist

### DIFF
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/generate"
+	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/internal/validate"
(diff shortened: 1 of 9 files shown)


internal/access/credential.go
```diff
@@ -72,9 +72,24 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
 	return nil
 }
 
-func updateCredential(c *gin.Context, user *models.Identity, newPassword string, isSelf bool) error {
+func updateCredential(c *gin.Context, user *models.Identity, oldPassword, newPassword string, isSelf, requireOldPassword bool) error {
 	db := getDB(c)
 
+	// Users have to supply their old password to change their existing password
+	if isSelf && requireOldPassword {
```
BruceMacD, line 79: Should we move this check to a wrapper function rather than adding another boolean flag to the function parameters? 

Ex:
```
func changePassword(c *gin.Context, user *models.Identity, oldPassword, newPassword string, isSelf) {
  if isSelf {
    // require old password
  }
  updateCredential(c, user, newPassword, isSelf)
}
func updateCredential(c *gin.Context, user *models.Identity, oldPassword, newPassword string, isSelf) {
  // do the actual updating directly here
}
```

internal/access/credential.go
```diff
@@ -72,9 +72,24 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
...
+		userCredential, err := data.GetCredential(db, data.ByIdentityID(user.ID))
+		if err != nil {
+			return fmt.Errorf("existing credential: %w", err)
+		}
+
+		// compare the stored hash of the user's password and the hash of the presented password
+		err = bcrypt.CompareHashAndPassword(userCredential.PasswordHash, []byte(oldPassword))
+		if err != nil {
+			// this probably means the password was wrong
```
BruceMacD, line 88: we should debug or trace level log the actual error here in case we need to debug this

internal/access/credential.go
dnephin: Ya, seems like we can remove both the `oldPassword` and `requireOldPassword` arguments by doing this check in `access.UpdatePassword`.

internal/access/credential.go
```diff
@@ -72,9 +72,24 @@ func UpdateCredential(c *gin.Context, user *models.Identity, newPassword string)
...
+			return fmt.Errorf("%w: invalid oldPassword", internal.ErrBadRequest)
```
dnephin, line 89: To return a structured error here, you can use `internal/validate.Error`, `checkPasswordRequirements` is an example of that.

Do we want to return a different error for missing oldPassword vs oldPassword did not match?

pdevine: We do. Added a different error for it.

internal/access/credential.go
pdevine: Fixed.

internal/access/credential.go
pdevine: Added a trace level log.

BruceMacD (approved):
looks good to me

Issue #3435, opened 2022-10-12
openapi doc generation handling of embedded structs


dnephin:
I noticed this with the list grants response:

```
      "ListGrantsResponse": {
          "items": {
            "items": {
              "properties": {
```

It seems like it's not reading the `json:,inline` struct tag, or not handling embedded structs correctly. I don't think there should be two levels of `items`.

dnephin:
Looks like this is still a bug.

PR #3436, opened 2022-10-12
Add API handler support for ListGrants blocking request

dnephin:
## Summary

Branched from #3400

This PR adds the API support for long polling of list grants requests. Best viewed by individual commit.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -2,6 +2,8 @@ package api
 
 import (
 	"bytes"
+	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -132,6 +134,10 @@ type ListResponse[T any] struct {
 	PaginationResponse `json:",inline"`
 	Count              int `json:"count"`
 	Items              []T `json:"items"`
+
+	// LastUpdateIndex is set to the latest update index for any request made
+	// with the updateIndex query parameter.
+	LastUpdateIndex `json:"-"`
 }
 
 func NewListResponse[T, M any](items []M, pr PaginationResponse, fn func(item M) T) *ListResponse[T] {
@@ -146,6 +152,19 @@ func NewListResponse[T, M any](items []M, pr PaginationResponse, fn func(item M)
 	}
 
 	return result
+}
+
+type LastUpdateIndex struct {
+	Index int64 `json:"-"`
+}
+
+func (l *LastUpdateIndex) setValuesFromHeader(header http.Header) error {
+	if idx := header.Get("Last-Update-Index"); idx != "" {
+		var err error
+		l.Index, err = strconv.ParseInt(idx, 10, 64)
+		return err
+	}
+	return nil
 }
 
 // PEM is a base64 encoded string, commonly used to store certificates and
--- a/internal/access/grant.go
+++ b/internal/access/grant.go
@@ -63,18 +63,13 @@ func ListGrants(c *gin.Context, opts data.ListGrantsOptions, lastUpdateIndex int
 		return ListGrantsResponse{}, err
 	}
 
-	var resp ListGrantsResponse
-
-	// TODO: validate that only supported query parameters are set, and that at least
-	// one of the required parameters are set with lastUpdateIndex
-	// TODO: change request timeout for these requests
 	listenOpts := data.ListenForGrantsOptions{
 		ByDestination: opts.ByDestination,
 		OrgID:         rCtx.DBTxn.OrganizationID(),
 	}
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ postgres:
 		--tmpfs=/var/lib/postgresql/data \
 		-p 127.0.0.1:15432:5432 \
 		postgres:14-alpine -c fsync=off -c full_page_writes=off \
-			-c listen_addresses=127.0.0.1 -c max_connections=100
+			-c max_connections=100
```
dnephin, line 41: Merged into main as part of #3430, but I haven't rebased this yet

internal/server/testdata/openapi3.json
```diff
@@ -305,6 +305,83 @@
...
       },
+      "ListGrantsResponse": {
+        "properties": {
+          "count": {
+            "format": "int",
+            "type": "integer"
+          },
+          "items": {
+            "items": {
+              "properties": {
```
dnephin, line 316: Opened #3435 for this, it seems strange

internal/server/data/data.go
```diff
@@ -194,16 +194,20 @@ func (t *Transaction) GormDB() *gorm.DB {
...
 func (t *Transaction) Rollback() error {
-	if t.committed.Load() {
+	if t.completed.Load() {
 		return nil
 	}
-	return t.DB.Rollback().Error
+	err := t.DB.Rollback().Error
+	if err == nil {
+		t.completed.Store(true)
+	}
```
dnephin, line 203: Added this so that calling rollback multiple times does not error. This allows a request handler to end a transaction early to keep transaction lifetime short.

api/types.go
```diff
@@ -132,6 +134,10 @@ type ListResponse[T any] struct {
 	PaginationResponse `json:",inline"`
 	Count              int `json:"count"`
 	Items              []T `json:"items"`
+
+	// LastUpdateIndex is set to the latest update index for any request made
+	// with the updateIndex query parameter.
+	LastUpdateIndex `json:"-"`
```
dnephin, line 140: I'm going to look at moving this out of `ListResponse` again. I tried once, but I ran into problems that I now can't remember.

internal/access/grant.go
```diff
@@ -31,18 +41,103 @@ func ListGrants(c *gin.Context, opts data.ListGrantsOptions) ([]models.Grant, er
...
+		return ListGrantsResponse{}, err
+	}
+
+	if lastUpdateIndex == 0 {
+		result, err := data.ListGrants(rCtx.DBTxn, opts)
+		return ListGrantsResponse{Grants: result}, err
+	}
+
+	// Close the request scoped txn to avoid long-running transactions.
+	if err := rCtx.DBTxn.Rollback(); err != nil {
```
BruceMacD, line 62: Should we commit this transaction rather than rolling it back? If we modify the database somewhere else it seems potentially unexpected that it gets rolled back here when there is no error.

internal/server/routes.go
```diff
@@ -234,14 +235,23 @@ func wrapRoute[Req, Res any](a *API, routeID routeIdentifier, route route[Req, R
 			return err
 		}
 
-		if err := tx.Commit(); err != nil {
+		completeTx := tx.Commit
+		if route.txnOptions != nil && route.txnOptions.ReadOnly {
+			// use rollback to avoid an error when the request handler already completed the txn
```
BruceMacD, line 240: In what case would the transaction already be committed here?

internal/access/grant.go
dnephin: This is in `ListGrants` so it should always be a read-only db connection. `get` in `routes.go` is used to create this route, and it sets `txnOptions` to `ReadOnly: true`. Any writes would fail before this point I think.  

We shouldn't be writing anything in `GET` requests, except for the "middleware" stuff at the beginning of a request, which we do in a separate transaction.

I'm not sure if there is a difference between `Commit` nothing , or `Rollback` nothing, but I guess rollback does communicate to the reader that a bit better.

internal/server/routes.go
dnephin: The transaction is not necessarily committed, but it should be completed (rollback or commit).

The one use case we have now is the one you commented on above. Long running transactions are bad for database performance (it consumes resources to keep the transaction open), so we need to end the transaction before blocking.

I should probably make a similar change for the `debug/pprof` endpoint, so that we don't keep a transaction open while profiling and tracing.

BruceMacD (approved):
thanks for answering my questions, this looks good to me

PR #3437, opened 2022-10-12
Remove getDB and hasAuthorization

dnephin:
## Summary

`getDB` was deprecated a while ago, when we added `RequestContext`. We left it for a bit to make sure we had the right interface for `RequestContext`. It seems to be working well, so now we can remove it.

We had 3 different ways to authorize a request (`hasAuthorization`, `RequireInfraRole`, and `Can` which was recently renamed to `IsAuthorized`).  We've been replacing the use of `hasAuthorization` with a check in the caller to compare the requested ID to the authorized user. This PR replaces the last few callers to use the pattern we have in other place. It also changes `isIdentitySelf` to accept a `RequestContext` and remove the error return (since it always returned `nil`).  This reduces the number of ways to authorize a request to 2. We should be able to reduce it to a single function in the future.

### DIFF
--- a/internal/access/access.go
+++ b/internal/access/access.go
@@ -11,24 +11,6 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-// hasAuthorization checks if a caller is the owner of a resource before checking if they have an approprite role to access it
-func hasAuthorization(c *gin.Context, requestedResource uid.ID, isResourceOwner func(rCtx RequestContext, requestedResourceID uid.ID) (bool, error), oneOfRoles ...string) (data.GormTxn, error) {
-	rCtx := GetRequestContext(c)
-	owner, err := isResourceOwner(rCtx, requestedResource)
-	if err != nil {
-		return nil, fmt.Errorf("owner lookup: %w", err)
-	}
-
-	if owner {
-		return rCtx.DBTxn, nil
-	}
-
-	if err := IsAuthorized(rCtx, oneOfRoles...); err != nil {
-		return nil, err
-	}
-	return rCtx.DBTxn, nil
-}
-
 const ResourceInfraAPI = "infra"
 
 // RequireInfraRole checks that the identity in the context can perform an action on a resource based on their granted roles
--- a/internal/access/access.go
+++ b/internal/access/access.go
@@ -11,24 +11,22 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-// TODO: replace calls to this function with GetRequestContext once the
-// data interface has stabilized.
-func getDB(c *gin.Context) *data.Transaction {
-	return GetRequestContext(c).DBTxn
-}
-
 // hasAuthorization checks if a caller is the owner of a resource before checking if they have an approprite role to access it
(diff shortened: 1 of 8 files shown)


internal/access/organization.go
```diff
@@ -35,16 +29,18 @@ func ListOrganizations(c *gin.Context, name string, pg *data.Pagination) ([]mode
...
 func GetOrganization(c *gin.Context, id uid.ID) (*models.Organization, error) {
-	roles := []string{models.InfraSupportAdminRole}
-
-	// If the user is in the org, allow them to call this endpoint, otherwise they must be
-	// an InfraSupportAdmin.
-	db, err := hasAuthorization(c, id, isOrganizationSelf, roles...)
-	if err != nil {
-		return nil, HandleAuthErr(err, "organizations", "get", models.InfraSupportAdminRole)
+	rCtx := GetRequestContext(c)
+	if user := rCtx.Authenticated.User; user != nil && user.OrganizationID == id {
```
dnephin, line 33: I made one small change here from the implementation we had in `isOrganizationSelf`. It now checks the authorized `user.OrganizationID` instead of the authenticated `organization.ID`.

On unauthenticated routes the authenticated organization is still set, so the authenticated `user.OrganizationID` is a slightly safer pattern.  In this case it doesn't really matter because the route required authentication.

internal/access/identity.go
```diff
@@ -15,19 +15,24 @@ import (
...
+	return identity != nil && identity.ID == userID
 }
 
 func GetIdentity(c *gin.Context, id uid.ID) (*models.Identity, error) {
-	roles := []string{models.InfraAdminRole, models.InfraViewRole, models.InfraConnectorRole}
-	db, err := hasAuthorization(c, id, isIdentitySelf, roles...)
-	if err != nil {
-		return nil, HandleAuthErr(err, "user", "get", roles...)
+	rCtx := GetRequestContext(c)
+	if isIdentitySelf(rCtx, id) {
```
BruceMacD, line 25: Should this match the same pattern as `credentials.go`? It's cleaner to me.
```
if !isSelf {
		err := IsAuthorized(rCtx, models.InfraAdminRole)
		if err != nil {
			return HandleAuthErr(err, "user", "update", models.InfraAdminRole)
		}
	}
```

internal/access/access.go
```diff
@@ -11,24 +11,22 @@ import (
...
 	if owner {
-		return getDB(c), nil
+		return rCtx.DBTxn, nil
 	}
 
-	return RequireInfraRole(c, oneOfRoles...)
+	if err := IsAuthorized(rCtx, oneOfRoles...); err != nil {
+		return nil, err
+	}
+	return rCtx.DBTxn, nil
```
BruceMacD: Not blocking, but I liked that the previous implementation of this defaulted to failure, while this function defaults to success.

internal/access/identity.go
dnephin: Ya, I like that. I think in the other places we have this we have the empty `if` block, but we can fix those as we find them

internal/access/access.go
dnephin: Hmm, I don't think I follow. I deleted this `hasAuthorization` in the next commit.  Is this about `hasAuthorization` or `IsAuthorized`?

internal/access/identity.go
dnephin: Done!

BruceMacD (approved):

PR #3438, opened 2022-10-13
feat: welcome screen for first time users

pdevine:
## Summary

This PR contains a welcome screen which users can learn about the basics of how to use Infra.

Screen Shot 2022-10-12 at 6 21 02 PM

It includes:
* a link to instructions on how to set up the CLI;
* directions on how to use the `infra login` and `infra use` commands; and
* a `kubectl` command

Missing in this draft PR is way to automatically bring the welcome screen up, and also, some way to _finish_ the instructions and not have the screen display again. Ideally there would also probably be checkboxes which denote that the user has completed the steps, although we can potentially iterate on this after the first version. Also, there is the assumption that the user has been given at least one grant; right now the screen is choosing the first _non-infra_ grant as the destination to use for the `infra use` command.

### DIFF
--- a/ui/pages/destinations/welcome.js
+++ b/ui/pages/destinations/welcome.js
@@ -1,0 +1,169 @@
+import Head from 'next/head'
+import Link from 'next/link'
+import { useState } from 'react'
+import copy from 'copy-to-clipboard'
+import {
+  CheckIcon,
+  DuplicateIcon,
+  ExternalLinkIcon,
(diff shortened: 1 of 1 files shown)


BruceMacD (commented):
I think it could be a bit confusing that this screen covers the infrastructure UI completely, maybe making the first time instructions a modal that allow visibility to the actual infrastructure page in the background would help show the user that there is a full list of infrastructure they can access.

PR #3439, opened 2022-10-13
fix: include organization_id in where clause of update queries

dnephin:
If someone is able to guess the ID of a row or the ID is accidentally shared it's possible a non-authorized user could update the row using the ID.

This PR fixes the bug by setting `WHERE organization_id = ?` so that only a user authorized for that org can perform updates.

Also adds a bunch more godoc for `query.go`.

### DIFF
--- a/internal/server/data/data.go
+++ b/internal/server/data/data.go
@@ -362,6 +362,9 @@ func list[T models.Modelable](tx GormTxn, p *Pagination, selectors ...SelectorFu
 func save[T models.Modelable](tx GormTxn, model *T) error {
 	db := tx.GormDB()
 	setOrg(tx, model)
+	if isOrgMember(model) {
+		db = ByOrgID(tx.OrganizationID())(db)
+	}
 	err := db.Save(model).Error
 	return handleError(err)
 }
--- a/internal/server/data/query.go
+++ b/internal/server/data/query.go
@@ -10,6 +10,7 @@ import (
 )
 
 type Table interface {
+	// Table returns the name of the database table.
 	Table() string
 	// Columns returns the names of the table's columns. Columns must return
 	// a slice literal where every item in the slice is a string literal.
@@ -32,7 +33,10 @@ type Insertable interface {
 }
 
 type Updatable interface {
-	Insertable
+	Table
+	// Values returns the values for all fields. The values must be in the same
+	// order as the column names returned by Columns.
+	Values() []any
 	// Primary returns the value for the field that is mapped to the primary key
 	// of the table.
 	Primary() uid.ID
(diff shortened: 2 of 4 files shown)


internal/server/data/data.go
```diff
@@ -362,6 +362,9 @@ func list[T models.Modelable](tx GormTxn, p *Pagination, selectors ...SelectorFu
 func save[T models.Modelable](tx GormTxn, model *T) error {
 	db := tx.GormDB()
 	setOrg(tx, model)
+	if isOrgMember(model) {
+		db = ByOrgID(tx.OrganizationID())(db)
+	}
```
dnephin, line 367: I think this is the bit we were missing. `setOrg` makes it look like we're doing the right thing, but that only sets the org_id of the new row, it doesn't force the query to only operate on an ID of the correct org.

I looked back to see if this was accidentally removed at some point, but I didn't find it, so I think it was something we missed while adding org support.

internal/server/data/query.go
```diff
@@ -103,12 +122,19 @@ func update(tx WriteTxn, item Updatable) error {
 	query.B(item.Table())
 	query.B("SET")
 	query.B(columnsForUpdate(item), item.Values()...)
-	query.B("WHERE deleted_at is null AND id = ?;", item.Primary())
+	query.B("WHERE deleted_at is null")
+	query.B("AND id = ?", item.Primary())
+	query.B("AND organization_id = ?;", tx.OrganizationID())
```
dnephin, line 127: This is the same fix for the new SQL query approach.

internal/server/data/data.go
ssoroka: right. setting the org is enough for inserts, but not for updates. As is it would effectively allow you to steal records from another org to your org.

internal/server/data/query.go
ssoroka: doesn't this need a check to see if the model has an org id?

dnephin: Ah, yes it does. I guess the tables that use it right now all have an OrgID. I'll add that check

dnephin: Done!

jmorganca (approved):

internal/server/data/query.go
```diff
@@ -103,12 +122,21 @@ func update(tx WriteTxn, item Updatable) error {
 	query.B(item.Table())
 	query.B("SET")
 	query.B(columnsForUpdate(item), item.Values()...)
-	query.B("WHERE deleted_at is null AND id = ?;", item.Primary())
+	query.B("WHERE deleted_at is null")
```
BruceMacD, line 125: Could hiding deleted at on update cause issues with restoring deleted rows?

BruceMacD (approved):

internal/server/data/query.go
dnephin: We don't have a way to restore deleted rows through the API, right? If we wanted to do that we'd have to write a migration or do it directly in the database?

My rational for `deleted_at is null` is that we don't really want someone to be able to updated the `deleted_at` field of a row that was already deleted.

If we ever support un-delete from the API we will definitely need to change this.

PR #3441, opened 2022-10-13
Remove duplicate OrganizationID in RequestContext

dnephin:
One of the unfortunate trade-offs with `RequestContext` was that we end up storing the OrganizationID in two places. Once at `Authenticated.Organization.ID`, and the other at `DBTxn.OrganizationID()`.

I attempted to resolve this previously, but ran into problems because a transaction needs to exist in packages that don't have access to `RequestContext`, or before the Organization ID is known.

This is another attempt to address this problem. The first commit allows a transaction to accept a `MetadataSource` to provide the metadata, which is implemented by `access.Authenticated`. This way we end up using the same field to provide the value in both places.

The second commit renames `WithOrgID` to `WithMetadata` so that we don't forget to add additional parameters when we add more metadata.

### DIFF
--- a/internal/access/request.go
+++ b/internal/access/request.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
 )
 
 const RequestContextKey = "requestContext"
@@ -28,4 +29,11 @@ type Authenticated struct {
 	AccessKey    *models.AccessKey
 	User         *models.Identity
 	Organization *models.Organization
+}
+
+func (n Authenticated) OrganizationID() uid.ID {
+	if org := n.Organization; org != nil {
+		return org.ID
+	}
+	return 0
 }
--- a/internal/access/access_test.go
+++ b/internal/access/access_test.go
@@ -26,13 +26,14 @@ func setupDB(t *testing.T) *data.DB {
 	return db
 }
 
-func loginAs(db *data.Transaction, user *models.Identity, org *models.Organization) (*gin.Context, *data.Transaction) {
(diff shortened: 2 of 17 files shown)


ssoroka (approved):

dnephin:
After rebasing with main I ran into test failures because of #3439 . `data.ValidateRequestAccessKey` did not have a `MetadaSource` so it was crashing. The latest commit fixes it.

internal/access/request.go
```diff
@@ -29,3 +30,10 @@ type Authenticated struct {
 	User         *models.Identity
 	Organization *models.Organization
 }
+
+func (n Authenticated) OrganizationID() uid.ID {
+	if org := n.Organization; org != nil {
+		return org.ID
+	}
+	return 0
```
BruceMacD, line 38: Should we return something that isn't possible (like `-1`) just to be safe?

BruceMacD (commented):
I find this change a bit more confusing because the `WithMetadata` function only ever sets the org ID, but the naming makes that much less clear now

dnephin:
> I find this change a bit more confusing because the WithMetadata function only ever sets the org ID, but the naming makes that much less clear now

I think we're planning on adding something like `CreatedBy() uid.ID` to the Metadata so that we can set that ID in the data layer. I think we can wait to merge this until we have a second piece of metadata.

internal/access/request.go
dnephin: `0` should be just as safe as `-1`, neither should make any rows. I think we can make the `uid.ID.Value` return nil instead of the 0 int, which would help it fail when the column is set to not nil.

dnephin:
I'll make this a draft until we have a need for `CreatedBy`. As I've thought about that more, I think it's possible we won't need it, and we should just set that in the caller before  calling `data.CreateX`.

dnephin:
We can revisit this sometime later

PR #3442, opened 2022-10-13
feat: users scim

### LINKED ISSUE #3378: SCIM Inbound User Provisioning

BruceMacD:
## Summary
This change includes endpoints to create, update, get, patch and delete users from an identity provider using SCIM.

## Checklist

## Related Issues

Resolves #3378

### DIFF
--- a/api/scim.go
+++ b/api/scim.go
@@ -1,6 +1,9 @@
 package api
 
-import "github.com/infrahq/infra/internal/validate"
+import (
+	"github.com/infrahq/infra/internal/validate"
+	"github.com/infrahq/infra/uid"
+)
 
 type SCIMUserName struct {
 	GivenName  string `json:"givenName"`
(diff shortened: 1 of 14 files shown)


internal/server/data/provideruser.go
```diff
@@ -78,22 +90,68 @@ func CreateProviderUser(db GormTxn, provider *models.Provider, ident *models.Ide
...
 	}
 	providerUser.LastUpdate = time.Now().UTC()
 
+	// check if the user's email has changed
+	identity, err := GetIdentity(tx, ByID(providerUser.IdentityID), Preload("Providers"))
+	if err != nil {
+		return fmt.Errorf("check identity before provider update: %w", err)
+	}
+
+	if identity.Name != providerUser.Email {
```
mxyng, line 107: It seems a workaround for the email conflict is to save the alternate email to only the `ProviderUser` record and report it as an alias

BruceMacD: That would be a potentially large change so I think we should address it in a follow-up.

internal/server/data/provideruser.go
```diff
@@ -164,6 +201,31 @@ func ListProviderUsers(tx ReadTxn, opts ListProviderUsersOptions) ([]models.Prov
...
+	query.B(pu.Table())
+	query.B("SET")
+	query.B("active = ?, last_update = ?", providerUser.Active, providerUser.LastUpdate)
+	query.B("WHERE provider_id = ? AND identity_id = ?", providerUser.ProviderID, providerUser.IdentityID)
+	query.B("RETURNING")
+	query.B(columnsForSelect(pu))
+
+	err := tx.QueryRow(query.String(), query.Args...).Scan(pu.ScanFields()...)
+	if err != nil {
+		if strings.Contains(err.Error(), "no rows in result set") {
```
dnephin, line 221: Can we use `handleReadError` to translate the error? If that works maybe we should combine `handlerReadEror` into `handlerError` so that we can use that one function everywhere.

internal/server/data/provideruser.go
```diff
@@ -199,7 +261,34 @@ func GetProviderUser(tx ReadTxn, providerID, identityID uid.ID) (*models.Provide
 	return (*models.ProviderUser)(pu), nil
 }
 
-func SyncProviderUser(ctx context.Context, tx WriteTxn, user *models.Identity, provider *models.Provider, oidcClient providers.OIDCClient) error {
+func ProvisionProviderUser(tx GormTxn, user *models.ProviderUser) error {
```
dnephin, line 264: I noticed this PR uses `Provision` instead of `Create`. In the RFC they seem to use `Create` as well (https://www.rfc-editor.org/rfc/rfc7644#section-3.3). Are there other scim docs that use `Provision` , or could we use `Create` for the `api` type and these functions?

dnephin (approved):
Looks good! Some questions/suggestions on naming and linking to docs

api/scim.go
```diff
@@ -65,3 +68,56 @@ type ListProviderUsersResponse struct {
 	StartIndex   int        `json:"startIndex"`
 	ItemsPerPage int        `json:"itemsPerPage"`
 }
+
+type SCIMUserProvisionRequest struct {
```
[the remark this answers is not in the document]
BruceMacD: That's the one, added some comment referencing relevant sections.

internal/server/data/identity.go
```diff
@@ -33,7 +33,7 @@ func (i *identitiesTable) ScanFields() []any {
 	return []any{&i.CreatedAt, &i.CreatedBy, &i.DeletedAt, &i.ID, &i.LastSeenAt, &i.Name, &i.OrganizationID, &i.UpdatedAt, &i.VerificationToken, &i.Verified}
 }
 
-func AssignIdentityToGroups(tx WriteTxn, user *models.Identity, provider *models.Provider, newGroups []string) error {
+func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.Provider, newGroups []string) error {
```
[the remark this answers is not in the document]
BruceMacD: Had to revert this transaction to a `GormTxn` because some of the nested calls need a `GormTxn` rather than a `WriteTransaction` still after rebasing

internal/server/data/provideruser.go
BruceMacD: I renamed most of the functions to "Create", but I left the data level as "provision", there is another "CreateProviderUser" function that has different functionality (it relies on the identity already existing)

PR #3443, opened 2022-10-13
Improve blocking requests and add more test coverage

dnephin:
## Summary

Branched from #3436 

Best viewed by individual commit. This PR makes the following changes:
* disables pagination when the request includes `lastUpdateIndex`, so that all the results are returned on an update
* extends the request timeout from 1 minute to 5 minutes for request with `lastUpdateIndex > 0`, and makes both timeouts configurable from server options
* adds a test for concurrent requests
* adds tests for blocking requests, to show they unblock on unpdate
* fixes a bug when a query for grants returned no results

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,8 @@ issues:
       text: "\"io/ioutil\" has been deprecated"
     - linters: [errcheck]
       text: "tx\\.Rollback"
+    - linters: [noctx]
+      path: "_test\\.go"
 
 linters:
   enable:
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -734,7 +734,7 @@ func removeDeletedIdentityProviderUsers() *migrator.Migration {
 
 func addUpdateIndexAndGrantNotify() *migrator.Migration {
 	return &migrator.Migration{
-		ID: "2022-09-21T13:50",
+		ID: "2022-10-17T18:00",
 		Migrate: func(tx migrator.DB) error {
 			if _, err := tx.Exec(`
 				CREATE SEQUENCE IF NOT EXISTS seq_update_index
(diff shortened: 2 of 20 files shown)


.golangci.yml
```diff
@@ -27,6 +27,8 @@ issues:
       text: "\"io/ioutil\" has been deprecated"
     - linters: [errcheck]
       text: "tx\\.Rollback"
+    - linters: [noctx]
+      path: "_test\\.go"
```
dnephin, line 31: Using a context for `http.Request` isn't always necessary in tests, so I think we can ignore issues from this linter in test files.

internal/server/grants_test.go
```diff
@@ -492,30 +492,18 @@ func TestAPI_ListGrants(t *testing.T) {
...
+						{User: idOther, Privilege: "connector", Resource: "res1.ns2"},
+					},
+					Count: 3,
 				}
-				assert.DeepEqual(t, grants.Items, expected, cmpAPIGrantShallow)
+				assert.DeepEqual(t, grants, expected, cmpAPIGrantShallow)
 				assert.Equal(t, resp.Result().Header.Get("Last-Update-Index"), "10004")
 			},
 		},
-		"with latest update index and timeout": {
```
dnephin, line 504: I moved this test case to `TestAPI_ListGrants_ExtendedRequestTimeout`. The trick I was using previously to set a context from the in-process client no longer works because of how we extend the request timeout.

BruceMacD (commented):
looks good, will approve once tests pass

internal/server/data/migrations.go
```diff
@@ -734,7 +734,7 @@ func removeDeletedIdentityProviderUsers() *migrator.Migration {
 
 func addUpdateIndexAndGrantNotify() *migrator.Migration {
 	return &migrator.Migration{
-		ID: "2022-09-21T13:50",
+		ID: "2022-10-17T18:00",
```
dnephin, line 737: This migration is not in prod yet, but it did run in dev. I'm changing it here instead of creating a new one because these new changes should effectively replace all of the old migration.

BruceMacD (approved):

dnephin:
I pushed one more commit which adds more tests for waiting on a notify, and also to cover the case where an unrelated grant changes (to show it doesn't unblock).

PR #3444, opened 2022-10-13
Update the connector to use blocking requests for ListGrants

dnephin:
## Summary

This PR updates the connector to use blocking requests for `ListGrants`. The connector will block until there is an update or timeout. Also adds log messages for the sync operation.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -345,14 +345,16 @@ func (c Client) DeleteProvider(id uid.ID) error {
 
 func (c Client) ListGrants(ctx context.Context, req ListGrantsRequest) (*ListResponse[Grant], error) {
 	return get[ListResponse[Grant]](ctx, c, "/api/grants", Query{
-		"user":          {req.User.String()},
-		"group":         {req.Group.String()},
-		"resource":      {req.Resource},
-		"destination":   {req.Destination},
-		"privilege":     {req.Privilege},
-		"showInherited": {strconv.FormatBool(req.ShowInherited)},
-		"showSystem":    {strconv.FormatBool(req.ShowSystem)},
-		"page":          {strconv.Itoa(req.Page)}, "limit": {strconv.Itoa(req.Limit)},
+		"user":            {req.User.String()},
+		"group":           {req.Group.String()},
+		"resource":        {req.Resource},
+		"destination":     {req.Destination},
+		"privilege":       {req.Privilege},
+		"showInherited":   {strconv.FormatBool(req.ShowInherited)},
+		"showSystem":      {strconv.FormatBool(req.ShowSystem)},
+		"page":            {strconv.Itoa(req.Page)},
+		"limit":           {strconv.Itoa(req.Limit)},
+		"lastUpdateIndex": {strconv.FormatInt(req.LastUpdateIndex, 10)},
 	})
 }
 
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -80,7 +80,7 @@ type KubernetesOptions struct {
 
 // connector stores all the dependencies for the connector operations.
 type connector struct {
-	k8s         *kubernetes.Kubernetes
+	k8s         kubeClient
 	client      apiClient
 	destination *api.Destination
 	certCache   *CertCache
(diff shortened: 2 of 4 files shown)


internal/connector/connector.go
```diff
@@ -403,23 +422,58 @@ func getEndpointHostPort(k8s *kubernetes.Kubernetes, opts Options) (types.HostPo
...
+				Int64("updateIndex", latestIndex).
+				Msgf("no updated grants from server")
+			return nil
+		case err != nil:
+			return fmt.Errorf("list grants: %w", err)
+		}
+		logging.L.Info().
+			Int64("updateIndex", latestIndex).
+			Int("numGrants", len(grants.Items)).
+			Msgf("received grants from server")
```
dnephin, line 447:
```suggestion
			Msg("received grants from server")
```

internal/connector/connector.go
```diff
@@ -403,23 +422,58 @@ func getEndpointHostPort(k8s *kubernetes.Kubernetes, opts Options) (types.HostPo
...
+			Destination:     con.destination.Name,
+			BlockingRequest: api.BlockingRequest{LastUpdateIndex: latestIndex},
+		})
+		var apiError api.Error
+		switch {
+		case errors.As(err, &apiError) && apiError.Code == http.StatusGatewayTimeout:
+			// a timeout is expected when there are no changes
+			logging.L.Info().
+				Int64("updateIndex", latestIndex).
+				Msgf("no updated grants from server")
```
dnephin, line 439:
```suggestion
				Msg("no updated grants from server")
```

internal/connector/connector.go
```diff
@@ -403,23 +422,58 @@ func getEndpointHostPort(k8s *kubernetes.Kubernetes, opts Options) (types.HostPo
...
+			return nil
+		case err != nil:
+			return fmt.Errorf("list grants: %w", err)
+		}
+		logging.L.Info().
+			Int64("updateIndex", latestIndex).
+			Int("numGrants", len(grants.Items)).
```
dnephin, line 446:
```suggestion
			Int("grants", len(grants.Items)).
```

mxyng (approved):

api/client.go
```diff
@@ -297,17 +305,24 @@ func (c Client) DeleteProvider(id uid.ID) error {
...
+		"showInherited":   {strconv.FormatBool(req.ShowInherited)},
+		"showSystem":      {strconv.FormatBool(req.ShowSystem)},
+		"page":            {strconv.Itoa(req.Page)},
+		"limit":           {strconv.Itoa(req.Limit)},
+		"lastUpdateIndex": {strconv.FormatInt(req.LastUpdateIndex, 10)},
+	}
+	return request[EmptyRequest, ListResponse[Grant]](ctx, c, http.MethodGet, "/api/grants", q, nil)
+}
+
+func (c Client) ListGrantsOld(req ListGrantsRequest) (*ListResponse[Grant], error) {
```
BruceMacD, line 324: Is this just left here to clean-up in a different PR?

internal/cmd/grants.go
```diff
@@ -90,7 +91,7 @@ func newGrantsListCmd(cli *CLI) *cobra.Command {
 				listReq.Group = group.ID
 			}
 
-			grants, err := listAll(client.ListGrants, listReq)
+			grants, err := listAll(client.ListGrantsOld, listReq)
```
BruceMacD, line 94: Why not use the new ListGrants here?

internal/connector/connector.go
```diff
@@ -77,13 +80,36 @@ type KubernetesOptions struct {
...
-	k8s         *kubernetes.Kubernetes
-	client      *api.Client
+	k8s         kubeClient
+	client      apiClient
 	destination *api.Destination
 	certCache   *CertCache
 	options     Options
 }
 
+type apiClient interface {
```
BruceMacD, line 90: What advantage does creating a new interface for this give us over using the api.Client?

api/client.go
dnephin: Ya, this was temporary so that I didn't have to change all the list methods. I ended up doing that in e44127f138b450d1322479d29cfc8a9eb24ee69c. I've rebased this PR so this `ListGrantsOld` is gone now.

internal/cmd/grants.go
dnephin: `listAll` requires a function that doesn't accept context.Context. I fixed that in e44127f138b450d1322479d29cfc8a9eb24ee69c so that we do now use `ListGrants` here.

internal/connector/connector.go
dnephin: Good question! I made these interface so that we can use a fake in tests. That way I can easily write a test to show that this loop behave correctly when there are errors from the API. 

We could have used a full `httptest.Server` to return error messages, but I think that ends up being a bit more work. A fakeClient was a bit easier.

A secondary benefit is that we can easily see which API calls the connector is using from this interface, without having to trace all of the code.

dnephin:
I rebased this PR to resolve the conflicts and remove the temporary changes to `api.Client` (those changes are now in main).

PR #3446, opened 2022-10-14
feat: device flow

ssoroka:
## Summary

Device flow

## Checklist

## Related Issues

Resolves #


api/deviceflow.go
```diff
@@ -0,0 +1,28 @@
...
+type ApproveDeviceFlowRequest struct {
+	UserCode string `json:"userCode" example:"BDSD-HQMK"`
+}
+
+type StartDeviceFlowRequest struct {
+	ClientID string `form:"client_id"`
+}
+
+type DeviceFlowResponse struct {
+	DeviceCode          string `json:"deviceCode" example:"NGU4QWFiNjQ5YmQwNG3YTdmZMEyNzQ3YzQ1YSA" note:"a long string the device will use to exchange for an access token"`
```
BruceMacD, line 12:
```suggestion
	DeviceCode          string `json:"deviceCode" example:"NGU4QWFiNjQ5YmQwNG3YTdmZMEyNzQ3YzQ1YSA" note:"a code that a device will use to exchange for an access key after device login is approved"`
```

api/deviceflow.go
```diff
@@ -0,0 +1,28 @@
...
+	VerificationURI     string `json:"verificationURI" example:"https://infrahq.com/device" note:"This is the URL the user needs to enter into their browser to start logging in"`
+	UserCode            string `json:"userCode" example:"BDSD-HQMK" note:"This is the text the user will enter at the URL above"`
```
BruceMacD, line 14:
```suggestion
	UserCode            string `json:"userCode" example:"BDSD-HQMK" note:"This is the text the user will enter at the verification URI"`
```

In case the order changes.

api/deviceflow.go
```diff
@@ -0,0 +1,28 @@
+package api
+
+type ApproveDeviceFlowRequest struct {
+	UserCode string `json:"userCode" example:"BDSD-HQMK"`
+}
+
+type StartDeviceFlowRequest struct {
+	ClientID string `form:"client_id"`
```
BruceMacD, line 8: We shouldn't include this client ID where we don't support third party login using this method. Seems like we only ignore this.

internal/server/data/deviceflowauthrequest.go
```diff
@@ -0,0 +1,140 @@
...
+func (d deviceFlowAuthRequestUpdateTable) Columns() []string {
+	return []string{"id", "client_id", "user_code", "device_code", "approved", "access_key_id", "expires_at",
+		"updated_at", "deleted_at"}
+}
+
+func (d deviceFlowAuthRequestUpdateTable) Values() []any {
+	return []any{d.ID, d.ClientID, d.UserCode, d.DeviceCode, d.Approved, d.AccessKeyID, d.ExpiresAt, d.UpdatedAt, d.DeletedAt}
+}
+
+func validateDeviceFlowAuthRequest(dfar *models.DeviceFlowAuthRequest) error {
```
BruceMacD, line 47: Should we just remove this validation for now where it doesnt do anything?

internal/server/data/migrations.go
```diff
@@ -714,3 +715,30 @@ func addProviderUserSCIMFields() *migrator.Migration {
 		},
 	}
 }
+
+func addDeviceFlowAuthRequestTable() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-10-06T14:58",
+		Migrate: func(tx migrator.DB) error {
+			_, err := tx.Exec(`
+				CREATE TABLE IF NOT EXISTS device_flow_auth_requests (
```
BruceMacD, line 724: We could add an org ID here from the initial request, could be nice where we are thinking of making org ID a standard thing

internal/server/data/migrations.go
```diff
@@ -714,3 +715,30 @@ func addProviderUserSCIMFields() *migrator.Migration {
...
+					approved 				 bool DEFAULT NULL,
+					access_key_id 	 bigint,
+					access_key_token text,
+					expires_at 			 timestamp with time zone,
+					created_at 			 timestamp with time zone,
+					updated_at 			 timestamp with time zone,
+					deleted_at 			 timestamp with time zone
+				);
+
+				CREATE UNIQUE INDEX IF NOT EXISTS idx_dfar_user_code on device_flow_auth_requests (user_code) WHERE (deleted_at IS NULL);
```
BruceMacD, line 738: This index should probably also include an org ID or device code, clashes could be frequent if we dont delete the device code after it expires.

internal/server/deviceflow.go
```diff
@@ -0,0 +1,148 @@
...
+func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.PollDeviceFlowRequest) (*api.DevicePollResponse, error) {
+	rctx := getRequestContext(c)
+	dfar, err := access.FindDeviceFlowAuthRequest(rctx, req.ClientID, req.DeviceCode)
+	if err != nil {
+		return nil, err
+	}
+	logging.Debugf("Found record")
+
+	if dfar.ExpiresAt.Before(time.Now()) {
+		logging.Debugf("it's expired")
```
BruceMacD, line 65:
```suggestion
```

internal/server/deviceflow.go
```diff
@@ -0,0 +1,148 @@
...
+		return &api.DevicePollResponse{
+			Status:     "expired",
+			DeviceCode: dfar.DeviceCode,
+		}, nil
+	}
+
+	if dfar.Approved != nil && !*dfar.Approved {
+		logging.Debugf("it was rejected")
```
BruceMacD, line 73:
```suggestion
```

internal/server/models/encryption.go
```diff
@@ -17,18 +17,39 @@ var SymmetricKey *secrets.SymmetricKey
 // SkipSymmetricKey is used for tests that specifically want to avoid field encryption
 var SkipSymmetricKey bool
 
-func (s EncryptedAtRest) Value() (driver.Value, error) {
+func (s EncryptedAtRest) Encrypt() (string, error) {
```
BruceMacD, line 20: Are these encryption changes related to device flow or just done in passing?

BruceMacD (commented):
Looking good, my main feedback is around removing client ID and requiring both the device code and the user code to get the access key back.

ui/pages/destinations/add.js
```diff
@@ -97,7 +97,7 @@ export default function DestinationsAdd() {
         extensionDeadline: '720h',
       }),
     })
-    const key = await res.json()
```
hoyyeva, line 100: what is` jsonBody()`? what is that difference between `res.json()` and `jsonBody(rest) `?

api/deviceflow.go
```diff
@@ -0,0 +1,28 @@
...
+type DeviceFlowResponse struct {
+	DeviceCode          string `json:"deviceCode" example:"NGU4QWFiNjQ5YmQwNG3YTdmZMEyNzQ3YzQ1YSA" note:"a long string the device will use to exchange for an access token"`
+	VerificationURI     string `json:"verificationURI" example:"https://infrahq.com/device" note:"This is the URL the user needs to enter into their browser to start logging in"`
+	UserCode            string `json:"userCode" example:"BDSD-HQMK" note:"This is the text the user will enter at the URL above"`
+	ExpiresInSeconds    int16  `json:"expiresIn" example:"1800" note:"The number of seconds that this set of values is valid"`
+	PollIntervalSeconds int8   `json:"interval" example:"5" note:"the number of seconds the device should wait between polling to see if the user has finished logging in"`
+}
+
+type PollDeviceFlowRequest struct {
+	ClientID   string `json:"clientID"`
```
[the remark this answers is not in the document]
ssoroka: device codes are significantly long to avoid guessing.

internal/server/data/migrations.go
ssoroka: in the future we want to be able to do device flow requests without knowing the org, so putting it here would be a problem

internal/server/data/deviceflowauthrequest.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (d *deviceFlowAuthRequestTable) ScanFields() []any {
+	return []any{d.ID, d.ClientID, d.UserCode, d.DeviceCode, d.Approved, d.AccessKeyID, d.ExpiresAt, d.CreatedAt, d.UpdatedAt}
+}
+
+// deviceFlowAuthRequestUpdateTable is used to update the DeviceFlowAuthRequest. It excludes
+// the CreatedAt field, because that field is not part of the input to
+// UpdateDeviceFlowAuthRequest.
+type deviceFlowAuthRequestUpdateTable models.DeviceFlowAuthRequest
```
dnephin, line 32: I think we don't need this as long as there is no update , or if the updates use the existing data as a base for the update.

`data.UpdateDeviceFlowAuthRequest` doesn't appear to be called yet, so not sure if we'll need that later, or if we can remove it.

internal/server/data/migrations.go
ssoroka: chances of collision are n in 20^8, where n is the number of existing records.
I think instead we'll just delete them frequently. working on adding something for that.

internal/server/deviceflow.go
```diff
@@ -0,0 +1,148 @@
...
+		VerificationURI:     fmt.Sprintf("https://%s/device", host),
+		UserCode:            userCode,
+		ExpiresInSeconds:    DeviceCodeExpirySeconds,
+		PollIntervalSeconds: 5,
+	}, nil
+}
+
+// can error with one of authorization_pending, access_denied, expired_token, slow_down
+func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.PollDeviceFlowRequest) (*api.DevicePollResponse, error) {
+	rctx := getRequestContext(c)
```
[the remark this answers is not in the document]
ssoroka: this is what the device code is doing.

internal/server/models/encryption.go
ssoroka: Was going to use this to encrypt the access key stored, but that causes problems when we switch to per-org encryption keys, so I ended up leaving it out. We could discard these changes or keep them, since it provides a manual way to encrypt and decrypt these fields, which we will likely need as we move away from gorm.

internal/server/routes.go
```diff
@@ -112,6 +112,11 @@ func (s *Server) GenerateRoutes() Routes {
 	get(a, noAuthnNoOrg, "/api/server-configuration", a.GetServerConfiguration)
 	post(a, noAuthnNoOrg, "/api/forgot-domain-request", a.RequestForgotDomains)
 
+	// Device flow
+	post(a, noAuthnNoOrg, "/api/device", a.StartDeviceFlow)
```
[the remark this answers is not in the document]
ssoroka: in the future we want to be able to do device flow requests without knowing the org, so putting it here would be a problem

ui/pages/destinations/add.js
ssoroka: see the implementation here https://github.com/infrahq/infra/pull/3446/files#diff-350a2c2ecbe2db05755f218618618e394994910966e81d4fed13d9d4689e646c

ui/pages/device/index.js
```diff
@@ -0,0 +1,93 @@
...
+                      error ? 'border-red-500' : 'border-gray-300'
+                    }`}
+                  />
+                  {error && (
+                    <p className='absolute top-full mt-1 text-xs text-red-500'>
+                      {error}
+                    </p>
+                  )}
+                </div>
+                <button
```
hoyyeva, line 79: we should probably add the type of button here
`<button  type='submit' ...
`

ui/pages/device/index.js
```diff
@@ -0,0 +1,93 @@
...
+                  disabled={codeEntered}
+                  className='mt-4 mb-2 flex w-full cursor-pointer justify-center rounded-md border border-transparent bg-blue-500 py-2 px-4 font-medium text-white shadow-sm hover:bg-blue-700 focus:outline-none focus:ring-2 focus:ring-blue-500 focus:ring-offset-2 sm:text-sm'
```
hoyyeva, line 81: we might want to add the styling when it is disabled
`className='... disabled:cursor-not-allowed disabled:opacity-30'`

ui/pages/device/index.js
```diff
@@ -0,0 +1,93 @@
...
+                  >
+                    Device Code
+                  </label>
+                  <input
+                    required
+                    autoFocus
+                    placeholder='AAAA-BBBB'
+                    type='text'
+                    name='code'
+                    style={{"textTransform": 'uppercase'}}
```
hoyyeva, line 66: you probably don't need this here, what you can do instead is to add `uppercase` to `className`

ui/pages/device/index.js
```diff
@@ -0,0 +1,93 @@
...
+                    value={code}
+                    onChange={e => setCode(e.target.value)}
+                    className={`mt-1 block w-full rounded-md shadow-sm focus:border-blue-500 focus:ring-blue-500 lg:text-lg ${
```
hoyyeva, line 69: `lg:text-lg` .. it seems big, what is it look like visually?

api/deviceflow.go
```diff
@@ -0,0 +1,26 @@
+package api
+
+type ApproveDeviceFlowRequest struct {
+	UserCode string `json:"userCode" example:"BDSD-HQMK"`
+}
+
+type StartDeviceFlowRequest struct {
```
BruceMacD, line 7: You can remove this and use `EmptyRequest` instead

internal/cmd/login.go
```diff
@@ -488,6 +499,63 @@ func attemptTLSRequest(options loginCmdOptions) error {
...
+	defer spinner.Stop()
+
+	var spinnerCount int = 0
+	var pollResp *api.DevicePollResponse
+loop:
+	for {
+		select {
+		case <-spinner.C:
+			spinnerCount++
+			fmt.Printf("  %s\r", string(spinChars[spinnerCount%len(spinChars)]))
```
BruceMacD, line 532: fresh 👍

BruceMacD (approved):
Thanks for addressing my feedback, this looks good to me

internal/server/data/deviceflowauthrequest.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (d *deviceFlowAuthRequestTable) ScanFields() []any {
+	return []any{d.ID, d.UserCode, d.DeviceCode, d.Approved, d.AccessKeyID, d.ExpiresAt, d.CreatedAt, d.UpdatedAt}
+}
+
+func validateDeviceFlowAuthRequest(dfar *models.DeviceFlowAuthRequest) error {
+	err := validate.Error{}
+	validationRules := []validate.ValidationRule{
+		validate.String("user_code", dfar.UserCode, 8, 8, validate.DeviceFlowUserCode),
+		validate.String("device_code", dfar.DeviceCode, 38, 38, validate.AlphaNumeric),
```
dnephin, line 36: This validation should be on the `api` type so that we validate the input earlier in the request flow. 

Some of these are checking values that are not user input, so a 400 error would not be correct in most of these cases, but the error being returned here will become a 400 response code.

ssoroka: I'll wrap the error so it's not a 400. These values are generated server-side, so there's less value in validating them from the client, but I can include that too.

api/deviceflow.go
```diff
@@ -0,0 +1,37 @@
...
+
+type DevicePollResponse struct {
+	Status        string         `json:"status,omitempty" note:"can be one of pending, rejected, expired, confirmed"`
+	DeviceCode    string         `json:"deviceCode,omitempty" example:""`
+	LoginResponse *LoginResponse `json:"login,omitempty"`
+}
+
+func (adfr *ApproveDeviceFlowRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.String("user_code", adfr.UserCode, 8, 9, append(validate.DeviceFlowUserCode, validate.CharRange{Low: '-', High: '-'})),
```
dnephin, line 35: This `append` is not safe, it could introduce a very subtle bug. When this is called there's a good chance that it will modify the array backing `validate.DeviceFlowUserCode`. If any other code does a similar append with a different second argument it will modify the array, which modifies this slice as well. We'd end up with data races and also the wrong values in the slice.

The rule I generally follow is:

> if the first argument to append is a variable, the result of append should only be assigned to the same variable as the first argument to append

To fix this you can invert the arguments:

```go
append([]validate.CharRange{{Low: '-', High: '-'}}, validate.DeviceFlowUserCode...)
```

internal/cmd/login.go
```diff
@@ -533,16 +600,13 @@ func promptLoginOptions(cli *CLI, client *api.Client) (loginMethod loginMethod,
...
-		return localLogin, nil, nil
-	}
-
 	var options []string
 	for _, p := range providers {
 		options = append(options, fmt.Sprintf("%s (%s)", p.Name, p.URL))
 	}
 
 	options = append(options, "Login with username and password")
+	options = append(options, "Login with device code")
```
jmorganca, line 609: Rename to `Login with your browser`

ui/pages/device/index.js
```diff
@@ -0,0 +1,94 @@
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+
+import DashboardLayout from '../../components/layouts/dashboard'
+
+export default function DeviceShow() {
```
jmorganca, line 6:
```suggestion
export default function Device() {
```

ui/pages/device/index.js
```diff
@@ -0,0 +1,94 @@
...
+      setError(e.message)
+    }
+
+    return false
+  }
+
+  return (
+    <div className='flex min-h-[280px] w-full flex-col items-center px-10 py-8'>
+      <>
+        <h1 className='text-base font-bold leading-snug'>Device Code</h1>
```
jmorganca, line 39:
```suggestion
        <h1 className='text-base font-bold leading-snug'>Confirm Log In</h1>
```

ui/pages/device/index.js
```diff
@@ -0,0 +1,94 @@
...
+        {codeEntered ? (
+          <p className='text-s my-3 flex max-w-[260px] flex-1 flex-col items-center justify-center text-center text-gray-600'>
+            You&apos;ve authorized the new device. You can close this window.
```
jmorganca, line 42:
```suggestion
            You are logged in. You may now close this window.
```

ui/pages/device/index.js
```diff
@@ -0,0 +1,94 @@
...
+          </p>
+        ) : (
+          <>
+            <h2 className='my-1.5 mb-4 max-w-md text-center text-xs text-gray-500'>
+              This code gives a new device access to your account. Never accept
+              a device code from someone else.
```
jmorganca, line 48:
```suggestion
              Please confirm this is the code displayed in your terminal.
```

ui/pages/device/index.js
```diff
@@ -0,0 +1,94 @@
...
+              onSubmit={onSubmit}
+              className='relative flex w-full max-w-sm flex-1 flex-col justify-center'
+            >
+              <div className='my-2 w-full'>
+                <label
+                  htmlFor='code'
+                  className='text-2xs font-medium text-gray-700'
+                >
+                  Device Code
+                </label>
```
jmorganca, line 60:
```suggestion

```

ui/pages/device/index.js
```diff
@@ -0,0 +1,94 @@
...
+              </button>
+            </form>
+          </>
+        )}
+      </>
+    </div>
+  )
+}
+
+DeviceShow.layout = page => <DashboardLayout>{page}</DashboardLayout>
```
jmorganca, line 94: While this can be addressed in another PR, we should use a different layout that does not include any sidenav or top bar. This could be done via:

1. Removing this layout
2. Making sure the content is centered on the screen
3. Ensuring users are logged in – see the use of `useUser` here https://github.com/infrahq/infra/blob/main/ui/components/layouts/dashboard.js#L91

PR #3464, opened 2022-10-17
fix: migration for update_index

dnephin:
Without this migration we can't list grants because it fails to load a nil into an int64.

The test here fails with the following when the migration implementation was missing:

```
sql: Scan error on column index 0, name "update_index": converting NULL to int64 is unsupported
```

It passes with the new migration


mxyng (approved):

BruceMacD (approved):

dnephin:
I updated the default value to 2. We return a `Last-Update-Index=1` when there are no results, so that the client doesn't need to know about this special case. It can keep pass us the header value and it will block.  

By using a value of 2 a client that passed `Last-Update-Index=1` will still receive all the grants that existed before the migration.

PR #3465, opened 2022-10-17
fix: minor ui bugs

hoyyeva:
## Summary
- fixed sorting on destination grant form after first add
- fixed password error message on reset password input fields
- fixed esc key clear input value when input type is `search`
- allowed input to have spaces beginning and the end of the string, and trim them on submit

### DIFF
--- a/ui/pages/groups/index.js
+++ b/ui/pages/groups/index.js
@@ -57,6 +57,11 @@ function AddGroupsDialog({ setOpen }) {
                 autoFocus
                 spellCheck='false'
                 type='search'
+                onKeyDown={e => {
+                  if (e.key === 'Escape') {
+                    e.preventDefault()
+                  }
+                }}
                 value={name}
                 onChange={e => setName(e.target.value)}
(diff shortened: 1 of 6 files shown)


ui/pages/groups/index.js
```diff
@@ -57,6 +57,11 @@ function AddGroupsDialog({ setOpen }) {
                 autoFocus
                 spellCheck='false'
                 type='search'
+                onKeyDown={e => {
+                  if (e.key === 'Escape') {
```
BruceMacD, line 61: nit:
In IE/Edge this may be 'Esc' rather than 'Escape'
```suggestion
                  if (e.key === 'Escape' || e.key === 'Esc') {
```

BruceMacD (approved):

ui/pages/groups/index.js
hoyyeva: oh i see. what if we compare the `keyCode` instead - the keycode is the same through out all the browsers

BruceMacD: I think that would work too

hoyyeva: oh nvm .. keycode is actually deprecated.

PR #3470, opened 2022-10-18
fix: remove authinfo for context

mxyng:
## Summary

kubeconfig.AuthInfo no longer follows the name of the contexts so when kubeconfig entries are removed, the authinfo remains. Instead of removing entries based on the name of the contexts, use the name of the individual references, e.g. context.Cluster and context.AuthInfo


BruceMacD (approved):
good catch

PR #3471, opened 2022-10-18
maintain: convert identities to direct sql

BruceMacD:
## Summary
This change updates the identities database operations to use direct SQL rather than Gorm. This is the first conversion I'm aware of that supports pre-loads, so I'd appreciate feedback there.

## Checklist


go.mod
```diff
@@ -168,6 +168,7 @@ require (
 	github.com/ugorji/go/codec v1.2.6 // indirect
 	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c // indirect
 	go.uber.org/atomic v1.10.0 // indirect
+	golang.org/x/exp v0.0.0-20221012211006-4de253d81b95
```
BruceMacD, line 171: Using this library in pre-loads to easily get map keys. It seems generally useful when working with maps.

internal/server/data/identity.go
```diff
@@ -112,51 +131,329 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
...
+	return insert(tx, (*identitiesTable)(identity))
 }
 
-func GetIdentity(db GormTxn, selectors ...SelectorFunc) (*models.Identity, error) {
-	return get[models.Identity](db, selectors...)
+type GetIdentityOptions struct {
+	ByID             uid.ID
+	ByName           string
+	PreloadGroups    bool
+	PreloadProviders bool
```
dnephin, line 145: What do you think about calling these `LoadGroups` and `LoadProviders` ?  `Preload` I think is a Gorm term that I was not really familiar with. In the future we might want to implement these using joins instead of separately selecting everything. We might only need to join across to the id and name (to avoid loading the full model of providers and groups). Calling it something other than `Preload` might help indicate that it's not necessarily the same as what gorm was doing.

If it's a hassle to change I think it's fine this way, we can change it later.

internal/server/data/identity.go
```diff
@@ -112,51 +131,329 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
...
+	identities := &identitiesTable{}
+	query := querybuilder.New("SELECT")
+	query.B(columnsForSelect(identities))
+	if opts.Pagination != nil {
+		query.B(", count(*) OVER()")
+	}
+	query.B("FROM")
+	query.B(identities.Table())
+	if opts.ByGroupID != 0 {
+		query.B("JOIN identities_groups ON identities_groups.identity_id = id")
```
dnephin, line 234: minor: could move the `ByGroupID` from the where clause to here, so that it's all handled in one place.

internal/server/data/identity.go
```diff
@@ -112,51 +131,329 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
...
+}
+
+type ListIdentityOptions struct {
+	ByID             uid.ID
+	ByIDs            []uid.ID
+	ByNotIDs         []uid.ID
+	ByName           string
+	ByNotName        string
+	ByGroupID        uid.ID
+	CreatedBy        uid.ID
```
dnephin, line 218: In other places I think this was called `ByCreatedBy`, probably good to be consistent here, since they do the same thing.

internal/server/data/identity.go
```diff
@@ -112,51 +131,329 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
...
+	query.B("WHERE deleted_at IS NULL AND organization_id = ?", tx.OrganizationID())
+	if opts.ByID != 0 {
+		query.B("AND id = ?", opts.ByID)
+	}
+	if len(opts.ByIDs) > 0 {
+		query.B("AND id IN (?)", opts.ByIDs)
+	}
+	if len(opts.ByNotIDs) > 0 {
+		query.B("AND id NOT IN (?)", opts.ByNotIDs)
+	}
```
dnephin, line 245: I think `NotIDs` is potentially dangerous on its own. In other places I've made it required to use `ByCreatedBy`, otherwise `NotIDs` is ignored. I believe the only place we use `NotIDs` is with `ByCreatedBy`, right ?

dnephin (approved):
Looks great! A few minor suggestions for consistency and about options we can remove, but nothing blocking a merge

internal/server/data/data.go
```diff
@@ -531,7 +532,7 @@ func InfraProvider(db GormTxn) *models.Provider {
 // InfraConnectorIdentity returns the connector identity for the organization set
 // in the db context.
 func InfraConnectorIdentity(db GormTxn) *models.Identity {
```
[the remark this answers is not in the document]
dnephin: Ah, I guess we can't do this until #3210 merges, because of the loading of groups and providers.

internal/server/data/data_test.go
```diff
@@ -91,16 +94,16 @@ func TestPaginationSelector(t *testing.T) {
 			assert.Equal(t, alphabeticalIdentities[i+(p.Page-1)*p.Limit], actual[i].Name)
 		}
 
-		p.Page = 2
-		actual, err = ListIdentities(db, &p, NotName(models.InternalInfraConnectorIdentityName))
+		p.Page = 2 // TODO: does this work
```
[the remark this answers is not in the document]
BruceMacD: Whoops, meant to remove this comment before PR. Thanks for pointing it out.

internal/access/passwordreset.go
```diff
@@ -16,7 +16,11 @@ func PasswordResetRequest(c *gin.Context, email string, ttl time.Duration) (toke
 	rCtx := GetRequestContext(c)
 	db := rCtx.DBTxn
 
-	users, err := data.ListIdentities(db, &data.Pagination{Limit: 1}, data.ByName(email))
+	opts := data.ListIdentityOptions{
+		Pagination: &data.Pagination{Limit: 1},
+		ByName:     email,
+	}
+	users, err := data.ListIdentities(db, opts)
```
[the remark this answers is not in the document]
BruceMacD: Removed it in this PR

internal/server/data/identity.go
BruceMacD: Made this change, not a hassle at all

internal/server/data/identity.go
BruceMacD: I left this one out just because it made the where clause more complicated

internal/server/data/identity.go
```diff
@@ -112,51 +131,329 @@ func AssignIdentityToGroups(tx GormTxn, user *models.Identity, provider *models.
...
 }
 
-func SaveIdentity(db GormTxn, identity *models.Identity) error {
-	return save(db, identity)
+type DeleteIdentitiesOptions struct {
+	ByID         uid.ID
+	ByIDs        []uid.ID
+	ByNotIDs     []uid.ID
+	CreatedBy    uid.ID
+	ByName       string
```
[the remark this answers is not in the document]
BruceMacD: Removed `ByName`

PR #3472, opened 2022-10-18
Use the same pattern for repeating operations in agent, server, and connector

dnephin:
## Summary

In #3444 I need to add exponential backoff when the `sycnWithServer` operation fails. I'd also like to be able to test the `sycnWithServer` function without sleeping.  To do that, I need an interface that looks like this:

```go
type waiter interface {
    Reset()
    Wait(context.Context) error
}
```

With that interface I can provide a fake implementation that waits no time, and it gives us a hook to check the backoff is being reset on successful operation.

We already had two functions for this in `internal/repeat`. All 3 of these end up being roughly the same pattern.

This PR removes the existing `repeat.Start`, and `repeat.InGroup` and replaces them with a new `repeat.Waiter` which implements the interface above. This allows us to use the same pattern in multiple places. Our previous implementations combined 3 concerns (sleeping, looping, and concurrency with a goroutine), the new implementation only deals with sleeping and failure counting. Looping is handled by the caller (generally just a `for` loop with no condition), and concurrency is also handled by the  caller, generally with an `errgroup`. This allows us to use the same `errgroup` for all the operations in a process (in this PR I moved the telemetry goroutine into the existing errgroup).

In this PR we use the new `repeat.Waiter`:
1. with a constant backoff for the telemetry goroutine
2. with an exponential backoff for the agent goroutine

And in #3444 I'll be using it in the connector `syncWithServer` goroutine. I expect we can also use it in the connector `sycnDestination` goroutine as well.

There are also a few changes to the agent behaviour, which I will comment on inline.


go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/Netflix/go-expect v0.0.0-20220104043353-73e0943537d2
 	github.com/alicebob/miniredis/v2 v2.23.0
+	github.com/cenkalti/backoff/v4 v4.1.3
```
dnephin, line 38: The v3 version of this library was already a transitive dependency, but I figured we may as well use the latest major version.

This library does more than we need. I'm not a fan of the `Retry` functions it provides, but I think the `BackOff` interface and the implementations are nice. We could pretty easily inline some of the implementations or replace the library with our own implementations, but for now it seems like a good fit.

internal/cmd/agent.go
```diff
@@ -36,27 +30,40 @@ func newAgentCmd() *cobra.Command {
...
+			group.Go(func() error {
+				backOff := &backoff.ExponentialBackOff{
+					InitialInterval:     time.Minute,
+					MaxInterval:         5 * time.Minute,
+					RandomizationFactor: 0.05,
+					Multiplier:          1.5,
+					// MaxElapsedTime is the duration the infra agent should
+					// continue to run without a successful sync. After this time
+					// the agent will exit.
+					MaxElapsedTime: 15 * time.Minute,
```
dnephin, line 47: Instead of a fixed 3 failures, this new implementation does the following:
1. starts exponential backoff from 1 minute up to 5 minutes
2. after 15 minutes of no successful operations it will exit

Does 15 minutes seem like a reasonable delay? I'm happy to change these settings.

It uses a 5% randomization to prevent thundering herds when recovering from an outage, and a 1.5 multiplier to add time each interval.

mxyng: I didn't bother setting an exponential back off for this since 1 minute intervals are infrequent enough that consecutive failures won't cause excessive noise. The problem with exponential back off is that more than a few failures produces a too long wait between attempts. At that point, it's likely better to exit

dnephin: I'm hoping we'll be able to use long polling here pretty soon, so that we can significantly decrease the wait  time between operations.  Until then, I could reduce the MaxInterval to 2 minutes, and change toe MaxElapsedTime to ~5 minutes. That should be pretty close to the old behaviour.

internal/server/server.go
```diff
@@ -238,6 +240,16 @@ func (s *Server) Run(ctx context.Context) error {
 	return err
 }
 
+func runTelemetryHeartbeat(ctx context.Context, tel *Telemetry) error {
+	waiter := repeat.NewWaiter(backoff.NewConstantBackOff(time.Hour))
+	for {
+		tel.EnqueueHeartbeat()
+		if err := waiter.Wait(ctx); err != nil {
+			return err
```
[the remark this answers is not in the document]
dnephin: Yes, that's right. `waiter.Wait` only returns an error when the context is cancelled (or when the backoff returns Stop, but the constant backoff used here never returns Stop). We want this goroutine to exit when the context is cancelled, that's the signal the process is being shutdown. 

If one of the goroutines in the `errgroup` return a non-nil error, it will cancel the context, which indicates to the other goroutines that they should exit as well.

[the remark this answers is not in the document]
dnephin: I forgot to say, this is the same behaviour we had before. When the context was cancelled the goroutine would exit in `repeat.Start`.

BruceMacD (approved):

internal/cmd/agent.go
dnephin: Ok, I've adjusted these values to be closer to the previous behaviour.

PR #3477, opened 2022-10-19
Fix context cancellation in blocking requests

dnephin:
## Summary

I noticed this problem while testing blocking requests in dev.

PR #3443 tried one approach for using a longer timeout in blocking requests using `extendedContext`. That worked to extend the deadline, but breaks regular cancellation (ex: when a client disconnects).

This PR takes a different approach. By removing the `TimeoutMiddleware` and moving that timeout into `wrapWroute` we can capture the original context set on the request by `http.Server`.  This allows us to create a new context without losing cancellation on the original context.

With this approach we get both a longer deadline, and proper cancellation when a client disconnects.

### DIFF
--- a/internal/server/grants_test.go
+++ b/internal/server/grants_test.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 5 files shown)


internal/server/grants_test.go
```diff
@@ -731,6 +732,39 @@ func TestAPI_ListGrants_ExtendedRequestTimeout(t *testing.T) {
 	assert.Equal(t, resp.Code, http.StatusGatewayTimeout, (*responseDebug)(resp))
 }
 
+func TestAPI_ListGrants_ExtendedRequestTimeout_CancelledByClient(t *testing.T) {
```
dnephin, line 735: This test fails on `main` and passes with these changes.

internal/server/middleware_test.go
```diff
@@ -59,30 +59,6 @@ func issueToken(t *testing.T, db data.GormTxn, identityName string, sessionDurat
 	return body
 }
 
-func TestRequestTimeoutError(t *testing.T) {
```
dnephin, line 62: I removed this test because the function it was testing was removed, and the API timeout behaviour is tested by `TestRequestTimeout` in `routes_test.go`.

internal/server/middleware_test.go
```diff
@@ -59,30 +59,6 @@ func issueToken(t *testing.T, db data.GormTxn, identityName string, sessionDurat
...
-		time.Sleep(110 * time.Millisecond)
-
-		assert.ErrorIs(t, c.Request.Context().Err(), context.DeadlineExceeded)
-
-		c.Status(200)
-	})
-	router.ServeHTTP(httptest.NewRecorder(), httptest.NewRequest("GET", "/", nil))
-}
-
-func TestRequestTimeoutSuccess(t *testing.T) {
```
dnephin, line 75: I removed this test because the function it was testing was removed, and the behaviour is covered by basically every API test we have.

BruceMacD (approved):

PR #3482, opened 2022-10-20
improve: change device flow input and layout to separate boxes with no sidebar

ssoroka:
## Summary
Screen Shot 2022-10-20 at 3 31 06 PM

### DIFF
--- a/ui/pages/device/index.js
+++ b/ui/pages/device/index.js
@@ -1,5 +1,5 @@
 import { useRouter } from 'next/router'
-import { useState } from 'react'
+import React, { useState } from 'react'
 
 import DashboardNoSidebarLayout from '../../components/layouts/dashboard-no-sidebar'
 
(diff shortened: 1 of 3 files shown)


ui/pages/device/index.js
```diff
@@ -33,6 +32,48 @@ export default function Device() {
     return false
   }
 
+  async function setCodeSegment(segment, pos) {
+    var codeCopy = code || ''
+    for (; codeCopy.length < 8; ) {
```
BruceMacD, line 37: What is this function doing? Adding spaces to a the code to populate the remaining boxes in the UI?

BruceMacD (approved):
looks good

ui/pages/device/index.js
ssoroka: yeah, it covers some really corner cases where you try to fill the last box first, etc.

PR #3483, opened 2022-10-20
feat: ui generate scim access key for provider

### LINKED ISSUE #3378: SCIM Inbound User Provisioning

BruceMacD:
## Summary
Create an access key that a provider can use for SCIM in the UI.

Screen Shot 2022-10-20 at 6 10 14 PM
Screen Shot 2022-10-20 at 6 11 32 PM
Screen Shot 2022-10-20 at 6 11 42 PM
Screen Shot 2022-10-20 at 6 11 50 PM

## Checklist

## Related Issues

Resolves #3378

### DIFF
--- a/ui/pages/providers/[id].js
+++ b/ui/pages/providers/[id].js
@@ -11,7 +11,7 @@ import RemoveButton from '../../components/remove-button'
 import Notification from '../../components/notification'
 import SCIMKey from '../../components/scim-key'
 
-function SCIMKeyDialog(props) {
+function SCIMKeyDialog({ provider, setOpen }) {
   const [scimAccessKey, setSCIMAccessKey] = useState('')
   const [error, setError] = useState('')
 
@@ -20,7 +20,7 @@ function SCIMKeyDialog(props) {
     setError('')
 
     try {
-      const keyName = props.provider.name + '-scim'
+      const keyName = provider.name + '-scim'
 
       // delete any existing access key for this provider
       await fetch(`/api/access-keys?name=${keyName}`, {
@@ -31,7 +31,7 @@ function SCIMKeyDialog(props) {
       const res = await fetch('/api/access-keys', {
         method: 'POST',
         body: JSON.stringify({
-          userID: props.provider.id,
+          userID: provider.id,
           name: keyName,
           ttl: '87600h',
           extensionDeadline: '720h',
@@ -59,7 +59,7 @@ function SCIMKeyDialog(props) {
                 <div className='relative mt-4'>
                   <h2 className='mt-5 text-sm'>
                     Generating a new SCIM access key will revoke any existing
-                    SCIM access key for this identity provider.
+                    SCIM access keys for this identity provider.
                   </h2>
                   <h2 className='mt-5 text-sm'>Do you wish to continue?</h2>
                 </div>
@@ -67,7 +67,7 @@ function SCIMKeyDialog(props) {
               <div className='flex flex-row items-center justify-end space-x-3'>
                 <button
                   type='button'
-                  onClick={() => props.setOpen(false)}
+                  onClick={() => setOpen(false)}
                   className='inline-flex items-center rounded-md border border-gray-300 bg-white px-4 py-2 text-xs font-medium text-gray-700 shadow-sm hover:bg-gray-100'
                 >
                   Cancel
@@ -206,7 +206,7 @@ export default function ProvidersEditDetails() {
               <Dialog
                 as='div'
                 className='relative z-50'
-                onClose={() => router.replace('/providers')}
+                onClose={() => setKeyDialogOpen(false)}
               >
                 <Transition.Child
                   as={Fragment}
@@ -279,7 +279,11 @@ export default function ProvidersEditDetails() {
                 className='px-6 py-5 text-left first:pr-6 first:pl-0'
               >
                 <div className='text-2xs text-gray-400'>{g.label}</div>
-                <span className='text-sm font-medium text-gray-800'>
+                <span
+                  className={`text-sm ${
+                    g.font ? g.font : 'font-medium'
+                  } text-gray-800`}
+                >
                   {g.value}
                 </span>
               </div>
--- a/ui/pages/providers/[id].js
+++ b/ui/pages/providers/[id].js
@@ -100,7 +100,7 @@ function SCIMKeyDialog(props) {
                 <button
                   className={`absolute right-2 top-2 rounded-md border border-black/10 bg-white px-2 py-2 text-black/40 backdrop-blur-xl hover:text-black/70`}
                   onClick={() => {
-                    copy(key)
+                    copy(scimAccessKey)
                     setKeyCopied(true)
                     setTimeout(() => setKeyCopied(false), 2000)
                   }}
--- a/ui/pages/providers/[id].js
+++ b/ui/pages/providers/[id].js
@@ -1,13 +1,133 @@
-import { useEffect, useRef, useState } from 'react'
+import { Fragment, useEffect, useRef, useState } from 'react'
+import { Transition, Dialog } from '@headlessui/react'
 import { useRouter } from 'next/router'
+import copy from 'copy-to-clipboard'
 import Head from 'next/head'
 import Link from 'next/link'
 import useSWR, { useSWRConfig } from 'swr'
+import { DuplicateIcon, CheckIcon } from '@heroicons/react/outline'
 import dayjs from 'dayjs'
 
 import Dashboard from '../../components/layouts/dashboard'
 import RemoveButton from '../../components/remove-button'
 import Notification from '../../components/notification'
+
+function SCIMKeyDialog(props) {
+  const [scimAccessKey, setSCIMAccessKey] = useState('')
+  const [error, setError] = useState('')
+  const [keyCopied, setKeyCopied] = useState(false)
+
+  async function onSubmit(e) {
+    e.preventDefault()
+
+    try {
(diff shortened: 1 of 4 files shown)


ui/pages/providers/add.js
```diff
@@ -109,11 +180,38 @@ export default function ProvidersAddDetails() {
...
+          extensionDeadline: '720h',
+        }),
+      })
+
+      const data = await res.json()
+
+      if (!res.ok) {
+        throw data
+      }
+
```
hoyyeva, line 207: I believe @ssoroka added global function `jsonBody` to handle this - so you can just do `await jsonBody(res)`

BruceMacD: Good catch, updated this to use this function

ui/pages/providers/add.js
```diff
@@ -302,6 +435,20 @@ export default function ProvidersAddDetails() {
                   </p>
                 )}
               </div>
+              <div>
+                <input
+                  id='scim-checkbox'
+                  type='checkbox'
+                  value=''
+                  class='h-4 w-4 rounded border-gray-300 bg-gray-100 text-blue-600 focus:ring-2 focus:ring-blue-500 dark:border-gray-600 dark:bg-gray-700 dark:ring-offset-gray-800 dark:focus:ring-blue-600'
```
hoyyeva, line 443: why do we need to styling the dark mode?

ui/pages/providers/index.js
```diff
@@ -24,7 +24,7 @@ export default function Providers() {
         <h1 className='py-1 font-display text-xl font-medium'>Providers</h1>
         <Link href='/providers/add'>
           <a className='inline-flex items-center rounded-md border border-transparent bg-black px-4 py-2 text-xs font-medium text-white shadow-sm hover:cursor-pointer hover:bg-gray-800'>
-            Connect provider
+            Connect Provider
```
hoyyeva, line 27: then maybe we also need to change the `Connect cluster` to `Connect Cluster` on the infrastructure page too

ui/pages/providers/add.js
```diff
@@ -155,6 +249,45 @@ export default function ProvidersAddDetails() {
       <Head>
         <title>Add Identity Provider - {kind}</title>
       </Head>
+      <header className='my-6 flex items-center justify-between'>
+        {/* SCIM key dialog */}
+        <Transition.Root show={keyDialogOpen} as={Fragment}>
```
hoyyeva, line 254: maybe we should componentized this - because we are using the same one for the other SCIM key dialog and also it looks very similar to the add group / user dialog too.

ui/pages/providers/add.js
```diff
@@ -1,15 +1,81 @@
...
+            >
+              <span class='font-medium'>Error:</span> {props.errorMsg}
+            </div>
+          )}
+        </section>
+
+        {/* Finish */}
+        <section className={`my-10 flex justify-between`}>
+          <Link href='/providers'>
+            <a className='flex-none items-center self-center rounded-md border border-transparent bg-black px-4 py-2 text-2xs font-medium text-white shadow-sm hover:bg-gray-800'>
```
hoyyeva, line 69: we should make the button to align to the right side of the dialog

ui/pages/providers/add.js
```diff
@@ -1,15 +1,81 @@
...
+              class='mb-4 rounded-lg bg-red-100 p-4 text-sm text-red-700 dark:bg-red-200 dark:text-red-800'
+              role='alert'
+            >
+              <span class='font-medium'>Error:</span> {props.errorMsg}
+            </div>
+          )}
+        </section>
+
+        {/* Finish */}
+        <section className={`my-10 flex justify-between`}>
```
hoyyeva, line 67: it can just be `<section className='my-10 flex justify-between'>`

ui/pages/providers/add.js
```diff
@@ -1,15 +1,81 @@
...
+                  Use this access key to configure your identity provider for
+                  inbound SCIM provisioning
+                </p>
+              </div>
+              <div className='group relative my-4 flex'>
+                <pre className='w-full overflow-auto rounded-lg bg-gray-50 px-5 py-4 text-xs leading-normal text-gray-800'>
+                  {props.accessKey}
+                </pre>
+                <button
+                  className={`absolute right-2 top-2 rounded-md border border-black/10 bg-white px-2 py-2 text-black/40 backdrop-blur-xl hover:text-black/70`}
```
hoyyeva, line 41: it seems like we can just do `<button
                  className='...'`

hoyyeva: also there should be type for the button, like `<button type='button'`

ui/pages/providers/add.js
```diff
@@ -1,15 +1,81 @@
...
+  CheckIcon,
+  InformationCircleIcon,
+  XIcon,
+} from '@heroicons/react/outline'
 
 import { providers } from '../../lib/providers'
 
 import Dashboard from '../../components/layouts/dashboard'
 
+function SCIMKeyDialog(props) {
```
hoyyeva, line 20: instead of passing in props, can we actually list what we are passing in ? It is kind of confusing just leave it as `props`

ui/pages/providers/[id].js
```diff
@@ -1,14 +1,130 @@
...
+                  className='inline-flex items-center rounded-md border border-transparent bg-black px-4 py-2 text-xs font-medium text-white shadow-sm hover:bg-gray-800'
+                >
+                  Continue
+                </button>
+              </div>
+            </form>
+          </section>
+        ) : (
+          <>
+            <section>
```
Comment (hoyyeva) [inline ui/pages/providers/[id].js:85]: we should be able to refactor this to make it reusable since we have the same function in `../add.js` too

ui/pages/providers/[id].js
```diff
@@ -1,14 +1,130 @@
...
 import Notification from '../../components/notification'
 
+function SCIMKeyDialog(props) {
+  const [scimAccessKey, setSCIMAccessKey] = useState('')
+  const [error, setError] = useState('')
+  const [keyCopied, setKeyCopied] = useState(false)
+
+  async function onSubmit(e) {
+    e.preventDefault()
+
```
Comment (hoyyeva) [inline ui/pages/providers/[id].js:22]: do we want to reset the error state here too?

ui/pages/providers/[id].js
```diff
@@ -1,14 +1,130 @@
...
+    try {
+      let keyName = props.provider.name + '-scim'
```
Comment (hoyyeva) [inline ui/pages/providers/[id].js:24]: can we change this to `const`?

ui/pages/providers/[id].js
```diff
@@ -120,9 +237,55 @@ export default function ProvidersEditDetails() {
...
+                    </Transition.Child>
+                  </div>
+                </div>
+              </Dialog>
+            </Transition.Root>
           </h1>
 
           <div className='my-3 flex space-x-2 md:my-0'>
+            <button
+              onClick={() => setKeyDialogOpen(true)}
```
Comment (hoyyeva) [inline ui/pages/providers/[id].js:284]: we need to add `type='button'` to the button component

ui/pages/providers/add.js
BruceMacD: Good eye. This was from the component I was basing it off, missed it in this case. Removed it.

ui/pages/providers/index.js
BruceMacD: Lowercase was the convention so I changed this back to `Connect provider`

ui/pages/providers/add.js
BruceMacD: Moved this out into a component in this PR, was fairly easy so I thought I may as well do it in the first go

ui/pages/providers/add.js
BruceMacD: fixed

ui/pages/providers/add.js
BruceMacD: I fixed this and the other instances I found

ui/pages/providers/add.js
BruceMacD: fixed in the component

ui/pages/providers/[id].js
Comment (BruceMacD) [reply ui/pages/providers/[id].js:21]: good catch

ui/pages/providers/[id].js
```diff
@@ -151,11 +279,7 @@ export default function ProvidersEditDetails() {
                 className='px-6 py-5 text-left first:pr-6 first:pl-0'
               >
                 <div className='text-2xs text-gray-400'>{g.label}</div>
-                <span
-                  className={`text-sm ${
-                    g.font ? g.font : 'font-medium'
-                  } text-gray-800`}
-                >
+                <span className='text-sm font-medium text-gray-800'>
```
Comment (hoyyeva) [inline ui/pages/providers/[id].js:282]: why did we remove the `g.font`?

ui/pages/providers/[id].js
```diff
@@ -8,6 +9,85 @@ import dayjs from 'dayjs'
 import Dashboard from '../../components/layouts/dashboard'
 import RemoveButton from '../../components/remove-button'
 import Notification from '../../components/notification'
+import SCIMKey from '../../components/scim-key'
+
+function SCIMKeyDialog(props) {
```
Comment (hoyyeva) [inline ui/pages/providers/[id].js:14]: can we list the props?

Comment (BruceMacD) [reply ui/pages/providers/[id].js]: missed this one, fixed now

ui/pages/providers/[id].js
Comment (BruceMacD) [reply ui/pages/providers/[id].js]: Thought I added this on accident, restored

hoyyeva (approved):

PR #3484, opened 2022-10-20
Make pprof endpoint a blocking request

dnephin:
And close the transaction before blocking to prevent long running transactions.

### DIFF
--- a/internal/server/debug.go
+++ b/internal/server/debug.go
@@ -12,7 +12,7 @@ import (
 	"github.com/infrahq/infra/internal/server/models"
 )
 
-var pprofRoute = route[api.EmptyRequest, *api.EmptyResponse]{
+var pprofRoute = route[pprofRequest, *api.EmptyResponse]{
 	handler: pprofHandler,
 	routeSettings: routeSettings{
 		omitFromTelemetry:          true,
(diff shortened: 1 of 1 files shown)


internal/server/debug.go
```diff
@@ -22,10 +22,21 @@ var pprofRoute = route[api.EmptyRequest, *api.EmptyResponse]{
 	},
 }
 
-func pprofHandler(c *gin.Context, _ *api.EmptyRequest) (*api.EmptyResponse, error) {
+type pprofRequest struct{}
+
+func (pprofRequest) IsBlockingRequest() bool {
```
dnephin, line 27: This method is used in `wrapRoute` to indicate the request should get a longer timeout (5 minutes). It will allow us to get profile data for a longer period of time instead of the 1 minute timeout for regular requests.

BruceMacD (approved):
looks good, but I don't understand what this change accomplishes?

dnephin:
Fair question. It does 2 things:
1. the `IsBlockingRequest` method allows us to gather profilers over a longer period of time. I think the default is 30s, but if you specify `s=120` now we should be able to gather a trace or a profile for longer than 1 minute. Previously we'd get a timeout at 1 minute I think. Not sure if we got the data back after that minute, or just an error
2. closing the transaction is a minor thing. It should very slightly reduce the load on the database by not having a long running transaction. I think it may also return the connection used by that transaction to the DB connection pool, which lets another request use it.

Both small things, but if we ever need to use these endpoints to debug a production issue we might want to gather multiple profiles at the same time (cpu, trace, mem, blocking, etc). Having them all not holding onto a transaction and a connection should be nice.

PR #3491, opened 2022-10-21
Convert provider queries to use sql

dnephin:
## Summary

This PR converts the provider data functions to SQL, and adds more test coverage of each function. 

### DIFF
--- a/internal/server/data/provider.go
+++ b/internal/server/data/provider.go
@@ -38,11 +38,11 @@ func validateProvider(p *models.Provider) error {
 	}
 }
 
-func CreateProvider(db GormTxn, provider *models.Provider) error {
+func CreateProvider(tx WriteTxn, provider *models.Provider) error {
 	if err := validateProvider(provider); err != nil {
 		return err
 	}
-	return add(db, provider)
+	return insert(tx, (*providersTable)(provider))
 }
 
 func GetProvider(db GormTxn, selectors ...SelectorFunc) (*models.Provider, error) {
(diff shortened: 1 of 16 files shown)


dnephin:
I've added the missing test cases, so this is now ready for review.

internal/server/data/provider.go
```diff
@@ -20,32 +40,147 @@ func validateProvider(p *models.Provider) error {
...
+	// be excluded. In other words, these IDs will not be deleted, even if they
+	// match CreatedBy.
+	// Can only be used with CreatedBy.
+	NotIDs []uid.ID
+}
+
+func DeleteProviders(db GormTxn, opts DeleteProvidersOptions) error {
+	var toDelete []models.Provider
+	if opts.ByID != 0 {
+		if provider, _ := GetProvider(db, GetProviderOptions{ByID: opts.ByID}); provider != nil {
```
BruceMacD, line 172: Why not just add this ID directly to `toDelete`?

BruceMacD (approved):
One comment but looks good

internal/server/data/provider.go
dnephin: Great question! I had it that way originally, but changed it in e8cc797acacf7e84a2b85d2a83c2d8770f784f44.

The problem with just using the ID was that if someone guessed an ID it would allow them to delete some provider_user entries from a different org.

I think another way to deal with this would be to first delete the provider, and use `RETURNING id`. That would give us all the provider IDs instead of having to call `ListProviders`. I'll explore that option in a follow up PR.

PR #3492, opened 2022-10-21
Start adding context.Context to api.Client methods

dnephin:
## Summary

To test blocking requests we need to be able to cancel the blocking operation, which requires that the `api.Client` method accepts a `context.Context`. I need this for the connector tests in #3444

This PR starts the work of adding a `context.Context` argument to all `api.Client` methods.

The first commits prepares for the new argument by splitting up `api.request`. This function already had more than enough arguments. Splitting up into `buildRequest`, `encodeRequestBody`, and `request` allows us to reduce the number of arguments to each, and make `buildRequest` a method on `Client`.

The second commit adds the `context.Context` argument to `buildRequest` and the http method helper functions, using `context.TODO` as a placeholder.

The last commit adds the `context.Context` argument to all `List` methods. We have to do all of these together because of the `listAll` helper that deals with required pagination in the CLI.  The only blocking request we have right now is `ListGrants`, so it's primarily list operations that need this, although technically all requests do benefit from cancellation.

In follow up PRs I think we can add the `context.Context` argument to the rest of the `Get` operations, and then to all the write and delete operations.

I've added `context.Background` where it seems reasonable to start a new context. In a bunch of places I used `context.TODO` to avoid having to change multiple functions to pass along a context. We can fix those in follow up PRs.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -79,8 +79,14 @@ func ErrorStatusCode(err error) int32 {
 	return 0
 }
 
-func (c *Client) buildRequest(method string, path string, query Query, body io.Reader) (*http.Request, error) {
-	req, err := http.NewRequest(method, fmt.Sprintf("%s%s", c.URL, path), body)
+func (c *Client) buildRequest(
+	ctx context.Context,
+	method string,
+	path string,
+	query Query,
+	body io.Reader,
+) (*http.Request, error) {
+	req, err := http.NewRequestWithContext(ctx, method, fmt.Sprintf("%s%s", c.URL, path), body)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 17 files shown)


api/types.go
```diff
@@ -13,7 +14,7 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-type Query map[string][]string
+type Query url.Values
```
dnephin, line 17: This is the same underlying type (`url.Values` is also a `map[string][]string`). I think it might help indicate how the `Query` type is used by declaring it as a `url.Values`.

api/client.go
```diff
@@ -168,41 +165,85 @@ type readsResponseHeader interface {
...
+	ctx := context.TODO()
+	httpReq, err := client.buildRequest(ctx, http.MethodPost, path, nil, body)
+	if err != nil {
+		return nil, err
+	}
+	return request[Res](client, httpReq)
+}
+
+func encodeRequestBody(req any) (io.Reader, error) {
+	b, err := json.Marshal(req)
```
BruceMacD, line 190: I think we should keep the nil check here, it's valid to make a post request with no body:
```
if req != nil {
		b, err := json.Marshal(req)
		if err != nil {
			return nil, fmt.Errorf("marshal request: %w", err)
		}

		return bytes.NewReader(b), nil
	}
return nil, nil // not sure if this would cause issues elsewhere
```

BruceMacD (commented):
Minor comment about the nil request body, but otherwise looks good. Will approve once that is resolved.

api/client.go
dnephin: I've added the nil check here

BruceMacD (approved):

PR #3506, opened 2022-10-25
fix: namespaces access with small tweak

hoyyeva:
## Summary
- allow namespaces to have `cluter-admin` role
- fix check icon styling on select roles


BruceMacD (approved):
looks good

PR #3511, opened 2022-10-27
feat: Add Authentication to OpenAPI Gen for all the endpoints

technovangelist:
## Summary

Today when you import the openapijson doc into a tool like Paw or Postman, you have to add Authorization: Bearer <key> to every header. This fixes the generator to add authorization to each api endpoint
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -397,6 +397,20 @@ func buildRequest(r reflect.Type, op *openapi3.Operation, method string) {
 			},
 		},
 	})
+	
+	op.AddParameter(&openapi3.Parameter{
+		Name:     "Authorization",
+		In:       "header",
+		Required: true,
+		Schema: &openapi3.SchemaRef{
+			Value: &openapi3.Schema{
+				Example:     "Bearer ACCESSKEY",
+				Format:      `Bearer\s[\d|a-z]{10}-[\d|a-z]{24}-`,
+				Type:        "string",
+				Description: "Bearer followed by your access key",
+			},
+		},
+	})
 
 	schema := &openapi3.Schema{
 		Type:       "object",
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -398,7 +398,7 @@ func buildRequest(r reflect.Type, op *openapi3.Operation, method string) {
 			},
 		},
 	})
-	var noLoginOperations = []string{"Login", "Signup"}
+	var noLoginOperations = []string{"Login", "Signup", "RequestPasswordReset", "VerifiedPasswordReset", "GetPassword", "ListProviders", "GetSettings", "Version", "GetServerConfiguration", "RequestForgotDomains"}
 
 	if !slices.Contains(noLoginOperations, op.OperationID) {
 		op.AddParameter(&openapi3.Parameter{
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -397,7 +397,7 @@ func buildRequest(r reflect.Type, op *openapi3.Operation, method string) {
 			},
 		},
 	})
-	
+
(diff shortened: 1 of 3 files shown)


internal/server/openapi.go
```diff
@@ -398,6 +398,20 @@ func buildRequest(r reflect.Type, op *openapi3.Operation, method string) {
...
 	})
 
+	op.AddParameter(&openapi3.Parameter{
+		Name:     "Authorization",
+		In:       "header",
+		Required: true,
+		Schema: &openapi3.SchemaRef{
+			Value: &openapi3.Schema{
+				Example:     "Bearer ACCESSKEY",
+				Format:      `Bearer\s[\d|a-z]{10}-[\d|a-z]{24}-`,
```
dnephin, line 408: I think it's a dot between the two sections, isn't it?
```suggestion
				Format:      `Bearer\s[\d|a-z]{10}\.[\d|a-z]{24}-`,
```

internal/server/openapi.go
```diff
@@ -398,6 +398,20 @@ func buildRequest(r reflect.Type, op *openapi3.Operation, method string) {
 		},
 	})
 
+	op.AddParameter(&openapi3.Parameter{
+		Name:     "Authorization",
+		In:       "header",
+		Required: true,
```
dnephin, line 404: It's only required for some paths. We don't have the `route` all the way through to this function yet, and I think doing that is a larger change to this code. Should we set this to `false` until we can correctly set `true/false` on each path?

internal/server/openapi.go
technovangelist: it is. doh!

internal/server/openapi.go
technovangelist: bummer

internal/server/openapi.go
technovangelist: Added the . and got rid of those dashes

internal/server/openapi.go
technovangelist: So is it just Login and Signup?

I tried 

```
var noLoginOperations = []string{"Login", "Signup"}
if !slices.Contains(noLoginOperations, op.OperationID) {
  op.AddParameter...
```

Would that solve the problem?

technovangelist: Found the list: Login, RequestPasswordReset, VerifiedPasswordReset, GetProvider, ListProviders, GetSettings, Signup, Version, GetServerConfiguration, RequestForgotDomains

internal/server/openapi.go
```diff
@@ -397,6 +398,23 @@ func buildRequest(r reflect.Type, op *openapi3.Operation, method string) {
 			},
 		},
 	})
+	var noLoginOperations = []string{"Login", "Signup", "RequestPasswordReset", "VerifiedPasswordReset", "GetPassword", "ListProviders", "GetSettings", "Version", "GetServerConfiguration", "RequestForgotDomains"}
```
BruceMacD, line 401:
```suggestion
	var noLoginOperations = []string{"Login", "Signup", "RequestPasswordReset", "VerifiedPasswordReset", "GetPassword", "ListProviders", "GetSettings", "Version", "GetServerConfiguration", "RequestForgotDomains", "StartDeviceFlow", "GetDeviceFlowStatus"}
```

BruceMacD (approved):

BruceMacD:
I ran `go test ./internal/server -update` and pushed the change, that should get the tests passing and make this ready to merge.

dnephin (approved):
I think we'll eventually want to use the `route` to look up if authn is required or not, but I think this works well for now. Including this header in the API reference docs is a big improvement.

PR #3525, opened 2022-10-31
Remove unused pki package

dnephin:
## Summary

This package has not been used in months. Let's remove it so it doesn't cause confusion. We can restore it from git in the future if we need it.


ssoroka (changes requested):
Seems like you're cutting a bit too much. KeyPair is used.

ssoroka (approved):

PR #3526, opened 2022-10-31
Move test helpers to helpers_test.go

dnephin:
@pdevine mentioned that they can be hard to find sometime, so this PR moves the test helpers I know of to `helpers_test.go`.

I think the risk with this approach is that contributors can start to overload these helpers and tests become very difficult to work with because there are huge layers of abstraction that you have to understand before you can understand the test.

I think the helpers we have now are ok. We should strive to keep this file small, and keep less-generic helpers next to the specific tests that warrant them. In this PR I moved a few helpers or vars into tests, because those were very specific to a test, and should not be re-used.

### DIFF
--- a/internal/server/helpers_test.go
+++ b/internal/server/helpers_test.go
@@ -1,0 +1,274 @@
+package server
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	gocmp "github.com/google/go-cmp/cmp"
+	"gotest.tools/v3/assert"
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/generate"
+	"github.com/infrahq/infra/internal/server/data"
+	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
+)
+
+type organizationData struct {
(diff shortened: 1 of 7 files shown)


internal/server/helpers_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+func createGroups(t *testing.T, db data.GormTxn, groups ...*models.Group) {
+	t.Helper()
+	for i := range groups {
+		err := data.CreateGroup(db, groups[i])
+		assert.NilError(t, err, groups[i].Name)
+	}
+}
+
+func createUser(t *testing.T, srv *Server, routes Routes, email string) *api.CreateUserResponse {
```
mxyng, line 152: It might be out of scope but what's the difference between `createUser` and `createIdentities`? It looks like `createUser` creates a user with API while `createIdentities` creates a user with data. Something to consider since these look very similar on the surface

ssoroka (approved):

internal/server/helpers_test.go
dnephin: Ya, there is some cleanup to do here.

I think `createIdentities` is a good pattern. It's a convenient way of creating a user with data that is relevant for the test case.

`createUser` is different from most since it uses the API to create things, and only lets you set the email address.

PR #3527, opened 2022-10-31
Extend the threshold for Destination.LastSeenAt connected

dnephin:
Long polling timeout is 5 minutes, so we need a bit more time to keep the destination showing as connected between long polling requests.

### DIFF
--- a/internal/server/models/destination.go
+++ b/internal/server/models/destination.go
@@ -24,9 +24,7 @@ type Destination struct {
 
 func (d *Destination) ToAPI() *api.Destination {
 	connected := false
-	// TODO: this should be configurable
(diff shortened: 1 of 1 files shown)


internal/server/models/destination.go
```diff
@@ -24,9 +24,7 @@ type Destination struct {
 
 func (d *Destination) ToAPI() *api.Destination {
 	connected := false
-	// TODO: this should be configurable
-	// https://github.com/infrahq/infra/issues/2505
```
dnephin, line 28: I commented on this issue that with long-polling we can't realistically make this configurable without making other changes to how we track LastSeenAt.

ssoroka (approved):

mxyng (approved):

PR #3528, opened 2022-10-31
Add endpoint-specific fields to API request log

dnephin:
## Summary

This PR introduces a hook for adding fields to the API request log. It also uses the new hook in list grants to add a `lastUpdatedIndex` and `numGrants` fields. I think we may also want to use this with SCIM endpoints, so that we can track which provider is using which fields. I'm sure we'll find other use cases for adding fields to the request log as we add new endpoints.

I think maybe we can find a better way to include these fields, but this seemed to work for now. If anyone has a suggestion for a better interface, I'd be happy to incorporate it.

### DIFF
--- a/internal/access/request.go
+++ b/internal/access/request.go
@@ -3,6 +3,8 @@ package access
 import (
 	"net/http"
 
+	"github.com/rs/zerolog"
+
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 )
@@ -20,6 +22,10 @@ type RequestContext struct {
 	// start transactions. Most routes should use DBTxn and should not use
 	// DataDB directly.
 	DataDB *data.DB
+
+	// Response is a mutable field. It can be modified by API handlers to add
+	// new response metadata.
+	Response *ResponseMetadata
 }
 
 // Authenticated stores data about the authenticated user. If the AccessKey or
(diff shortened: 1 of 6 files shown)


internal/server/grants.go
```diff
@@ -25,6 +26,13 @@ func (r ListGrantsResponse) SetHeaders(h http.Header) {
 }
 
 func (a *API) ListGrants(c *gin.Context, r *api.ListGrantsRequest) (*ListGrantsResponse, error) {
+	rCtx := getRequestContext(c)
+
+	rCtx.AddLogFields = append(rCtx.AddLogFields, func(event *zerolog.Event) {
+		event.Int64("lastUpdateIndex", r.LastUpdateIndex)
+	})
```
dnephin, line 33: I add this separately from `numGrants` so that if the request errors and exits early we still get this context added to the log entry.

BruceMacD (commented):
What log level do these get output at?

dnephin:
Looks like INFO level: https://github.com/infrahq/infra/blob/b243a3a/internal/server/logging.go#L57

BruceMacD (approved):

dnephin:
I made a small change to the implementation. Instead of storing the slice directly on `RequestContext`,  I moved it to a new `Response` field. I think that helps keep it separate from other immutable fields that are inputs to the handlers.

ssoroka (approved):

internal/access/request.go
```diff
@@ -20,6 +22,10 @@ type RequestContext struct {
 	// start transactions. Most routes should use DBTxn and should not use
 	// DataDB directly.
 	DataDB *data.DB
+
+	// Response is a mutable field. It can be modified by API handlers to add
+	// new response metadata.
+	Response *ResponseMetadata
```
ssoroka, line 28: consider making this a non-pointer

dnephin: I believe this will need to be a pointer so that we can mutate it without needing to call `c.Set()` to update the value stored in the gin.Context.  

A pointer receiver may work, but I was hoping that making this a pointer make it more obvious that this was a mutable value.  

I'll think about it some more .

dnephin:
I pushed a commit to fix the lint failure on main.

I don't think we've found an alternative without adding methods to `RequestContext` which I'd really like to avoid. By keeping the methods off the top level it should be move obvious how the methods are supposed to be used (to read data from the request, or to add metadata associated with the response).  

I'm going to merge it this way. It should be super easy to change at any time. I don't see us adding all that many fields to our logs anytime soon.

PR #3529, opened 2022-10-31
Fix device flow status url when baseDomain is not set

dnephin:
## Summary

Also remove some unused device flow types and vars.
Also add some TODOs to think about. I'll comment on most of these changes inline. Most of them are minor style changes.

I'm happy to omit any of these from this PR if we want to separate them out.


api/client.go
```diff
@@ -221,7 +221,7 @@ func (c Client) StartDeviceFlow() (*DeviceFlowResponse, error) {
 	return post[EmptyRequest, DeviceFlowResponse](c, "/api/device", nil)
 }
 
-func (c Client) PollDeviceFlow(req *PollDeviceFlowRequest) (*DevicePollResponse, error) {
+func (c Client) GetDeviceFlowStatus(req *PollDeviceFlowRequest) (*DevicePollResponse, error) {
```
dnephin, line 224: I renamed this to be consistent with the API handler name, and to better match the URL. The name `Poll` could suggest that this call itself does polling.

api/deviceflow.go
```diff
@@ -6,6 +6,12 @@ type ApproveDeviceFlowRequest struct {
 	UserCode string `json:"userCode" example:"BDSD-HQMK"`
 }
 
+func (adfr *ApproveDeviceFlowRequest) ValidationRules() []validate.ValidationRule {
```
dnephin, line 9: This is a minor style suggestion. Moved this up from below so that the method follows immediately after the type, and before other types. This keeps the type and its methods together.

api/login.go
```diff
@@ -55,10 +55,3 @@ type LoginResponse struct {
 	Expires                Time   `json:"expires"`
 	OrganizationName       string `json:"organizationName,omitempty"`
 }
-
-type LoginDeviceFlow struct {
```
dnephin, line 59: This type was not used. I assume it was left over from a previous revision.

internal/cmd/login.go
```diff
@@ -523,8 +523,7 @@ func deviceFlowLogin(client *api.Client, cli *CLI) (*api.LoginResponse, error) {
 	defer spinner.Stop()
 
 	var spinnerCount int = 0
-	var pollResp *api.DevicePollResponse
-loop:
```
dnephin, line 527: I was able to remove both the loop label and the var by returning directly from the `case` instead of breaking from the case and returning at the bottom. I think this makes the function easier to read because you don't have to jump from the break to the return.

internal/server/deviceflow.go
```diff
@@ -39,7 +39,6 @@ retry:
 		ExpiresAt:  time.Now().Add(DeviceCodeExpirySeconds * time.Second),
 	})
 	if err != nil {
-		// TODO: on duplicate record jump back to line 21.
```
dnephin, line 42: I believe the `goto` below here resolves this TODO

internal/server/deviceflow.go
```diff
@@ -50,6 +49,10 @@ retry:
 	if rctx.Authenticated.Organization != nil {
 		host = rctx.Authenticated.Organization.Domain
 	}
+	if host == "" {
+		// Default to the request hostname when in single tenant mode
+		host = fmt.Sprintf("https://%v", rctx.Request.URL.Host)
+	}
```
dnephin, line 55: This is the fix for the url being wrong in single tenant mode.

TODO: write a test case to over this.

internal/server/deviceflow.go
```diff
@@ -91,6 +95,7 @@ func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.PollDeviceFlowRequest
 				Name:      dfar.AccessKey.IssuedForName,
 				AccessKey: dfar.AccessKeyToken,
 				Expires:   api.Time(dfar.AccessKey.ExpiresAt),
+				// TODO: set OrganizationName for consistency with other login methods
```
dnephin, line 98: Reminder to myself to resolve this TODO (unless someone knows of a reason we shouldn't).

internal/server/models/deviceflowauthrequest.go
```diff
@@ -8,13 +8,20 @@ import (
...
-	AccessKeyToken string // to be filled in once approved
-	ExpiresAt      time.Time
+	UserCode   string
+	DeviceCode string
 
-	// can be preloaded if there's an AccessKeyID
-	AccessKey *AccessKey
+	// TODO: remove Approved field? There's no way to reject a request right now
+	// and AccessKeyID != nil indicates approved.
+	Approved    *bool
```
dnephin, line 16: Does this sound right? 

If we decide to move the access key creation to the "get request status" part of the flow then we'll need to keep this field, but we'll be able to remove 2 or 3 of the access key fields below.

internal/server/models/deviceflowauthrequest.go
```diff
@@ -8,13 +8,20 @@ import (
...
+	AccessKeyID uid.ID
+
+	// AccessKeyToken is set once the request is approved.
+	// TODO: use EncryptedAtRest
+	AccessKeyToken string
+
+	ExpiresAt time.Time
+
+	// AccessKey will be populated by some queries, but is never used on writes.
+	AccessKey *AccessKey `db:"-"`
```
ssoroka, line 26: `db:"-"` is a gorm thing, right?

ssoroka (commented):
looks good

api/client.go
dnephin: Good catch. I renamed the request type to `DeviceFlowStatusRequest` and response to `DeviceFlowStatusResponse`. After those renames I can't find anything else that using `PollDeviceFlow`

internal/access/deviceflow.go
```diff
@@ -20,6 +20,7 @@ func FindDeviceFlowAuthRequest(ctx RequestContext, deviceCode string) (*models.D
 		return nil, err
 	}
 
+	// TODO: do this in the data function
```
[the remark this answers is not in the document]
dnephin: I don't follow. The `data` query function would be the callee (not the caller), right?

I was thinking of addressing this TODO whenever we replace the selectors used in `data.GetDeviceFlowAuthRequest`.  We could have `data.GetDeviceFlowAuthRequest` use a `LEFT JOIN` to access key. So if the AccessKeyID is set we'll receive all the fields to populate the `DeviceFlowAuthRequest.AccessKey` field..

internal/server/models/deviceflowauthrequest.go
dnephin: Nope, Gorm uses `gorm` as the field tag. This is the field tag we're using for generating the table methods in `TestGenerateTableMethods`.  I added because it can be a good indicator that this field is not always set, but I'm happy to omit it if we don't plan on using the method generation on this type.

internal/server/deviceflow.go
dnephin: I added the test for this in the same commit with this fix. There was a bug here (the scheme was being added twice).

internal/server/deviceflow.go
dnephin: I started writing a test for this and noticed the user name in the response is also missing. So I'm going to fix these in a separate PR.

BruceMacD (approved):

internal/access/deviceflow.go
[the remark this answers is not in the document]
ssoroka: oh I think I misread that.

PR #3530, opened 2022-10-31
Add a gauge metric for active requests

dnephin:
The timing for a request is not emitted until after the request ends, so when requests are slow or there are many blocking requests we don't have a good picture of the current request load. This new gauge metric tells us how many requests are being handled.

I tested this manually with some blocking requests. I think an automated test for this might be possible, but we don't have a pattern for testing metrics yet.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -47,11 +47,19 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
 		Buckets:   prometheus.ExponentialBuckets(0.001, 2, 15),
 	}, []string{"host", "method", "path", "status"})
 
-	registry.MustRegister(requestDuration)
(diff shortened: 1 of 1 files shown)


metrics/metrics.go
```diff
@@ -60,6 +67,7 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
 			"path":   c.FullPath(),
 			"status": strconv.Itoa(c.Writer.Status()),
 		}).Observe(time.Since(t).Seconds())
+		requestCount.Dec()
```
ssoroka, line 70: might want a defer to make sure this is called on panics

dnephin: Good call! I made that change

BruceMacD (approved):

mxyng (approved):

PR #3531, opened 2022-10-31
maintain: automatically add grant when cluster added

### LINKED ISSUE #3258: When connecting a kubernetes cluster, the user should be given `cluster-admin` access

BruceMacD:
- when an admin adds a cluster to infra through the UI automatically grant them cluster-admin access on finish

## Summary
It is confusing that someone must grant themselves access to a cluster they just added to Infra. This automatically gives a user "cluster-admin" access upon adding a cluster.

## Checklist

## Related Issues

Resolves #3258

### DIFF
--- a/ui/pages/destinations/add.js
+++ b/ui/pages/destinations/add.js
@@ -104,23 +104,6 @@ export default function DestinationsAdd() {
     setSubmitted(true)
   }
 
-  async function finish() {
-    // grant the person that added this cluster 'cluster-admin' access automatically
-    await fetch('/api/grants', {
-      method: 'POST',
-      body: JSON.stringify({
-        user: user.id,
-        privilege: 'cluster-admin',
-        resource: name,
-      }),
-    })
-    mutate()
-    // redirect to the root destinations page
-    router.replace({
-      pathname: '/destinations',
-    })
-  }
-
   const command = `helm repo add infrahq https://helm.infrahq.com \nhelm repo update \nhelm upgrade --install infra-connector infrahq/infra --set connector.config.server=${window.location.host} --set connector.config.name=${name} --set connector.config.accessKey=${accessKey}`
 
   if (!isAdmin) {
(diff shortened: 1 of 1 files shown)


ui/pages/destinations/add.js
```diff
@@ -238,11 +255,12 @@ export default function DestinationsAdd() {
...
             </div>
           )}
-          <Link href='/destinations'>
-            <a className='flex-none items-center self-center rounded-md border border-transparent bg-black px-4 py-2 text-2xs font-medium text-white shadow-sm hover:bg-gray-800'>
-              Finish
-            </a>
-          </Link>
+          <button
+            className='inline-flex items-center rounded-md border border-transparent bg-black px-4 py-2 text-2xs font-medium text-white shadow-sm hover:cursor-pointer hover:bg-gray-800 disabled:cursor-not-allowed disabled:opacity-30'
+            onClick={finish}
```
hoyyeva, line 260: Oh! this is such a good idea!!
nit: we don't really need a finish function. we can put all that in `onClick={async() => { ... }}`

hoyyeva: and we need a type in button tag

BruceMacD: thanks for the feedback, made these changes

ui/pages/destinations/add.js
```diff
@@ -238,11 +238,28 @@ export default function DestinationsAdd() {
...
-          <Link href='/destinations'>
-            <a className='flex-none items-center self-center rounded-md border border-transparent bg-black px-4 py-2 text-2xs font-medium text-white shadow-sm hover:bg-gray-800'>
-              Finish
-            </a>
-          </Link>
+          <button
+            className='inline-flex items-center rounded-md border border-transparent bg-black px-4 py-2 text-2xs font-medium text-white shadow-sm hover:cursor-pointer hover:bg-gray-800 disabled:cursor-not-allowed disabled:opacity-30'
+            type='button'
+            onClick={async () => {
+              // grant the person that added this cluster 'cluster-admin' access automatically
```
mxyng, line 245: I don't think it's safe to assume the individual adding the cluster should automatically get any grants, particularly `cluster-admin`.  It's possible this elevates their privileges beyond what they already had. I suggest not merging this change.

BruceMacD: If they're an admin in Infra then they could just grant themselves cluster admin access after adding the cluster, so this is not a privilege escalation.

hoyyeva (approved):

PR #3533, opened 2022-10-31
Populate names in device flow status API response

dnephin:
## Summary

This PR adds the user name and organization name to the device flow login response, so that all the fields in the response are correctly populated.

Previously the `GetAccessKey` operation did not populate the `IssusedForName`. This PR fixes the function to populate that field.

Also adds a lookup of the organization from the `AccessKey.OrganizationID`, so that we can include the name in the response.

### DIFF
--- a/internal/access/deviceflow.go
+++ b/internal/access/deviceflow.go
@@ -20,8 +20,6 @@ func FindDeviceFlowAuthRequest(ctx RequestContext, deviceCode string) (*models.D
 		return nil, err
 	}
 
-	// TODO: do these lookups in the data function when data.GetDeviceFlowAuthRequest
-	// is converted to SQL
 	if dfar.AccessKeyID > 0 {
 		dfar.AccessKey, err = data.GetAccessKey(ctx.DBTxn, data.GetAccessKeysOptions{ByID: dfar.AccessKeyID})
 		if err != nil {
--- a/internal/access/deviceflow.go
+++ b/internal/access/deviceflow.go
@@ -20,9 +20,17 @@ func FindDeviceFlowAuthRequest(ctx RequestContext, deviceCode string) (*models.D
 		return nil, err
 	}
 
-	// TODO: do this in the data function
+	// TODO: do these lookups in the data function when data.GetDeviceFlowAuthRequest
+	// is converted to SQL
 	if dfar.AccessKeyID > 0 {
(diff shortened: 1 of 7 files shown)


internal/server/models/deviceflowauthrequest.go
```diff
@@ -24,4 +24,7 @@ type DeviceFlowAuthRequest struct {
 
 	// AccessKey will be populated by some queries, but is never used on writes.
 	AccessKey *AccessKey `db:"-"`
+
+	// Organization will be populated by some queries, but is never used on writes.
+	Organization *Organization `gorm:"-" db:"-"`
```
dnephin, line 29: I had to add the `gorm` tag here as well, without it I was getting errors. I think because the `data.GetDeviceFlowAuthRequest` function uses gorm for selects, so we have to exclude the field.  I'm not sure why the `AccessKey` field above does not have the same problem.

internal/access/deviceflow.go
```diff
@@ -20,12 +20,18 @@ func FindDeviceFlowAuthRequest(ctx RequestContext, deviceCode string) (*models.D
 		return nil, err
 	}
 
-	// TODO: do this in the data function
+	// TODO: do these lookups in the data function when data.GetDeviceFlowAuthRequest
+	// is converted to SQL
 	if dfar.AccessKeyID > 0 {
 		dfar.AccessKey, err = data.GetAccessKey(ctx.DBTxn, data.GetAccessKeysOptions{ByID: dfar.AccessKeyID})
```
BruceMacD, line 26: Not part of this change but it seems a bit odd that the `DeviceFlowAuthRequest` has a separate `AccessKeyID` and `AccessKey`

BruceMacD (approved):

internal/server/deviceflow.go
```diff
@@ -112,7 +112,7 @@ func (a *API) ApproveDeviceAdd(c *gin.Context, req *api.ApproveDeviceFlowRequest
 		return nil, internal.ErrExpired
 	}
 
-	if dfar.AccessKey != nil {
+	if dfar.AccessKeyID != 0 {
```
dnephin, line 115: This fixes a bug. `AccessKey` is always nil because we don't populate it in the query used above.

internal/server/models/deviceflowauthrequest.go
dnephin: Shouldn;t need this anymore, I'll remove it

PR #3536, opened 2022-11-01
fixes auth header around open api docs

ssoroka:
## Summary

Auth header in openapi docs was a manual list that needed to be edited. this updates it so that it's detected and generated automatically, so impossible to screw up.
also fixed:
- regex for auth tokens
- this resolved a bug with one endpoint saying it needed auth which it did not.

## Checklist

## Related Issues

Resolves #


BruceMacD (approved):
Looks good to me, this even fixed a bug in the docs already

PR #3537, opened 2022-11-02
maintain: replace set-output with GITHUB_OUTPUT

mxyng:
## Summary

`set-output` is deprecated. Replace it with `GITHUB_OUTPUT`. Update the custom actions for the same reason.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


BruceMacD (approved):
Kinda surprising to me that the dist directories updated from this change, but seems good

mxyng:
> Kinda surprising to me that the dist directories updated from this change, but seems good

The dist directories for the actions were updated because I updated the package dependencies. The actions are then built into a single package complete with its dependencies. Otherwise, GitHub has a hard (impossible) time running them in CI

dnephin:
I added the `action/fuzz` label to trigger that workflow, to test it's still working with this change. If it goes green LGTM.

PR #3540, opened 2022-11-02
Use SQL for organizations queries

dnephin:
## Summary

This PR converts the `data` query functions for organizations to SQL, and adds test coverage for the functions.

### DIFF
--- a/internal/server/data/organization.go
+++ b/internal/server/data/organization.go
@@ -7,12 +7,33 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-// CreateOrganization creates a new organization and sets the current db context to execute on this org
+type organizationsTable models.Organization
+
+func (organizationsTable) Table() string {
+	return "organizations"
+}
+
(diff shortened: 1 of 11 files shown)


internal/server/data/organization.go
```diff
@@ -56,28 +79,83 @@ func CreateOrganization(tx GormTxn, org *models.Organization) error {
...
+	query.B("FROM organizations")
+	query.B("WHERE deleted_at is NULL")
+
+	switch {
+	case opts.ByID != 0:
+		query.B("AND id = ?", opts.ByID)
+	case opts.ByDomain != "":
+		query.B("AND domain = ?", opts.ByDomain)
+	default:
+		return nil, fmt.Errorf("an ID is required to get organization")
```
BruceMacD, line 100:
```suggestion
		return nil, fmt.Errorf("an ID or domain is required to get organization")
```

BruceMacD (approved):

PR #3541, opened 2022-11-02
Use SQL for settings queries

dnephin:
## Summary

Branched from #3540 

Convert `data` queries for settings to SQL, and add  test cases.

### DIFF
--- a/internal/server/data/migrations_test.go
+++ b/internal/server/data/migrations_test.go
@@ -904,6 +904,12 @@ func parseTime(t *testing.T, s string) time.Time {
 	return v
 }
 
+func runStep(t *testing.T, name string, fn func(t *testing.T)) {
+	if !t.Run(name, fn) {
+		t.FailNow()
+	}
+}
+
 // testCaseLine is motivated by this Go proposal https://github.com/golang/go/issues/52751.
 // That issue has additional context about the problem this solves.
 func testCaseLine(name string) testCaseLabel {
--- a/internal/server/data/settings.go
+++ b/internal/server/data/settings.go
@@ -8,26 +8,40 @@ import (
 
 	"gopkg.in/square/go-jose.v2"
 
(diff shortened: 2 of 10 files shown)


internal/server/data/migrations_test.go
```diff
@@ -904,6 +904,12 @@ func parseTime(t *testing.T, s string) time.Time {
 	return v
 }
 
+func runStep(t *testing.T, name string, fn func(t *testing.T)) {
+	if !t.Run(name, fn) {
```
BruceMacD, line 908: what does this do?

internal/server/data/settings.go
```diff
@@ -36,52 +50,48 @@ func initializeSettings(tx GormTxn, orgID uid.ID) (*models.Settings, error) {
...
-		return nil, err
+		return err
 	}
 
-	settings = &models.Settings{
+	settings := &models.Settings{
 		OrganizationMember: models.OrganizationMember{OrganizationID: orgID},
 		PrivateJWK:         models.EncryptedAtRest(secs),
 		PublicJWK:          pubs,
+		LengthMin:          8,
```
BruceMacD, line 65: Where was this enforced before, or was it missed?

internal/server/data/migrations_test.go
dnephin: I moved this from another file because it's now only used in `migrations_test.go`. That's why it's showing up in the diff. It was previously in another file.

This runs a function similar to `t.Run`, however if that test function fails it stops the parent test case. Normally a failure in `t.Run` will continue to run the rest of the parent test case. 

This is useful for multi-step tests, where you want some of the benefits of `t.Run` (using `t.Cleanup` to cleanup when the subtest exits, and the additional context of the `name` in the failure message), without the behaviour of continuing to run the remaining subtests.

BruceMacD (approved):
thanks for the explanations, looks good

PR #3542, opened 2022-11-02
device flow cleanup

ssoroka:
## Summary

- remove expired records
- encrypt device flow access key
- add background jobs

## Checklist

### DIFF
--- a/docs/api/openapi3.json
+++ b/docs/api/openapi3.json
@@ -1153,7 +1153,7 @@
       "url": "https://www.elastic.co/licensing/elastic-license"
     },
     "title": "Infra API",
-    "version": "99.99.99999"
+    "version": "0.0.0"
   },
   "paths": {
     "/api/access-keys": {
@@ -1167,7 +1167,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -1268,7 +1268,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -1405,7 +1405,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -1541,7 +1541,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -1648,7 +1648,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -1775,7 +1775,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -1926,7 +1926,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2031,7 +2031,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2136,7 +2136,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2298,7 +2298,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2383,7 +2383,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2497,7 +2497,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2599,7 +2599,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2702,7 +2702,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -2893,7 +2893,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -3090,7 +3090,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -3235,7 +3235,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -3340,7 +3340,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
@@ -3447,7 +3447,7 @@
             "required": true,
             "schema": {
               "description": "Version of the API being requested",
-              "example": "99.99.99999",
+              "example": "0.0.0",
               "format": "\\d+\\.\\d+\\(.\\d+)?(-.\\w(+\\w)?)?",
               "type": "string"
             }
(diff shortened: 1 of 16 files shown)


internal/server/data/deviceflowauthrequest.go
```diff
@@ -58,35 +57,48 @@ func CreateDeviceFlowAuthRequest(tx WriteTxn, dfar *models.DeviceFlowAuthRequest
...
-func GetDeviceFlowAuthRequest(db GormTxn, selectors ...SelectorFunc) (*models.DeviceFlowAuthRequest, error) {
-	return get[models.DeviceFlowAuthRequest](db, selectors...)
+type SelectDeviceFlowAuthRequestOptions struct {
+	ByID         uid.ID
+	ByDeviceCode string
+	ByUserCode   string
+}
+
+func GetDeviceFlowAuthRequest(tx GormTxn, opts SelectDeviceFlowAuthRequestOptions) (*models.DeviceFlowAuthRequest, error) {
+	rec := &deviceFlowAuthRequestTable{}
```
BruceMacD, line 67: Should this fail if none of id, device code, or user code are specified?

internal/server/jobs/jobs.go
```diff
@@ -0,0 +1,20 @@
...
+	"time"
+
+	"github.com/infrahq/infra/internal/server/data"
+)
+
+func RemoveOldDeviceFlowRequests(ctx context.Context, tx data.WriteTxn, lastRunAt, currentTime time.Time) error {
+	return data.DeleteExpiredDeviceFlowAuthRequests(tx)
+}
+
+func RemoveExpiredAccessKeys(ctx context.Context, tx data.WriteTxn, lastRunAt, currentTime time.Time) error {
```
BruceMacD, line 14: Seems like most of these arguments aren't used

internal/server/server.go
```diff
@@ -203,6 +203,8 @@ func New(options Options) (*Server, error) {
 
 	configureEmail(options)
 
+	server.SetupBackgroundJobs()
```
dnephin, line 206: Background jobs should be started by `Run`, not by `New`. This is really important for testing.

Should be easy to move this to `Run`.

internal/server/data/deviceflowauthrequest.go
```diff
@@ -58,35 +57,48 @@ func CreateDeviceFlowAuthRequest(tx WriteTxn, dfar *models.DeviceFlowAuthRequest
 	return insert(tx, (*deviceFlowAuthRequestTable)(dfar))
 }
 
-func GetDeviceFlowAuthRequest(db GormTxn, selectors ...SelectorFunc) (*models.DeviceFlowAuthRequest, error) {
-	return get[models.DeviceFlowAuthRequest](db, selectors...)
```
dnephin, line 62: Thanks for replacing this use of gorm!

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,77 @@
...
+type BackgroundJobFunc func(ctx context.Context, tx data.WriteTxn, lastRunAt, currentTime time.Time) error
+
+func (s *Server) SetupBackgroundJobs() {
+	s.registerJob(jobs.RemoveOldDeviceFlowRequests, 10*time.Minute)
+	s.registerJob(jobs.RemoveExpiredAccessKeys, 1*time.Minute)
+	s.registerJob(jobs.RemoveExpiredPasswordResetTokens, 15*time.Minute)
+}
+
+func (s *Server) registerJob(job BackgroundJobFunc, every time.Duration) {
+	ctx, cancel := context.WithCancel(context.Background())
```
dnephin, line 33: The context needs to be passed in, otherwise we can't cancel these goroutines in tests.

```suggestion
func (s *Server) registerJob(ctx context.Context, job BackgroundJobFunc, every time.Duration) {
	ctx, cancel := context.WithCancel(ctx)
```

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,77 @@
...
+	s.routines = append(s.routines, routine{
+		run:  jobWrapper(ctx, s.db, job, every),
+		stop: cancel,
+	})
+}
+
+func jobWrapper(ctx context.Context, tx data.WriteTxn, job BackgroundJobFunc, every time.Duration) func() error {
```
dnephin, line 40: This is probably one of the rare places we want to use `data.DB` instead of the interface, so that a job can choose to use a transaction if it wants to.
```suggestion
func jobWrapper(ctx context.Context, tx *data.DB, job BackgroundJobFunc, every time.Duration) func() error {
```

internal/server/jobs/jobs.go
```diff
@@ -0,0 +1,20 @@
+package jobs
```
dnephin, line 1: Do we need a separate package for this? This seems like a pretty natural fit for the `internal/server` package, since these are jobs performed by our api server.

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,77 @@
...
+			defer func() {
+				if err := recover(); err != nil {
+					logging.Errorf("background job %s panic: %s", getFuncName(job), err)
+				}
+			}()
+
+			startAt := time.Now().UTC()
+			logging.Debugf("background job %s starting", getFuncName(job))
+
+			// TODO: consider attaching the context to the tx so long-running executions don't hold up app shutdown
```
dnephin, line 58: I think comments above should address this. We need to pass in the shared context, and starting a transaction with `db.Begin` will set the context in the transaction.

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,77 @@
...
+			err := job(ctx, tx, lastRunAt, startAt)
```
dnephin, line 59: I think Bruce commented on this as well, but it seems like these two times are not used.

`startAt` does not seem necessary. A job can call `time.Now` itself to get the current time.

`lastRunAt` I could see being useful, but again a job should be able to track that for itself by using a struct or closure for the job function. 

I think we benefit from keeping this interface small, so I think we should remove these two args from the interface.

dnephin (commented):
Looks good

internal/server/backgroundjobs.go
dnephin: Oh, I guess the signature for `routine.run`  also needs to change to accept a `ctx context.Context`. That way we can pass the correct context to the jobs.

internal/server/backgroundjobs.go
dnephin: I think we don't need a `stop` function for these, If we change `route.run` to accept a `context.Context`, then we can pass that in directly when it's called.

docs/api/openapi3.json
```diff
@@ -229,7 +229,7 @@
           }
         }
       },
-      "DevicePollResponse": {
```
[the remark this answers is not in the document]
ssoroka: seems like this update was missed elsewhere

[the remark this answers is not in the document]
dnephin: I think we accidentally added this file, but we don't have any checks for it, like we do the test file.  We should remove one of the duplicate docs, and make sure the CI check exists for the one we didn't remove. Something for a different PR.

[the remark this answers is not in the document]
dnephin: I've got a fix for this in #3551

internal/server/jobs/jobs.go
ssoroka: if you want more complex scheduling you may need them.

internal/server/jobs/jobs.go
ssoroka: Perhaps not, but it keeps server from calling data directly. It's its own distinct concept that has nothing to do with serving requests.

internal/server/jobs/jobs.go
dnephin: Maybe, or maybe not, we don't know. If we need them we can add them. Maybe we'll need something totally different. It's a mistake to try and guess and what we need before we need it.

internal/server/backgroundjobs.go
ssoroka: a job could do it itself, but this seems like a fine thing to give to the job runner. Why should every job that wants it reimplement time tracking logic? Can always remove it later if no jobs end up using it.

internal/server/jobs/jobs.go
dnephin: The server package already calls `data` directly in two places: `Server.loadConfig`, and in `middleware.go`. This indirection hurts readability for no benefit.

internal/server/backgroundjobs.go
dnephin: Right now no jobs need time tracking logic. This is technical debt that will need to be cleaned up, why introduce it in the first place?

dnephin: Personally I would leave a lot more up to the individual job. There's no real benefit in abstracting things like a for loop. Better to repeat a little bit to make the code more obvious. We could have just a function that accepts a time.Duration and does the select, and leave the rest up to the job. That's roughly the pattern I used in https://github.com/infrahq/infra/blob/main/internal/repeat/waiter.go and it allows the code to be used in a few different ways without having a bunch of parameters.

I think this PR strikes a pretty nice balance. It don't think it hurts to combine the for loop and the select and recovery all together, that seems quite reasonable. But the time tracking is not necessary and distracts from the parts that are being used.

ssoroka:
@dnephin resolved your comments on the function arguments

internal/server/data/passwordresettoken_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"gotest.tools/v3/assert"
+
+	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
+)
+
+func TestRemoveExpiredPasswordResetTokens(t *testing.T) {
+	tx := setupDB(t)
+	prt, err := CreatePasswordResetToken(tx, &models.Identity{Model: models.Model{ID: uid.New()}}, 0)
+	assert.NilError(t, err)
```
dnephin, line 15: Not a blocker, but I've been trying to create records that won't be deleted , to check that the filter behaves properly and leaves non-expired records. I think that'd be a good improvement to this test.

internal/server/data/deviceflowauthrequest_test.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func TestDeleteExpiredDeviceFlowAuthRequests(t *testing.T) {
+	tx := setupDB(t)
+	dfar := &models.DeviceFlowAuthRequest{
+		UserCode:   "BCDFGHJK",
+		DeviceCode: "abcdefghijklmnopqrstuvwxyz123456789000",
+		ExpiresAt:  time.Now().Add(-1),
+	}
+	err := CreateDeviceFlowAuthRequest(tx, dfar)
```
dnephin, line 19: Here as well, it would be great to include non-expired rows, to show they are not deleted.

internal/server/data/deviceflowauthrequest.go
```diff
@@ -58,35 +58,52 @@ func CreateDeviceFlowAuthRequest(tx WriteTxn, dfar *models.DeviceFlowAuthRequest
...
+type SelectDeviceFlowAuthRequestOptions struct {
```
dnephin, line 61: So far we've always named these options struct to match the function that accepts them

```suggestion
type GetDeviceFlowAuthRequestOptions struct {
```

any reason this one needs to be different?

internal/server/data/access_key_test.go
```diff
@@ -679,3 +679,21 @@ func TestGetAccessKeyByID(t *testing.T) {
...
 }
+
+func TestRemoveExpiredAccessKeys(t *testing.T) {
+	tx := setupDB(t)
+	ak := &models.AccessKey{
+		Name:      "foo expiry",
+		IssuedFor: uid.New(),
+		ExpiresAt: time.Now().Add(-1),
+	}
+	_, err := CreateAccessKey(tx, ak)
```
dnephin, line 690: Same here for test data

internal/server/data/access_key.go
```diff
@@ -304,3 +304,12 @@ func ValidateRequestAccessKey(tx *Transaction, authnKey string) (*models.AccessK
 
 	return t, nil
 }
+
+func RemoveExpiredAccessKeys(tx WriteTxn) error {
+	query := querybuilder.New("UPDATE access_keys")
+	query.B("SET deleted_at = ?", time.Now().UTC())
+	query.B("WHERE expires_at <= ?", time.Now().UTC())
```
dnephin, line 311: Not a blocker, but I wonder if we want to add a buffer between when they expire and when we soft-delete them. So that if someone accidentally sets a short expiration time they can see the key exists but is expired, instead of being confused about why the key they created doesn't show up anymore.

The job only runs every 12 hours, so that's a bit of a buffer, but keys that expired a second ago when the job runs would still be deleted. Might be worth subtracting an hour or something to this time.

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,70 @@
...
+				logging.Debugf("background job %s successful, elapsed: %s", getFuncName(job), time.Since(startAt))
+			}
+		}
+
+		for {
+			select {
+			case <-t.C:
+				jobWithRescue()
+			case <-ctx.Done():
+				return nil // time to quit.
```
dnephin, line 66: We need to stop the ticker here to prevent a goroutine leak.

```suggestion
				t.Stop()
				return nil // time to quit.
```

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,70 @@
...
+				return
+			}
+			defer func() {
+				if err := recover(); err != nil {
+					logging.Errorf("background job %s panic: %s", getFuncName(job), err)
+				}
+			}()
+
+			startAt := time.Now().UTC()
+			logging.Debugf("background job %s starting", getFuncName(job))
```
dnephin, line 51: `getFuncName` is called in 4 places. It'd be nice to capture this once in a variable, ideally before the `for` loop, to avoid using reflection for every log line.

We pay the cost here even when debug is disabled because this is not done lazily.

internal/server/backgroundjobs.go
```diff
@@ -0,0 +1,70 @@
...
+
+			err := job(ctx, tx)
+			if err != nil {
+				logging.Errorf("background job %s error: %s", getFuncName(job), err.Error())
+			} else {
+				logging.Debugf("background job %s successful, elapsed: %s", getFuncName(job), time.Since(startAt))
```
dnephin, line 57: I think this could be valuable at `INFO` level, so that we can match up the runs of background jobs with other events.

dnephin (approved):
Nice! Looks good. A few suggestions, but nothing blocking. Would be fine as follow ups.

I still think that `SetupBackgroundJobs` and the `jobs` package create unnecessary indirection that hurts readability, but not worth blocking over it.

Issue #3096, opened 2022-09-01
Connector Internal Request Error Logs


BruceMacD:
## Describe the bug
The connector has been reported to log some connection issues in some cases:
```
ReverseProxy read error during body copy: unexpected EOF
```

```
failed to lookup endpoint: Get \"REDACTED": dial tcp REDACTED: connect: connection refused"
```

## Expected behavior
If these are errors we should log some info on how to fix them, otherwise we should suppress the noise.

## Environment

- connector version 0.14.4

PR #3543, opened 2022-11-02
fix: add http connector service to serve health check

mxyng:
## Summary

The connector has a single HTTPS service which is problematic for most health checkers as they tend to default to TCP. This causes an incomplete TLS handshake which shows up in logs as `http: TLS handshake error from <addr>:<port>: EOF`. Adding a HTTP service mitigates this since the health checker does not attempt an TLS handshake

### DIFF
--- a/helm/charts/infra/templates/connector/deployment.yaml
+++ b/helm/charts/infra/templates/connector/deployment.yaml
@@ -61,6 +61,9 @@ spec:
 {{- toYaml .Values.connector.volumeMounts | nindent 12 }}
 {{- end }}
           ports:
+            - name: http
+              containerPort: {{ .Values.connector.config.addr.http }}
+              protocol: TCP
             - name: https
               containerPort: {{ .Values.connector.config.addr.https }}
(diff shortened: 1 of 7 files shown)


mxyng:
Helm lint failure is related to https://github.com/infrahq/infra/pull/3544

BruceMacD:
Resolves #3096?

BruceMacD (approved):
lgtm once tests pass (looks like it will pass if re-run)

dnephin (approved):
I've seen `TestConnector_Run` flake a few times now, I'l look into that

dnephin:
OH, it's not a test flake, it's actually failing correctly I think. The test needs to be updated to set the address to `:0`.

dnephin:
I pushed a commit to fix the test failure and fix shutdown of the connector. The pattern in the connector is not quite as nice as what we have in the `server`. We have to add an explicit shutdown. We should look to use the `server.routine` pattern here as well at some point.

PR #3545, opened 2022-11-02
Return 304 on blocking request timeout, and add label for blocking request

dnephin:
## Summary

This PR addresses two problems with monitoring of blocking API requests:

1. changes the http response status code from 504 gateway timeout, to 304 not modified, when a blocking request times out
2. add a `blocking` label to both HTTP metrics, so that we can filter out `blocking=true` in the monitors for API request latency

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"net/http"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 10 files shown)


metrics/metrics.go
```diff
@@ -45,32 +46,42 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
...
+		blocking := blockingRequestLabel(c.Request)
+
+		count := requestCount.With(prometheus.Labels{"blocking": blocking})
+		count.Inc()
+		defer count.Dec()
 
 		t := time.Now()
 		c.Next()
 
 		requestDuration.With(prometheus.Labels{
```
mxyng, line 69: Does it even make sense to have request duration metrics for blocking requests? Perhaps this can be omitted if `blocking=true`?

mxyng (approved):

metrics/metrics.go
```diff
@@ -45,32 +46,42 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
...
+			"status":   strconv.Itoa(c.Writer.Status()),
+			"blocking": blocking,
 		}).Observe(time.Since(t).Seconds())
 	}
 }
 
+func blockingRequestLabel(req *http.Request) string {
+	rawIndex := req.URL.Query().Get("lastUpdateIndex")
+	updateIndex, _ := strconv.ParseInt(rawIndex, 10, 64)
+	return strconv.FormatBool(updateIndex > 0)
```
dnephin, line 82: Instead of parsing this as an int, I think I'm going to simplify this:

```suggestion
	return strconv.FormatBool(rawIndex != "0")
```

Any non-zero value should be fine to consider this a blocking request.

dnephin: hmm, although I still need to consider the `""` case as well

PR #3546, opened 2022-11-02
Use SQL for password reset token queries

dnephin:
## Summary

This PR converts the two data queries used for password reset tokens to use SQL.

I also made the following changes:
1. when fetching the token, delete it immediately. Previously we were waiting until later to delete it. Since we run all of these operations in the same transaction there is no need to do this in separate queries. We can delete the token and use `RETURNING` to return the fields we care about at the same time. This allows us to remove one of the data functions.
1. remove `models.PasswordResetToken`. With a couple small changes to the return values this type became an implementation detail of the `data` package. The caller only needs the `user.ID` and the `token`, which we can return without a type. The model was replaced with an unexported struct in the `data` package.


internal/server/data/passwordresettoken.go
```diff
@@ -4,63 +4,86 @@ import (
...
 		ID:         uid.New(),
 		Token:      token,
 		IdentityID: user.ID,
 		ExpiresAt:  time.Now().Add(ttl).UTC(),
 	}
 
 	tries++
-	if err = save(db, prt); err != nil {
-		// TODO: must use errors.As for error types
-		if tries <= 3 && errors.Is(err, UniqueConstraintError{}) {
```
dnephin, line 34: I also resolved this TODO at the same time. Previously the `errors.Is` would never return true because the `err` would have the `Table` and `Column` field set.

dnephin (commented):
There does not appear to be any tests for these in the `data` package, but the number of possible inputs is small, so hopefully the behaviour is well covered by the API tests.

internal/server/data/passwordresettoken.go
```diff
@@ -4,63 +4,86 @@ import (
...
+
+func (p *passwordResetToken) ScanFields() []any {
+	return []any{&p.ExpiresAt, &p.ID, &p.IdentityID, &p.OrganizationID, &p.Token}
+}
+
+func (p *passwordResetToken) OnInsert() error {
+	return nil
+}
+
+func CreatePasswordResetToken(db GormTxn, user *models.Identity, ttl time.Duration) (string, error) {
```
dnephin, line 43: TODO: check user.ID and ttl are not zero values.

The old model had struct tags for this, but they were not used by anything since we've removed go-playground/validator

dnephin:
I added some direct test coverage of these two functions and fixed the conflicts.

internal/access/passwordreset.go
```diff
@@ -5,20 +5,19 @@ import (
 
 	"github.com/gin-gonic/gin"
 
-	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 )
 
-func PasswordResetRequest(c *gin.Context, email string, ttl time.Duration) (token string, user *models.Identity, err error) {
+func PasswordResetRequest(c *gin.Context, email string, ttl time.Duration) (string, *models.Identity, error) {
```
ssoroka, line 12: unnamed string return arguments are less clear than named ones.

internal/server/data/data.go
```diff
@@ -265,20 +264,6 @@ func initialize(db *DB) error {
 	return tx.Commit()
 }
 
-func getDefaultSortFromType(t interface{}) string {
```
ssoroka, line 268: did we lose default sorts?

internal/server/data/passwordresettoken_test.go
```diff
@@ -0,0 +1,55 @@
...
+			tx := txnForTestCase(t, db, db.DefaultOrg.ID)
+
+			token, err := CreatePasswordResetToken(tx, 8222, -5*time.Second)
+			assert.NilError(t, err)
+			assert.Assert(t, token != "")
+
+			_, err = GetUserIDForPasswordResetToken(tx, token)
+			assert.ErrorIs(t, err, internal.ErrExpired)
+		})
+	})
```
ssoroka, line 54: might be worth adding a case where the record is gone because the background job trimmed it already.

ssoroka (approved):

internal/access/passwordreset.go
dnephin: Ya, especially true when there are multiple return values with the same type. Usually when they have different types it's pretty clear.

It's unfortunate that naming them also extends the scope of the variable (to any place in the function, instead of only after declaration). I can add these back

internal/server/data/data.go
dnephin: Every list function has a sort, but there's no central function that uses reflection to try and guess at which field to use.

internal/server/data/passwordresettoken_test.go
dnephin: I believe that case is covered by the second call in `deletes token`. It's calling the function with a token that was already deleted.

internal/server/data/passwordresettoken.go
```diff
@@ -2,65 +2,96 @@ package data
...
-	prts, err := list[models.PasswordResetToken](db, &Pagination{Limit: 1}, func(db *gorm.DB) *gorm.DB {
-		return db.Where("token = ?", token)
-	})
-	if err != nil {
-		return nil, err
-	}
+// ClaimPasswordResetToken deletes the password reset token, and returns the
+// user ID that was associated with the token. Returns an error if the token
+// does not exist or has expired.
+func ClaimPasswordResetToken(tx WriteTxn, token string) (uid.ID, error) {
```
dnephin, line 80: A push one more commit which renames this function from `Get` to `Claim`. Now that it does a delete, I think calling it `Get` could easily be misleading.

PR #3547, opened 2022-11-02
feat: access keys ui

pdevine:
## Summary

This PR includes a screen for allowing users to create their own access keys through the UI.

## 📷 📹 

https://user-images.githubusercontent.com/63033505/200917139-4887c2c5-a339-48c5-afec-2c8eaaf2f889.mov


ui/components/calendar.js
```diff
@@ -0,0 +1,315 @@
...
+]
+
+const daysOfWeek = [
+  { id: 0, title: 'S', value: 'Sunday' },
+  { id: 1, title: 'M', value: 'Monday' },
+  { id: 2, title: 'T', value: 'Tuesday' },
+  { id: 3, title: 'W', value: 'Wednesday' },
+  { id: 4, title: 'T', value: 'Thursday' },
+  { id: 5, title: 'F', value: 'Friday' },
+  { id: 6, title: 'S', value: 'Sunday' },
```
BruceMacD, line 27:
```suggestion
  { id: 6, title: 'Su', value: 'Sunday' },
```

ui/pages/settings/add/access-key.js
```diff
@@ -0,0 +1,345 @@
...
+                <>
+                  <div>
+                    This access key must be used at least once every{' '}
+                    <span className='font-semibold text-gray-900'>30 days</span>
+                    ,
+                  </div>
+                  <div>
+                    and will expire on{' '}
+                    <span className='font-semibold text-gray-900'>
+                      {moment()
```
BruceMacD, line 309: It feels a bit odd to me that this updates every time I type in the name and the page re-renders. Should we just be more vague and give the date without the time?

Ex: `February 13th, 2023` rather than `10:54:58 am, February 13th 2023`

We can display a more accurate expiry after creation if we want to be exact

ui/components/calendar.js
```diff
@@ -0,0 +1,315 @@
+import moment from 'moment'
+import React, { useRef, useEffect, useState } from 'react'
+import { ChevronLeftIcon, ChevronRightIcon } from '@heroicons/react/outline'
+
+const monthName = [
```
BruceMacD, line 5: Is the logic in this file adapted from a different time library? Seems like there is some functionality here we don't need. If we adapted it, it would be nice to credit the original library in a comment at the top of the file and mention their license.

ui/pages/settings/add/access-key.js
```diff
@@ -0,0 +1,345 @@
...
+import * as ReactDOM from 'react-dom'
+import copy from 'copy-to-clipboard'
+import moment from 'moment'
+
+import { useUser } from '../../../lib/hooks'
+
+import Dashboard from '../../../components/layouts/dashboard'
+
+const EXPIRATION_RATE = [
+  { name: '30 days', value: '720h', default: true },
```
jmorganca, line 21: Instead of using `default` we could probably use the index

ui/pages/settings/add/access-key.js
hoyyeva: Patrick suggested to add the time. What do you think @pdevine

BruceMacD (approved):

PR #3549, opened 2022-11-03
feature: grants by name

pdevine:
## Summary

Allow `PATCH /api/grants` to use user names as well as user/group IDs

## Checklist

### DIFF
--- a/api/grant.go
+++ b/api/grant.go
@@ -146,7 +146,8 @@ func (r ListGrantsRequest) SetPage(page int) Paginatable {
 type GrantRequest struct {
 	User      uid.ID `json:"user"`
 	Group     uid.ID `json:"group"`
-	Name      string `json:"name"`
+	UserName  string `json:"userName"`
+	GroupName string `json:"groupName"`
 	Privilege string `json:"privilege" example:"view" note:"a role or permission"`
 	Resource  string `json:"resource" example:"production" note:"a resource name in Infra's Universal Resource Notation"`
 }
@@ -155,8 +156,9 @@ func (r GrantRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
 		validate.RequireOneOf(
 			validate.Field{Name: "user", Value: r.User},
-			validate.Field{Name: "name", Value: r.Name},
+			validate.Field{Name: "userName", Value: r.UserName},
 			validate.Field{Name: "group", Value: r.Group},
+			validate.Field{Name: "groupName", Value: r.GroupName},
 		),
 		validate.Required("privilege", r.Privilege),
 		validate.Required("resource", r.Resource),
--- a/api/grant.go
+++ b/api/grant.go
@@ -146,6 +146,7 @@ func (r ListGrantsRequest) SetPage(page int) Paginatable {
 type GrantRequest struct {
 	User      uid.ID `json:"user"`
 	Group     uid.ID `json:"group"`
+	Name      string `json:"name"`
 	Privilege string `json:"privilege" example:"view" note:"a role or permission"`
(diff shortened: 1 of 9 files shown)


api/grant.go
```diff
@@ -146,6 +146,7 @@ func (r ListGrantsRequest) SetPage(page int) Paginatable {
 type GrantRequest struct {
 	User      uid.ID `json:"user"`
 	Group     uid.ID `json:"group"`
+	Name      string `json:"name"`
```
BruceMacD, line 149:
```suggestion
	UserName      string `json:"userName"`
```

Just specifying `name` looks like naming the grant to me.

internal/access/identity.go
```diff
@@ -34,6 +34,23 @@ func GetIdentity(c *gin.Context, id uid.ID) (*models.Identity, error) {
 	return data.GetIdentity(rCtx.DBTxn, data.GetIdentityOptions{ByID: id, LoadProviders: true})
 }
 
+func GetIdentityByName(c *gin.Context, name string) (*models.Identity, error) {
```
BruceMacD, line 37: Suggestion: I wonder if we could move the `GetIdentityOptions` to the caller in `server/users.go` and use one `access.GetIdentity` function that take the `GetIdentityOptions` as an argument.

internal/server/grants.go
```diff
@@ -145,33 +145,55 @@ func (a *API) UpdateGrants(c *gin.Context, r *api.UpdateGrantsRequest) (*api.Emp
...
+		subject = uid.NewIdentityPolymorphicID(identity.ID)
 	case r.User != 0:
 		subject = uid.NewIdentityPolymorphicID(r.User)
 	case r.Group != 0:
 		subject = uid.NewGroupPolymorphicID(r.Group)
 	}
 
+	switch {
+	case subject == "":
+		return nil, fmt.Errorf("%w: must specify name, user, or group", internal.ErrBadRequest)
```
BruceMacD, line 187:
```suggestion
		return nil, fmt.Errorf("%w: must specify user name, user, or group", internal.ErrBadRequest)
```

As opposed to group name

internal/server/grants.go
```diff
@@ -145,33 +145,55 @@ func (a *API) UpdateGrants(c *gin.Context, r *api.UpdateGrantsRequest) (*api.Emp
...
-func getGrantFromGrantRequest(r api.GrantRequest) *models.Grant {
+func getGrantFromGrantRequest(c *gin.Context, r api.GrantRequest) (*models.Grant, error) {
 	var subject uid.PolymorphicID
 
 	switch {
+	case r.Name != "":
+		// lookup user name
+		identity, err := access.GetIdentityByName(c, r.Name)
+		if err != nil {
+			return nil, fmt.Errorf("%w: couldn't find user %s", internal.ErrBadRequest, r.Name)
```
BruceMacD, line 176: This could also be a result of insufficient permissions I believe

internal/access/identity.go
dnephin: I like this suggestion. We've done this in a few other access functions already and it works well.

api/grant.go
dnephin: Ya, this needs to be `UserName`, because groups also have a name, so even `SubjectName` is ambiguous.

mxyng: It seems odd this references user by id or name but group only by id. Does it also need `GroupName`?

internal/server/grants.go
pdevine: I switched this to look for if the userName (and now also groupName) was not found. It will still pass back the insufficient permissions which isn't tbh the best error given it might not be clear *why* it's trying to look up an identity (or group).

Ideally we would automatically give everyone the View permission though by default which would solve this most of the time.

BruceMacD (approved):
Looks good to go to me now

PR #3550, opened 2022-11-03
fix: use httptest.NewRequest where appropriate

mxyng:
## Summary

`httptest.NewRequest` is more appropriate use with `ServeHTTP` as its intended to simulate an incoming request rather than a client request (`http.NewRequest`). This is critical as one of the differences between server and client requests has caused problems on [two](https://github.com/infrahq/infra/pull/2945) [separate](https://github.com/infrahq/infra/pull/3548) occasions.

Instances where the request is used with a `http.Client` are left unchanged.

### DIFF
--- a/internal/server/redirect_test.go
+++ b/internal/server/redirect_test.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"context"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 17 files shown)


BruceMacD (approved):
Good catch

internal/server/redirect_test.go
```diff
@@ -24,12 +23,10 @@ func TestVerifyAndRedirect_Works(t *testing.T) {
...
 	assert.Assert(t, !user.Verified)
 
-	redirectURL := "https://example.com/hello"
+	redirectURL := "/hello"
 	url := wrapLinkWithVerification(redirectURL, "example.com", user.VerificationToken)
 
-	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, url, nil)
-	assert.NilError(t, err)
-
+	req := httptest.NewRequest(http.MethodGet, url, nil)
```
dnephin, line 29: This change is a bit surprising. It seems like `wrapLinkWithVerification` is expecting a full url, so passing in just a path doesn't seem correct.

`redirectURL` is used in a comparison below, I guess that's why the value changed?  

We always pass `wrapLinkWithVerification` a full URL in our production code paths, so we should probably do that here as well.

dnephin (commented):
Good fix!

internal/server/redirect_test.go
mxyng: `httptest.NewRequest` uses `example.com` as request host if the URL is a relative path which is why this has less impact than expected

Issue #3261, opened 2022-09-19
Enhanced Google identity provider support


jmorganca:
- [x] Sign up with Google
- [ ] Sync groups and group membership
- [x] Invited users can also accept an invite and sign up with Google

BruceMacD:
Group sync out of scope for now.

PR #3553, opened 2022-11-03
feat: allow specifying email domains that can login with provider

BruceMacD:
- add optional allowed domains list to providers
- check allowed domains list on login with provider
- sanitize list providers response when not authenticated

## Summary
Allow optionally specifying which email domains are allowed to login with an identity provider. This is the first step in social login, which will require allowed domains be specified.

## Checklist

## Related Issues

Part of #3261

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ lint:
 	golangci-lint run $(LINT_ARGS)
 
 .PHONY: docs
-docs: docs/api/openapi3.json
+docs:
 	go run ./internal/docgen
 
 .PHONY: docs/api/openapi3.json
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ lint:
 	golangci-lint run $(LINT_ARGS)
 
 .PHONY: docs
-docs:
+docs: docs/api/openapi3.json
 	go run ./internal/docgen
 
 .PHONY: docs/api/openapi3.json
--- a/api/provider.go
+++ b/api/provider.go
@@ -48,6 +48,19 @@ func ValidateAllowedDomains(value []string) validate.StringSliceRule {
 		Value:     value,
 		Name:      "allowedDomains",
 		MaxLength: 20,
+		ItemRule: validate.StringRule{
+			Name:      "allowedDomains.values",
+			MaxLength: 254,
+			CharacterRanges: []validate.CharRange{
+				validate.AlphabetLower,
+				validate.AlphabetUpper,
+				validate.Numbers,
+				validate.Dash,
+				validate.Dot,
+				validate.Underscore,
+			},
+			FirstCharacterRange: validate.AlphaNumeric,
+		},
 	}
 }
 
--- a/api/provider.go
+++ b/api/provider.go
@@ -20,24 +20,26 @@ func (r ProviderAPICredentials) ValidationRules() []validate.ValidationRule {
 }
 
 type Provider struct {
-	ID       uid.ID   `json:"id"`
-	Name     string   `json:"name" example:"okta"`
-	Created  Time     `json:"created"`
-	Updated  Time     `json:"updated"`
-	URL      string   `json:"url" example:"infrahq.okta.com"`
-	ClientID string   `json:"clientID" example:"0oapn0qwiQPiMIyR35d6"`
-	Kind     string   `json:"kind" example:"oidc"`
-	AuthURL  string   `json:"authURL" example:"https://example.com/oauth2/v1/authorize"`
-	Scopes   []string `json:"scopes" example:"['openid', 'email']"`
+	ID             uid.ID   `json:"id"`
+	Name           string   `json:"name" example:"okta"`
+	Created        Time     `json:"created"`
+	Updated        Time     `json:"updated"`
+	URL            string   `json:"url" example:"infrahq.okta.com"`
+	ClientID       string   `json:"clientID" example:"0oapn0qwiQPiMIyR35d6"`
+	Kind           string   `json:"kind" example:"oidc"`
+	AuthURL        string   `json:"authURL" example:"https://example.com/oauth2/v1/authorize"`
+	Scopes         []string `json:"scopes" example:"['openid', 'email']"`
+	AllowedDomains []string `json:"allowedDomains" example:"['example.com', 'infrahq.com']"` // only returned when authenticated
 }
 
 type CreateProviderRequest struct {
-	Name         string                  `json:"name" example:"okta"`
-	URL          string                  `json:"url" example:"infrahq.okta.com"`
-	ClientID     string                  `json:"clientID" example:"0oapn0qwiQPiMIyR35d6"`
-	ClientSecret string                  `json:"clientSecret" example:"jmda5eG93ax3jMDxTGrbHd_TBGT6kgNZtrCugLbU"`
-	Kind         string                  `json:"kind" example:"oidc"`
-	API          *ProviderAPICredentials `json:"api"`
+	Name           string                  `json:"name" example:"okta"`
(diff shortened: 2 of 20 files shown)


docs/api/openapi3.json
```diff
@@ -1153,7 +1169,7 @@
       "url": "https://www.elastic.co/licensing/elastic-license"
     },
     "title": "Infra API",
-    "version": "0.0.0"
+    "version": "99.99.99999"
```
BruceMacD, line 1172: Not sure if this version updating is expected

dnephin: It's not, did you generate this file manually, or was it done with the test? The test should fail because it's expecting version 0.

I think we should not tell anyone to use `9.9.9` because that effectively disables versioning for them.

internal/server/testdata/openapi3.json
```diff
@@ -0,0 +1,6957 @@
+{
```
dnephin, line 1: This file was removed. Bad rebase with the changes in #3551 ?

docs/api/openapi3.json
BruceMacD: I used `make docs`. I'll update that command if we shouldnt be generating it anymore

dnephin: Oh, I did not realize that was still being done.  I guess that's why I had moved it to a test-only location, to not conflict with the file that we generate at runtime for a release.

I think removing it from `make docs` for now makes sense. We'll figure out how to integrate it into the docs later.

BruceMacD: Fixed in the new commit

internal/server/testdata/openapi3.json
BruceMacD: Removed 👍

api/provider.go
```diff
@@ -59,13 +61,14 @@ type PatchProviderRequest struct {
...
-	ClientID     string                  `json:"clientID" example:"0oapn0qwiQPiMIyR35d6"`
-	ClientSecret string                  `json:"clientSecret" example:"jmda5eG93ax3jMDxTGrbHd_TBGT6kgNZtrCugLbU"`
-	Kind         string                  `json:"kind" example:"oidc"`
-	API          *ProviderAPICredentials `json:"api"`
+	ID             uid.ID                  `uri:"id" json:"-"`
+	Name           string                  `json:"name" example:"okta"`
+	URL            string                  `json:"url" example:"infrahq.okta.com"`
+	ClientID       string                  `json:"clientID" example:"0oapn0qwiQPiMIyR35d6"`
+	ClientSecret   string                  `json:"clientSecret" example:"jmda5eG93ax3jMDxTGrbHd_TBGT6kgNZtrCugLbU"`
+	AllowedDomains []string                `json:"allowedDomains" example:"['example.com', 'infrahq.com']"`
```
dnephin, line 69: I think this needs another validation rule for `CreateProviderRequest` and `UpdateProviderRequest` to not allow commas (since we're using CommaSeparatedString, and maybe also set a maximum allowed length ? 

We could also require an `@` sign in every value, since that's needed later.

dnephin (commented):
Looks good! 

Let's chat more about the change to `GetProvider` and `ListProvider`.

internal/server/authn/oidc.go
```diff
@@ -45,6 +48,18 @@ func (a *oidcAuthn) Authenticate(ctx context.Context, db *data.Transaction, requ
 		return AuthenticatedIdentity{}, fmt.Errorf("exhange code for tokens: %w", err)
 	}
 
+	if len(provider.AllowedDomains) > 0 {
+		// get the domain of the email
+		at := strings.LastIndex(email, "@") // get the last @ since the email spec allows for multiple @s
+		if at == -1 {
+			return AuthenticatedIdentity{}, fmt.Errorf("%s is an invalid email address", email)
+		}
+		domain := email[at+1:]
```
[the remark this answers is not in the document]
BruceMacD: Im going to leave this one as is since technically the email spec allows an @ in an email, even though I've never seen it. Getting the last @ means we are getting the actual domain.

internal/access/provider.go
```diff
@@ -22,12 +22,34 @@ func CreateProvider(c *gin.Context, provider *models.Provider) error {
...
 func GetProvider(c *gin.Context, id uid.ID) (*models.Provider, error) {
 	rCtx := GetRequestContext(c)
-	return data.GetProvider(rCtx.DBTxn, data.GetProviderOptions{ByID: id})
+	provider, err := data.GetProvider(rCtx.DBTxn, data.GetProviderOptions{ByID: id})
+	if err != nil {
+		return nil, err
+	}
+	// if the caller is not authenticated for this org do not return the allowed domains, it is sensitive info
+	reqKey := rCtx.Authenticated.AccessKey
+	if reqKey == nil || reqKey.OrganizationID != rCtx.DBTxn.OrganizationID() {
```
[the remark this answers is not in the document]
dnephin: I wrote this up here #3599

Makefile
```diff
@@ -51,9 +51,5 @@ lint:
 	golangci-lint run $(LINT_ARGS)
 
 .PHONY: docs
-docs: docs/api/openapi3.json
+docs:
 	go run ./internal/docgen
-
-.PHONY: docs/api/openapi3.json
-docs/api/openapi3.json:
-	go run ./internal/openapigen $@
```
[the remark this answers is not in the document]
BruceMacD: put this back

internal/access/provider.go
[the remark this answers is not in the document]
BruceMacD: Remove this and added a TODO for #3599

internal/server/config_test.go
```diff
@@ -452,6 +452,7 @@ func TestLoadConfigWithProviders(t *testing.T) {
 		Kind:               models.ProviderKindOIDC, // the kind gets the default value
 		AuthURL:            "example.com/oauth2/default/v1/token",
 		Scopes:             []string{"openid", "email"},
+		AllowedDomains:     []string{},
```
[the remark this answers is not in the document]
BruceMacD: thanks, make this change

api/provider.go
BruceMacD: Created StringSliceValidation

api/provider.go
```diff
@@ -32,23 +32,33 @@ type Provider struct {
...
 }
 
 var kinds = []string{"oidc", "okta", "azure", "google"}
 
+func ValidateAllowedDomains(value []string) validate.StringSliceRule {
+	return validate.StringSliceRule{
+		Value:     value,
+		Name:      "allowedDomains",
+		MaxLength: 20,
+	}
```
ssoroka, line 51: might be nice to validate something about the contents of the items.

internal/server/data/migrations.go
```diff
@@ -883,3 +884,14 @@ func addDestinationKind() *migrator.Migration {
 		},
 	}
 }
+
+func addAllowedDomainsToProvidersTable() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-11-03T13:00",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `ALTER TABLE providers ADD COLUMN IF NOT EXISTS allowed_domains text DEFAULT ''`
```
ssoroka, line 892: could consider making this a jsonb field, though that would not work with `CommaSeparatedStrings`.. so it would need some extra work.

ssoroka (approved):

internal/server/data/migrations.go
BruceMacD: Let's do this in a separate change, not allowing commas in the domains is ok for now, jsonb would be nice for lookup though

PR #3560, opened 2022-11-04
Use 499 status code for context cancelled.

dnephin:
## Summary

So that these errors don't show up as 500 errors in our metrics and logs.

There's no standard response code for this, but Nginx uses 499 so let's do the same.


BruceMacD (approved):
Let's get this one in to fix the incident monitoring

PR #3563, opened 2022-11-04
fix: add http connector service

mxyng:
## Summary

Iteration on https://github.com/infrahq/infra/pull/3543. Separate out the Helm changes as it will cause problems if image and chart are deploy out of sync. Filter service port to pick out the `https` service port instead of using the first, which will be `http`. `http` will not work as the destination port since it does not serve the proxy

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -340,6 +340,7 @@ var runConnector = connector.Run
 func defaultConnectorOptions() connector.Options {
 	return connector.Options{
 		Addr: connector.ListenerOptions{
+			HTTP:    ":80",
 			HTTPS:   ":443",
 			Metrics: ":9090",
 		},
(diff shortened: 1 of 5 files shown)


internal/cmd/cmd.go
```diff
@@ -340,6 +340,7 @@ var runConnector = connector.Run
 func defaultConnectorOptions() connector.Options {
 	return connector.Options{
 		Addr: connector.ListenerOptions{
+			HTTP:    ":80",
```
BruceMacD, line 343: Should we allow HTTP listeners from the connector? Seems like we wouldnt want that.

mxyng: The HTTP service serves only a health endpoint. This is useful in the common case of external load balancers in public cloud where the load balancer health check often defaults to TCP checks which produces EOF errors when sent to the HTTPS service. By exposing a HTTP endpoint, the connector won't complain when the external load balancer hits it with a TCP handshake. It does not serve the reverse proxy so there's no security risk; sending a request path other than `/healthz` will return 404

BruceMacD (approved):

PR #3564, opened 2022-11-04
fix(helm): http connector service

mxyng:
## Summary

Split from https://github.com/infrahq/infra/pull/3563 to avoid out-of-sync deployment issues. Should be merged after #3563 has been merged and released


BruceMacD (approved):
Looks good once #3563 in

mxyng:
Removing changes to liveness and readiness checks so it doesn't break deployments using existing images

PR #3567, opened 2022-11-04
Fix response body for unknown paths

### LINKED ISSUE #3161: Non-existent API routes returns an HTML response

dnephin:
Fixes #3161

This problem is most obvious when a new client is making requests to an old server that does not yet have the new API path. Previously the response would be an HTML page from the UI. With this fix it should be a JSON API error response, which is much easier to debug.

The test for this was updated to enable the UI proxy, so that the test covers the same scenario as production. I had to also switch from `httptest.ResponseRecorder` to `httptest.Server`, because the response recorder does not work with the proxy (it would panic).

I found https://stackoverflow.com/questions/36357791/gin-router-path-segment-conflicts-with-existing-wildcard, which confirms this is a difficult problem to solve. One of the options there is to use a middleware (which is actually our problem), but it gave me the idea of just ignoring these paths in our middleware.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -257,22 +257,27 @@ func runTelemetryHeartbeat(ctx context.Context, tel *Telemetry) error {
 }
 
 func registerUIRoutes(router *gin.Engine, opts UIOptions) {
-	if opts.ProxyURL.Host != "" {
-		remote := opts.ProxyURL.Value()
-		proxy := httputil.NewSingleHostReverseProxy(remote)
-		proxy.Director = func(req *http.Request) {
-			req.Host = remote.Host
-			req.URL.Scheme = remote.Scheme
-			req.URL.Host = remote.Host
-		}
-		proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
-
-		router.Use(func(c *gin.Context) {
-			proxy.ServeHTTP(c.Writer, c.Request)
-			c.Abort()
-		})
+	if opts.ProxyURL.Host == "" {
 		return
 	}
+	remote := opts.ProxyURL.Value()
+	proxy := httputil.NewSingleHostReverseProxy(remote)
+	proxy.Director = func(req *http.Request) {
+		req.Host = remote.Host
+		req.URL.Scheme = remote.Scheme
+		req.URL.Host = remote.Host
+	}
+	proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
+
+	router.Use(func(c *gin.Context) {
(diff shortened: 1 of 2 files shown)


internal/server/server.go
```diff
@@ -257,22 +257,27 @@ func runTelemetryHeartbeat(ctx context.Context, tel *Telemetry) error {
...
+		req.URL.Host = remote.Host
+	}
+	proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
+
+	router.Use(func(c *gin.Context) {
+		// Don't proxy /api/* paths
+		if strings.HasPrefix(c.Request.URL.Path, "/api/") {
+			c.Next()
+			return
+		}
```
dnephin, line 277: Maybe not the most elegant solution, but this is the only one I could find. See the stackoverflow I linked in the description for a very long answer about possible workarounds to this problem.

internal/server/server.go
```diff
@@ -257,22 +257,27 @@ func runTelemetryHeartbeat(ctx context.Context, tel *Telemetry) error {
...
-			req.URL.Scheme = remote.Scheme
-			req.URL.Host = remote.Host
-		}
-		proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
-
-		router.Use(func(c *gin.Context) {
-			proxy.ServeHTTP(c.Writer, c.Request)
-			c.Abort()
-		})
+	if opts.ProxyURL.Host == "" {
```
dnephin, line 260: I swapped this condition to a guard so that we can return early. Best viewed with [Hide Whitespace](https://github.com/infrahq/infra/pull/3567/files?diff=split&w=1)

internal/server/server.go
BruceMacD: This seems like a fine solution to me

BruceMacD (commented):
What's the actual error returned now?

dnephin:
>What's the actual error returned now?

It's a 404 response code, but with a full page of HTML in the response body. I guess it's whatever the UI sends for its 404 page.

BruceMacD (approved):

mxyng (approved):

PR #3583, opened 2022-11-07
Update tests that use gorm to use the stdlib interface

dnephin:
## Summary

This PR updates a bunch of tests that were using the gorm interface for queries to use the stdlib interface for queries.

### DIFF
--- a/internal/server/data/migrator/helpers.go
+++ b/internal/server/data/migrator/helpers.go
@@ -15,7 +15,7 @@ func HasTable(tx DB, name string) bool {
 		SELECT count(*)
 		FROM information_schema.tables
 		WHERE table_schema = CURRENT_SCHEMA()
-		AND table_name = ? AND table_type = 'BASE TABLE'
+		AND table_name = $1 AND table_type = 'BASE TABLE'
 	`
 	if err := tx.QueryRow(stmt, name).Scan(&count); err != nil {
 		logging.L.Warn().Err(err).Msg("failed to check if table exists")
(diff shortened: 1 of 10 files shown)


internal/server/data/migrator/helpers.go
```diff
@@ -15,7 +15,7 @@ func HasTable(tx DB, name string) bool {
 		SELECT count(*)
 		FROM information_schema.tables
 		WHERE table_schema = CURRENT_SCHEMA()
-		AND table_name = ? AND table_type = 'BASE TABLE'
+		AND table_name = $1 AND table_type = 'BASE TABLE'
```
dnephin, line 18: This commit removes gorm entirely from the `internal/server/data/migrator` package. With gorm removed we have to use the PostgreSQL placeholder in queries which is `$n` instead of `?`.

We'll have to make this change in the `data` package as well, but that's going to be a bigger effort. It was easy enough to change the few queries in this package now.

internal/server/data/migrator/migrator_test.go
```diff
@@ -362,46 +355,25 @@ func runDBTests(t *testing.T, fn func(t *testing.T, db DB)) {
...
-		databases = append(databases, dbDriver{dialect: "postgres", driver: pg.Dialector})
+		databases = append(databases, dbDriver{dialect: "postgres", dsn: pg.DSN})
 	}
 
 	for _, database := range databases {
 		// Ensure defers are not stacked up for each DB
-		t.Run(database.driver.Name(), func(t *testing.T) {
-			db, err := gorm.Open(database.driver, &gorm.Config{})
+		t.Run(database.dialect, func(t *testing.T) {
+			db, err := sql.Open("pgx", database.dsn)
```
dnephin, line 364: This is where we switch from gorm to stdlib for opening the DB (only in the migrator tests for now).

BruceMacD (approved):

PR #3584, opened 2022-11-07
Remove gorm from models

dnephin:
## Summary

Making this a draft because it depends on #3542 (which removes the last use of `data.get` and `GormTxn`). Once that merges I'll rebase this PR and it should be ready.

This PR:
* removes `gorm` struct tags from models, and changes the deletedAt field to use the stdlib
* removes the `Modelable` which is no longer used. We have interface for more specific things implemented by models like `Updatable` and `orgMember`
* removes `GormTxn` and replaces all references to it with `ReadTxn` or `WriteTxn`, and removes the `GormDB` method from `Transaction` and `DB` now that nothing uses it


internal/server/data/deviceflowauthrequest.go
```diff
@@ -58,8 +56,9 @@ func CreateDeviceFlowAuthRequest(tx WriteTxn, dfar *models.DeviceFlowAuthRequest
 	return insert(tx, (*deviceFlowAuthRequestTable)(dfar))
 }
 
-func GetDeviceFlowAuthRequest(db GormTxn, selectors ...SelectorFunc) (*models.DeviceFlowAuthRequest, error) {
-	return get[models.DeviceFlowAuthRequest](db, selectors...)
+func GetDeviceFlowAuthRequest(tx ReadTxn) (*models.DeviceFlowAuthRequest, error) {
+	// FIXME:
+	return nil, nil
```
dnephin, line 61: This is just a temporary patch until #3542 is merged.

Issue #3585, opened 2022-11-07
Last steps for removing Gorm


dnephin:
This is a continuation of #2415 

After #3583 and #3584 are merged (done), the remaining tasks for removing gorm are:

1. Replace `internal/logging/database.go` (our shim for sending gorm query logs to zerolog) with instrumentation in `data.Transaction` for query logging. We should be able to use essentially the same implementation as our shim.
2. Switch from using `?` as the query placeholder to the postgreSQL standard placeholder of `$n` where n is the index of the argument in the argument list. 
3. replace `gorm.Open` with `sql.Open`, this should be an easy change once the above tasks are complete
4. after completed the above tasks, we are getting errors about converting queries that use  `IN (?)` statements. The database driver does not support encoding a slice into a single `?` placeholder. We can use a workaround of `any` and `all` functions ([docs](https://www.postgresql.org/docs/current/functions-comparisons.html#id-1.5.8.30.16)), or we could create our own helper that adds the necessary placeholders to the query string and expand out all the values in the slice:

    ```go
    query.B("id IN")
    query.B(placeholdersForInComparision(ids), ids...)
    ```
    
    `placeholdersForInComparision` would produce a string that looks like `(?, ?, ?, ...)` where the number of `?` is the length of ids.

Tasks 1 to 3 done in #3643

PR #3587, opened 2022-11-07
Add query logging without gorm

dnephin:
Completes task 1 in #3585
Branched from #3584

In preparation for removing gorm, this will give us query logging without gorm.

I mostly copied the existing implementation, but there are a few small differences which I will comment on inline


internal/server/data/data.go
```diff
@@ -390,3 +405,35 @@ func InfraConnectorIdentity(db ReadTxn) *models.Identity {
...
+
+const slowQueryThreshold = time.Second
+
+func logQuery(query string, err error, startedAt time.Time, rows int64) {
+	level := zerolog.TraceLevel
+
+	elapsed := time.Since(startedAt)
+	switch {
+	case errors.Is(err, sql.ErrNoRows), errors.Is(err, gorm.ErrRecordNotFound):
+		level = zerolog.WarnLevel
```
dnephin, line 417: Previously we were using `ErrorLevel` for queries that errored with "No Rows". I've changed it to WARN here, since a not found doesn't seem like a critical error.

internal/server/data/data.go
```diff
@@ -390,3 +405,35 @@ func InfraConnectorIdentity(db ReadTxn) *models.Identity {
...
+	case elapsed > slowQueryThreshold:
+		level = zerolog.WarnLevel
+	}
+
+	query = normalizeQueryString(query)
```
dnephin, line 422: Gorm was normalizing the query, I guess because it was parsing the full query and rewriting it for the appropriate dialect.  Since that doesn't happen anymore here I've done some basic normalization to remove extra whitespace.

internal/server/data/data.go
```diff
@@ -390,3 +405,35 @@ func InfraConnectorIdentity(db ReadTxn) *models.Identity {
...
+	logging.L.WithLevel(level).
+		CallerSkipFrame(2). // logQuery + tx.{Query,Exec}
+		Int64("rows", rows).
+		Str("query", query).
+		Dur("elapsed", elapsed).
+		Msg("DB query")
```
dnephin, line 428: Added a non-empty message so that we can find these easily by searching for `DB query`.

internal/server/data/data.go
```diff
@@ -390,3 +405,35 @@ func InfraConnectorIdentity(db ReadTxn) *models.Identity {
...
+		level = zerolog.WarnLevel
+	case elapsed > slowQueryThreshold:
+		level = zerolog.WarnLevel
+	}
+
+	query = normalizeQueryString(query)
+	logging.L.WithLevel(level).
+		CallerSkipFrame(2). // logQuery + tx.{Query,Exec}
+		Int64("rows", rows).
+		Str("query", query).
```
mxyng, line 426: One thing we need to be mindful of when logging SQL queries is to omit query parameters lest they leak secrets or other information

dnephin: Right! This is the first argument to `Tx.{Query, QueryRow, Exec}` which can never include those values (because it would be a potential SQL injection attack if they did).

dnephin:
These changes are included in https://github.com/infrahq/infra/pull/3643, so going to close this

PR #3590, opened 2022-11-07
maintain: make userid field for access keys consistent

pdevine:
## Summary

Change `user_id` to `userID` in `GET /api/access-keys?userID=` to make it consistent with the rest of the API.

## Checklist

Note that this will break the old API if something is using it, however, it's unlikely that a lot of callers (or any, really) are actually using it right now.

### DIFF
--- a/api/access_key.go
+++ b/api/access_key.go
@@ -17,7 +17,7 @@ type AccessKey struct {
 }
 
 type ListAccessKeysRequest struct {
-	UserID      uid.ID `form:"user_id"`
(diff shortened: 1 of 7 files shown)


jmorganca (approved):
Thanks for fixing this 🙏 !

dnephin (commented):
Nice fix!  

This is a breaking change, right? Existing CLI versions will break when they try to list keys for a user.

Should we create a handler for the old version make upgrades smooth?

PR #3591, opened 2022-11-07
Add kind field to destinations

dnephin:
## Summary

So that we can support more than one kind of destination.

This PR adds api, data, and CLI support for `destination.kind`. I'm not sure what (if anything) needs to change for the UI.

I made this a non-breaking change by making the API field optional and defaulting to kubernetes (since that's the only destination that exists right now). In the future we can make it required in the API as well.

### DIFF
--- a/api/destination.go
+++ b/api/destination.go
@@ -9,6 +9,7 @@ type Destination struct {
 	ID         uid.ID                `json:"id"`
 	UniqueID   string                `json:"uniqueID" form:"uniqueID" example:"94c2c570a20311180ec325fd56"`
 	Name       string                `json:"name" form:"name"`
+	Kind       string                `json:"kind"`
 	Created    Time                  `json:"created"`
 	Updated    Time                  `json:"updated"`
 	Connection DestinationConnection `json:"connection"`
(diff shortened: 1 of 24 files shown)


api/destination.go
```diff
@@ -61,6 +64,9 @@ func (r CreateDestinationRequest) ValidationRules() []validate.ValidationRule {
 		validate.Required("uniqueID", r.UniqueID),
 		validateDestinationName(r.Name),
 		validate.Required("name", r.Name),
+		// Allow "" for versions 0.16.1 and prior
+		// TODO: make this required in the future
+		validate.Enum("kind", r.Kind, []string{"kubernetes", "ssh", ""}),
```
jmorganca, line 69: We may want to keep this optional (and set it to `kubernetes` if unset for backwards compat). To make infra extensible it may be helpful for users to be able to set this to an arbitrary string (maybe with some format limitations)

jmorganca (approved):
Looks good!

api/destination.go
jmorganca: SGTM - well put

PR #3598, opened 2022-11-08
feat: ui social login

BruceMacD:
## Summary
This change allows for a base login redirect so that we can redirect the user to their org after doing social login (ex: "login with google").

Decided to merge this separately from some other changes I'm working on that allow configuring social login through the UI.

## Checklist

## Related Issues

Part of #3261

### DIFF
--- a/ui/lib/login.js
+++ b/ui/lib/login.js
@@ -27,8 +27,7 @@ export function saveToVisitedOrgs(domain, orgName) {
 }
 
 export function currentBaseDomain() {
-  let domain = window.location.host
-  let parts = domain.split('.')
+  let parts = window.location.host.split('.')
   if (parts.length > 2) {
     parts.shift() // remove the org
     domain = parts.join('.') // join the last two parts of the domain
@@ -37,8 +36,7 @@ export function currentBaseDomain() {
 }
 
 export function currentOrg() {
-  let domain = window.location.host
-  let parts = domain.split('.')
+  let parts = window.location.host.split('.')
   if (parts.length > 2) {
     return parts.shift() // this is the org
   }
--- a/ui/lib/login.js
+++ b/ui/lib/login.js
@@ -30,9 +30,8 @@ export function currentBaseDomain() {
   let parts = window.location.host.split('.')
   if (parts.length > 2) {
     parts.shift() // remove the org
-    domain = parts.join('.') // join the last two parts of the domain
   }
-  return domain
+  return parts.join('.') // return the domain without the org
 }
 
 export function currentOrg() {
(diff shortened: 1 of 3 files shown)


ui/lib/login.js
```diff
@@ -25,3 +25,35 @@ export function saveToVisitedOrgs(domain, orgName) {
...
+
+export function currentBaseDomain() {
+  let domain = window.location.host
+  let parts = domain.split('.')
+  if (parts.length > 2) {
+    parts.shift() // remove the org
+    domain = parts.join('.') // join the last two parts of the domain
+  }
+  return domain
+}
```
hoyyeva, line 37: we can maybe change the domain and parts to `const`
and instead of assigning domain to parts.join('.') we can just return parts.join('.')

ui/lib/login.js
```diff
@@ -25,3 +25,35 @@ export function saveToVisitedOrgs(domain, orgName) {
...
+
+export function currentOrg() {
+  let domain = window.location.host
+  let parts = domain.split('.')
```
hoyyeva, line 41: we probably don't ned `let domain = window.location.host` we can just do 
`const parts = window.localhost.host.split('.')`

ui/pages/login/redirect.js
```diff
@@ -0,0 +1,71 @@
...
+      const cookies = new Cookies()
+      // start with the base of the login callback URL we will redirect to
+      let callbackURL =
+        currentBaseDomain() +
+        '/login/callback' +
+        '?code=' +
+        code +
+        '&state=' +
+        state
+      let org = cookies.get('finishLogin') // the org to redirect to is stored in this cookie
```
hoyyeva, line 24: is org a constant value?

ui/pages/login/redirect.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }, [code, state, router])
+
+  if (!isReady) {
+    return null
+  }
+
+  return (
+    <div className='my-32 flex h-full w-full items-center justify-center'>
+      <svg
```
hoyyeva, line 50: is this a loader image? We have a loader component call `<Loader />` can we use that?

BruceMacD: Nice, using this component now

ui/pages/login/redirect.js
BruceMacD: it is, updated

ui/lib/login.js
BruceMacD: that is nicer 👍

BruceMacD:
Superseded by #3657

PR #3602, opened 2022-11-09
fix: replace `CREATE OR REPLACE TRIGGER` syntax with a prior `DROP TRIGGER` clause

jmorganca:


pdevine (approved):

.github/workflows/ci-core.yaml
```diff
@@ -38,9 +38,13 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version: [12, 14]
```
dnephin, line 43: I think we probably want to also test against 15, to catch issues before we upgrade, but not critical we do that now.

dnephin (approved):
Looks good

PR #3604, opened 2022-11-09
Fix update destination

### LINKED ISSUE #3594: Initial connected status takes 6 minutes
It takes 6 minutes for a connected cluster to show "connected"

dnephin:
## Summary

Start with the existing data so that we don't lose values for fields that
are not updated. We have 3 non-update fields now: CreatedAt,
LastSeenAt, and Kind.

Also remove the extra logic in the data layer, we no longer need this
now that we're handling it properly in the API.  This allows the
update operation for destination to look like all the other update
operations.

Also adds test cases for the `UpdateDestination` API endpoint.

## Related Issues

Resolves #3594

### DIFF
--- a/internal/access/destination.go
+++ b/internal/access/destination.go
@@ -18,7 +18,7 @@ func CreateDestination(c *gin.Context, destination *models.Destination) error {
 	return data.CreateDestination(db, destination)
 }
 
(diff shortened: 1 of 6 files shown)


BruceMacD (approved):
Looks good, hopefully this fixes the slow connector issue

PR #3608, opened 2022-11-10
fix: go test ci

mxyng:
## Summary

GitHub's required status check need exact job name matches. In case of `strategy.matrix`, the job name gets modified by the matrix such that it no longer matches the configured required status check. Instead, rename the job to `test-matrix` and create a new job matching the required status check name dependent on `text-matrix` results.


dnephin (approved):
LGTM! seems like this should fix it

PR #3615, opened 2022-11-10
Fix default orgs created by migration

dnephin:
## Summary

Any org created by the `addDefaultOrganization` never had a `created_by` or `domain` column set (they were both null in the database). Apparently `gorm` was ignoring columns with null values. Now that we're not using gorm any org created by that migration would fail to load.

This PR adds a new migration that sets empty values for those fields. The migration test failed before the migration, and passes afterward.


internal/server/data/migrations_test.go
```diff
@@ -854,6 +854,13 @@ DELETE FROM settings WHERE id=24567;
 				// schema changes are tested with schema comparison
 			},
 		},
+		{
+			label: testCaseLine("2022-11-10T17:30"),
+			expected: func(t *testing.T, tx WriteTxn) {
+				_, err := GetOrganization(tx, GetOrganizationOptions{ByID: defaultOrganizationID})
+				assert.NilError(t, err)
```
dnephin, line 861: We don't have to create the org in `setup`, because it's already created by an earlier migration.

internal/server/data/migrations.go
```diff
@@ -895,3 +896,16 @@ func addAllowedDomainsToProvidersTable() *migrator.Migration {
...
 	}
 }
+
+func fixDefaultOrgCreatedByMigration() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-11-10T17:30",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `
+				UPDATE organizations SET created_by=1 WHERE created_by is null;
+				UPDATE organizations SET domain='' WHERE domain is null`
```
dnephin, line 906: This `SET domain=''` will fail if there are multiple organizations with a null domain. That shouldn't ever be the case because the only way to get a null in that field was from the migration that created a single org.

jmorganca (approved):

PR #3616, opened 2022-11-11
maintain: fix the apimigrator for nested structs

pdevine:
## Summary

This change fixes an issue with the api migrations which would hide any nested structs which are part of the request struct. This includes all of the pagination structs. If you add a migration, any version which is _before_ the version you specify will be impacted, which means pagination will no longer work unless the client making the request specifies a _newer_ API version.

## Checklist

### DIFF
--- a/internal/server/access_keys_test.go
+++ b/internal/server/access_keys_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	gocmp "github.com/google/go-cmp/cmp"
 	"gotest.tools/v3/assert"
 
 	"github.com/infrahq/infra/api"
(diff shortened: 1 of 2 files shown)


internal/server/apimigrator.go
```diff
@@ -105,50 +105,59 @@ func rewriteRequired(c *gin.Context, migrationVersion *semver.Version) bool {
...
+
+	var rebuildRequestQuery func(t reflect.Value)
+
+	rebuildRequestQuery = func(v reflect.Value) {
+		t := v.Type()
+		for i := 0; i < v.NumField(); i++ {
+			f := v.Field(i)
+			if v.Field(i).Kind() == reflect.Struct {
+				// follow the rabbit hole for nested structs
+				rebuildRequestQuery(f)
```
pdevine, line 117: This recursively unpacks any nested structs (like `api.PaginationRequest`) so that they will also be included in the `query` and `body` variables.

jmorganca (approved):

internal/server/access_keys_test.go
```diff
@@ -205,6 +205,28 @@ func TestAPI_ListAccessKeys(t *testing.T) {
...
+
+		routes.ServeHTTP(resp, req)
+		assert.Equal(t, resp.Code, http.StatusOK)
+
+		keys := api.ListResponse[api.AccessKey]{}
+		err = json.Unmarshal(resp.Body.Bytes(), &keys)
+		assert.NilError(t, err)
+
+		// TODO: replace this with a more strict assertion using DeepEqual
+		assert.Equal(t, keys.PaginationResponse, api.PaginationResponse{Limit: 1, Page: 1, TotalCount: 2, TotalPages: 2})
```
BruceMacD, line 222: You can this with cmp.Options as seen in this test file:
https://github.com/infrahq/infra/blob/main/internal/server/data/access_key_test.go

Daniel may have a better suggestion though.

internal/server/apimigrator.go
BruceMacD, line 117: Thoughts on tracking the depth of this recursion and failing if it goes too deep?

internal/server/access_keys_test.go
pdevine: I did expand on this, but didn't use cmp.Options{}.

internal/server/apimigrator.go
pdevine: I went ahead and added something that will limit this to a depth of 5. If it exceeds it, it will log an error and then pop back out of the stack.

BruceMacD (approved):

PR #3617, opened 2022-11-11
fix(ui): trim leading and trailing whitespace in group names

mxyng:
## Summary

leading whitespaces are particularly problematic because it makes it possible to create group names that look empty, e.g. " ". a side effect is group names with " " (one whitespace) and "  " (multple consecutive whitespaces) will cause a 409 Conflict on creation since the server trims inputs.

non-leading or non-trailing whitespaces are left as is so it's possible to have repeats in t between other characters

### DIFF
--- a/ui/pages/groups/index.js
+++ b/ui/pages/groups/index.js
@@ -22,7 +22,7 @@ function AddGroupsDialog({ setOpen }) {
     try {
       const res = await fetch('/api/groups', {
         method: 'POST',
-        body: JSON.stringify({ name }),
+        body: JSON.stringify({ name: name.trim() }),
       })
 
       const group = await jsonBody(res)
@@ -59,7 +59,9 @@ function AddGroupsDialog({ setOpen }) {
                   }
                 }}
                 value={name}
-                onChange={e => setName(e.target.value)}
+                // trim leading whitespace on input. trailing whitespace will be
+                // trimmed on submit
+                onChange={e => setName(e.target.value.trimLeft())}
                 className={`mt-1 block w-full rounded-md shadow-sm focus:border-blue-500 focus:ring-blue-500 sm:text-sm ${
                   error ? 'border-red-500' : 'border-gray-300'
                 }`}


ui/pages/groups/index.js
```diff
@@ -59,7 +60,7 @@ function AddGroupsDialog({ setOpen }) {
                   }
                 }}
                 value={name}
-                onChange={e => setName(e.target.value)}
+                onChange={e => setName(e.target.value.replace(/^\s/g, ''))}
```
mxyng, line 63: On trim leading spaces since we want users to enter spaces into the input field. Trailing spaces will be trimmed on submit

hoyyeva (approved):

ui/pages/groups/index.js
BruceMacD: We should put this comment in the code too. Regex isn't intuitive imo

ui/pages/groups/index.js
```diff
@@ -22,7 +22,7 @@ function AddGroupsDialog({ setOpen }) {
     try {
       const res = await fetch('/api/groups', {
         method: 'POST',
-        body: JSON.stringify({ name }),
+        body: JSON.stringify({ name: name.trim() }),
```
BruceMacD, line 25: nit:

Rather than trimming the name here and in the error you could do one `setName(name.trim())` before the try/catch block.

ui/pages/groups/index.js
```diff
@@ -32,6 +32,7 @@ function AddGroupsDialog({ setOpen }) {
       router.replace(`/groups/${group.id}`)
     } catch (e) {
       setError(e.message)
+      setName(name.trim())
```
jmorganca, line 35: If we hit an error we should leave the input fields unchanged – otherwise it's a strange experience!

ui/pages/groups/index.js
```diff
@@ -59,7 +60,9 @@ function AddGroupsDialog({ setOpen }) {
                   }
                 }}
                 value={name}
-                onChange={e => setName(e.target.value)}
+                // trim leading whitespace on input. trailing whitespace will be
+                // trimmed on submit
+                onChange={e => setName(e.target.value.replace(/^\s/g, ''))}
```
jmorganca, line 65:
```suggestion
                onChange={e => setName(e.target.value.trimLeft())}
```

ui/pages/groups/index.js
mxyng: I would normally agree but if left unchanged, the input value will be different than what's being sent in the request.

PR #3620, opened 2022-11-11
Add SSHLoginName to User

dnephin:
## Summary

Branched from #3595

This PR adds `SSHLoginName` to the user table, so that we have an identifier that is a valid linux username that we can use as the target user for SSH.

Also, when should we create this username for the user. In this PR we do it in 3 places:
1. for users that doesn't exist yet, we'll create the username when the user is created
2. for existing users we create it when they login
3. for existing users that may already be logged in, we create it when they upload their first public key

I'm not sure how well this strategy will work, because the username has to exist before they `infra login` and before they are granted access to an ssh destination.

Another option might be to do it as part of the migration, but it's kind of slow because we have to iterate over all users. 

TODO:
* [x] replace non-alphanumeric and non-ascii characters in the username with allowed characters
* [x] add a list of reserved names and don't allow anyone to get a username of one of these reserved names
* [x] test coverage for everything


api/user.go
```diff
@@ -23,6 +23,7 @@ type User struct {
 	Name          string          `json:"name"`
 	ProviderNames []string        `json:"providerNames,omitempty"`
 	PublicKeys    []UserPublicKey `json:"publicKeys,omitempty"`
+	SSHUsername   string          `json:"sshUsername"`
```
mxyng, line 26: There will be other cases where a username will be needed so it doesn't make sense to create a column specific to ssh. It'd be better to create a column called `username` or `login`

dnephin: I agree, I don't really like the name `sshUsername`, and I would like to find something better.

If we call it `username` then I think we'd want it to be the users primary username throughout the system. The only problem I see with that is then we're forced to restrict these to the same rules of Linux usernames, which may be very restrictive.

Let's keep thinking about the right term to use here. Some other ideas:
* `linuxUsername`, `unixUsername`, or `posixUsername` - more specific, but maybe we'll want to use this for windows SSH or remote connections as well?
* (i'm out of ideas, but I think we can find something better)

mxyng: I don't think we need to enforce Linux username rules in this field. We can set some sensible rules and transform them to Linux username if necessary. Same with Windows.

[`useradd(8)`](https://man7.org/linux/man-pages/man8/useradd.8.html) has reasonable rules:

```
       Usernames must start with a lower case letter or an underscore,
       followed by lower case letters, digits, underscores, or dashes.
       They can end with a dollar sign. In regular expression terms:
       [a-z_][a-z0-9_-]*[$]?

       Usernames may only be up to 32 characters long.```

Comment (dnephin) [reply api/user.go:26]: I don't see how that will work. The entire purpose of this field was to establish safe and unique usernames ahead of time. If it doesn't do that, I don't see why we would add this field. We already have the first part of their email address as a proxy for this value.

We need these usernames to exist ahead of time so that the username in the SSH request already exists on the target system.

If we don't create these usernames ahead of time, how do we provision a Linux account for users before they SSH, and share that username with the SSH client so that it knows which username to use?

File: internal/server/data/migrations.go [hunk 2]
```diff
@@ -935,8 +935,9 @@ ALTER TABLE ONLY user_public_keys
 
 CREATE UNIQUE INDEX IF NOT EXISTS idx_user_public_keys_user_fingerprint ON user_public_keys
     USING btree (fingerprint) WHERE (deleted_at IS NULL);
+
+ALTER TABLE identities ADD COLUMN IF NOT EXISTS ssh_username text
```
dnephin, line 939: Reminder to self, this is not safe if I merge #3595 first which added this migration. I'll need to create a new one.

internal/server/data/identity.go
```diff
@@ -22,15 +24,16 @@ func (i identitiesTable) Table() string {
...
 
 func (i identitiesTable) Values() []any {
-	return []any{i.CreatedAt, i.CreatedBy, i.DeletedAt, i.ID, i.LastSeenAt, i.Name, i.OrganizationID, i.UpdatedAt, i.VerificationToken, i.Verified}
+	sshUsername := sql.NullString{String: i.SSHUsername, Valid: i.SSHUsername != ""}
+	return []any{i.CreatedAt, i.CreatedBy, i.DeletedAt, i.ID, i.LastSeenAt, i.Name, i.OrganizationID, sshUsername, i.UpdatedAt, i.VerificationToken, i.Verified}
 }
 
 func (i *identitiesTable) ScanFields() []any {
-	return []any{&i.CreatedAt, &i.CreatedBy, &i.DeletedAt, &i.ID, &i.LastSeenAt, &i.Name, &i.OrganizationID, &i.UpdatedAt, &i.VerificationToken, &i.Verified}
+	return []any{&i.CreatedAt, &i.CreatedBy, &i.DeletedAt, &i.ID, &i.LastSeenAt, &i.Name, &i.OrganizationID, (*optionalString)(&i.SSHUsername), &i.UpdatedAt, &i.VerificationToken, &i.Verified}
```
dnephin, line 36: There's a new pattern here for handling of possibly null values. 

In `Values` we use `sql.NullString` which allows us to save a null to the database if `SSHUsername` is the empty string.

In` ScanFields` we use a wrapper around `sql.NullString` to allow reading a null value as the empty string.

This lefts us avoid ` *string` field while still supporting nulls in the database.

Nulls in the database here are important because there's a unique constraint on this field. Multiple null values are allowed, but only a single row would allow the empty string. 

This is being done so that we don't have to generate every username immediately on deploy. We can generate a username as users login or add public keys.

We could avoid this by generating all the usernames in a migration on deploy, but I'm not sure if that's a good option.

internal/server/data/identity.go
```diff
@@ -135,7 +138,135 @@ func CreateIdentity(tx WriteTxn, identity *models.Identity) error {
...
+	"uuidd":                  {},
+	"vpopmail":               {},
+	"www-data":               {},
+	"xfs":                    {},
+
+	"git":    {},
+	"apache": {},
+	"nginx":  {},
+	"docker": {},
+	// TODO: what other users are created by commonly installed packages?
```
dnephin, line 264: I suspect there are others I should include here

mxyng (approved):

internal/server/data/identity_test.go
```diff
@@ -102,7 +219,7 @@ func TestGetIdentity(t *testing.T) {
 		assert.NilError(t, err)
 
 		var (
-			bond   = models.Identity{Name: "jbond@infrahq.com"}
+			bond   = models.Identity{Name: "jbond@infrahq.com", SSHLoginName: "jbond"}
```
jmorganca, line 222: 🤵‍♂️

jmorganca (approved):
Good stuff!

internal/server/data/migrations.go
```diff
@@ -951,28 +952,16 @@ func updateAccessKeysTimeoutColumn() *migrator.Migration {
...
-				SELECT EXISTS (
-					SELECT 1 
-					FROM information_schema.columns 
-					WHERE table_name='access_keys' AND column_name='extension_deadline'
-				);
-			`
-			var exists bool
-			if err := tx.QueryRow(stmt).Scan(&exists); err != nil {
-				return fmt.Errorf("%w: check extension_deadline column", err)
+			if !migrator.HasColumn(tx, "access_keys", "extension_deadline") {
```
BruceMacD, line 955: thanks for cleaning this up

internal/server/data/migrations.go
```diff
@@ -1003,7 +992,22 @@ ALTER TABLE ONLY user_public_keys
...
     USING btree (fingerprint) WHERE (deleted_at IS NULL);
 `
-			// TODO: is this the right index for joining to the user table?
+			_, err := db.Exec(stmt)
+			return err
+		},
+	}
+}
+
+func addUserSSHLoginName() *migrator.Migration {
```
BruceMacD, line 1001: As you mentioned in the description of this PR, I would personally rather see existing users get assigned SSH usernames in this migration even though it is heavy. It means we don't have to add checks for empty SSH usernames at other places in ways that we have to clean up in the future.

internal/server/data/identity.go
BruceMacD: We could avoid this new type by assigning the SSH username in the migration, correct?

dnephin: Ya, I think we can avoid this if we migrate all users immediately

internal/server/data/migrations.go
dnephin: I think our current approach to migrations won't really scale. At some point we're going to need to migrate data in batches separate from our deployment. I'm not sure exactly when that will be.

It's probably ok to do this migration all at deployment time. I'll check the number of users we have right now in the DB to try and get some idea of how safe this is.

dnephin: Ok, I added a commit that does the generation of `SSHLoginName` in the migration. There's a couple unfortunate things, but overall it is simpler and safer than having to do it later. Eventually we will need to start migrating data separately from deploys.

dnephin:
I fixed a problem with the new index for `ssh_login_name`. I was missing a `WHERE (deleted_at is null);`

BruceMacD (approved):
lgtm

PR #3621, opened 2022-11-11
Do not require uniqueID for a destination

dnephin:
## Summary

Related to #3595, #3620

This PR makes `destination.UniqueID` an optional field. We need to keep it for existing connectors, but after some time we can look to remove it.  

This PR also allows connectors to send either the uniqueID or name for updating the LastSeenAt status of the destination. A new `Infra-Destination-Name` header is added for accepting the name instead of uniqueID.

### DIFF
--- a/api/destination.go
+++ b/api/destination.go
@@ -61,7 +61,6 @@ type CreateDestinationRequest struct {
 
 func (r CreateDestinationRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
-		validate.Required("uniqueID", r.UniqueID),
 		validateDestinationName(r.Name),
 		validate.Required("name", r.Name),
 		// Allow "" for versions 0.16.1 and prior
@@ -83,7 +82,6 @@ type UpdateDestinationRequest struct {
 
 func (r UpdateDestinationRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
-		validate.Required("uniqueID", r.UniqueID),
 		validate.Required("id", r.ID),
 		validate.Required("name", r.Name),
 		validateDestinationName(r.Name),
--- a/docs/api/openapi3.json
+++ b/docs/api/openapi3.json
@@ -1740,7 +1740,6 @@
                   }
                 },
                 "required": [
-                  "uniqueID",
                   "name"
                 ],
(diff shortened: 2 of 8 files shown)


internal/connector/connector.go
```diff
@@ -124,12 +124,7 @@ func Run(ctx context.Context, options Options) error {
 	logging.L.Debug().Str("uniqueID", checkSum).Msg("Cluster uniqueID")
 
 	if options.Name == "" {
-		autoname, err := k8s.Name(checkSum)
-		if err != nil {
-			logging.Errorf("k8s name error: %s", err)
-			return err
-		}
-		options.Name = autoname
+		return fmt.Errorf("destination name is required")
```
BruceMacD, line 127: Should we give a name to this connector automatically rather than making this as a breaking change? We may be able to do it with simpler logic than the `k8s.Name()` function.

internal/connector/connector.go
```diff
@@ -180,7 +175,7 @@ func Run(ctx context.Context, options Options) error {
 			Transport: httpTransportFromOptions(options.Server),
 		},
 		Headers: http.Header{
-			"Infra-Destination": {checkSum},
+			"Infra-Destination-Name": {options.Name},
```
BruceMacD, line 178: Could this cause existing connectors to be registered as a new instance if they didn't have a name before?

Ex:
1. Stand-up existing connector without a name, it gets a unique ID.
2. Upgrade the connector to specify a name (the same name as the unique ID).
3. Could this result in a duplicate destination?

jmorganca (approved):
LGTM

dnephin:
I'm going to revert a bunch of this change and only keep the part that makes uniqueID optional. The current kube connector can continue to use uniqueID and auto-detect the name. We can fix that problem later. All I need for now is to not require uniqueID.

dnephin (commented):
I've reduced this PR to now include only 2 changes:

1. make uniqueID not required when creating a destination
2. accept `destination.Name` in the header sent by connectors to update the `lastSeenAt` time of the destination.

I removed the changes to the kubernetes connector, we can look at those some other time. 

I'll leave this open for another day if anyone else wants to review the trimmed down diff.

mxyng (approved):

PR #3622, opened 2022-11-11
Add SSH connector and `infra ssh auth-keys`

dnephin:
## Summary

Branched from #3665

This PR starts with some light refactoring of the connector to allow it to be used  for multiple destination kinds.

This PR adds the ssh connector and `infra ssh auth-keys` for authenticating and authorizing the ssh requests.

Also changes the existing kubernetes connector to use `name` instead of `uniqueID` for both the `Infra-Destination-Name` header used to update the `LastSeenAt` time of the destination, and for `createOrUpdateDestination` (to query for the ID of the destination).

See infrahq/internal#27 for more details


api/user.go
```diff
@@ -16,19 +16,22 @@ func (r GetUserRequest) ValidationRules() []validate.ValidationRule {
...
-	Name          string   `json:"name"`
-	ProviderNames []string `json:"providerNames,omitempty"`
+	ID            uid.ID          `json:"id"`
+	Created       Time            `json:"created"`
+	Updated       Time            `json:"updated"`
+	LastSeenAt    Time            `json:"lastSeenAt"`
+	Name          string          `json:"name"`
+	ProviderNames []string        `json:"providerNames,omitempty"`
+	PublicKeys    []UserPublicKey `json:"publicKeys,omitempty"`
+	SSHUsername   string          `json:"sshUsername"`
```
jmorganca, line 26: Is there a way we can keep this less specific to SSH? Other destinations will have similar limitations in usernames (e.g. example Kubernetes service accounts are restricted to the Kubernetes resource naming schema).

jmorganca (commented):
This is looking pretty fantastic. Nice work!

api/user.go
dnephin: Ya, there is discussion about this over in https://github.com/infrahq/infra/pull/3620#pullrequestreview-1179545073

I'd like to find a better name for this field

jmorganca (approved):

internal/cmd/sshauthkeys.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	cmd.Flags().StringVar(&opts.configFilename, "config-file",
+		"/etc/infra/connector.yaml", "Path to connector config file")
+	return cmd
+}
+
+func setupLogger(cli *CLI) zerolog.Logger {
+	out := []io.Writer{cli.Stderr}
+	// TODO: log to stderr if this fails?
```
BruceMacD, line 63: writing to a file could be an option too, if all else fails

internal/cmd/sshauthkeys.go
```diff
@@ -0,0 +1,175 @@
...
+	return nil
+}
+
+func verifyUsernameAndFingerprint(
+	ctx context.Context,
+	logger zerolog.Logger,
+	client *api.Client,
+	opts sshAuthKeysOptions,
+) (*api.User, error) {
+	users, err := client.ListUsers(ctx, api.ListUsersRequest{
```
BruceMacD, line 120: Does this mean the connector has to call the server every time it verifies a username/fingerprint? Does this happen frequently or just when a session is opened?

internal/connector/connector.go
```diff
@@ -57,6 +73,25 @@ type ServerOptions struct {
 	TrustedCertificate types.StringOrFile
 }
 
+func (o Options) APIClient() *api.Client {
+	url := o.Server.URL
+	if !strings.HasPrefix(url, "http") {
```
BruceMacD, line 78: Couldn't this result in
`https://example.com` -> `https://https://example.com`

BruceMacD (approved):

internal/cmd/sshauthkeys.go
dnephin: A fallback logfile seems like a good idea

internal/cmd/sshauthkeys.go
dnephin: Not the connector, but `infra sshd auth-keys` does yes. It will make two API calls per connect , one for the user, and one for grants. It only happens when a new SSH connection is opened.

internal/connector/connector.go
dnephin: I don't think so, `https://...` also has a prefix of `http`, so that will skip this branch

dnephin: I've fixed this in the latest commit. I changed `options.Server.URL` to `cmd/types.URL` so that it's parsed earlier, and that lets us set the schema to `https` here without having to parse it.

BruceMacD (approved):

dnephin:
Note: 
This PR changes the existing kubernetes connector to use `name` instead of `uniqueID` for both the `Infra-Destination-Name` header used to update the `LastSeenAt` time of the destination, and for `createOrUpdateDestination` (to query for the ID of the destination).

Issue #3592, opened 2022-11-07
Checkbox does not have a check mark


jmorganca:
<img width="163" alt="Screenshot 2022-11-07 at 4 29 54 PM" src="https://user-images.githubusercontent.com/251292/200419528-83673288-e6cd-4dfd-87d2-757368208063.png">

PR #3624, opened 2022-11-11
fix: checkbox check does not show when it is checked

hoyyeva:
## Summary
the checkbox's check does not show when it is checked on production - the reason why is because we should use event.target.checked instead of `value`.

## Related Issues

#3592


BruceMacD (approved):
thanks for fixing this

jmorganca (approved):
Great! Thanks for fixing

PR #3643, opened 2022-11-15
Remove the last of Gorm

### LINKED ISSUE #3585: Last steps for removing Gorm

dnephin:
## Summary

Resolves #3585. Best viewed by individual commit.

This PR starts by replacing `gorm.Open` with `sql.Open` so that we're using the standard library directly. That change required a few small changes to our `data.DB` , and `data.Transaction` to call the stdlib interface instead of the gorm one.

Once that was done it revealed a few tests that were still using gorm, so the second commit replaces gorm with stdlib sql in those tests.  It also fixes a few places where we constructed a `data.Transaction` without calling `db.Begin`. That should not be allowed, so I added a TODO to find a way to avoid that.

The next commit adds a function to replace `?` in queries with `$n` (which is what postgres supports). That was pretty easy and is essentially a much simpler version of what gorm was doing. In the future we may want to convert our queries to the postgres placeholders.

The next commit fixes a test failure related to context cancellation. Details in that commit message.

Another commit handles `IN (?)` queries. Gorm was translating `IN (?)` into a format that the driver accepts (probably something like `(?, ?, ?, ...)`. I found https://github.com/jackc/pgx/issues/334 which is an issue for the same problem. It seems that the driver doesn't have any automatic way to convert a slice into placeholders, so this commit adds a `queryInClause` function that produces the `(?, ?, ?, ...)` string for us.

The last couple of commits replace the query logging we had with gorm, and remove gorm from `go.mod`.


internal/logging/database.go
```diff
@@ -1,56 +0,0 @@
...
-
-	L.WithLevel(level).
-		CallerSkipFrame(3).
-		Int64("rows", rows).
-		Str("query", sql).
-		Dur("elapsed", time.Since(begin)).
-		Err(err).
-		Msg("")
-}
-
```
ssoroka, line 51: looks like we lose the slow query logger.

internal/server/data/data.go
```diff
@@ -78,44 +77,69 @@ func NewDB(dbOpts NewDBOptions) (*DB, error) {
...
+	return row
+}
+
+func rewriteQueryPlaceholders(query string, num int) string {
+	var counter int
+	var buf strings.Builder
+	buf.Grow(len(query))
+
+	for _, r := range query {
+		if r != '?' {
```
ssoroka, line 128: not all ? are necessarily safe to assume as placeholders unless you're parsing the sql.

internal/server/data/grant.go
```diff
@@ -176,11 +176,11 @@ func ListGrants(tx ReadTxn, opts ListGrantsOptions) ([]models.Grant, error) {
 			for _, id := range groupIDs {
 				subjects = append(subjects, uid.NewGroupPolymorphicID(id).String())
 			}
-			query.B(`AND subject IN (?)`, subjects)
+			query.B(`AND subject = any(?)`, subjects)
```
ssoroka, line 179: maybe query.B should just look for `IN (?)` paired with a slice and auto-convert it to `IN (?, ?, ?)` etc. Feels very weird to force the dev to use different sql. At the very least we can auto-convert it to the `any` syntax if that is identical.

internal/logging/database.go
dnephin: It's still there `case elapsed > slowQueryThreshold` in `data.go`. I'll drop a comment on that line.

internal/server/data/data.go
dnephin: Which cases are not safe to replace?  If someone is putting user input in this part of the query then it's a potential SQL injection, so that should never be the case. When would we write a `?` into a query that was not a placeholder?

internal/server/data/data.go
```diff
@@ -390,3 +422,35 @@ func InfraConnectorIdentity(db ReadTxn) *models.Identity {
...
+		level = zerolog.WarnLevel
+	}
+
+	query = normalizeQueryString(query)
+	logging.L.WithLevel(level).
+		CallerSkipFrame(2). // logQuery + tx.{Query,Exec}
+		Int64("rows", rows).
+		Str("query", query).
+		Dur("elapsed", elapsed).
+		Msg("DB query")
```
dnephin, line 445: Here's the slow query logging

internal/server/data/data.go
dnephin: I had a read over the gorm source code, and from what I can tell this is all it was doing (albeit with a ton more abstraction and complexity that makes it hard to see). It's possible I read it wrong.

dnephin: My understanding is that this is safe because:
1. Any user input values can not be in that part of the query
2. Postgres identifiers can not contain a ? (https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)
3. Even if we did run into a problem that we haven't accounted for, it would be from a query we wrote in the source code. So we'd notice it in CI when someone wrote a test for it. So worst case we could address the problem then. And maybe at that point we'll already have replaced all of our queries with `$n` literals, which will allow us to delete this code, and it won't be a problem.

dnephin: We chatted about this a bit offline this week. It is possible that we write a query that uses something with `LIKE = '?..'`, however we can always rewrite that query to `LIKE = ?` with a value of `"?..."`, and since that query needs to be part of the source (not user input), we should catch that problem in tests pretty easily.

internal/server/data/grant.go
dnephin: Ok, I've implemented `queryInClause` which looks like the second example above, and removed all of the `= any(?)` and `<> all(?)`.

internal/server/backgroundjobs.go
```diff
@@ -31,38 +32,42 @@ func (s *Server) registerJob(ctx context.Context, job BackgroundJobFunc, every t
 	})
 }
 
-func jobWrapper(ctx context.Context, tx *data.DB, job BackgroundJobFunc, every time.Duration) func() error {
-	tx = &data.DB{DB: tx.WithContext(ctx), DefaultOrgSettings: tx.DefaultOrgSettings, DefaultOrg: tx.DefaultOrg}
```
dnephin, line 35: This construction of an artificial transaction no longer works for a couple reasons:
1. Our `data.DB` does not include a `context.Context`, because a `data.DB` is not request scoped. So storing a context on that struct would be very prone to errors. I think we should avoid that.
2. Our `data.Transaction` struct can no longer be created with any arbitrary `DB` object. It must be an `sql.Tx`, which means it must be an actual transaction.

I had to change the background jobs to create a new transaction on each run, and commit it or roll it back when they are done. Those changes are below here.

I also had to move the logging out of `jobWithRescue` into the `select` so that I could log any error that might be returned from `jobWithRescue`.

It looks like there is not any test coverage of this function, or the jobs, so I guess i'll follow this up with some tests.

dnephin:
Ok, I believe all the TODOs are resolved, so I've marked this ready for review.

internal/server/data/data.go
```diff
@@ -78,44 +77,71 @@ func NewDB(dbOpts NewDBOptions) (*DB, error) {
 // DB wraps the underlying database and provides access to the default org,
 // and settings.
 type DB struct {
-	*gorm.DB // embedded for now to minimize the diff
+	DB *sql.DB
```
jmorganca, line 80: 🎉

jmorganca (approved):

internal/server/backgroundjobs.go
dnephin: Added the tests in #3701 and everything seems to be ok.

PR #3645, opened 2022-11-15
fix: migrate Link with a child

mxyng:
## Summary

NextJS 13 renders <Link> as <a> so an <a> child will trigger a runtime exception. This updates the UI by removing the <a> child using migration suggested by https://nextjs.org/docs/messages/invalid-new-link-with-extra-anchor


mxyng:
Closing for #3644

PR #3648, opened 2022-11-15
Update cobra to official version

dnephin:
We were using a fork for group support. That feature is now in an official release so we can stop using our fork. This will allow dependabot to notify us of updates.

Once this is merged we can archive https://github.com/infrahq/cobra

I tested this with the following

```
$ go build .
$ ./infra --help > before
# update cobra
$ go build .
$ ./infra help > after
$ go diff -u before after
```

Added a test using `golden` to show that it still looks the same.


jmorganca (approved):
Fantastic thanks great catch!

dnephin:
I made that change to use constants, and added a test using `golden`.

PR #3650, opened 2022-11-15
maintain: remove unused get grant endpoint

### LINKED ISSUE #3156: Removed unused GET `/api/grants/:id` API endpoint
We never use the `GetGrant` (get grant by ID) API endpoint. We always list grants and delete by ID. This makes sense because there's no way of getting an ID without listing grants, and once you have a list you have all the data. There's nothing else to get.

We should be able to remove it, as it's not used by the UI or CLI.

BruceMacD:
## Checklist

## Related Issues

Resolves #3156


dnephin (approved):
LGTM, thanks for doing this cleanup!

PR #3651, opened 2022-11-15
maintain: api doc categories

### LINKED ISSUE #3597: API endpoints with bad tags
## Describe the bug

There are a bunch of endpoints with a misc tag when they should have more descriptive tags. All the Authentication endpoints have the Authentication tag. Destinations are correct. Grants are correct. But all the password reset, settings, signup, version, device flow, organizations, server config, etc are miscellaneous. This shows up in the openapi3.json file and thus in the generated docs.

## Expected behavior

We should set the correct tags so that the api docs are easier to read

## Screenshots
CleanShot 2022-11-08 at 07 34 24

## Environment

```
$ infra version
```

```
$ kubectl version
```

BruceMacD:
- move all api endpoints into specific categories
- hide the sign-up endpoint from the api docs

## Summary
A lot of new functionality was getting grouped under the `Misc` category in our API docs. This is my attempt at cleaning things up.

## Related Issues

Resolves #3597

### DIFF
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -76,6 +76,15 @@ type WellKnownJWKResponse struct {
 	Keys []jose.JSONWebKey `json:"keys"`
 }
 
+func (a *API) SignupRoute() route[api.SignupRequest, *api.SignupResponse] {
+	return route[api.SignupRequest, *api.SignupResponse]{
+		handler: a.Signup,
+		routeSettings: routeSettings{
+			omitFromDocs: true,
+		},
+	}
+}
+
 func (a *API) Signup(c *gin.Context, r *api.SignupRequest) (*api.SignupResponse, error) {
 	if !a.server.options.EnableSignup {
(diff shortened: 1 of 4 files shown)


internal/server/handlers.go
```diff
@@ -76,6 +76,15 @@ type WellKnownJWKResponse struct {
...
+func (a *API) SignupRoute() route[api.SignupRequest, *api.SignupResponse] {
+	return route[api.SignupRequest, *api.SignupResponse]{
+		handler: a.Signup,
+		routeSettings: routeSettings{
+			omitFromDocs: true,
+		},
+	}
+}
+
 func (a *API) Signup(c *gin.Context, r *api.SignupRequest) (*api.SignupResponse, error) {
```
pdevine, line 88: I think this can be `signup()` now since it doesn't need to be exported.

pdevine (approved):
lgtm to me.

Issue #3118, opened 2022-09-02
Simplify access key query logic related to expiry time


dnephin:
Related to https://github.com/infrahq/infra/pull/3094#discussion_r961690050

Currently to exclude expired access keys from `ListAccessKeys` we have to check 6 conditions:
* expiry > now
* expiry not null
* expiry not zero
* extension deadline > now
* extension deadline not null
* extension deadline not zero

However in `RemoveExpiredAccessKeys` all we do is check `expires_at <= ?`, which it turns out should be sufficient, because:

1. neither field can be null. If they had null values then our code would refuse to scan them into a `time.Time`
2. the extension deadline does not matter. It can actually be a time later than `expires_at` (I believe) , because we keep adding `extension` time to it, without checking if that time is greater than the expiry (which is ok because expiry always take precedence). We even have code that relies on this now (https://github.com/infrahq/infra/pull/3654#discussion_r1024218163).
3. `expires_at` must always be set, we enforce a non-zero value when we create an access key

So give that, we should be able to simplify the query logic to simply `expiry > now AND extension_deadline > now`. Although we do need to make sure that we're always setting `extension_deadline == expiry` instead of leaving it 0.

mxyng:
We need to clarify the semantics of extension deadline. As implemented today, the extension deadline is a time window within which the access key must be used otherwise it gets invalidated. This doesn't align with my idea of what an extension deadline means. In my mind, the extension deadline, or maybe some other name, is an extension of key lifetime whenever it's used.

Example, a key with an expiry of 1d and an extension of 1h will increase its expiry by 1h every time the key is used. The key will never expire as long as it's used within a 1h period. If the key isn't used within the 1h window, it expires and the client must get a new key.

In this interpretation, the extension deadline could be null/zero but it should not be part of the `is_expired` check since the extension deadline updates the `expires_at` value.

dnephin:
That improvement is tracked in #2089. I agree we should look at improving the terminology here, and it's good to call out that these two issues could impact each other.

PR #3654, opened 2022-11-16
Change Access Keys API + CLI

pdevine:
## Summary

This change makes a number of quality of life changes to both the access keys API and CLI.

* Change access keys to make their names unique to a specific user, instead of for an entire org
* Make `infra keys list` default to the current user
* Add an `--all` flag to `infra keys list` for admins which can list all access keys in the org
* Make `infra keys add` not require a user name
* Add a `--user=` argument to `infra keys add` to be consistent with `infra keys list`
* Add a `--connector` flag to `infra keys add` to create the key for a connector
* Make `infra keys remove` to be specific to the current user
* Add a `--user=` argument similar to `infra keys list` and `infra keys add`
* Changed the API to call `DELETE /api/access-keys/:id` instead of `DELETE /api/access-keys`
* Added `LastUsed` field to the API and updated `infra keys list` to show the last time a specific key was used

## Checklist

Note: This change should be backwards compatible, but there are a number of caveats. I didn't remove the `DELETE /api/access-keys` endpoint because older CLI clients will still need to use it, and it's not 100% clear how to accomplish this with the current API migrator. Also, if a client upgrades and hits an older version of the server, it will have issues displaying the `LAST USED` column because the API won't populate it.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -590,7 +590,8 @@ infra keys remove KEY [flags]
 #### Options
 
 ```console
-      --force   Exit successfully even if access key does not exist
+      --force         Exit successfully even if access key does not exist
+      --user string   The name of the user who owns the key
 ```
 
 **Additional options**
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -524,6 +524,7 @@ infra keys list [flags]
 #### Options
 
 ```console
+      --all            Show keys for all users
       --show-expired   Show expired access keys
       --user string    The name of a user to list access keys for
 ```
@@ -543,7 +544,7 @@ Create an access key
 Create an access key for a user or a connector.
 
 ```bash
-infra keys add USER|connector [flags]
+infra keys add [flags]
 ```
 
 #### Examples
@@ -551,19 +552,25 @@ infra keys add USER|connector [flags]
 ```bash
 
 # Create an access key named 'example-key' for a user that expires in 12 hours
-$ infra keys add user@example.com --ttl=12h --name example-key
+$ infra keys add --ttl=12h --name example-key
 
 # Create an access key to add a Kubernetes connection to Infra
-$ infra keys add connector
+$ infra keys add --connector
+
+# Set an environment variable with the newly created access key
+$ MY_ACCESS_KEY=$(infra keys add -q --name my-key)
 
 ```
 
 #### Options
 
 ```console
+      --connector                     Create the key for the connector
       --extension-deadline duration   A specified deadline that the access key must be used within to remain valid (default 720h0m0s)
       --name string                   The name of the access key
+  -q, --quiet                         Only display the access key
       --ttl duration                  The total time that the access key will be valid for (default 720h0m0s)
+      --user string                   The name of the user who will own the key
 ```
 
 **Additional options**
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -82,7 +82,7 @@ func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
 		if len(keys) > 0 {
 			key = &keys[0]
 		} else {
-			return fmt.Errorf("no key named '%s' found", name)
+			return fmt.Errorf("%w: no key named '%s' found", internal.ErrNotFound, name)
 		}
 	}
 
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -60,9 +60,30 @@ func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string, 
 func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
 	rCtx := GetRequestContext(c)
 
-	key, err := data.GetAccessKey(rCtx.DBTxn, data.GetAccessKeysOptions{ByID: id, ByName: name})
-	if err != nil {
-		return err
+	var key *models.AccessKey
+	var err error
+
+	if id != 0 {
+		key, err = data.GetAccessKey(rCtx.DBTxn, data.GetAccessKeysOptions{ByID: id})
+		if err != nil {
+			return err
+		}
+	} else {
+		// if the specific key isn't specified, look up the key by name for the current user
+		opts := data.ListAccessKeyOptions{
+			IncludeExpired: false,
+			ByIssuedForID:  rCtx.Authenticated.User.ID,
+			ByName:         name,
+		}
+		keys, err := data.ListAccessKeys(rCtx.DBTxn, opts)
+		if err != nil {
+			return err
+		}
+		if len(keys) > 0 {
+			key = &keys[0]
+		} else {
+			return fmt.Errorf("no key named '%s' found", name)
+		}
 	}
 
 	if key.IssuedFor == rCtx.Authenticated.User.ID {
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -168,7 +168,26 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
 				return err
 			}
 
-			err = client.DeleteAccessKeyByName(ctx, args[0])
+			config, err := currentHostConfig()
+			if err != nil {
+				return err
+			}
+
+			keyName := args[0]
+
+			userID := config.UserID
+			keys, err := listAll(ctx, client.ListAccessKeys, api.ListAccessKeysRequest{Name: keyName, UserID: userID})
+			if err != nil {
+				return err
+			}
+
+			if len(keys) == 0 {
+				return Error{
+					Message: "No access key named: " + keyName,
+				}
+			}
+
+			err = client.DeleteAccessKey(ctx, keys[0].ID)
 			if err != nil {
 				if api.ErrorStatusCode(err) == 403 {
 					logging.Debugf("%s", err.Error())
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -40,17 +40,18 @@ type keyCreateOptions struct {
 	UserName          string
 	TTL               time.Duration
 	ExtensionDeadline time.Duration
+	Connector         bool
+	Quiet             bool
 }
 
 func newKeysAddCmd(cli *CLI) *cobra.Command {
 	var options keyCreateOptions
-	var connector bool
-	var quiet bool
 
 	cmd := &cobra.Command{
 		Use:   "add",
 		Short: "Create an access key",
 		Long:  `Create an access key for a user or a connector.`,
+		Args:  NoArgs,
 		Example: `
 # Create an access key named 'example-key' for a user that expires in 12 hours
 $ infra keys add --ttl=12h --name example-key
@@ -83,7 +84,7 @@ $ MY_ACCESS_KEY=$(infra keys add -q --name my-key)
 			userID := config.UserID
 
 			// override the user setting if the user wants to create a connector access key
-			if connector {
+			if options.Connector {
 				options.UserName = "connector"
 			}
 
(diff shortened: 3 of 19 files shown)


docs/reference/cli.md
```diff
@@ -542,27 +543,33 @@ Create an access key
...
 
 ```bash
 
 # Create an access key named 'example-key' for a user that expires in 12 hours
-$ infra keys add user@example.com --ttl=12h --name example-key
+$ infra keys add --ttl=12h --name example-key
 
 # Create an access key to add a Kubernetes connection to Infra
-$ infra keys add connector
+$ infra keys add --connector
```
BruceMacD, line 557: We will also have to update `kubernetes.md` and `NOTES.txt` with this new format

BruceMacD (commented):
Looks good, will approve once the loose ends in the docs are resolved

internal/cmd/keys.go
```diff
@@ -111,24 +118,35 @@ $ infra keys add connector
...
-			expMsg.WriteString(format.ExactDuration(options.TTL))
-			if !resp.Expires.Equal(resp.ExtensionDeadline) {
-				expMsg.WriteString(", and must be used every ")
-				expMsg.WriteString(format.ExactDuration(options.ExtensionDeadline))
-				expMsg.WriteString(" to remain valid")
-			}
-			cli.Output("Issued access key %q for %q", resp.Name, userName)
-			cli.Output(expMsg.String())
-			cli.Output("")
+			if !quiet {
```
dnephin, line 121: I'd suggest a guard clause here to keep this more readable (and also reduce the diff):

```go
if quiet {
    cli.Output(resp.AccessKey)
    return nil
}

...
```

internal/cmd/keys.go
```diff
@@ -150,7 +168,26 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
...
-			err = client.DeleteAccessKeyByName(ctx, args[0])
+			config, err := currentHostConfig()
+			if err != nil {
+				return err
+			}
+
+			keyName := args[0]
+
+			userID := config.UserID
+			keys, err := listAll(ctx, client.ListAccessKeys, api.ListAccessKeysRequest{Name: keyName, UserID: userID})
```
dnephin, line 179: Is `infra keys remove` losing a way for admins to remove access keys of other users? Previously an admin could list all the keys and remove any keys in their org. It looks like the remove part is no longer possible. 

I suspect we don't want to lose that.

Now that key names are not unique across the org, I guess we need a `--user` flag here so the admin can select the user that owns the key.

internal/cmd/keys_test.go
```diff
@@ -92,19 +92,13 @@ func TestKeysAddCmd(t *testing.T) {
...
-		err := Run(ctx, "keys", "add", "--ttl=400h", "--extension-deadline=5h", "my-user")
+		err := Run(ctx, "keys", "add", "--ttl=400h", "--extension-deadline=5h", "--user=my-user")
 		assert.NilError(t, err)
 
 		req := <-ch
 		assert.Equal(t, req.Name, "") // filled by server
 		assert.Equal(t, withNewline(bufs.Stdout.String()), expectedKeysAddOutput)
 	})
-
-	t.Run("without required arguments", func(t *testing.T) {
```
dnephin, line 103: This is still a good test case, we can change it to "with unexpected arguments" to show that it should fail if you give it any positional arguments.  I believe that will fail right now since `Args: NoArgs` is missing.

I think it would be great to add tests cases for the other parameter combinations (ex: no `--user` flag), but maybe we should convert this test to use a real API sever first. Would be fine in a follow up.

internal/cmd/testdata/TestKeysListCmd/list_all
```diff
@@ -1,4 +1,4 @@
-  NAME        ISSUED FOR  CREATED       EXPIRES           EXTENSION DEADLINE  
-  front-door  admin       24 hours ago  never             never               
-  side-door   admin       24 hours ago  6 hours from now  26 hours from now   
-  storage     clerk       20 hours ago  6 hours from now  never               
+  NAME        ISSUED FOR  CREATED       LAST USED  EXPIRES           EXTENSION DEADLINE  
+  front-door  admin       24 hours ago  never      never             never
```
dnephin, line 2: Maybe as a follow up, once we've converted the test to use a real API server, but I guess we should add some data that has a last used value, so that we see it formatted in this expected output.

internal/server/models/access_key.go
```diff
@@ -46,6 +46,7 @@ func (ak *AccessKey) ToAPI() *api.AccessKey {
 		ID:                ak.ID,
 		Name:              ak.Name,
 		Created:           api.Time(ak.CreatedAt),
+		LastUsed:          api.Time(ak.UpdatedAt),
```
dnephin, line 49: On first read this seemed wrong, but I guess it works in some cases because of the extension deadline. If the key has an extension deadline then we'll update it on use, so `UpdatedAt` will be updated.

However I believe there are 2 cases where that may not work:
* if the key doesn't have an extension deadline then we don't update it, which means `LastUsed` will be wrong (it'll show the created time always)
* if we get to the period of time where the extension deadline has reached the expiration, so it can't be extended anymore. This could potentially be a long period of time if the extension deadline is large. I think today that may not be an issue because I'm not sure we're handling this correctly, but once we fix that bug it'll be a problem here.

I think we may need a new field for `LastUsed` to accurately show the used time.

internal/server/data/migrations_test.go
```diff
@@ -850,6 +850,15 @@ INSERT INTO providers(id, name) VALUES (12345, 'okta');
...
 			},
 		},
+		{
+			label: testCaseLine("2022-11-15T10:00"),
+			expected: func(t *testing.T, tx WriteTxn) {
+				_, err := tx.Exec(`INSERT INTO access_keys(id, issued_for, name) VALUES(10000, 12345, 'foo')`)
+				assert.NilError(t, err)
+				_, err = tx.Exec(`INSERT INTO access_keys(id, issued_for, name) VALUES(10001, 12346, 'foo')`)
+				assert.NilError(t, err)
+			},
```
dnephin, line 860: Nice test coverage!

I'd expect this test to fail because it's leaving data in the tables after the test ends. I'm not sure why that safeguard is not failing the test, I'll have to look into that.

For this PR, I think this needs a `cleanup:` function set here, to delete these two rows.

dnephin (commented):
Looks good! Some minor suggestions, but I think there are a few things here that we should fix before merge. I left some comments inline, but I noticed one more thing.

`access.DeleteAccessKeys` does a get query with only a name, which is no longer valid with these changes. The query could return multiple results. I think that needs to be changed to use `ListAccessKeys`. I think `data.GetAccessKey` needs to change to require both an `IssusedFor` and a `ByName` (or just an `ID`) because otherwise it's a broken query.

If we're no longer using the `DELETE /api/access-keys` (no id in the path) endpoint, we should create a github issue to look at removing it in the future (unless of course we find another need for it over the next month or so). 

Test coverage of this change is pretty light, and what we do have uses a fake handler. It would be great to get some test coverage of these CLI commands with the new behaviour using a real API server as a follow up to this PR.

dnephin:
I suspect we can fix the API migration if we use #2006, but I'd have to look at the query parameter some more to get a better idea of exactly what we need.

internal/cmd/keys.go
```diff
@@ -77,11 +80,15 @@ $ infra keys add connector
...
-			var userID uid.ID
-			if userName == config.Name { // user is requesting their own stuff
-				userID = config.UserID
-			} else {
-				user, err := getUserByNameOrID(client, userName)
+			userID := config.UserID
+
+			// override the user setting if the user wants to create a connector access key
+			if connector {
+				options.UserName = "connector"
```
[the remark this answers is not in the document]
pdevine: Right now we use the fixed word `connector` as an argument, so this effectively changes it so that `--connector` replaces it (i.e. `infra keys add --connector` vs `infra keys add connector`).  We don't expose the concept of machine users like we did before (although we may want to revisit that decision at some point), but effectively the connector is just a machine user, so both `--user=connector` and `--connector` will work.

internal/cmd/keys.go
pdevine: Good catch. The `--user` flag should be here too.

internal/server/models/access_key.go
pdevine: This is interesting, however I don't think we _currently_ would hit either of those cases.
1. Right now we _require_ the user to include the `extensionDeadline` if you're creating the key through the API, otherwise you'll get a 400. This could be abused by us if someone creates an access key outside of the API, so it feels a little bit brittle.
2. Once the expiry is hit, the key won't work anymore. The extensionDeadline is ignored, so `UpdatedAt` will never get updated, but I think this is the correct value (i.e. the last time the access key was successfully used). I can see an argument for changing it if we want to record attempted use instead of actual use.

For now, I'll leave some comments in the code.

internal/cmd/testdata/TestKeysListCmd/list_all
pdevine: I'll make an issue for this. I also want to potentially change the `ISSUED FOR` column to not appear unless you're using the `--all` flag, or the `--user=` argument.

pdevine:
Thanks for the review @dnephin ! I think I've addressed each of the comments.

Ideally we will deprecate `DELETE /api/access-keys` in favour of `DELETE /api/access-keys/:id`, however we currently are calling that endpoint through the CLI. For now I've changed the `DELETE /api/access-keys` endpoint to _only_ allow deleting an access key which belongs to the current user. If an admin wants to delete a key for a different user through the API, they will have to use the other endpoint. Through the CLI, they can delete by key name, but will need to specify a user name.

internal/cmd/keys.go
```diff
@@ -111,6 +119,11 @@ $ infra keys add connector
 				return err
 			}
 
+			if options.Quiet {
+				cli.Output(resp.AccessKey)
+				return nil
+			}
```
mxyng, line 125: I'm not convinced this needs a quiet option right now. The output is simple enough that any non essential outputs, like the pretext, should be printed to `stderr` while the essential outputs, in this case `resp.AccessKey` should be printed to `stdout`. The examples will remain the same since only stdout will be captured

internal/server/models/access_key.go
dnephin: There are ways  to create an access key without an extension deadline I think.

`Signup` and `Server.loadConfig` both create an access key without an extension deadline, it can't be extended.

For 2, I don't think I explained it well. The issue is not that the key is expired, it's that the key can't be extended anymore. For example:

1. Extension deadline set to Dec 1st 12:00, with an extension value of 30 days
2. After Nov 1st, we shouldn't be making the expiry later than Dec 1st 12:00, because then the expiry would be past the extension deadline.

Today that's what we do, which means we are forced to look at two different values to determine if the key is valid or not. See #3118

During that period of time (nov 1st to dec 1st) the key won't be extended anymore, so the lastSeenAt will be wrong. In this example it's only 30 days, but it could be much longer if the extension value is longer.

internal/server/data/migrations_test.go
dnephin: Ah, I figured out why this wasn't failing. The transaction we give to `testCase.expected` is always rolled back, so we don't technically need the `cleanup` function in this case (but it doesn't hurt to show the pattern).

Normally we would do inserts in `setup`, but this is a different case. I forgot that the transaction given to `expected` was never committed.

internal/access/access_key.go
```diff
@@ -60,9 +60,30 @@ func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string,
...
+			ByName:         name,
+		}
+		keys, err := data.ListAccessKeys(rCtx.DBTxn, opts)
+		if err != nil {
+			return err
+		}
+		if len(keys) > 0 {
+			key = &keys[0]
+		} else {
+			return fmt.Errorf("no key named '%s' found", name)
```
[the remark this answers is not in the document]
pdevine: This is a good catch. I also realized that there was no validation happening on the name of the key to delete, so I've included that as well.

PR #3589, opened 2022-11-07
feat: social login

BruceMacD:
## Summary
This change enables organizations to use our shared social login. Specifically this change allows the CLI to login using the shared social login. My next follow-up change will complete the social login feature by updating the UI to allow for social login also.

Configuring the server config file looks like this:
```
server:
  config:
    socialProviders:
      - name: google
        url: accounts.google.com
        clientID:  abc
        clientSecret: 123
        kind: google
```

## Checklist

- [x] Wrote appropriate unit tests
- [x] Considered security implications of the change
- [x] Updated associated docs where necessary
- [x] Updated associated configuration where necessary
- [x] Change is backwards compatible if it needs to be (user can upgrade without manual steps?)
- [x] Nothing sensitive logged
- [x] Considered data migrations for smooth upgrades

## Related Issues

Part of #3261

PR #3657, opened 2022-11-16
feat: google social login

BruceMacD:
## Summary
This change enables organizations to use our shared social login. This change combines #3589 and  #3598, but configures Google social login through server flags rather than the config file.

Configuring Google social login now looks like this:
```
# infra.yaml
server:
  config:
    googleClientID: 123.apps.googleusercontent.com
    googleClientSecret: GOCSPX-xyz
```

## Checklist

## Related Issues

Part of #3261

### DIFF
--- a/internal/cmd/server.go
+++ b/internal/cmd/server.go
@@ -80,6 +80,8 @@ func newServerCmd() *cobra.Command {
 	cmd.Flags().Duration("session-extension-deadline", 0, "A user must interact with Infra at least once within this amount of time for their session to remain valid")
 	cmd.Flags().Bool("enable-signup", false, "Enable one-time admin signup")
 	cmd.Flags().String("base-domain", "", "base-domain for the server, eg example.com")
+	cmd.Flags().String("google-client-id", "", "Client ID of the Google client used for social login")
+	cmd.Flags().String("google-client-secret", "", "Client secret of the Google client used for social login")
 
 	return cmd
 }
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -95,7 +95,7 @@ func (a *API) signup(c *gin.Context, r *api.SignupRequest) (*api.SignupResponse,
 		// this should be caught by request validation before we hit this
 		return nil, fmt.Errorf("%w: %s", internal.ErrBadRequest, err)
 	}
-	if allowedSocialLoginDomain == "gmail.com" {
+	if allowedSocialLoginDomain == "gmail.com" || allowedSocialLoginDomain == "googlemail.com" {
 		// the admin will have to manually specify this later, default to no allowed domains
 		allowedSocialLoginDomain = ""
 	}
--- a/internal/server/handlers.go
+++ b/internal/server/handlers.go
@@ -202,9 +202,19 @@ func (a *API) Login(c *gin.Context, r *api.LoginRequest) (*api.LoginResponse, er
 
 		loginMethod = authn.NewPasswordCredentialAuthentication(r.PasswordCredentials.Name, r.PasswordCredentials.Password)
 	case r.OIDC != nil:
-		provider, err := access.GetProvider(c, r.OIDC.ProviderID)
-		if err != nil {
-			return nil, fmt.Errorf("invalid identity provider: %w", err)
+		var provider *models.Provider
+		if r.OIDC.ProviderID == 0 {
+			if a.server.Google == nil {
+				return nil, fmt.Errorf("%w: google login is not configured, provider id must be specified for oidc login", internal.ErrBadRequest)
+			}
+			// default to Google social login
+			provider = a.server.Google
+		} else {
+			var err error
+			provider, err = access.GetProvider(c, r.OIDC.ProviderID)
+			if err != nil {
+				return nil, fmt.Errorf("invalid identity provider: %w", err)
+			}
 		}
 
 		providerClient, err := a.providerClient(c, provider, r.OIDC.RedirectURL)
@@ -217,9 +227,7 @@ func (a *API) Login(c *gin.Context, r *api.LoginRequest) (*api.LoginResponse, er
 			r.OIDC.RedirectURL,
 			r.OIDC.Code,
 			providerClient,
-			// TODO: specify allowed login domains from rCtx.Authenticated.Organization.AllowedDomains
-			//		 in social login PR #3657
-			[]string{},
+			rCtx.Authenticated.Organization.AllowedDomains,
 		)
 		if err != nil {
 			return nil, err
--- a/internal/server/providers.go
+++ b/internal/server/providers.go
@@ -30,6 +30,11 @@ func (a *API) ListProviders(c *gin.Context, r *api.ListProvidersRequest) (*api.L
 		return nil, err
 	}
 
+	// if social login is configured, also return that option
+	if a.server.Google != nil {
+		providers = append(providers, *a.server.Google)
+	}
+
 	result := api.NewListResponse(providers, PaginationToResponse(p), func(provider models.Provider) api.Provider {
 		return *provider.ToAPI()
 	})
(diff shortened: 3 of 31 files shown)


ui/pages/login/redirect.js
```diff
@@ -0,0 +1,52 @@
...
+      // start with the base of the login callback URL we will redirect to
+      let callbackURL =
+        currentBaseDomain() +
+        '/login/callback' +
+        '?code=' +
+        code +
+        '&state=' +
+        state
+      const org = cookies.get('finishLogin') // the org to redirect to is stored in this cookie
+      if (org !== undefined && org !== '') {
```
hoyyeva, line 26: why don't we just do `if(org)`?

hoyyeva (commented):
just a small comment, everything else the UI part looks good to me!

internal/server/handlers.go
```diff
@@ -85,18 +85,43 @@ func (a *API) SignupRoute() route[api.SignupRequest, *api.SignupResponse] {
...
+		URL:          "accounts.google.com",
+		ClientID:     a.server.options.GoogleClientID,
+		ClientSecret: models.EncryptedAtRest(a.server.options.GoogleClientSecret),
+		CreatedBy:    models.CreatedBySystem,
+		Kind:         models.ProviderKindGoogle,
+		Managed:      true,
+	}
+	if err := a.setProviderInfoFromServer(c, google); err != nil {
+		return nil, err
+	}
```
dnephin, line 104: This is to set `AuthURL` and `Scopes`, is that right?

Does this always set the same value every time it's called, or would each org end up with different values for those fields?

internal/server/providers.go
```diff
@@ -156,17 +160,45 @@ func (a *API) UpdateProvider(c *gin.Context, r *api.UpdateProviderRequest) (*api
...
 }
 
+// overrideManagedProviderFields checks if an existing provider is managed on update and overrides managed fields
+func overrideManagedProviderFields(c *gin.Context, p *models.Provider) error {
+	existing, err := access.GetProvider(c, p.ID)
+	if err != nil {
+		return err
+	}
+	if existing.Managed {
+		// can't update managed client ID, secret, URL, or API details
```
dnephin, line 187: Which fields can be updated? Is it only the name?

Would it be easier and safer to explicitly allow updates of specific fields (`Name`), rather than default to allow and having to remember to exclude the fields here every time we add a field?

internal/server/server.go
```diff
@@ -51,6 +51,9 @@ type Options struct {
 	// Redis contains configuration options to the cache server.
 	Redis redis.Options
 
+	GoogleClientID     string
+	GoogleClientSecret string
```
dnephin, line 55: minor: I've been trying to keep any new config settings off the top level config. If we are planning on adding more managed providers maybe we can put this under a `ManagedProviders` field, and have:

```go
type ManagedProviders struct {
	GoogleClientID     string
	GoogleClientSecret string
}
```

internal/server/data/organization.go
```diff
@@ -50,6 +50,14 @@ func CreateOrganization(tx WriteTxn, org *models.Organization) error {
 		return fmt.Errorf("failed to create infra provider: %w", err)
 	}
 
+	// social login is enabled by default when configured
+	if googleProvider != nil {
+		googleProvider.OrganizationMember = models.OrganizationMember{OrganizationID: org.ID}
```
dnephin, line 55: As a safeguard maybe we should also set `googleProvider.Managed = true` here, to prevent accidentally saving it as not-managed.

 I think that would also require the test for `CreateUpdated` be updated to check that `Managed = true` was set, which would be nice.

dnephin (commented):
Looks good!

If I understand this correctly we're creating a row in the `providers` table for every org, with roughly the same data in each row.

If we ever need to update our client id or client secret are we going to need to update every org separately? Could we potentially not store the clientID and ClientSecret we get from the config for these managed providers and always read those values from the `server.Options`? It seems like that might make it easier for us to change them later.

I guess the disadvantage to that approach is that reading providers in `GetProvider` and `ListProvider` will need to do a bit more work to add the ClientID from the config.

internal/access/organization.go
```diff
@@ -41,13 +41,13 @@ func GetOrganization(c *gin.Context, id uid.ID) (*models.Organization, error) {
 	return data.GetOrganization(rCtx.DBTxn, data.GetOrganizationOptions{ByID: id})
 }
 
-func CreateOrganization(c *gin.Context, org *models.Organization) error {
+func CreateOrganization(c *gin.Context, org *models.Organization, google *models.Provider) error {
```
mxyng, line 44: How do you plan on scaling this to other global IDPs, e.g. GitHub?

internal/server/server.go
BruceMacD: We dont have plans to add more for the time being, let's move it under a struct if we do

internal/server/handlers.go
BruceMacD: This moved to being statically set when we load the Google login from the server parameters

internal/server/providers.go
BruceMacD: This went away with one Google provider being shared for all orgs

internal/server/data/organization.go
BruceMacD: This managed field went away with one Google provider being shared for all orgs

internal/access/organization.go
BruceMacD: We dont have plans for those at the moment

internal/cmd/server.go
```diff
@@ -80,6 +80,8 @@ func newServerCmd() *cobra.Command {
 	cmd.Flags().Duration("session-extension-deadline", 0, "A user must interact with Infra at least once within this amount of time for their session to remain valid")
 	cmd.Flags().Bool("enable-signup", false, "Enable one-time admin signup")
 	cmd.Flags().String("base-domain", "", "base-domain for the server, eg example.com")
+	cmd.Flags().String("google-client-id", "", "Client ID of the Google client used for social login")
+	cmd.Flags().String("google-client-secret", "", "Client secret of the Google client used for social login")
```
mxyng, line 84: Probably don't need these as command line arguments.

ui/lib/login.js
```diff
@@ -11,17 +11,38 @@ export function saveToVisitedOrgs(domain, orgName) {
...
+
+export function currentBaseDomain() {
+  let parts = window.location.host.split('.')
+  if (parts.length > 2) {
+    parts.shift() // remove the org
+  }
+  return parts.join('.') // return the domain without the org
+}
+
+export function currentOrg() {
```
mxyng, line 29: Why not use `/api/organizations/self` which contains the subdomain?

ui/lib/login.js
```diff
@@ -11,17 +11,38 @@ export function saveToVisitedOrgs(domain, orgName) {
...
-
     cookies.set('orgs', visitedOrgs, {
       path: '/',
-      domain: `.${cookieDomain}`,
+      domain: `.${currentBaseDomain()}`,
     })
   }
 }
+
+export function currentBaseDomain() {
```
mxyng, line 21: Should use `/api/server-configuration` which has the base domain and fall back to `window.location` if it's empty.

ui/lib/login.js
BruceMacD: They're not logged in yet, and they can't use this endpoint without authentication

internal/cmd/server.go
BruceMacD: We're looking at trying to move away from config file and more towards server flags for server config stuff

jmorganca (approved):
Overall looks great! Excited to see this land.

ui/lib/login.js
BruceMacD: Made this change

PR #3658, opened 2022-11-17
Move `connector` and `use` commands into their own files

dnephin:
## Summary

Most of our CLI commands are in a file that matches the name of the command (ex: `infra login` is in `login.go`). This PR moves the `connector` command, and the `use` command into `connector.go` , and `use.go`. 

Also move the tests for those commands into `connector_test.go` and `use_test.go`. There was one other function I noticed that was only used in `server.go`, so I also moved it out of `cmd.go`. 

All of the changes in this PR should be exclusively moving functions into different files with copy/paste. Other than imports changing there should not be any other changes.

I'm doing this now ahead of expanding the scope of the connector to support the SSH connector.


internal/cmd/server.go
```diff
@@ -126,3 +127,17 @@ var runServer = func(ctx context.Context, srv *server.Server) error {
 
 // newServer is a shim for testing.
 var newServer = server.New
+
+func canonicalPath(path string) (string, error) {
```
dnephin, line 131: moved from `cmd.go` since it is only used in this file

BruceMacD (approved):

PR #3659, opened 2022-11-17
fix: create one kubeconfig context per destination

### LINKED ISSUE #3611: Only create one kubeconfig context per cluster
We can continue to set the namespace in the context if they only have access to one namespace, otherwise we can leave it blank and they'll have to specify the namespace to kubectl.

mxyng:
## Summary

create only one kubeconfig context per destination regardless if there may be multiple namespaces granted to the user.

update `use` to set the `Namespace` field of the target context.

Resolves #3611

### DIFF
--- a/internal/cmd/kubernetes.go
+++ b/internal/cmd/kubernetes.go
@@ -27,33 +27,33 @@ func clientConfig() clientcmd.ClientConfig {
 func kubernetesSetContext(cluster, namespace string) error {
 	config := clientConfig()
 
-	kubeconfig, err := config.RawConfig()
+	kubeConfig, err := config.RawConfig()
 	if err != nil {
 		return err
 	}
 
 	name := strings.TrimPrefix(cluster, "infra:")
 
-	if namespace != "" {
-		name = fmt.Sprintf("%s:%s", name, namespace)
-	}
-
 	// set friendly name based on user input rather than internal format
 	friendlyName := strings.ReplaceAll(name, ":", ".")
 
-	context := fmt.Sprintf("infra:%s", name)
-	if _, ok := kubeconfig.Contexts[context]; !ok {
+	contextName := fmt.Sprintf("infra:%s", name)
+	kubeContext, ok := kubeConfig.Contexts[contextName]
+	if !ok {
 		return fmt.Errorf("context not found: %v", friendlyName)
 	}
 
-	kubeconfig.CurrentContext = context
+	kubeContext.Namespace = namespace
 
-	if err := clientcmd.WriteToFile(kubeconfig, config.ConfigAccess().GetDefaultFilename()); err != nil {
+	kubeConfig.CurrentContext = contextName
+	kubeConfig.Contexts[contextName] = kubeContext
+
+	configFile := config.ConfigAccess().GetDefaultFilename()
+	if err := safelyWriteConfigToFile(kubeConfig, configFile); err != nil {
 		return err
 	}
 
 	fmt.Fprintf(os.Stderr, "Switched to context %q.\n", friendlyName)
-
 	return nil
 }
 
@@ -85,11 +85,16 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
 		return err
 	}
 
-	keep := make(map[string]bool)
+	type clusterContext struct {
+		Namespace string
+		URL       string
+		CA        []byte
+	}
+
+	infraContexts := make(map[string]clusterContext)
 
 	for _, g := range grants {
 		parts := strings.Split(g.Resource, ".")
-
 		cluster := parts[0]
 
 		var namespace string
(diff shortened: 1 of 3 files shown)


internal/cmd/kubernetes.go
```diff
@@ -85,80 +85,78 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
...
 		}
 
-		if !exists {
+		if infraContext.URL == "" {
 			continue
 		}
 
-		u, err := urlx.Parse(url)
+		contextName := "infra:" + cluster
+		if _, ok := infraContexts[contextName]; ok && namespace != "" {
```
mxyng, line 125: If there's a non-namespace grant, set that as the default context. Otherwise, pick a namespace

BruceMacD (commented):
Looks good once checks pass

dnephin (commented):
I have a hard time following `writeKubeconfig`.

Do we have a test case for each of these cases that you could point me at?
1. a user with access to multiple namespaces gets a single context with no namespace set
2. a user with access to a single namespace gets a single context with the namespace set
3. a user with access to the full cluster (not a single namespace) gets a single context with no namespace set

mxyng:
Good call out on the tests. It was definitely missing some cases. I've refactored the test so it should be easier to follow and added the missing cases

internal/cmd/kubernetes_test.go
```diff
@@ -48,123 +43,123 @@ func TestWriteKubeconfig(t *testing.T) {
...
-func TestWriteKubeconfig_UserNamespaceOverride(t *testing.T) {
-	home := t.TempDir()
-	t.Setenv("HOME", home)
-	t.Setenv("USERPROFILE", home)
+	t.Run("MultipleNamespaces", func(t *testing.T) {
+		expectedContexts := map[string]*clientcmdapi.Context{
+			"infra:connected": {
+				AuthInfo:  "user",
+				Cluster:   "infra:connected",
+				Namespace: "namespace",
```
dnephin, line 107: I thought the expectation would be for `Namespace: ""`, since this user has access to multiple namespaces. Is this test expecting the right thing?

dnephin (commented):
Nice! The tests look great now! I'll try and give `writeKubeconfig` a more detailed read tomorrow. One question about the tests

internal/cmd/kubernetes_test.go
mxyng: `Namespace: ""` is equivalent to `Namespace: "default"` which the user may not have access to. It makes sense to set `Namespace` to something as long as the user doesn't have some form of global access. We can probably introduce heuristics to make the selection smarter but for now, it'll pick the first namespace user has access to.

dnephin: I guess if one of the namespaces they have access to is default, then it would make sense for that to be the namespace. If they don't have access to default then any arbitrary one seems reasonable.

internal/cmd/kubernetes.go
```diff
@@ -85,80 +85,82 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
...
 				break
 			}
 		}
 
-		if !exists {
+		if infraContext.URL == "" {
+			continue
+		}
+
+		contextName := "infra:" + cluster
```
jmorganca, line 128: nit: could check this and `continue` earlier in the function

internal/cmd/kubernetes.go
```diff
@@ -85,80 +85,82 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
 		return err
 	}
 
-	keep := make(map[string]bool)
+	type clusterContext struct {
```
jmorganca, line 88:
```suggestion
	type context struct {
```

internal/cmd/kubernetes.go
```diff
@@ -85,80 +85,82 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
...
+		Namespace string
+		URL       string
+		CA        []byte
+	}
+
+	infraContexts := make(map[string]clusterContext)
```
jmorganca, line 94:
```suggestion
	contexts := make(map[string]context)
```

internal/cmd/kubernetes.go
```diff
@@ -85,80 +85,82 @@ func writeKubeconfig(user *api.User, destinations []api.Destination, grants []ap
...
+		if _, ok := infraContexts[contextName]; ok && namespace != "" {
 			continue
 		}
 
-		u, err := urlx.Parse(url)
+		infraContext.Namespace = namespace
+		infraContexts[contextName] = infraContext
+	}
+
+	for contextName, infraContext := range infraContexts {
```
jmorganca, line 137:
```suggestion
	for name, context := range contexts {
```

jmorganca (commented):
Nit: some of the naming could be simplified/tightened

jmorganca:
Looks good! Some super minor comments

internal/cmd/kubernetes.go
mxyng: I wanted to avoid clobbering the `context` package

jmorganca: Good call!

PR #3665, opened 2022-11-17
Add SSH commands and integrate with login

dnephin:
## Summary

Branched from #3620

This PR adds 2 things:

1. `infra login` now checks that `~/.ssh/config` exists and includes `Match exec "infra ssh hosts"`. If it doesn't, the file is created and these two lines are appended:

    ```
    Match exec "infra ssh hosts %h %p"
        Include ~/.ssh/infra/config
    ```

2. Adds the `infra ssh hosts` command does a few things:
   1. tries to match the hostname and port from the `ssh [-p port] <hostname>` command run by the user to the list of destinations the user has access to. If any of them match, proceeds with the following steps. If none match then it returns immediately (which indicates to `ssh` that it should not include the `~/.ssh/infra/config` file).
   2. checks if a key pair exists in `~/.ssh/infra/key`, and if it doesn't generates the key pair and uploads the public key to the infra API
   3. writes the host keys of the target destination to `~/.ssh/infra/known_hosts`
   4. writes a config file to `~/.ssh/infra/config` which tells ssh about the key pair, the known hosts file, the port, and the username to use in the SSH connection.

The `infra ssh hosts` command will be run first. This gives us a nice hook to generate all the files in the `~/.ssh/infra` directory each time the user tries to connect to a new server.

There are a few things to do as follow ups, but this PR has test coverage for the new command, so I think it's in a reasonable state to be reviewed and merged.

### DIFF
--- a/internal/cmd/cmd_test.go
+++ b/internal/cmd/cmd_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal/certs"
+	"github.com/infrahq/infra/internal/server"
 	"github.com/infrahq/infra/uid"
 )
 
@@ -79,6 +80,29 @@ func newTestClientConfig(srv *httptest.Server, user api.User) ClientConfig {
 				Host:               srv.Listener.Addr().String(),
 				TrustedCertificate: string(certs.PEMEncodeCertificate(srv.Certificate().Raw)),
 				AccessKey:          "the-access-key",
+				Expires:            api.Time(time.Now().Add(time.Hour)),
+				Current:            true,
+			},
+		},
+	}
+}
+
+func newTestClientConfigForServer(srv *server.Server, user api.User, acKey string) ClientConfig {
+	if user.Name == "" {
+		user.Name = "testuser@example.com"
+	}
+	if user.ID == 0 {
+		user.ID = uid.New()
+	}
+	return ClientConfig{
+		ClientConfigVersion: clientConfigVersion,
+		Hosts: []ClientHostConfig{
+			{
+				UserID:             user.ID,
+				Name:               user.Name,
+				Host:               srv.Addrs.HTTPS.String(),
+				TrustedCertificate: string(srv.Options().TLS.Certificate),
+				AccessKey:          acKey,
 				Expires:            api.Time(time.Now().Add(time.Hour)),
 				Current:            true,
 			},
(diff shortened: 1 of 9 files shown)


internal/cmd/ssh.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+
+	if err := os.Rename(tmp.Name(), filename); err != nil {
+		return err
+	}
+	cli.Output(`
+Your SSH config at %v has been created or updated to use 'infra ssh hosts' for
```
jmorganca, line 301:
```suggestion
Your SSH config at %v has been updated to connect to Infra SSH destinations.
```

My suggestion is we don't talk about `infra ssh hosts` since it's a hidden command. At the same time, we want to make sure users don't feel Infra is too invasive.

jmorganca (commented):
Overall this looks great! Small suggestion on the SSH output

jmorganca (approved):
Approving since functionally looks great. Let's see if we can update the output to be a bit more friendly since it refers to a hidden command

internal/cmd/ssh.go
```diff
@@ -0,0 +1,388 @@
...
+		// TODO: match destination name as well?
+	}
+	return nil
+}
+
+func splitHostPortSSH(hostname string) (host, port string) {
+	var err error
+	host, port, err = net.SplitHostPort(hostname)
+	if err != nil {
+		return hostname, "22"
```
BruceMacD, line 114:
```suggestion
		return hostname, "22" // default port for ssh client connections is 22
```

internal/cmd/cmd_test.go
```diff
@@ -86,6 +87,29 @@ func newTestClientConfig(srv *httptest.Server, user api.User) ClientConfig {
...
+		user.ID = uid.New()
+	}
+	return ClientConfig{
+		ClientConfigVersion: clientConfigVersion,
+		Hosts: []ClientHostConfig{
+			{
+				UserID:             user.ID,
+				Name:               user.Name,
+				Host:               srv.Addrs.HTTPS.String(),
+				TrustedCertificate: string(srv.Options().TLS.Certificate),
```
BruceMacD, line 104: why not just use `srv.Options.TLS.Certificate`?

BruceMacD (approved):

internal/cmd/cmd_test.go
dnephin: `srv.options` is not exported, so we can't access it from here (different package).

PR #3667, opened 2022-11-17
maintain: return idpauth as struct

BruceMacD:
## Summary
This is just a clean-up, no logical changes. While working on the social login I noticed that the amount of values returned from OIDC login was unwieldy, refactoring them into their own struct instead. 

## Checklist


dnephin (approved):
Looks good!

Issue #3664, opened 2022-11-17
Lots of info is missing in the openapi3 json doc for parameters


technovangelist:
## Describe the bug

For the docs generated by openapi, we need to have a description for all the parameters, otherwise folks won't be able to use the docs. But there is a lot of missing info. I started putting together a table of whats missing, but it will be a huge table. Can we make sure everything has a Description, Example, and Format?

This table has an X for everything that is missing in the first few hundred lines of the json doc

| schema             | parameter name               | description | example | format |
| ------------------------ | ---------------------------- | ----------- | ------- | ------ |
| CreateAccessKeyResponse  | accessKey                    | X           | X       | X      |
|                          | id                           | X           |         |        |
|                          | issuedFor                    | X           |         |        |
|                          | name                         | X           | X       | X      |
|                          | provider                     | X           |         |        |
| CreateGrantResponse      | group                        | X           |         |        |
|                          | id                           | X           |         |        |
|                          | priviledge                   |             | X       | X      |
|                          | resource                     |             | X       | X      |
|                          | user                         | X           |         |        |
|                          | wasCreated                   | X           | X       | X      |
| CreateTokenResponse      | token                        | X           | X       | X      |
| CreateUserResponse       | id                           | X           |         |        |
|                          | name                         | X           | X       | X      |
|                          | oneTimePassword              | X           | X       | X      |
| Destination              | connected                    | X           | X       | X      |
|                          | connection.ca                | X           |         | X      |
|                          | connection.url               | X           |         | X      |
|                          | name                         | X           | X       | X      |
|                          | resources                    | X           | X       | X      |
|                          | roles                        | X           | X       | X      |
|                          | uniqueID                     | X           |         | X      |
|                          | version                      | X           | X       | X      |
| DeviceFlowResponse       | deviceCode                   |             |         | X      |
|                          | userCode                     |             |         | X      |
|                          | verificationURI              |             |         | X      |
| DeviceFlowStatusResponse | deviceCode                   | X           |         | X      |
|                          | login.accessKey              | X           | X       | X      |
|                          | login.name                   | X           | X       | X      |
|                          | login.organizationName       | X           | X       | X      |
|                          | login.passwordUpdateRequired | X           | X       | X      |
|                          | login.userId                 | X           |         |        |
|                          | status                       |             | X       | X      |
| Grant                    | group                        | X           |         |        |
|                          | id                           | X           |         |        |
|                          | privilege                    |             | X       | X      |
[...]

PR #3668, opened 2022-11-17
fix: Document the Destination struct

technovangelist:

## Summary

When openapi generates the json document describing the api, a lot of stuff is missing. This pr fixes whats missing from the Destination object.

## Checklist

## Related Issues

#3664


api/destination.go
```diff
@@ -6,21 +6,21 @@ import (
 )
 
 type Destination struct {
-	ID         uid.ID                `json:"id"`
-	UniqueID   string                `json:"uniqueID" form:"uniqueID" example:"94c2c570a20311180ec325fd56"`
-	Name       string                `json:"name" form:"name"`
-	Kind       string                `json:"kind"`
+	ID         uid.ID                `json:"id" note:"ID of the destination" example:"7a0b06b30F"`
```
ssoroka, line 9: this example isn't a valid ID, there are no `0` in the id set. has to be 11 characters from the set of `123456789abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ`

api/destination.go
```diff
@@ -6,21 +6,21 @@ import (
...
+	UniqueID   string                `json:"uniqueID" form:"uniqueID" note:"Unique ID generated by the connector" example:"94c2c570a20311180ec325fd56"`
+	Name       string                `json:"name" form:"name" note:"Name of the destination" example:"production-cluster"`
+	Kind       string                `json:"kind" note:"Kind of destination. eg. kubernetes or ssh or postgres" example:"kubernetes"`
 	Created    Time                  `json:"created"`
 	Updated    Time                  `json:"updated"`
-	Connection DestinationConnection `json:"connection"`
+	Connection DestinationConnection `json:"connection" note:"Object that includes the URL and CA for the destination" example:"{'url':'localhost:8443', 'ca':'clustercertificate'}"`
```
ssoroka, line 15: this example should probably be covered by the DestinationConnection struct

ssoroka (changes requested):

api/destination.go
technovangelist: does that mean the format for the ID is wrong? that shows up as "[\\da-zA-HJ-NP-Z]{1,11}". i think \d means 0-9 not 1-9

api/destination.go
technovangelist: that doesn't show up in the openapi json doc

api/destination.go
technovangelist: but updated the string to use other digits

technovangelist: and pr #3671 fixes it

BruceMacD (commented):
Looks good, will approve once the docs are re-generated.

dnephin (approved):
LGTM!

jmorganca (approved):

PR #3674, opened 2022-11-18
improve: generate and publish openapi spec in release

mxyng:
## Summary

Generate the OpenAPI spec and publish it as part of the release.


technovangelist:
should we also have the latest always be saved as openapi3.json? So it always creates openapi-version.json and openapi.json. openapi.json just gets overwritten each time.

.github/workflows/cd-binaries.yaml
```diff
@@ -62,7 +64,7 @@ jobs:
       - run: |
           # fail fast if the release does not exist
           gh release view v${{ needs.prepare.outputs.release-name }} || exit 1
-          gh release upload v${{ needs.prepare.outputs.release-name }} *.txt *.zip *.deb *.rpm
+          gh release upload v${{ needs.prepare.outputs.release-name }} *.txt *.zip *.deb *.rpm *.json
```
BruceMacD, line 67: Should we be more specific of the json file we are uploading? The zip, rpm, and deb files are all clearly built things, but a JSON file could be used elsewhere (ex: some kind of configuration file). It seems this is scoped to output from the build steps though so maybe its ok.

dnephin: Matching the `openapi-*.json` pattern might be good so that it documents which file we are expecting to upload.

dnephin (approved):
LGTM

PR #3675, opened 2022-11-18
improve: replace local oidc flow with a streamlined `infra login` that uses browser login

jmorganca:
## Summary

Simplifies the `infra login` command to offer two ways to authenticate:

1. Browser (i.e. "device flow") for identity provider-based login
2. Username & password
3. Access key

The "local" OIDC login was removed

This change isn't backwards compatible with older versions of the Infra Server (< 0.16.0) which is within +/- 1 minor version of this CLI. A warning has been added for users to upgrade the server.

## Checklist

## Next steps
* Have a default server argument so `infra login` "just works"


docs/reference/cli.md
```diff
@@ -9,46 +9,32 @@ title: CLI Reference
...
-# Login with an access key
-$ export INFRA_ACCESS_KEY=1M4CWy9wF5.fAKeKEy5sMLH9ZZzAur0ZIjy
-$ infra login
-
-# Login with pre-set provider and server
-$ export INFRA_SERVER=example.infrahq.com
-$ export INFRA_PROVIDER=google
-$ infra login
+# Login with username (will prompt for password)
+$ infra login --user me@example.com
```
BruceMacD, line 24: nice

internal/cmd/local_server.go
```diff
@@ -1,118 +0,0 @@
-package cmd
```
BruceMacD, line 1: We need to delete the `cmd/pages` directory also. It contains the pages that the local server displayed.

internal/cmd/login.go
```diff
@@ -516,12 +442,14 @@ func deviceFlowLogin(ctx context.Context, client *api.Client, cli *CLI) (*api.Lo
 		return nil, err
 	}
 
+	url := resp.VerificationURI + "?code=" + resp.UserCode
+
 	// display to user
-	cli.Output("Navigate to " + resp.VerificationURI + " and enter the following code:\n")
+	cli.Output("Navigate to " + url + " and enter the following code:\n")
```
BruceMacD, line 448: If they enter the url as declared above they dont have to enter the code.

BruceMacD (approved):

docs/reference/cli.md
```diff
@@ -9,46 +9,32 @@ title: CLI Reference
...
 ```
 
 #### Options
 
 ```console
-      --key string                       Login with an access key
       --no-agent                         Skip starting the Infra agent in the background
       --non-interactive                  Disable all prompts for input
-      --provider string                  Login with an identity provider
+  -p, --password string                  Password
```
dnephin, line 33: I think generally it's considered a bad idea to specify the password on the command line, because it can be see by anyone with access to `/proc`, and it also ends up in files like the bash history file (and probably others) that may be world readable.

For non-interactive the environment variable works well. That's still visible in `/proc`, but at least won't end up in log files in all cases.

`psql` and probably others have a `--password` flag to force a prompt. I think that'd be a nice way to do this. 
1. If someone sets `--user <username>` they only get prompted for password
2. If someone sets `--password` they get prompted for both
3. If they set both it's no different than `--user <username>`

I'd also suggest we remove `-p`. We don't need short flags for most operations, and it in the future we might want it for something else.

```suggestion
      --password string                  Prompt for a password.
```

docs/manage/idp/azure-ad.md
```diff
@@ -25,26 +25,21 @@ infra providers add azure \
 3. Click **New registration**
 4. Register the application:
    - For **Name** write **Infra**
-   - For **Redirect URI** select `Web` and add `http://localhost:8301`
+   - For **Redirect URI** select `Web` and add `https://<INFRA_SERVER_HOST>/login/callback`
```
dnephin, line 28: Do we explain what `INFRA_SERVER_HOST` is anywhere in these docs? Might be a good follow up to say something about it (ex: generally `<org>.infrahq.com`).

internal/cmd/login.go
```diff
@@ -84,12 +63,40 @@ $ infra login`,
 				options.Server = args[0]
 			}
 
+			if options.Server == "" {
+				return Error{
+					Message: "No server specified. Usage: infra login SERVER",
+				}
+			}
```
dnephin, line 70: If we're planning on doing something like "lookup the last server from their local config when not set" in a follow up, then I think this diff makes sense.

If we're not planning that, maybe we should change `Args: MaxArgs(1)` above to instead be `Args: ExactArgs(1)` to avoid this custom handling?

dnephin (commented):
Looks good! This is a nice simplification.

My only concern would be about the `--password` flag accepting the password

internal/cmd/local_server.go
jmorganca: Thanks!

docs/reference/cli.md
jmorganca: I opted to just remove the `--password` flag, as it's the same as providing `--user`.

internal/cmd/login.go
```diff
@@ -24,57 +23,42 @@ import (
 	"github.com/infrahq/infra/internal/certs"
 	"github.com/infrahq/infra/internal/cmd/types"
 	"github.com/infrahq/infra/internal/format"
-	"github.com/infrahq/infra/internal/generate"
 	"github.com/infrahq/infra/internal/logging"
 )
 
 type loginCmdOptions struct {
 	Server             string
-	AccessKey          string
```
mxyng, line 33: The access key story will likely need to be revisited. The CLI currently can take access key from an environment variable but requires an existing configuration. This is currently done by setting `INFRA_SERVER` and `INFRA_ACCESS_KEY` followed by `infra login` which uses the environment variables to set up these configurations. By removing the access key login option, we're removing authn with access key.

Now the workflow described above isn't ideal since `infra login` should not be necessary when using an access key but the goal of using an access key with the Infra CLI should still be valid.

mxyng (commented):
Great improvement! I'm concerned about removing access key, not necessarily as a login option but as a method of authenticating with the server. This will break existing deployment workflows

internal/cmd/login.go
```diff
@@ -84,12 +68,15 @@ $ infra login`,
 				options.Server = args[0]
 			}
 
+			if user, ok := os.LookupEnv("INFRA_USER"); ok {
+				options.User = user
+			}
```
dnephin, line 73: The `cliopts.DefaultsFromEnv` I add in #3710 may be able to replace this.

I think this code may be wrong because the env var value overrides the cli flag, where as the usual order is that CLI flags override the env var.

By removing this, `cliopts.DefaultsFromEnv` will set the value correctly from env var, but only when the CLI flag is not set to a value.

internal/cmd/login.go
```diff
@@ -24,57 +23,42 @@ import (
...
+# Login
+infra login example.infrahq.com
+
+# Login with username (will prompt for password)
+infra login example.infrahq.com --user foo@gmail.com
+
+# Login with username and password in (non-interactive)
+export INFRA_SERVER=example.infrahq.com
+export INFRA_USER=foo@gmail.com
+export INFRA_PASSWORD=p4ssw0rd
```
dnephin, line 59: There's no longer any `INFRA_PASSWORD` env var, right? Seems like that was removed along with the flag. Do we still want the password to be set exclusively from env var?

dnephin: Oops, it still exists, but was in a different place than I expected.

internal/cmd/login.go
```diff
@@ -105,10 +92,30 @@ func login(cli *CLI, options loginCmdOptions) error {
 		return err
 	}
 
-	if options.Server == "" {
-		if options.NonInteractive {
+	var password string
+	if p, ok := os.LookupEnv("INFRA_PASSWORD"); ok {
+		password = p
+	}
```
dnephin, line 98: Not a blocker, but I wonder if this should be next to where we read `INFRA_SERVER`, so that all the env var reading is in the same place.

internal/cmd/login.go
jmorganca: Great catch, thanks! Yes that's a bug – CLI flag should always take precedent if specified

internal/cmd/login.go
jmorganca: Good catch thx!

docs/manage/idp/azure-ad.md
```diff
@@ -56,4 +51,4 @@ infra providers add azure \
      - **User** > **User.Read** - Click **Add permissions**.
      - Click **Grant admin consent for Default Directory** and select **Yes** when prompted.
        ![Add API permissions](../../images/azure-setup/connect-users-azure-6.png)
-9. From the **Overview** tab copy the **Application (client) ID**, **Directory (tenant) ID**, and **Client Secret** values and provide them into Infra's Dashboard or CLI.
+8. From the **Overview** tab copy the **Application (client) ID**, **Directory (tenant) ID**, and **Client Secret** values and provide them into Infra's Dashboard or CLI.
```
mxyng, line 54: For markdown, I suggest just using `1.` for order lists since it will automatically get converted to the right number. Leaving it as `1.` reduces diffs like this as well as mental gymnastics of maintaining the ordering

internal/cmd/login.go
```diff
@@ -25,57 +24,48 @@ import (
...
+export INFRA_SERVER=example.infrahq.com
+export INFRA_ACCESS_KEY=2vrEbqFEUr.jtTlxkgYdvghJNdEa8YoUxN0
+infra login example.infrahq.com --user user@example.com
+
+# Login with username and password
+export INFRA_SERVER=example.infrahq.com
+export INFRA_USER=user@example.com
+export INFRA_PASSWORD=p4ssw0rd
+infra login
+		`,
```
mxyng, line 68: Is this line intentional? It's add an empty line to the generated docs which isn't ideal.

internal/cmd/login.go
```diff
@@ -25,57 +24,48 @@ import (
...
-# Login with pre-set provider and server
-$ export INFRA_SERVER=example.infrahq.com
-$ export INFRA_PROVIDER=google
-$ infra login`,
+		Use:     "login [SERVER]",
+		Short:   "Login to Infra",
 		Args:    MaxArgs(1),
 		GroupID: groupCore,
+		Example: `
+# Login
```
mxyng, line 52: This should be in the previous line otherwise there's an empty line in the generated docs

internal/cmd/login.go
```diff
@@ -218,12 +239,8 @@ func loginToInfra(cli *CLI, lc loginClient, loginReq *api.LoginRequest, noAgent
 	if err != nil {
 		if api.ErrorStatusCode(err) == http.StatusUnauthorized || api.ErrorStatusCode(err) == http.StatusNotFound {
 			switch {
-			case loginReq.AccessKey != "":
-				return &LoginError{Message: "your access key may be invalid"}
```
mxyng, line 222: Should this be removed? Access key looks unchanged so removing it may lead to some confusing errors

internal/cmd/login.go
jmorganca: Thanks!

internal/cmd/login.go
jmorganca: No – good catch!

internal/cmd/login.go
jmorganca: This is a fantastic catch, thanks! I've restored logging in with access keys since this is a simple way to authenticate as a non-human user.

internal/cmd/login.go
jmorganca: Thanks, I must have incorrectly committed this

dnephin:
CI is not running on this PR after the last push. Might need to force push something to trigger another run.

dnephin (approved):

internal/cmd/login.go
```diff
@@ -25,57 +24,47 @@ import (
...
+		Example: `# Login
+infra login example.infrahq.com
+
+# Login with username and password (prompt for password)
+infra login example.infrahq.com --user user@example.com
+
+# Login with access key
+export INFRA_SERVER=example.infrahq.com
+export INFRA_ACCESS_KEY=2vrEbqFEUr.jtTlxkgYdvghJNdEa8YoUxN0
+infra login example.infrahq.com --user user@example.com
```
BruceMacD, line 61: It looks like the user is ignored when an access key is specified, so it should be omitted from this example.

BruceMacD (approved):
The example for access key exchange should be fixed, but otherwise looks good.

internal/cmd/login.go
jmorganca: Good catch. Thanks!

PR #3676, opened 2022-11-18
feat: add last used column

hoyyeva:
## Summary
API added the new field `last used` for the access keys and it is also available on CLI. So we updated the access keys table on UI as well to expose the value of last used of access key

## 📷 
Screen Shot 2022-11-17 at 11 17 26 PM
Screen Shot 2022-11-17 at 11 17 10 PM

### DIFF
--- a/ui/pages/settings/index.js
+++ b/ui/pages/settings/index.js
@@ -269,7 +269,6 @@ export default function Settings() {
                 columns={[
                   {
                     cell: function Cell(info) {
-                      console.log(info)
                       return (
                         <div className='flex flex-col py-0.5'>
                           <div className='truncate text-sm font-medium text-gray-700'>
--- a/ui/pages/settings/index.js
+++ b/ui/pages/settings/index.js
@@ -269,11 +269,28 @@ export default function Settings() {
                 columns={[
                   {
                     cell: function Cell(info) {
+                      console.log(info)
                       return (
                         <div className='flex flex-col py-0.5'>
                           <div className='truncate text-sm font-medium text-gray-700'>
                             {info.getValue()}
                           </div>
+                          {info.row.original.created && (
+                            <div className='space-y-1 pt-2 text-3xs text-gray-500 sm:hidden'>
+                              created -{' '}
+                              <span className='font-semibold text-gray-700'>
+                                {moment(info.row.original.created).from()}
+                              </span>
+                            </div>
+                          )}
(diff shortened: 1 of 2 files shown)


ui/pages/settings/index.js
```diff
@@ -269,11 +269,28 @@ export default function Settings() {
                 columns={[
                   {
                     cell: function Cell(info) {
+                      console.log(info)
```
jmorganca, line 272: Remove me :-)

ui/pages/settings/index.js
```diff
@@ -269,11 +269,28 @@ export default function Settings() {
...
+                            <div className='space-y-1 pt-2 text-3xs text-gray-500 sm:hidden'>
+                              created -{' '}
+                              <span className='font-semibold text-gray-700'>
+                                {moment(info.row.original.created).from()}
+                              </span>
+                            </div>
+                          )}
+                          {info.row.original.lastUsed && (
+                            <div className='space-y-1 pt-2 text-3xs text-gray-500 sm:hidden'>
+                              last used -{' '}
```
jmorganca, line 288: note/nit: would make sure this doesn't wrap on smaller window sizes

jmorganca (commented):
LGTM once the `console.log` is removed

ui/pages/settings/index.js
```diff
@@ -274,6 +274,22 @@ export default function Settings() {
...
+                              <span className='font-semibold text-gray-700'>
+                                {moment(info.row.original.lastUsed).from()}
```
jmorganca, line 289: Do we know if `lastUsed` is always set? This may cause an error in `moment`

hoyyeva: this is only on the small screen and this wont random if `info.row.original.lastUsed` is not true

ui/pages/settings/index.js
```diff
@@ -299,6 +315,17 @@ export default function Settings() {
                     ),
                     accessorKey: 'created',
                   },
+                  {
+                    cell: info => (
+                      <div className='hidden sm:table-cell'>
+                        {info.getValue() ? moment(info.getValue()).from() : '-'}
+                      </div>
```
hoyyeva, line 322: and this is how we handle when its on the bigger screen where we see the value in the table - so if the value is not set then it will return `-`

jmorganca (approved):

PR #3677, opened 2022-11-18
maintain: update make docs to output correct version

BruceMacD:
## Summary
The `make docs` command was setting the version to the `99...` placeholder value from the code. Updating the command to use the correctly generated output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ docs: docs/api/openapi3.json
 
 .PHONY: docs/api/openapi3.json
 docs/api/openapi3.json:
-	go test ./internal/server -run TestWriteOpenAPIDocToFile -update
+	go run -ldflags '-s -X github.com/infrahq/infra/internal.Version=0.0.0' ./internal/openapigen $@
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -56,7 +56,7 @@ docs: docs/api/openapi3.json
 
 .PHONY: docs/api/openapi3.json
 docs/api/openapi3.json:
-	go run ./internal/openapigen $@
+	go test ./internal/server -run TestWriteOpenAPIDocToFile -update
```
dnephin, line 59: Another option that just came to mind is:

```suggestion
	go run -ldflags '-s -X github.com/infrahq/infra/internal.Version=0.0.0' ./internal/openapigen $@
```

I think this allows us to check that `internal/openapigen` still works ahead of a release. I tried it locally and it seems to generate the correct doc.

dnephin (approved):
Thanks for making this change!

I know we talked about this a little while ago. I forgot if I was supposed to make this change. My apologies if I lost track of it.

PR #3679, opened 2022-11-18
Revert "maintain: remove unused get grant endpoint (#3650)"

mxyng:
This reverts commit d4259e4e7657337d11c817516c2948809d7672a4.

## Summary

While `GetGrant` and `GET /api/grants/:id` is not used by code in this repo, it'll be used by the Terraform provider. Removing it break that use case.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -348,6 +348,10 @@ func (c Client) ListGrants(ctx context.Context, req ListGrantsRequest) (*ListRes
 	})
 }
 
+func (c Client) GetGrant(ctx context.Context, id uid.ID) (*Grant, error) {
+	return get[Grant](ctx, c, fmt.Sprintf("/api/grants/%s", id), Query{})
+}
+
 func (c Client) CreateGrant(ctx context.Context, req *GrantRequest) (*CreateGrantResponse, error) {
 	return post[CreateGrantResponse](ctx, c, "/api/grants", req)
 }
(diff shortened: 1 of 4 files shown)


dnephin:
#3156 had some context. The rational for removing it was that you already have all the detail about a grant from the create or list.

I had a quick look at https://github.com/infrahq/terraform-provider-infra/pull/1/files. It seems like the only time we call `GetGrant` is immediately after a `CreateGrant`. Is that right? `CreateGrantResponse` already has the full details of the grant. Is  there any reason to query for the grant again?

mxyng:
It's not quite accurate. `resourceUserGrantRead` and `resourceGroupGrantRead` are used to refresh the Terraform state. Terraform refreshes before most operations to determine if there's diff between the local state and remote state.

dnephin:
Ah yes, I see that now. So we would use `GetGrant`, because all we're storing in state is the `grant.ID` ? If we stored the natural identifier of `subject/resource/privilege` I guess we could use that with `ListGrant`. That might work better for the case where the grant ID changed, but otherwise the details of the grant have not.

Is there any way we could store more details about the grant to avoid the lookup by ID?

If not, I think it doesn't hurt to restore this endpoint, but I was hoping that in the future we could replace it with `GET /api/grants/:subject/:resource/:privilege`.

mxyng:
It does store the values for subject, resource, and privilege so it's possible to use `ListGrants`. I wanted to avoid doing a search when a more direct option is available.

dnephin (approved):
Ok, let's bring this back!

Issue #3661, opened 2022-11-17
UI should not show browser session access key and allow deletion


mchiang0610:
The UI should not show browser session access key and allow deletion. 

![image](https://user-images.githubusercontent.com/3325447/202499068-2d3ad116-ecd6-42b3-9d40-ce3ec4321eaf.png)

mchiang0610:
related to #3593

dnephin:
I opened #3681 which changes the API to refuse the delete if the access key being deleted is the same one that was used to authenticate the request. That change probably doesn't address this issue entirely (because we may also want to hide the button in the UI), but it's probably a nice safeguard to prevent other API clients from making the same mistake.

PR #3681, opened 2022-11-18
Refuse to delete the access key being used by the API request

dnephin:
## Summary

Related to part of #3661

This PR changes the `DELETE /api/access-keys/:id` API to refuse to delete the key, if it's the key that was used to authenticate the current request. We do something similar for users (we don't allow users to delete themselves).We ran into this problem earlier in the week, and I suspect we don't want to allow this operation generally.

I did a bit of cleanup in the tests to split the `ListAccessKeys` tests from the `DeleteAccessKey` tests. That's done in a second commit, so you can look at just the first commit to see the new test case that was added.

I'll comment on some of the other cleanup inline.

### DIFF
--- a/internal/access/access_key.go
+++ b/internal/access/access_key.go
@@ -57,9 +57,7 @@ func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string, 
 	return body, err
 }
 
-func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
-	rCtx := GetRequestContext(c)
-
+func DeleteAccessKey(rCtx RequestContext, id uid.ID, name string) error {
 	var key *models.AccessKey
 	var err error
 
@@ -89,10 +87,13 @@ func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
 	if key.IssuedFor == rCtx.Authenticated.User.ID {
 		// users can delete their own keys
 	} else {
-		_, err := RequireInfraRole(c, models.InfraAdminRole)
-		if err != nil {
+		if err := IsAuthorized(rCtx, models.InfraAdminRole); err != nil {
 			return HandleAuthErr(err, "access key", "delete", models.InfraAdminRole)
 		}
+	}
+
+	if rCtx.Authenticated.AccessKey.ID == key.ID {
+		return fmt.Errorf("%w: cannot delete the access key used by this request", internal.ErrBadRequest)
 	}
 
 	return data.DeleteAccessKeys(rCtx.DBTxn, data.DeleteAccessKeysOptions{ByID: key.ID})
--- a/internal/access/access_key_test.go
+++ b/internal/access/access_key_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gin-gonic/gin"
 	"gotest.tools/v3/assert"
 
 	"github.com/infrahq/infra/internal/server/data"
@@ -21,7 +22,13 @@ func TestAccessKeys_SelfManagement(t *testing.T) {
 	err = data.CreateIdentity(db, user)
 	assert.NilError(t, err)
 
-	c, _ := loginAs(&data.Transaction{DB: db.DB}, user, org)
+	c, _ := gin.CreateTestContext(nil)
+	tx := txnForTestCase(t, db).WithOrgID(org.ID)
+	rCtx := RequestContext{
+		DBTxn:         tx,
+		Authenticated: Authenticated{User: user, Organization: org},
+	}
+	c.Set(RequestContextKey, rCtx)
 
 	t.Run("can manage my own keys", func(t *testing.T) {
 		key := &models.AccessKey{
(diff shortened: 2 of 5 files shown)


internal/access/access_key.go
```diff
@@ -57,9 +57,7 @@ func CreateAccessKey(c *gin.Context, accessKey *models.AccessKey) (body string,
 	return body, err
 }
 
-func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
-	rCtx := GetRequestContext(c)
-
+func DeleteAccessKey(rCtx RequestContext, id uid.ID, name string) error {
```
dnephin, line 60: I changed from `gin.Context` to `RequestContext` here because the tests in `access/access_key_test` were failing due to this change. By accepting the `RequestContext` I can avoid mutating the `gin.Context`, which was being re-used by subsequent test cases.

I believe we have a few other access functions that already accept `RequestContext` directly, and long term I'd like to see us use `RequestContext` directly in both the API handlers and the access functions.

internal/access/access_key.go
```diff
@@ -89,12 +87,15 @@ func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
 	if key.IssuedFor == rCtx.Authenticated.User.ID {
 		// users can delete their own keys
 	} else {
-		_, err := RequireInfraRole(c, models.InfraAdminRole)
-		if err != nil {
+		if err := IsAuthorized(rCtx, models.InfraAdminRole); err != nil {
```
dnephin, line 90: `RequireInfraRole` calls `IsAuthorized`, so this is not a change in behaviour. The only difference between these functions is one accepts `gin.Context` and returns a DB, while the other accepts `RequestContext` and expects the caller to use the db from the `RequestContext`.

 I had to change from one to the other because I removed `gin.Context`.

internal/access/access_key.go
```diff
@@ -89,12 +87,15 @@ func DeleteAccessKey(c *gin.Context, id uid.ID, name string) error {
...
 			return HandleAuthErr(err, "access key", "delete", models.InfraAdminRole)
 		}
 	}
 
+	if rCtx.Authenticated.AccessKey.ID == key.ID {
+		return fmt.Errorf("%w: cannot delete the access key used by this request", internal.ErrBadRequest)
+	}
```
dnephin, line 97: This is the primary change to behaviour.  If anyone has suggestions to improve the wording I'm happy to make those changes.

internal/access/access_key_test.go
```diff
@@ -33,7 +40,9 @@ func TestAccessKeys_SelfManagement(t *testing.T) {
 		_, err = CreateAccessKey(c, key)
 		assert.NilError(t, err)
 
-		err = DeleteAccessKey(c, key.ID, "")
+		r := rCtx // shallow copy
+		r.Authenticated.AccessKey = &models.AccessKey{}
+		err = DeleteAccessKey(r, key.ID, "")
```
dnephin, line 45: These test changes were necessary because previously this test did not set `Authenticated.AccessKey` which caused a panic now that we're checking `Authenticated.AccessKey`. We will never encounter that scenario in production because an `AccessKey` is required for this endpoint.

internal/server/access_keys_test.go
```diff
@@ -243,53 +243,6 @@ func TestAPI_ListAccessKeys(t *testing.T) {
 		}
 	})
 
-	t.Run("delete by name", func(t *testing.T) {
```
dnephin, line 246: All of this below was moved to `TestAPI_DeleteAccessKey`. Look at the first commit to see the new test case before I extracted these cases into a separate function.

pdevine:
Two things on my mind about this change:
1. Are there any circumstances when it would make sense to remove the access key that you're currently using? (I can't really think of any, but maybe I'm being unimaginative here)
2. I think we still probably need to change something in the UI to give you a better prompt to not delete your current session.

I had been talking to @hoyyeva about this, and we had discussed setting a second cookie with the name of the access key that you're currently using. That would enable to the UI not even allow you to select the key to delete. I don't think that change would preclude this change though (belt and suspenders?).

internal/server/access_keys_test.go
pdevine: Thanks for moving this. I had looked at doing it in the last commit and it was getting a little crazy.

pdevine (approved):

dnephin:
> Are there any circumstances when it would make sense to remove the access key that you're currently using? (I can't really think of any, but maybe I'm being unimaginative here)

I was trying to think of that as well, but I don't think so. We have the same check for a user (trying to delete themselves), so this feels consistent to me. 

> I think we still probably need to change something in the UI to give you a better prompt to not delete your current session.

I agree, I commented on the issue saying the same: https://github.com/infrahq/infra/issues/3661#issuecomment-1320529700. I think something like this is a nice safeguard for other API clients (CLI, terraform, etc).

jmorganca:
LGTM! Nice fix

RE deleting the current access key: logging out should be the only use case and is handled separately.

jmorganca (approved):

PR #3701, opened 2022-11-21
Add test coverage for background job wrapper

dnephin:
## Summary

Branched from #3643

I made some changes to this code in #3643, so adding these tests to show that I didn't break anything

### DIFF
--- a/internal/server/backgroundjobs.go
+++ b/internal/server/backgroundjobs.go
@@ -10,13 +10,11 @@ import (
 	"github.com/infrahq/infra/internal/server/jobs"
 )
 
-// BackgroundJobFunc is the interface for implementing a new background job.
-//
-// errors will be logged but will not cause the app to crash.
-//
-// panics will be caught and logged
-//
-// jobs should gracefully exit if their context quits, eg ctx.Done() or ctx.Err()
+// BackgroundJobFunc is the interface for running periodic background jobs, like
+// garbage collection. Errors and panics from the job will be logged by
+// jobWrapper, and will not stop the goroutine running the job.
+// The job should exit when ctx is cancelled. Unlike most transactions, the
+// transaction passed to this job will not have an OrganizationID.
 type BackgroundJobFunc func(ctx context.Context, tx *data.Transaction) error
 
 func (s *Server) SetupBackgroundJobs(ctx context.Context) {
--- a/internal/server/backgroundjobs_test.go
+++ b/internal/server/backgroundjobs_test.go
@@ -1,0 +1,91 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"golang.org/x/sync/errgroup"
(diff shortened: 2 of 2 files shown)


internal/server/backgroundjobs.go
```diff
@@ -51,6 +51,7 @@ func jobWrapper(ctx context.Context, db *data.DB, job BackgroundJobFunc, every t
 			if err != nil {
 				return fmt.Errorf("failed to start transaction :%w", err)
 			}
+			tx = tx.WithOrgID(db.DefaultOrg.ID)
```
dnephin, line 54: I suspect that most of our jobs will perform queries that ignore the organization, but for testing it helped to have the org set.  The previous code would set a default org, so this restores the behaviour from before #3643.

internal/server/backgroundjobs_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	runOnce := func() {
+		chRun <- signal
+		<-chReady
+	}
+
+	var runID int
+	job := func(ctx context.Context, tx *data.Transaction) error {
+		chReady <- signal
+		<-chRun
```
dnephin, line 34: This allows us to test the loop one iteration at a time.

The background goroutine will signal when it's ready, and wait for a signal from the main test goroutine before running.

Each test case will call `runOnce` which tells the background goroutine to run, and wait for the loop to get back to `chReady` before making assertions about the run.

internal/server/backgroundjobs_test.go
```diff
@@ -0,0 +1,90 @@
...
+	})
+	runStep(t, "panic is recovered", func(t *testing.T) {
+		runOnce()
+	})
+	runStep(t, "cancel shutdowns the job", func(t *testing.T) {
+		runOnce()
+		start := time.Now()
+		cancel()
+		assert.NilError(t, g.Wait())
+		assert.Assert(t, time.Since(start) < 3*time.Second)
```
BruceMacD, line 82: Could this test really take 3 seconds or is that just a safe amount of buffer time?

internal/server/backgroundjobs.go
dnephin: Fair enough. I added it because the code I had refactored was doing this previously and I didn't want to change the behaviour yet.  

It's unfortunately somewhat risky I think, because so much of our code panics if we don't set an OrgID (which I'm not a fan of), and it's very difficult to test the integration of the "jobs" with this `jobWrapper`.

So we end up not having very good test coverage to catch when someone creates a job that expects an organization. I guess it's still better that this panics than to only use the default org. 

I'll remove this line and handle it in the test.

internal/server/backgroundjobs_test.go
dnephin: It's just a large buffer of time to be safe when things are slow with `-race` and in CI. On my machine this takes less than 10ms.

BruceMacD (approved):

PR #3704, opened 2022-11-21
maintain: move allowed login domains to a global org field

BruceMacD:
- move domains that are allowed to use social login to the org table rather than on specific providers
- remove allowed domains from providers
- set allowed domains to match the org admin on sign-up
- remove validation added for provider allowed domains

## Summary
Google social login is going to be global provider, not created in each org. To allow for this the `allowed domains` that can login via Google (social login) need to be stored on the org rather than an individual provider level.

## Checklist

## Related Issues

Part of #3261


internal/server/data/data.go
```diff
@@ -227,9 +227,10 @@ func initialize(db *DB) error {
 	switch {
 	case errors.Is(err, internal.ErrNotFound):
 		org = &models.Organization{
-			Model:     models.Model{ID: defaultOrganizationID},
-			Name:      "Default",
-			CreatedBy: models.CreatedBySystem,
+			Model:               models.Model{ID: defaultOrganizationID},
+			Name:                "Default",
+			CreatedBy:           models.CreatedBySystem,
+			AllowedLoginDomains: []string{},
```
BruceMacD, line 233: The default organization created through config will never need `AllowedLoginDomains` since social login is only available in the hosted Infra offering.

internal/server/data/migrations.go
```diff
@@ -923,3 +924,19 @@ func modifyAccessKeysIndex() *migrator.Migration {
...
+func moveAllowedDomainsToOrganizationsTable() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-11-21T11:00",
+		Migrate: func(tx migrator.DB) error {
+			stmt := `ALTER TABLE providers DROP COLUMN IF EXISTS allowed_domains`
+			if _, err := tx.Exec(stmt); err != nil {
+				return err
+			}
+
+			stmt = `ALTER TABLE organizations ADD COLUMN IF NOT EXISTS allowed_login_domains text DEFAULT ''`
```
BruceMacD, line 937: Not adding any allowed social login domains in this migration, we can expose it as a setting in the UI for existing orgs once other social login changes are in.

internal/server/handlers.go
```diff
@@ -199,7 +212,18 @@ func (a *API) Login(c *gin.Context, r *api.LoginRequest) (*api.LoginResponse, er
 			return nil, fmt.Errorf("update provider client: %w", err)
 		}
 
-		loginMethod = authn.NewOIDCAuthentication(r.OIDC.ProviderID, r.OIDC.RedirectURL, r.OIDC.Code, providerClient)
+		loginMethod, err = authn.NewOIDCAuthentication(
+			provider,
+			r.OIDC.RedirectURL,
+			r.OIDC.Code,
+			providerClient,
+			// TODO: specify allowed login domains from rCtx.Authenticated.Organization.AllowedLoginDomains
```
BruceMacD, line 220: Waiting to add the allowed domains until #3657 is in, or this could result in incorrect login failures

internal/server/organizations_test.go
```diff
@@ -157,17 +155,20 @@ func TestAPI_GetOrganization(t *testing.T) {
...
 					time.Now().UTC().Format(time.RFC3339),
 					srv.db.DefaultOrg.Domain,
+					srv.db.DefaultOrg.AllowedLoginDomains,
 				))
 				actual := jsonUnmarshal(t, resp.Body.String())
 
-				var cmpAPIOrganizationJSON = gocmp.Options{
+				cmpAPIOrganizationJSON := gocmp.Options{
 					gocmp.FilterPath(pathMapKey(`created`, `updated`), cmpApproximateTime),
+					gocmp.FilterPath(pathMapKey(`allowedLoginDomains`), cmpEquateEmptySlice),
```
BruceMacD, line 171: Without this comparator this test failed on `[]any{} != string([])`

api/organization.go
```diff
@@ -29,8 +30,9 @@ func (r ListOrganizationsRequest) ValidationRules() []validate.ValidationRule {
 }
 
 type CreateOrganizationRequest struct {
-	Name   string `json:"name"`
-	Domain string `json:"domain"`
+	Name                string   `json:"name"`
+	Domain              string   `json:"domain"`
+	AllowedLoginDomains []string `json:"allowedLoginDomains" note:"domains which can be used with social login for this organization" example:"['example.com', 'infrahq.com']"`
```
jmorganca, line 35: Does this need to be here for creation?

BruceMacD: I thought it may be useful when we directly create an org so that we dont have to configure a username/password for that org. I removed it for now though, we can add it when we need it.

api/organization.go
```diff
@@ -6,11 +6,12 @@ import (
...
-	Name    string `json:"name"`
-	Created Time   `json:"created"`
-	Updated Time   `json:"updated"`
-	Domain  string `json:"domain"`
+	ID                  uid.ID   `json:"id"`
+	Name                string   `json:"name"`
+	Created             Time     `json:"created"`
+	Updated             Time     `json:"updated"`
+	Domain              string   `json:"domain"`
+	AllowedLoginDomains []string `json:"allowedLoginDomains" note:"domains which can be used with social login for this organization" example:"['example.com', 'infrahq.com']"`
```
jmorganca, line 14:
```suggestion
	AllowedDomains []string `json:"allowedLoginDomains" note:"domains which can be used to log into this organization" example:"['example.com', 'infrahq.com']"`
```

`AllowedDomains` is a better name here, and will eventually be for any non-idp login so we may want to keep the comment general

internal/server/authn/oidc.go
```diff
@@ -48,14 +47,14 @@ func (a *oidcAuthn) Authenticate(ctx context.Context, db *data.Transaction, requ
 		return AuthenticatedIdentity{}, fmt.Errorf("exhange code for tokens: %w", err)
 	}
 
-	if len(provider.AllowedDomains) > 0 {
+	if len(a.AllowedLoginDomains) > 0 {
 		// get the domain of the email
 		at := strings.LastIndex(idpAuth.Email, "@") // get the last @ since the email spec allows for multiple @s
```
jmorganca, line 52: This seems to be the same logic as above `GetDomain`?

internal/server/authn/oidc.go
```diff
@@ -48,14 +47,14 @@ func (a *oidcAuthn) Authenticate(ctx context.Context, db *data.Transaction, requ
 		return AuthenticatedIdentity{}, fmt.Errorf("exhange code for tokens: %w", err)
 	}
 
-	if len(provider.AllowedDomains) > 0 {
+	if len(a.AllowedLoginDomains) > 0 {
```
jmorganca, line 50: Just to confirm, do we verify somewhere that this is the "social" provider and not a random 3rd party IDP?

jmorganca (commented):
Looking good! A few comments

internal/server/authn/oidc.go
BruceMacD: This was a hold-over from an earlier implementation, removed it

internal/server/authn/oidc.go
BruceMacD: The check for social login is in #3657, I removed this logic and added a TODO that I will resolve in that PR instead

mxyng:
Looks good overall. I'm curious about the decision to include the signup user's email domain by default. I would've expected to be a user configuration when they decide to enable social login.

BruceMacD:
> Looks good overall. I'm curious about the decision to include the signup user's email domain by default. I would've expected to be a user configuration when they decide to enable social login.

That is a viable option too for sure, I chose to use the admin's email at login for a consistent experience for the time being which doesnt require extra steps after sign-up.

mxyng (approved):

PR #3705, opened 2022-11-21
Revert "improve: generate and publish openapi spec in release"

jmorganca:
Reverts infrahq/infra#3674 as it is blocking v0.18.0


mchiang0610 (approved):
lgtm

PR #3710, opened 2022-11-22
Probably fix some test flakes

dnephin:
## Summary

This PR includes 3 small changes that I made while debugging [this flaky test](https://github.com/infrahq/infra/actions/runs/3501714088/jobs/5865598897). Full details in the commit messages.

1. Use a random schema name in most tests. This is possibly the main fix for the flake (at least the ones I was seeing locally), but I don't think I can explain how this would fix CI because there should not have been duplicate schema names in any CI run.
2. Disable the `infra agent` in tests. It would have been started by login tests. This should prevent the background process from running, and prevent request goroutines from outliving the test and keeping the old schema around, with old data in it. Now that we're using random schemas this should be less of a problem, but could still cause strange behaviour.
3. A small optimization to avoid encrypting an empty payload. This prevented the test flake because the field that was causing the "wrong key" error was an empty value, but I don't think that's the root cause of the problem.


dnephin:
#3665 passed in CI with these fixes. I'll try re-running them to see if it passes again. They were failing almost 50% of the time before this fix.

mxyng (approved):

PR #3711, opened 2022-11-23
maintain: default access key lifetime 1 year

BruceMacD:
## Summary
Change the default access key lifetime to be 1 year, with an inactivity timeout of 30 days. This means tokens that are actively in use will not unexpectedly fail in a short window.

Screen Shot 2022-11-23 at 9 51 37 AM

```
infra keys add
Issued access key "test@example.com-gWf2fJwD39"
This key will expire in 1 year, and must be used every 30 days to remain valid

Key: aaa.bbb
```

## Checklist

### DIFF
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -12,7 +12,10 @@ import (
 	"github.com/infrahq/infra/internal/logging"
 )
 
-const thirtyDays = 30 * (24 * time.Hour)
+const (
+	thirtyDays = 30 * (24 * time.Hour)
+	oneYear    = 366 * (24 * time.Hour) // leap year
+)
 
 func newKeysCmd(cli *CLI) *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


internal/cmd/keys.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/infrahq/infra/internal/logging"
 )
 
-const thirtyDays = 30 * (24 * time.Hour)
+const (
+	thirtyDays = 30 * (24 * time.Hour)
+	oneYear    = 366 * (24 * time.Hour) // leap year
```
jmorganca, line 17: `365` might be a better number here to keep it simple / familiar

ui/pages/settings/access-key/add.js
```diff
@@ -22,10 +22,10 @@ import Calendar from '../../../components/calendar'
 const CUSTOM_TITLE = 'custom...'
 
 const EXPIRATION_RATE = [
-  { name: '30 days', value: '720h' },
-  { name: '60 days', value: '1440h' },
-  { name: '90 days', value: '2160h' },
   { name: '1 year', value: '8766h' },
```
jmorganca, line 25: Let's keep it 30 days in the UI since this default is more obvious to the user (whereas in the CLI they see the default after submitting)

jmorganca (approved):

PR #3714, opened 2022-11-24
maintain: remove deprecated machine grant subject

mxyng:
## Summary

machine configuration has been deprecated since v0.13.0 so it's probably safe to delete.


jmorganca (approved):
Great stuff. Thanks for cleaning this up!

PR #3723, opened 2022-11-24
Remove unused JWT nonce field

jmorganca:
Fixes https://github.com/infrahq/infra/issues/2014


BruceMacD (approved):
seems like this should be good now

PR #3724, opened 2022-11-24
Use migrator helper to check for columns

dnephin:
The other query was not working for me. It was failing here: 

> internal/server/data TestMigrations/2022-11-22T14:00/run_again_to_check_idempotency (0.02s
> migrations_test.go:95: assertion failed: error is not nil: rename access key extension deadline: ERROR: column "extension" does not exist (SQLSTATE 42703)

I believe this fails for us locally because our database has multiple schemas. The previous query didn't include `table_schema`, so it would make schemas that were not the one being tested. That's why it passes in CI (because there is only 1 schema).

`migrator.HasColumn` uses `table_schema = CURRENT_SCHEMA()`.

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -950,28 +950,16 @@ func updateAccessKeysTimeoutColumn() *migrator.Migration {
 	return &migrator.Migration{
 		ID: "2022-11-22T14:00",
 		Migrate: func(tx migrator.DB) error {
-			stmt := `
-				SELECT EXISTS (
-					SELECT 1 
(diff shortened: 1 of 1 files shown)


jmorganca (approved):
LGTM, thanks for finding and fixing this. I've been seeing as well in the tests

BruceMacD (approved):

dnephin:
Ohh, I think thiswas failing because my local DB has multiple schemas, where as CI would only have the one. The old code would get the wrong value because it matched a different schema. I updated the description with that detail.

PR #3727, opened 2022-11-24
maintain: display login error

BruceMacD:
## Summary
If OIDC login failed for some reason, this spinner would just run forever. Update the UI to display an error if the login fails.

📸 
Screen Shot 2022-11-24 at 4 05 48 PM

## Checklist

### DIFF
--- a/ui/pages/login/callback.js
+++ b/ui/pages/login/callback.js
@@ -67,7 +67,7 @@ export default function Callback() {
   return (
     <>
       {error ? (
-        <p className='my-1 text-xs text-red-500'>{error}</p>
+        <p className='my-1 text-xs text-red-500'>An error occurred while logging in: {error}</p>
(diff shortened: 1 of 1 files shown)


ui/pages/login/callback.js
```diff
@@ -59,7 +64,15 @@ export default function Callback() {
     return null
   }
 
-  return <Loader className='h-20 w-20' />
+  return (
+    <>
+      {error ? (
+        <p className='my-1 text-xs text-red-500'>{error}</p>
```
jmorganca, line 70:
```suggestion
        <p className='my-1 text-xs text-red-500'>An error occurred while logging in: {error}</p>
```

jmorganca (approved):
Small suggestion on rendering the error otherwise LGTM

PR #3728, opened 2022-11-24
fix: access key api migration

mxyng:
## Summary

Fixing a few bugs related to the access key migration.

* The JSON tag for `ExtensionDeadline` is `inactivityTimeout` so the migration isn't actually doing anything
* The response for `MethodPost` should be `CreateAccessKeyResponse` and not `AccessKey`
* The response for `MethodGet` is missing which should return a list of `AccessKey`s

This bug highlights a limitation of the API migrator: the type of the new response (the input into the migrator) cannot be statically checked. It's possible the data being migrated is completely off

### DIFF
--- a/internal/server/access_keys_test.go
+++ b/internal/server/access_keys_test.go
@@ -315,26 +315,8 @@ func TestAPI_ListAccessKeys(t *testing.T) {
 		})
 	})
 
-	t.Run("latest", func(t *testing.T) {
-		resp := httptest.NewRecorder()
-		// nolint:noctx
-		req := httptest.NewRequest(http.MethodGet, "/api/access-keys", nil)
-		req.Header.Set("Authorization", "Bearer "+adminAccessKey(srv))
-		req.Header.Set("Infra-Version", "0.12.3")
-
-		routes.ServeHTTP(resp, req)
-		assert.Equal(t, resp.Code, http.StatusOK)
-
-		resp1 := &api.ListResponse[api.AccessKey]{}
-		err := json.Unmarshal(resp.Body.Bytes(), resp1)
-		assert.NilError(t, err)
-
-		assert.Assert(t, len(resp1.Items) > 0)
-	})
-
 	t.Run("no version header", func(t *testing.T) {
 		resp := httptest.NewRecorder()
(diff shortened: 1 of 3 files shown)


dnephin (commented):
Nice! I think we can test the change as well by adding a case to `TestAPI_CreateAccessKey`.

The `setup` function in the `testCase` can set the `Infra-Version` header to 0.18.1, which should exercise this code path, and show that the fix works.  

We can check the response by decoding to `map[string]interface{}`. There's a `jsonUnmarshal` test helper that is used by a few other tests in this package, that has an example of this comparison.

internal/server/access_keys_test.go
```diff
@@ -35,10 +37,17 @@ func TestAPI_CreateAccessKey(t *testing.T) {
 		body := tc.setup(t)
 
 		// nolint:noctx
-		req := httptest.NewRequest(http.MethodPost, "/api/access-keys", jsonBody(t, body))
+		req := httptest.NewRequest(http.MethodPost, "/api/access-keys", body)
 		req.Header.Set("Authorization", "Bearer "+adminAccessKey(srv))
 		req.Header.Set("Infra-Version", apiVersionLatest)
 
+		for k := range tc.headers {
+			req.Header.Del(k)
```
jmorganca, line 45: If we're always calling `req.Header.Set` below we can skip this Del call since Set should always [replace](https://pkg.go.dev/net/http#Header.Set) any values for that key

jmorganca (approved):
Minor comment otherwise LGTM

dnephin (approved):
LGTM!

I pushed one more commit that adds a test case for the other endpoint that changed here.

We use a slightly different pattern in other test cases for `Create` endpoints I think. A `body func(t *testing.T) io.Reader` for returning the body, and a `setup(t *testing.T, req *http.Request)` for setting request headers. 

But I think this works well too. We can change it later if we ever have a need to set other fields on the request.

internal/server/access_keys_test.go
mxyng: You're right. I actually thought I used `Header.Add()` which will require the `Header.Del` to ensure a proper override.

Issue #2186, opened 2022-06-03
Proposal: change how we accept secrets from config


dnephin:
### Background

Today we can accept secrets (sensitive values that should not be checked into git) from config, command line flags, or environment vars using a micro-format. The format is documented in https://infrahq.com/docs/install/configure/secrets, but I'll summarize here.

The format is `<secretBackend>:<identifier>`, for example:

* `plaintext:this-is-the-literal-secret`
* `env:FROM_THIS_ENV_VAR`
* `file:/path-to-a-file`
* `kubernetes:name-of-kube-secret/key-in-that-secret`
* `vault:path-to-secret-in-the-mount`
* `awsssm:name-of-secret`
* `awssecretsmanager:arn-or-name-of-secret`

In addition to these secret backends, a user can configure other names that can be used as prefixes. For example to accept a base64 encoded env var, the user would first configure their secret backend with:

```yaml
secrets:
   - kind: env
      name: base64env
      config:
         base64: true`
```

Then they'd specify the secret value using: `base64env:ENV_VAR_TO_READ`.

## Problems

This approach is very flexible and allows a user to provide a secret from many places, but it has some disadvantages.

(1) Indirection

When specifying a secret, the user has to be aware of not only the built-in names for secret providers, but also any that are configured in a separate place in their config.  In the most extreme case of indirection, an env var can specify a value of `env:SOME_OTHER_VAR`. Someone less familiar with this configuration may have a hard time understanding where the secret value is coming from.

(2) The values in the config are not validated until the secret is looked up

Late validation can make it challenging to do GitOps. Ideally it should be possible for our config to be validate statically without a running server. That allows someone to run validate on a config before deploying the new config. Late validation of the parts of the secret (both the backend name before the `:` and the identifier after the `:`) make this impossible.

(3) Obviousness

A user who is not familiar with this format might naturally set a value of `dbPassword: my:password`, and get an error back saying:

> provider "my" not found for field "dbPassword"

A user is required to not only know about this micro-format, but also some of the normalization rules that each provider may perform (currently undocumented).

  * the format of a kube secret with a `/` separating the name and the key in the secret
  * vault names modified to replace `/` and `:` with `_`
  * awsssm keys are modified to replace any non-alphanumeric characters (except for `.`, `-`, and `/`) with `_`

(4) Use in code

[...]

PR #3732, opened 2022-11-25
load google client secret from secrets

BruceMacD:
## Summary
In order to be able to load our client secret from a kubernetes secret in our infrastructure configuration we need the ability to load the Google client secret with the `secrets` library.

## Checklist

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -199,11 +199,15 @@ func New(options Options) (*Server, error) {
 	}
 
 	if options.GoogleClientID != "" {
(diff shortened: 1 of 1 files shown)


dnephin (commented):
Are you sure we need this?

I think we can load this value from the environment variable `INFRA_SERVER_GOOGLE_CLIENT_SECRET`, and helm can set the value of that environment variable from a secret.

Having to remember to call `secrets.GetSecret` on arbitrary `string` values before using them is very error prone, so I think we need to make some changes to this interface, or potentially remove it entirely (#2186).

PR #3735, opened 2022-11-25
Remove generated docs along with `docs` target from Makefile

jmorganca:
This removes the generated docs (as they can just be generated on demand by running the `docgen` package). Some of `CONTRIBUTING.md` was auto-formatted by my editor for consistency


mxyng (approved):
Thanks for cleaning up the docs. We can probably remove the `check-generated` job in `ci-core.yaml` since it doesn't seem to be checking anything anymore

jmorganca:
Thanks for the review. Still seems like that job checks the go modules

PR #3737, opened 2022-11-25
Use standard kubernetes loading instead of custom options in the connector

jmorganca:
This PR changes the connector to load Kubernetes cluster info via standard Kubernetes loading.

The benefits of this are:
1. Less options for `infra connector`
2. Familiar: it's how most tools that connect to a Kubernetes cluster load config info and credentials
3. Support for running the connector out-of-cluster using a local kubeconfig (e.g. via `infra connector ...`)
4. Support for non-token authentication to Kubernetes clusters (probably rare, but good for local development against tools such as Docker Desktop which uses a keypair)

This PR also cleans up some unused code for now that directly require the aws sdk (a big dependency), and adds a few missing CLI arguments to make running the connector (locally and elsewhere) without needing to write out a config file.

### DIFF
--- a/internal/cmd/connector.go
+++ b/internal/cmd/connector.go
@@ -68,11 +68,6 @@ func newConnectorCmd() *cobra.Command {
 	cmd.Flags().String("ca-cert", "", "Path to CA certificate file")
 	cmd.Flags().String("ca-key", "", "Path to CA key file")
 	cmd.Flags().Bool("server-skip-tls-verify", false, "Skip verifying server TLS certificates")
-	cmd.Flags().String("endpoint-addr-host", "", "Connector proxy host")
-	cmd.Flags().String("endpoint-addr-port", "", "Connector proxy port")
-	cmd.Flags().String("addr-http", "", "HTTP address to bind to")
-	cmd.Flags().String("addr-https", "", "HTTPS address to bind to")
-	cmd.Flags().String("addr-metrics", "", "HTTP address to bind to for metrics")
 
 	return cmd
 }
--- a/internal/cmd/connector.go
+++ b/internal/cmd/connector.go
@@ -68,6 +68,11 @@ func newConnectorCmd() *cobra.Command {
 	cmd.Flags().String("ca-cert", "", "Path to CA certificate file")
 	cmd.Flags().String("ca-key", "", "Path to CA key file")
 	cmd.Flags().Bool("server-skip-tls-verify", false, "Skip verifying server TLS certificates")
+	cmd.Flags().String("endpoint-addr-host", "", "Connector proxy host")
+	cmd.Flags().String("endpoint-addr-port", "", "Connector proxy port")
+	cmd.Flags().String("addr-http", "", "HTTP address to bind to")
+	cmd.Flags().String("addr-https", "", "HTTPS address to bind to")
+	cmd.Flags().String("addr-metrics", "", "HTTP address to bind to for metrics")
 
 	return cmd
 }
--- a/internal/cmd/connector_test.go
+++ b/internal/cmd/connector_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"path"
 	"reflect"
 	"sort"
 	"strings"
@@ -26,6 +27,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/certs"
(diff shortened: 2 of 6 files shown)


internal/cmd/connector_test.go
```diff
@@ -72,6 +75,15 @@ func TestConnector_Run(t *testing.T) {
 	ctx := context.Background()
 	runAndWait(ctx, t, srv.Run)
 
+	kubeconfig := path.Join(dir, "kubeconfig")
```
jmorganca, line 78: This is probably the only downside to using the `client-go` loading: a bit more pain in writing tests since we have to generate a full KubeConfig. It's not too bad

internal/cmd/connector.go
```diff
@@ -68,6 +68,11 @@ func newConnectorCmd() *cobra.Command {
 	cmd.Flags().String("ca-cert", "", "Path to CA certificate file")
 	cmd.Flags().String("ca-key", "", "Path to CA key file")
 	cmd.Flags().Bool("server-skip-tls-verify", false, "Skip verifying server TLS certificates")
+	cmd.Flags().String("endpoint-addr-host", "", "Connector proxy host")
+	cmd.Flags().String("endpoint-addr-port", "", "Connector proxy port")
+	cmd.Flags().String("addr-http", "", "HTTP address to bind to")
+	cmd.Flags().String("addr-https", "", "HTTPS address to bind to")
+	cmd.Flags().String("addr-metrics", "", "HTTP address to bind to for metrics")
```
mxyng, line 75: What are these extra command line options for?

mxyng (approved):
LGTM though @dnephin should also review

BruceMacD (approved):
this looks good to me also

internal/cmd/connector.go
dnephin: Might be nice to avoid these, especially if we're looking at remove the `addr-` ones with the proxy.

We can use the equivalent env vars if we want a way to set these without the config file.

internal/cmd/connector_test.go
```diff
@@ -72,6 +75,15 @@ func TestConnector_Run(t *testing.T) {
...
+	os.Setenv("KUBECONFIG", kubeconfig)
```
dnephin, line 79:
```suggestion
	t.Setenv("KUBECONFIG", kubeconfig)
```

So that this value is reset when the test ends.

internal/connector/connector.go
```diff
@@ -46,8 +46,6 @@ type Options struct {
 	// This value is sent to the infra API server to update the
 	// Destination.Connection.URL.
 	EndpointAddr types.HostPort
-
-	Kubernetes KubernetesOptions
```
dnephin, line 50: The `KubernetesOptions` struct is no longer referenced, you can delete it as well.

internal/kubernetes/kubernetes.go
```diff
@@ -32,16 +24,12 @@ type Kubernetes struct {
...
-		k.Config.BearerToken = authToken
-		return k, nil
-	}
+func NewKubernetes() (*Kubernetes, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.WarnIfAllMissing = false
 
-	config, err := rest.InClusterConfig()
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+	config, err := kubeConfig.ClientConfig()
```
dnephin, line 32: I thought this only read the `kubectl` config file. Don't we still need `rest.InClusterConfig` for running this in a pod?

dnephin (commented):
Nice! I like the cleanup, a few questions about how this works. 

If I understand these changes correctly it means that:

1. We will no longer attempt to lookup a destination name. It will either use what's provided by helm values (hopefully), or fall back to the hash of the CA.
2. Outside of a kube pod we would configure the connector by reading the standard kubectl config. When I tried to get this working in #3350 I think I wasn't able to read in the auth token, but I guess my config was not setup correctly. Did you get that working?

internal/cmd/connector_test.go
jmorganca: Fantastic. Thanks!

internal/connector/connector.go
jmorganca: 💥

internal/kubernetes/kubernetes.go
jmorganca: I will double check and manual test ahead of any merge, but should be okay: https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/clientcmd/merged_client_builder.go#L120

internal/cmd/connector.go
jmorganca: Sounds good. I'll omit them for now.

jmorganca:
> We will no longer attempt to lookup a destination name. It will either use what's provided by helm values (hopefully), or fall back to the hash of the CA.

Correct.

> Outside of a kube pod we would configure the connector by reading the standard kubectl config. When I tried to get this working in https://github.com/infrahq/infra/pull/3350 I think I wasn't able to read in the auth token, but I guess my config was not setup correctly. Did you get that working?

Yep this will read any `AuthInfo` in kubeconfigs it finds, or if in-cluster will read the mounted token

internal/kubernetes/kubernetes.go
dnephin: Awesome! That is definitely easier

dnephin (approved):

PR #3738, opened 2022-11-25
feat: fail when client is ahead of server version

BruceMacD:
## Summary
When the CLI version is ahead of the server version fail fast.

```
$ infra groups list
Error: client version (0.18.1) is ahead of server version (0.18.0), download the CLI version that matches your server, reference https://infrahq.com/docs/reference/self-hosting#cli-versions
```

## Checklist

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/ssoroka/slice"
 
 	"github.com/infrahq/infra/internal/logging"
@@ -464,4 +465,36 @@ func HandleConnError(err error) error {
 	}
 
 	return nil
+}
+
+func ServerCompatible(ctx context.Context, client *Client) error {
+	srv, err := client.GetServerVersion(ctx)
+	if err != nil {
+		logging.L.Debug().Err(err).Msg("check server version")
+		return nil // result inconclusive
+	}
+	ahead, err := versionAhead(client.Version, srv.Version)
+	if err != nil {
+		logging.L.Debug().Err(err).Msg("failed to compare client version to server")
+		return nil // result inconclusive
+	}
+	if ahead {
+		return fmt.Errorf("client version (%s) is ahead of server version (%s), please upgrade your server", apiVersion, srv.Version)
+	}
+	return nil
+}
+
+// versionAhead returns if the current semantic version is ahead of the comparison version
+func versionAhead(current, compare string) (bool, error) {
+	v, err := semver.NewVersion(current)
+	if err != nil {
+		return false, fmt.Errorf("could not create semantic current version: %w", err)
+	}
+
+	c, err := semver.NewVersion(compare)
+	if err != nil {
+		return false, fmt.Errorf("could not create semantic comparison version: %w", err)
+	}
+
+	return v.GreaterThan(c), nil
 }
--- a/api/client.go
+++ b/api/client.go
@@ -479,7 +479,7 @@ func ServerCompatible(ctx context.Context, client *Client) error {
 		return nil // result inconclusive
 	}
 	if ahead {
-		return fmt.Errorf("client version (%s) is ahead of server version (%s), please upgrade your server", apiVersion, srv.Version)
+		return fmt.Errorf("client version (%s) is ahead of server version (%s)", apiVersion, srv.Version)
 	}
 	return nil
 }
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -31,31 +30,6 @@ func Run(ctx context.Context, args ...string) error {
 	return cmd.ExecuteContext(ctx)
 }
 
-func mustBeLoggedIn() error {
-	if _, ok := os.LookupEnv("INFRA_ACCESS_KEY"); ok {
-		// user doesn't need to log in if supplying an access key
-		return nil
-	}
-
-	config, err := currentHostConfig()
-	if err != nil {
-		if errors.Is(err, ErrConfigNotFound) {
-			return Error{Message: "Not logged in; run 'infra login' before running this command"}
-		}
-		return fmt.Errorf("getting host config: %w", err)
-	}
-
-	// Check expired before checking isLoggedin, since if we check isLoggedIn first, we will never know if it's expired
-	if config.isExpired() {
-		return Error{Message: "Session expired; run 'infra login' to start a new session"}
-	}
-
-	if !config.isLoggedIn() {
-		return Error{Message: "Not logged in; run 'infra login' before running this command"}
-	}
-	return nil
-}
-
 func printTable(data interface{}, out io.Writer) {
 	table := tableprinter.New(out)
 
@@ -73,8 +47,14 @@ func printTable(data interface{}, out io.Writer) {
 	table.Print(data)
 }
 
-// Creates a new API Client from the current config
-func defaultAPIClient() (*api.Client, error) {
+type APIClientOpts struct {
+	Host      string
+	AccessKey string
+	Transport *http.Transport
+}
+
+// Creates API Client options from the current config
+func defaultClientOpts() (*APIClientOpts, error) {
 	config, err := currentHostConfig()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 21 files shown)


internal/cmd/cmd.go
```diff
@@ -294,3 +306,50 @@ Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
...
+		fmt.Println("Warning: Your client version is ahead of the server you are connected to.")
+		fmt.Printf("Client: %s\n", api.Version)
+		fmt.Printf("Server: %s\n", server.Version)
+		fmt.Println()
+	}
+}
+
+// versionAhead returns if the current semantic version is comparison version
+func versionAhead(current, compare string) (bool, error) {
+	if current == compare {
```
BruceMacD, line 338: Without this check equivalent versions return true, I believe its due to appending the `-0`.

cc: @jmorganca do we need that `-0` check? If not, this check can go.

internal/cmd/cmd.go
```diff
@@ -31,6 +33,19 @@ func Run(ctx context.Context, args ...string) error {
...
+func authenticatedPreRun() error {
+	if err := mustBeLoggedIn(); err != nil {
+		return err
+	}
+	api, err := defaultAPIClient()
+	if err != nil {
+		logging.L.Debug().Err(err).Msg("failed to create api client to check version")
+		return nil
+	}
+	warnVersion(context.Background(), api)
```
jmorganca, line 45: I believe we'd be better to put this in the API client so it catches all calls (authenticated, unauthenticated). This check could happen once per instance of the api client (e.g. on the first call) which should almost map 1-1 to a command.

internal/cmd/cmd.go
```diff
@@ -294,3 +306,50 @@ Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
 Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
 `
 }
+
+// warnVersion checks if the client version is ahead of the server and logs a warning if so
+func warnVersion(ctx context.Context, api *api.Client) {
```
jmorganca, line 311: We'll want to fail outright to avoid unintended consequences that could happen from only warning (e.g. ugly errors)

internal/cmd/cmd.go
BruceMacD: This will result in a lot of extra logging, many CLI commands are not 1-1, for example removing a group means getting listing the groups then deleting by IDs. Being in a pre-run also means that we don't have to add this unrelated logging to the client code which is used by connectors also, and ensures it is only called once.

dnephin: I don't think we have any place to put this in the API client because we don't really use a constructor for it, but we could put it in the `apiClient` func, which is called by `defaultAPIClient`. All the CLI commands should use that function

internal/cmd/cmd.go
jmorganca: It was important for "pre-release" versions, but I don't think it's applicable here.

jmorganca (approved):
Awesome work with this one!

mchiang0610:
adding note here as well per #3754 

https://infrahq.com/docs/reference/self-hosting#cli-versions

This is where we can point users to.

api/client.go
```diff
@@ -113,7 +117,15 @@ func (c *Client) buildRequest(
 	return req, nil
 }
 
-func request[Res any](client Client, req *http.Request) (*Res, error) {
+func request[Res any](ctx context.Context, client Client, req *http.Request) (*Res, error) {
+	// if this is a request to a self-hosted infra instance check that the client is not ahead of the server version
+	// this could result in unexpected behavior
+	// skip this check if it's a version request since that would result in infinitely calling request
+	if !client.IgnoreVersion && !strings.HasSuffix(req.URL.Host, ".infrahq.com") && !strings.HasSuffix(req.URL.Path, "/api/version") {
```
mxyng, line 124: Might be more readable as a switch/case here

```
switch {
case client.IgnoreVersion:
case strings.HasSuffix(req.URL.Host, "infrahq.com"):
case strings.HasSuffix(req.URL.Path, "/api/version"):
default:
    if err := serverIncompatible(ctx, client); err != nil {
        return nil, err
    }
}
```

api/client.go
```diff
@@ -457,3 +469,35 @@ func HandleConnError(err error) error {
 
 	return nil
 }
+
+func serverIncompatible(ctx context.Context, client Client) error {
```
mxyng, line 473: Would be better to name this `serverCompatible`

api/client.go
```diff
@@ -457,3 +469,35 @@ func HandleConnError(err error) error {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to check api semantic version: %w", err)
+	}
+
+	v, err := semver.NewVersion(current)
+	if err != nil {
+		return false, fmt.Errorf("could not create semantic server version: %w", err)
+	}
+
+	return c.Check(v), nil
```
dnephin, line 502: Could we use https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.GreaterThan instead of `NewConstraint` ?

internal/cmd/cmd.go
```diff
@@ -73,8 +73,8 @@ func printTable(data interface{}, out io.Writer) {
 	table.Print(data)
 }
 
-// Creates a new API Client from the current config
-func defaultAPIClient() (*api.Client, error) {
+// Creates a new API Client from the current config and command flags
+func defaultAPIClient(flags *pflag.FlagSet) (*api.Client, error) {
```
mxyng, line 77: This feels heavy handed for a single flag. It also precludes using environment variables to set `ignore-version`

internal/cmd/cmd.go
```diff
@@ -101,10 +101,18 @@ func defaultAPIClient() (*api.Client, error) {
 		server = envServer
 	}
 
-	return apiClient(server, accessKey, httpTransportForHostConfig(config)), nil
+	ignoreVersion := false
+	if flags != nil {
+		ignoreVersion, err = flags.GetBool("ignore-version")
```
mxyng, line 106: `ignore-version` doesn't convey the right message, maybe `skip-version-check`?

internal/cmd/cmd.go
BruceMacD: I did it this way because we would have to parse the flag everywhere otherwise. Added the environment variable though.

internal/cmd/destinations.go
```diff
@@ -47,7 +47,7 @@ func newDestinationsListCmd(cli *CLI) *cobra.Command {
 		Short:   "List connected destinations",
 		Args:    NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			client, err := defaultAPIClient()
+			client, err := defaultAPIClient(cmd.Flags())
```
jmorganca, line 50: Instead of passing `cmd.Flags()` would it make sense to just have `cli` include this state? E.g.

```
client, err := defaultAPIClient(cli.IgnoreVersion)
```

Ideally it's optional somehow (perhaps by making `defaultAPIClient` take an options struct). And then `rootPreRun` in `cmd.go` could parse the value and include it in the `CLI` struct that's passed around

internal/cmd/cmd.go
```diff
@@ -101,11 +103,25 @@ func defaultAPIClient() (*api.Client, error) {
...
+			return nil, fmt.Errorf("failed to read cmd flags for client: %w", err)
+		}
+	}
+
+	return apiClient(server, accessKey, httpTransportForHostConfig(config), ignoreVersion)
 }
 
-func apiClient(host string, accessKey string, transport *http.Transport) *api.Client {
-	return &api.Client{
+func apiClient(host string, accessKey string, transport *http.Transport, ignoreVersion bool) (*api.Client, error) {
```
jmorganca, line 123: Perhaps to help, we can merge this and above into a single constructor that takes an options struct:

```
func NewAPIClient(&APIClientOptions{
  Host: "...",
  accessKey: "...",
  FromConfig: true,
  IgnoreVersion: true
})
```

jmorganca: Even better, this would be simple & flat, without any constructors:

```
// uses default config
client := api.Client{}
```

```
// ignores version
client := api.Client{IgnoreVersion: true}
```

```
// custom for login
client := api.Client{URL}
```

internal/cmd/destinations.go
jmorganca: With above comment, we could have something like:

```
client := api.Client{IgnoreVersion: cli.IgnoreVersion}
```

api/client.go
```diff
@@ -457,3 +458,35 @@ func HandleConnError(err error) error {
...
+		logging.L.Debug().Err(err).Msg("check server version")
+		return nil // result inconclusive
+	}
+	ahead, err := versionAhead(client.Version, srv.Version)
+	if err != nil {
+		logging.L.Debug().Err(err).Msg("failed to compare client version to server")
+		return nil // result inconclusive
+	}
+	if ahead {
+		return fmt.Errorf("client version (%s) is ahead of server version (%s), please upgrade your server", apiVersion, srv.Version)
```
pdevine, line 474: Since it's kinda unrealistic to tell the user to upgrade their server, we should also include a note which says that they can downgrade their client, along with instructions on how to do that.

pdevine (approved):
Overall looks fine. I like making the version check more generic and getting rid of the device flow specific parts.

api/client.go
BruceMacD: Added the link:
```
Error: client version (0.18.1) is ahead of server version (99.99.99999), download the CLI version that matches your server, reference https://infrahq.com/docs/reference/self-hosting#cli-versions
```

Let me know what you think of the wording.

jmorganca: @BruceMacD can you edit the link to https://infrahq.com/docs/messages/cli-versions as a "short link" and I'll make sure the redirect goes to above

BruceMacD: ^ updated

dnephin (approved):
LGTM!

jmorganca (approved):
LGTM! Nice work on this

PR #3739, opened 2022-11-26
Don't create two access keys for device flow login

jmorganca:
This PR removes the intermediate access key created during a "headless" device flow authentication. The goal is to lower our "access key sprawl".

Instead, an access key is only created in the "last leg" of the device flow process. When a user approves a device flow request, Infra track's the ID of the currently signed in user to re-create a new access key with the same fields such as provider ID.

Left to do:
- [x] Add an access key scope for approving device flow requests. That way, a key resulting from the device flow won't be able continually approve new keys (extending the expiration indefinitely).


internal/server/models/deviceflowauthrequest.go
```diff
@@ -11,16 +11,8 @@ type DeviceFlowAuthRequest struct {
...
-	AccessKeyToken EncryptedAtRest
-
 	ExpiresAt time.Time
 
-	// AccessKey will be populated by some queries, but is never used on writes.
-	AccessKey *AccessKey `db:"-"`
-
-	// Organization will be populated by some queries, but is never used on writes.
-	Organization *Organization `db:"-"`
+	// ApprovedWith is the access key that approved this request
```
BruceMacD, line 16:
```suggestion
	// ApprovedWith is the ID of the access key that approved this request
```

This looks alarming otherwise.

internal/server/deviceflow.go
```diff
@@ -79,29 +79,94 @@ func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.DeviceFlowStatusReque
...
+		InactivityExtension: a.server.options.SessionInactivityTimeout,
+		Scopes:              models.CommaSeparatedStrings{models.ScopeAllowCreateAccessKey},
+	}
+
+	bearer, err := data.CreateAccessKey(rctx.DBTxn, accessKey)
+	if err != nil {
+		return nil, fmt.Errorf("%w: device flow create new key: %v", internal.ErrUnauthorized, err)
+	}
+
+	identity, err := data.GetIdentity(rctx.DBTxn, data.GetIdentityOptions{ByID: ak.IssuedFor})
```
BruceMacD, line 118: nit: might make more sense to do this before creating the access key. The end result is the same in both cases though where we would rollback the transaction on failure.

BruceMacD (commented):
I dont see any problems with this, but I would like @ssoroka to give it a look before approval where it's mostly his implementation.

internal/server/data/migrations_test.go
```diff
@@ -876,6 +876,12 @@ INSERT INTO providers(id, name) VALUES (12345, 'okta');
 				// schema changes are tested with schema comparison
 			},
 		},
+		{
+			label: testCaseLine(updateDeviceFlowAuthRequestsToUseApprovedByKeyID().ID),
+			expected: func(t *testing.T, db WriteTxn) {
+				// schema changes are tested with schema comparison
```
jmorganca, line 882: Note: a migration test is needed / coming

internal/cmd/login.go
```diff
@@ -154,116 +156,57 @@ func login(cli *CLI, options loginCmdOptions) error {
...
 
 	if loginRes.PasswordUpdateRequired {
 		fmt.Fprintf(cli.Stderr, "  Your password has expired. Please update your password.\n")
 
-	PROMPTLOGIN:
-		password, err := promptSetPassword(cli, loginReq.PasswordCredentials.Password)
-		if err != nil {
-			return err
-		}
+		for {
```
pdevine, line 189: I _think_ this is fine. It's definitely better w/ the `for` loop here than the `goto`. I'm trying to think if we would ever break out of here on too many attempts (I mean, the user can always CTRL-C, so we have to handle the case anyway).

internal/cmd/login.go
```diff
@@ -289,9 +232,24 @@ func loginToInfra(cli *CLI, lc loginClient, loginReq *api.LoginRequest, noAgent
 	}
 
 	fmt.Fprintf(cli.Stderr, "  Logged in as %s\n", termenv.String(loginRes.Name).Bold().String())
+
+	if options.InjectUserSSHConfig {
```
pdevine, line 236: is this bleed from @dnephin 's changes?

internal/cmd/login.go
```diff
@@ -289,9 +232,24 @@ func loginToInfra(cli *CLI, lc loginClient, loginReq *api.LoginRequest, noAgent
...
+		return updateUserSSHConfig(cli)
+	}
+
 	return nil
 }
 
+func equalHosts(x, y string) bool {
+	if x == y {
```
pdevine, line 244: I don't think you need this clause because the `strings.TrimPrefix()` part below should catch it anyway.

internal/server/data/deviceflowauthrequest.go
```diff
@@ -92,24 +88,29 @@ func GetDeviceFlowAuthRequest(tx ReadTxn, opts GetDeviceFlowAuthRequestOptions)
...
-func SetDeviceFlowAuthRequestAccessKey(tx WriteTxn, dfarID uid.ID, accessKey *models.AccessKey) error {
-	_, err := tx.Exec(`
-		UPDATE device_flow_auth_requests
-		SET 
-			access_key_id = ?,
-			access_key_token = ?
-		WHERE id = ?
-	`, accessKey.ID, models.EncryptedAtRest(accessKey.Token()), dfarID)
+func DeleteExpiredDeviceFlowAuthRequests(tx WriteTxn) error {
+	query := querybuilder.New("UPDATE device_flow_auth_requests")
```
pdevine, line 92: Shouldn't this be `DELETE FROM device_flow_auth_requests`...? I thought we wanted to hard delete here so that the table doesn't keep filling up (i.e. soft delete in `DeleteDeviceFlowAuthRequest()` and then reap in this function).

internal/server/deviceflow.go
```diff
@@ -67,9 +67,14 @@ retry:
 // flow login. The response status can be pending, rejected, expired, or confirmed.
 func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.DeviceFlowStatusRequest) (*api.DeviceFlowStatusResponse, error) {
 	rctx := getRequestContext(c)
-	dfar, err := access.FindDeviceFlowAuthRequest(rctx, req.DeviceCode)
+	db := rctx.DBTxn
+
+	dfar, err := data.GetDeviceFlowAuthRequest(db, data.GetDeviceFlowAuthRequestOptions{ByDeviceCode: req.DeviceCode})
```
pdevine, line 72: Why skip the access layer here? I think the handler will end up just passing through (i.e. the call here is unauthenticated), but we probably want to keep call through access to maintain the idiom (since everything else calls through it).

internal/server/deviceflow.go
```diff
@@ -79,29 +84,85 @@ func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.DeviceFlowStatusReque
...
+
+	// Delete the request so it can't be claimed twice
+	err = data.DeleteDeviceFlowAuthRequest(rctx.DBTxn, dfar.ID)
+	if err != nil {
+		return nil, fmt.Errorf("%w: device flow delete auth request: %v", internal.ErrUnauthorized, err)
+	}
+
 	return &api.DeviceFlowStatusResponse{
-		Status:     "pending",
+		Status:     "confirmed",
```
pdevine, line 152: small nit: this should probably be a constant

internal/server/deviceflow.go
```diff
@@ -79,29 +84,85 @@ func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.DeviceFlowStatusReque
 		}, nil
 	}
 
-	if dfar.AccessKey != nil {
+	if !dfar.Approved() {
 		return &api.DeviceFlowStatusResponse{
-			Status:     "confirmed",
+			Status:     "pending",
```
pdevine, line 89: this should probably be a constant (there's also "expired" above here)

internal/cmd/login.go
jmorganca: No, `Login` and `loginToInfra` have been merged. Since device flow isn't a login method I had to restructure the logic around getting the login access key

internal/server/data/deviceflowauthrequest.go
jmorganca: Yes good catch! Thanks. This must have been from in a bad rebase

internal/cmd/login.go
jmorganca: 💯  – this code was shuffled around and not added by this PR (unfortunately `git` didn't catch it). I went ahead and included that fix as part of this change!

internal/server/deviceflow.go
```diff
@@ -112,29 +171,10 @@ func (a *API) ApproveDeviceAdd(c *gin.Context, req *api.ApproveDeviceFlowRequest
...
-		ExpiresAt:           rctx.Authenticated.AccessKey.ExpiresAt,
-		InactivityExtension: 30 * days,
-		InactivityTimeout:   time.Now().UTC().Add(30 * days),
-	}
-
-	_, err = access.CreateAccessKey(c, accessKey)
-	if err != nil {
-		return nil, err
+	if dfar.Approved() {
+		return nil, internal.ErrNotFound
```
ssoroka, line 175: this removes endpoint idempotency

ssoroka (changes requested):
How does this solve the duplicate access key? Is that coming from somewhere else?

I think this is the wrong approach for reasons posted in comments.

internal/server/deviceflow.go
```diff
@@ -64,46 +64,105 @@ retry:
...
+	setCookie(c, cookie)
+
+	a.t.User(accessKey.IssuedFor.String(), user.Name)
+	a.t.OrgMembership(accessKey.OrganizationID.String(), accessKey.IssuedFor.String())
+	a.t.Event("login", accessKey.IssuedFor.String(), accessKey.OrganizationID.String(), Properties{"method": "deviceflow"})
+
+	// Update the request context so that logging middleware can include the userID
+	rctx.Authenticated.User = user
+	c.Set(access.RequestContextKey, rctx)
+
```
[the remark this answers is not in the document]
dnephin: These 2 lines will make it so that the log line emitted for this request includes a `userID`. Without this the userID is not set because the request is not authenticated until the device code is approved.

internal/server/deviceflow.go
```diff
@@ -64,46 +64,105 @@ retry:
...
+	accessKey := &models.AccessKey{
+		IssuedFor:     user.ID,
+		IssuedForName: user.Name,
+
+		// Share the same provider ID that was used to approve
+		ProviderID:          dfar.ProviderID,
+		ExpiresAt:           time.Now().UTC().Add(a.server.options.SessionDuration),
+		InactivityTimeout:   time.Now().UTC().Add(a.server.options.SessionInactivityTimeout),
+		InactivityExtension: a.server.options.SessionInactivityTimeout,
+		Scopes:              models.CommaSeparatedStrings{models.ScopeAllowCreateAccessKey},
```
ssoroka, line 110: device keys shouldn't be allowed to create more keys

internal/server/deviceflow.go
```diff
@@ -64,46 +64,105 @@ retry:
...
+	}
+
+	bearer, err := data.CreateAccessKey(rctx.DBTxn, accessKey)
+	if err != nil {
+		return nil, fmt.Errorf("%w: creating new access key: %v", internal.ErrUnauthorized, err)
+	}
```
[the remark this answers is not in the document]
ssoroka: clarification: ApproveDeviceFlow should fail if the approver's token is missing models.ScopeAllowCreateAccessKey

internal/server/deviceflow.go
```diff
@@ -112,29 +171,10 @@ func (a *API) ApproveDeviceAdd(c *gin.Context, req *api.ApproveDeviceFlowRequest
...
-	_, err = access.CreateAccessKey(c, accessKey)
-	if err != nil {
-		return nil, err
+	if dfar.Approved() {
+		return nil, internal.ErrNotFound
 	}
 
-	// update device flow auth request with the access key id
-	err = access.SetDeviceFlowAuthRequestAccessKey(rctx, dfar.ID, accessKey)
+	err = data.ApproveDeviceFlowAuthRequest(rctx.DBTxn, dfar.ID, rctx.Authenticated.User.ID, rctx.Authenticated.AccessKey.ProviderID)
```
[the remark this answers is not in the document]
ssoroka: clarification: ApproveDeviceFlow should fail if the current key is missing models.ScopeAllowCreateAccessKey

internal/server/deviceflow.go
dnephin: That seems like it's going to be a problem. Previously we were able to login from the CLI and create access keys. If device flow doesn't count as a login, then there's no way for someone to truly login via the CLI.

jmorganca: I've added a new scope for access keys that allows them to approve device flow requests. The issue in the PR was not that device flow keys could create access keys (they are equivalent to login keys), but that they can continue to "renew" themselves by approving subsequent device flow requests.

internal/server/deviceflow.go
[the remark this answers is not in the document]
pdevine: This is the same thing that happens in `server/authn/authn_method.go` inside of `Login()`, which is also an unauthenticated endpoint. The device flow is essentially the same thing as a login, only the _credential_ is the device code itself.

If we wanted to refactor this in the future, device flow really should just be another login method inside of `authn/`. I think the problem though is that code is a little confusing right now.

internal/server/models/access_key.go
```diff
@@ -13,8 +13,9 @@ var (
 )
 
 const (
-	ScopePasswordReset        string = "password-reset"
-	ScopeAllowCreateAccessKey string = "create-key"
+	ScopePasswordReset                 string = "password-reset"
+	ScopeAllowCreateAccessKey          string = "create-key"
+	ScopeAllowApproveDeviceFlowRequest string = "approve-device-flow-request"
```
dnephin, line 18: These don't seem like they are really different scopes. Are we assigning these two scopes in all the same cases?

Device flow is effectively a way of getting an access key in the CLI. Can we continue to use `ScopeAllowCreateAccessKey` for approving of the request?

internal/server/deviceflow.go
pdevine: On second thought, this should skip the access layer because it's essentially the same thing as the login handler (which also skips the access layer).

internal/server/deviceflow.go
pdevine: We should probably be idempotent here, but ultimately once the polling code does the login, this record ends up being deleted and the approval will fail. So really there is a <= 5s window where this would be an issue.

internal/server/routes.go
```diff
@@ -135,6 +130,11 @@ func (s *Server) GenerateRoutes() Routes {
 
 	add(a, noAuthnWithOrg, http.MethodGet, "/.well-known/jwks.json", wellKnownJWKsRoute)
 
+	// Device flow
+	post(a, noAuthnNoOrg, "/api/device", a.StartDeviceFlow)
+	post(a, noAuthnWithOrg, "/api/device/status", a.GetDeviceFlowStatus)
```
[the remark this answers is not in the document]
pdevine: You would need to know this to get the provider/user (which is different than the old implementation).

internal/server/deviceflow.go
dnephin: Definitely agree! I would go even further to say that all endpoints that do not require authorization should skip the access layer.

If the purpose of the access layer is "to notice when something is missing proper authorization", then a function in the `access` package that does not check authorization is actually defeating the purpose of this package.

internal/server/models/access_key.go
jmorganca: They are close, but access keys resulting from logging with device flow should be able to:
1. Create an access key (since the user ran `infra login`), they should be able to run `infra keys add`
2. But not approve new device flows, since this would mean the user could continue to approve their own device flow requests

dnephin: > But not approve new device flows, since this would mean the user could continue to approve their own device flow requests

Why is that a problem?

pdevine (approved):
lgtm

BruceMacD (approved):
Tested and validated that this revokes sessions on identity removal from the IDP

internal/cmd/login.go
```diff
@@ -153,85 +156,57 @@ func login(cli *CLI, options loginCmdOptions) error {
...
-		}
+		loginRes, err = lc.APIClient.Login(ctx, &api.LoginRequest{
+			PasswordCredentials: &api.LoginRequestPasswordCredentials{
+				Name:     options.User,
+				Password: options.Password,
+			},
+		})
+		if err != nil {
+			if api.ErrorStatusCode(err) == http.StatusUnauthorized || api.ErrorStatusCode(err) == http.StatusNotFound {
+				return &LoginError{Message: "your username or password may be invalid"}
```
dnephin, line 171: Not a blocker for merging this, but do you know what causes the `NotFound`? Seems like maybe we should be handling that as `StatusUnauthorized` so that someone can't detect when a user exists or not.

internal/server/data/migrations.go
```diff
@@ -1052,3 +1053,19 @@ func makeIdxEmailsProvidersUnique() *migrator.Migration {
...
+func deviceFlowAuthRequestsAddUserIDProviderID() *migrator.Migration {
+	return &migrator.Migration{
+		ID: "2022-12-03T13:14",
+		Migrate: func(tx migrator.DB) error {
+			_, err := tx.Exec(`
+				ALTER TABLE device_flow_auth_requests
+					DROP COLUMN if exists access_key_id,
+					DROP COLUMN if exists access_key_token,
+					ADD COLUMN IF NOT EXISTS user_id bigint,
+					ADD COLUMN IF NOT EXISTS provider_id bigint;
```
dnephin, line 1066: What happens to existing device flow records when we run this migration? I think they will break because user_id and provider_id are null.

I guess since these are short lived and not many people are using it right now that it's ok, but still important to know. We should expect any active device flow request to fail when this is deployed to prod. Might be good to try and deploy during off-peak.

internal/server/deviceflow.go
```diff
@@ -64,77 +64,112 @@ retry:
...
+// flow login. The response status can be pending, expired, or confirmed.
 func (a *API) GetDeviceFlowStatus(c *gin.Context, req *api.DeviceFlowStatusRequest) (*api.DeviceFlowStatusResponse, error) {
 	rctx := getRequestContext(c)
-	dfar, err := access.FindDeviceFlowAuthRequest(rctx, req.DeviceCode)
+
+	dfar, err := data.GetDeviceFlowAuthRequest(rctx.DBTxn, data.GetDeviceFlowAuthRequestOptions{ByDeviceCode: req.DeviceCode})
 	if err != nil {
-		return nil, err
+		if errors.Is(err, internal.ErrNotFound) {
+			return nil, err
```
dnephin, line 74: This potentially allows someone to scan for device flow requests. Rate limiting should help mitigate that, but a distributed attack could still make lots of guesses. I guess since we delete it once approved it doesn't help them much.

Do we need to differentiate this error, or can we always return unauthorized?

internal/server/routes.go
dnephin, line 135: I think we could use the org_id from either the user or provider that we have IDs for, right? I'm not sure if that's necessary for this PR, but it seems like we don't necessarily need to get the org from the request domain.

However it would require relaxing our requirement that `data.GetIdentity` sets an orgID in the transaction metadata. We'd need an option to `GetIdentity` to tell it to skip the org requirement.

internal/server/deviceflow.go
```diff
@@ -64,77 +64,112 @@ retry:
...
-const days = 24 * time.Hour
-
-func (a *API) ApproveDeviceAdd(c *gin.Context, req *api.ApproveDeviceFlowRequest) (*api.EmptyResponse, error) {
+func (a *API) ApproveDeviceFlow(c *gin.Context, req *api.ApproveDeviceFlowRequest) (*api.EmptyResponse, error) {
 	rctx := getRequestContext(c)
-	dfar, err := access.FindDeviceFlowAuthRequestForApproval(rctx, strings.Replace(req.UserCode, "-", "", 1))
+
+	if !rctx.Authenticated.AccessKey.Scopes.Includes(models.ScopeAllowCreateAccessKey) {
+		// require the device flow scope to approve access keys provided by login & signup keys, but not
+		// keys resulting from device flow itself
```
dnephin, line 157:
```suggestion
		// require the create access keys scope, because this approval allows the creation of an access key
```

internal/server/deviceflow.go
```diff
@@ -64,77 +64,112 @@ retry:
...
+		return nil, fmt.Errorf("%w: access key missing scope '%s'", internal.ErrUnauthorized, models.ScopeAllowCreateAccessKey)
```
dnephin, line 158: Just a note that we end up hiding all of these messages from the user: https://github.com/infrahq/infra/blob/8c6e5f361e561de77c6de330f6c77499d9d54140/internal/server/errors.go#L44-L47

We should find a way to allow us to send messages that we have written explicitly for these errors. I can create an issue for this.

internal/server/deviceflow_test.go
```diff
@@ -69,48 +92,54 @@ func TestDeviceFlow(t *testing.T) {
...
+	var loginuser api.User
+	resp = request(t, "GET", "http://"+org.Domain+"/api/users/self", statusResp.LoginResponse.AccessKey, nil, &loginuser)
+	assert.Equal(t, loginuser.Name, user.Name)
+
+	// valid access key
 	newKey := statusResp.LoginResponse.AccessKey
 	assert.Assert(t, len(newKey) > 0)
 	assert.Assert(t, strings.Contains(newKey, "."))
 
+	// TODO: add test for approving another device flow request
```
dnephin, line 141: Is this TODO for a follow up? Or did you want to add it here?

dnephin (commented):
LGTM! Some minor suggestions about the error code, but probably fine as follow ups.

There's a TODO in the test that might need to be removed or implemented.

internal/cmd/login.go
jmorganca: This was moved from elsewhere, but `/api/login` should never return 404 for user discovery reasons and I don't think the endpoint ever does, so I'll remove that check.

internal/server/deviceflow.go
dnephin: Created #3837

internal/server/deviceflow.go
jmorganca: Great catch. Let's just return 401 here, as all public endpoints should.

internal/server/routes.go
jmorganca: Indeed. Went with this for now given `GetIdentity` requires the org id in the transaction

internal/server/data/migrations.go
jmorganca: This is a great catch. I'll add to the migration to expire existing requests immediately so any mid-device-flow users see a message and can restart the flow.

The alternative would be to populate `user_id` and `provider_id` from the access key, but these are short-lived and it would be better that we get the correct provider_id (which is not the case with the transient access key in the current model).

dnephin (approved):
Good idea to expire all the active requests. That should be a smoother upgrade experience.

PR #3753, opened 2022-11-28
Add parseDuration that supports years, weeks, and days

### LINKED ISSUE #3750: cli: support parsing time.Duration of days, weeks, and years
### Problem

The `infra keys add` CLI command has flags for a duration that is commonly in days, weeks, or years. The stdlib [time.ParseDuration](https://pkg.go.dev/time#ParseDuration) only supports parsing and printing a duration in sub-seconds, seconds, minutes, and hours. 

We'd like to improve the UX of this command by accepting a duration in days, weeks, or years.

### Background

Our `internal/format` package has `HumanDuration` and `ExactDuration` for formatting a `time.Duration`. `HumanDuration` does support days/weeks/months/years. `ExactDuration` only supports seconds/minutes/hours.

There are some random packages out there that attempt to re-implement `time.ParseDuration` with day/week/year support. Re-implementation seems like the wrong approach when it could extend the implementation and call to `time.Duration` for the rest.

### Proposal

I propose these changes:

1. Add day/week/year support to `ExactDuration`, so that we can print the duration appropriately
2. Add a `Duration` type to `internal/cmd/types` that supports parsing a duration string with days, weeks, or years.

The parse can be pretty limited. We can support only `#d` for days, `#w` for weeks, and `#y` for years, without supporting multiple units in the string (ex: no `3d7h`).  If you want that granularity use hours, and the error message from parsing the string should indicate that. If the string does not contain `d`, `w`, or `y`, then we fallback to calling `time.ParseDuration` as we do today, to support all the strings parsable by `time.ParseDuration`.

dnephin:
## Summary

Resolves #3750

This PR implements the parse portion of #3750. It ended up being pretty easy to support years, weeks, and days along with hours, minutes, seconds, and sub-seconds. This implementation happened to also support negative without any extra work.

The only thing not supported is putting smaller units before larger ones. That means that the input string must be sorted from years, weeks, days, hours, minutes, seconds, sub-seconds. 

Making this a draft for now. 

TODO:
* [ ] more test coverage (missing value for unit, fuzz test for unexpected input)
* [ ] use this new type in the CLI
* [ ] support printing these units in `format.ExactDuration`

### DIFF
--- a/internal/cmd/types/duration.go
+++ b/internal/cmd/types/duration.go
@@ -1,0 +1,77 @@
+package types
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type Duration time.Duration
+
+func (d *Duration) String() string {
+	if d == nil {
+		return "0s"
+	}
+	return time.Duration(*d).String()
+}
+
+func (d *Duration) Set(s string) error {
+	v, err := parseDuration(s)
+	if err != nil {
+		return err
+	}
+	*d = Duration(v)
+	return nil
+}
+
+func (d *Duration) Type() string {
+	return "duration"
+}
+
+func parseDuration(raw string) (time.Duration, error) {
+	if raw == "" {
+		return 0, fmt.Errorf(`invalid duration ""`)
+	}
+
+	var result time.Duration
+
+	years, rest, ok := strings.Cut(raw, "y")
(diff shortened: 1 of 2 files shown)


jmorganca (commented):
Looking good!

PR #3760, opened 2022-11-29
feat: build for chocolatey

mxyng:
## Summary

Chocolatey isn't liking the version string created by goreleaser. This shouldn't be a problem for tagged releases but main builds will fail.


jmorganca:
Nice! This is a big improvement on requiring `scoop` (we can replace even the `scoop` package with this)

mxyng:
Closing this for now. While it's still valid, the timing isn't quite right for this change

PR #3763, opened 2022-11-29
proxy-less K8s connector

ssoroka:
## Summary

should be mostly working.
things left to do:

- [ ] handle selecting/running multiple connectors via configuration
- [ ] some manual testing

## Checklist


api/client.go
```diff
@@ -333,6 +333,17 @@ func (c Client) DeleteProvider(ctx context.Context, id uid.ID) error {
 	return delete(ctx, c, fmt.Sprintf("/api/providers/%s", id), Query{})
 }
 
+func (c Client) ListCredentialRequests(ctx context.Context, req ListCredentialRequest) (*ListCredentialRequestResponse, error) {
+	return get[ListCredentialRequestResponse](ctx, c, "/api/credentials", Query{
+		"destination":     []string{req.Destination},
```
BruceMacD, line 338: Looks like we will want to re-name the existing Credentials struct that stores password hashes for users

internal/access/credentialrequest.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	// commit the transaction so others can see the request
+	err = rCtx.DBTxn.Commit()
+	if err != nil {
+		return nil, err
+	}
+
+	// wait for the request to be filled; listen for the notify on the specific record req.ID
+	listener, err := data.ListenForNotify(c, rCtx.DataDB, data.ListenForNotifyOptions{
```
BruceMacD, line 59: Why do we wait for the request to be fulfilled rather than just finishing and letting the client do a long poll as a separate request? It seems possible to me that you may just want to request a credential then carry on until it is approved.

api/credential_request.go
```diff
@@ -0,0 +1,86 @@
...
+	LastUpdateIndex int64  `form:"lastUpdateIndex"`
+}
+
+type ListCredentialRequestResponse struct {
+	Items          []CredentialRequest `json:"items"`
+	MaxUpdateIndex int64               `json:"maxUpdateIndex"`
+}
+
+// CredentialRequest is the database CredentialRequest object, not a request/response struct
+type CredentialRequest struct {
```
BruceMacD, line 22: This is going to store credentials to access the destination where the connector is running?

internal/server/data/credentialrequest.go
```diff
@@ -0,0 +1,91 @@
...
+
+func GetCredentialRequest(tx ReadTxn, id, organizationID uid.ID) (*models.CredentialRequest, error) {
+	query := querybuilder.New("SELECT")
+	query.B("id, organization_id, expires_at, update_index, user_id, destination_id")
+	query.B("FROM credential_requests")
+	query.B("AND organization_id = ?", tx.OrganizationID())
+	query.B("AND id = ?", id)
+
+	r := &models.CredentialRequest{}
+	err := tx.QueryRow(query.String(), query.Args...).Scan(r.ID, r.OrganizationID, r.ExpiresAt, r.UpdateIndex, r.UserID, r.DestinationID)
```
dnephin, line 19: I think `Scan` requires pointers to fields

internal/server/data/credentialrequest.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/infrahq/infra/internal/server/data/querybuilder"
+	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
+)
+
+func GetCredentialRequest(tx ReadTxn, id, organizationID uid.ID) (*models.CredentialRequest, error) {
+	query := querybuilder.New("SELECT")
+	query.B("id, organization_id, expires_at, update_index, user_id, destination_id")
+	query.B("FROM credential_requests")
+	query.B("AND organization_id = ?", tx.OrganizationID())
```
dnephin, line 15: `WHERE organization_id` , right?

internal/server/data/notify.go
```diff
@@ -0,0 +1,123 @@
...
+			err := json.Unmarshal([]byte(payload), &grant)
+			if err != nil {
+				return err
+			}
+			destination, _, _ := strings.Cut(grant.Resource, ".")
+			if destination != opts.GrantsByDestination {
+				return errNotificationNoMatch
+			}
+			return nil
+		}
```
dnephin, line 120: I think this `switch` needs cases for the new options that set `isMatchingNotify`. If the channel name is sufficiently unique then I guess it can always return `true`. Alternatively `WaitForNotification` could assume `true` when `isMatchingNotify` is nil.

internal/server/models/credentialrequest.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"time"
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/uid"
+)
+
+// CredentialRequest represents a request from a user to get login credentials to a destination.
+type CredentialRequest struct {
```
dnephin, line 11: I wonder if we should call this one `DestinationCredential` and the original one `InfraCredential` or `LocalCredential`.

internal/server/routes.go
```diff
@@ -134,6 +135,9 @@ func (s *Server) GenerateRoutes() Routes {
 
 	add(a, noAuthnWithOrg, http.MethodGet, "/.well-known/jwks.json", wellKnownJWKsRoute)
 
+	// routes only interesting to connectors
+	get(a, authn, "/api/credentials", a.ListCredentialRequests)
```
dnephin, line 139: I expect the UI to want to list credentials as well, so that a user can revoke their own credentials, or an admin can revoke all of a users credentials.

dnephin (commented):
Nice progress on this! 

So far this is only creating credentials, right? (no removal of credentials yet).

I guess we'll want to split this up to get it merged in. I'd be happy to chat about that when you're ready for it.

api/credential_request.go
ssoroka: no, just on the api.

internal/server/routes.go
ssoroka: I don't think so; these are ephemeral things that expire quickly.

dnephin: Ah true, if these are very short lived (~5min) then remove isn't important

ssoroka:
closing to break into smaller PRs.

PR #3764, opened 2022-11-29
os packages

mxyng:
## Summary

- Update Linux packages to include systemd service
- Add archlinux package

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -29,7 +29,7 @@ builds:
 nfpms:
   - id: packages
     package_name: infra
-    file_name_template: "{{ .ProjectName }}-{{ .Version }}-{{ .Arch }}"
+    file_name_template: "{{ .ConventionalFileName }}"
     builds:
       - infra
     vendor: Infra Technologies, Inc.
@@ -39,18 +39,33 @@ nfpms:
     formats:
       - deb
       - rpm
-    bindir: /usr/bin
+    bindir: /usr/local/sbin
+    contents:
+      - src: package-files/systemd/infra.service
+        dst: /usr/lib/systemd/system/infra.service
+        type: config
+      - src: package-files/connector.example.yaml
+        dst: /etc/infra/connector.example.yaml
+        type: config
+    scripts:
+      preinstall: package-files/preinstall.sh
+      postinstall: package-files/postinstall.sh
+      preremove: package-files/preremove.sh
     overrides:
       deb:
-        file_name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Arch }}"
-      apk:
-        replacements:
-          amd64: x86_64
-          arm64: aarch64
+        dependencies:
+          - passwd
+          - procps
+        suggests:
+          - openssh-client
+          - openssh-server
       rpm:
-        replacements:
-          amd64: x86_64
-          arm64: aarch64
+        dependencies:
+          - shadow-utils
+          - procps-ng
+        suggests:
+          - openssh-clients
+          - openssh-server
 archives:
   - id: zip
     format: zip
--- a/package-files/postinstall.sh
+++ b/package-files/postinstall.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+HOMEDIR='/etc/infra'
+
+set -eu
+
+if [ -f "$HOMEDIR/connector.yaml" ]; then
+  if command -v systemctl >/dev/null && [ "$(systemctl is-system-running)" != "offline" ]; then
+    systemctl daemon-reload
+    systemctl restart infra
+  fi
+fi
--- a/package-files/preinstall.sh
+++ b/package-files/preinstall.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+HOMEDIR='/etc/infra'
+
+if command -v dpkg >/dev/null; then
+  NOLOGIN=/usr/sbin/nologin
+elif command -v rpm >/dev/null; then
+  NOLOGIN=/sbin/nologin
+fi
+
+set -eu
+
+status() { echo "$*" >&2; }
+
+if ! getent group infra >/dev/null; then
+  status 'creating group "infra"'
+  groupadd --system infra
+fi
+
+if ! getent passwd infra >/dev/null; then
+  status 'creating user "infra"'
+  useradd --system --shell "$NOLOGIN" --home-dir "$HOMEDIR" --no-create-home --no-user-group --comment 'Infra Agent' infra
+  usermod --group infra infra
+  usermod --lock infra
+fi
+
+if ! getent group infra-users >/dev/null; then
+  status 'creating group "infra-users"'
+  groupadd infra-users
+fi
--- a/package-files/preremove.sh
+++ b/package-files/preremove.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+set -eu
+
+systemctl_stop_service() {
+  if command -v systemctl >/dev/null; then
+    systemctl stop infra || true
+    systemctl disable infra || true
+  fi
+}
+
+dpkg_prerm() {
+  case $1 in
+    remove) systemctl_stop_service ;;
+  esac
+}
+
+rpm_prerm() {
+  case $1 in
+    0) systemctl_stop_service ;;
+  esac
+}
+
+if command -v dpkg >/dev/null; then
+  dpkg_prerm $1
+elif command -v rpm >/dev/null; then
+  rpm_prerm $1
+else
+  systemctl_stop_service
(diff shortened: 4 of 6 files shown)


package-files/systemd/infra.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Modern infrastructure access.
+Documentation=https://docs.infrahq.com
+Wants=network-online.service
+After=network-online.service
+ConditionPathExists=/etc/infra/config.yaml
```
dnephin, line 6: I think we should use `/etc/infra/connector.yaml`. The `infra` binary can run in different modes, (cli, server, connector, agent) and we probably want a different config file for each.

package-files/systemd/infra.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Modern infrastructure access.
```
dnephin, line 2: I guess this should be a description of the service rather than of Infra?

```suggestion
Description=Run the infra ssh connector
```

Or something like that?

package-files/preinstall.sh
```diff
@@ -0,0 +1,32 @@
...
+if ! getent passwd infra >/dev/null; then
+  status 'creating user "infra"'
+  useradd --system --shell "$NOLOGIN" --home-dir "$HOMEDIR" --no-create-home --no-user-group --comment 'Infra Agent' infra
+  usermod --group infra infra
+  usermod --lock infra
+fi
+
+if ! getent group infra-users >/dev/null; then
+  status 'creating group "infra-users"'
+  groupadd infra-users
```
dnephin, line 31: I've been using `infrausers`, but `infra-users` should work as well.

package-files/postinstall.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+HOMEDIR='/etc/infra'
+
+set -eu
+
+if command -v systemctl >/dev/null 2>&1; then
+  systemctl enable infra
+fi
```
dnephin, line 9: I don't think we should enable the `infra` service. This is going to make it impossible to re-use this package for installing the CLI or the server.

Let's make the package only install things, and let the user start the services once they are correctly configured. That's how all arch linux packages work, and I think it's a much better experience than I've seen from packages in other distros.

package-files/preremove.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+set -eu
+
+systemctl_stop_service() {
+  if command -v systemctl >/dev/null 2>&1; then
+    systemctl stop infra
+    systemctl disable infra
```
dnephin, line 8: Do we need `|| true` here, or some other way to ignore the error? The service isn't necessarily running or enabled.

dnephin (commented):
Looks good!

I guess there are a few more files like `/etc/infra/sshd_config` and `/etc/infra/connector.example.yaml` to add to this?

This is the sample config I've been using:

```yaml
kind: ssh

# Infra destination name
name: "" 

# host and optional port used to connect to this SSH server
endpointAddr: ""

server:
  accessKey: "" 
```

.goreleaser.yml
```diff
@@ -38,18 +38,37 @@ nfpms:
     formats:
       - deb
       - rpm
+      - archlinux
     bindir: /usr/bin
+    contents:
```
dnephin, line 43: Does `bindir` here need to change to `/usr/sbin` ? 

I think if that path is a symlink on any distro it's going to complain, but I'm not sure which distros symlink it.

package-files/preinstall.sh
dnephin: Ok,I've changed #3756 to use `infra-users`, so let's keep this.

package-files/systemd/infra.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
```
dnephin, line 1: Should we make this service `infra-connector`, instead of just `infra`? That would give us room to run other infra services.

mxyng: I think this is what we want. The trend has been to name connectors generally as `infra`

.goreleaser.yml
```diff
@@ -29,7 +29,7 @@ builds:
 nfpms:
   - id: packages
     package_name: infra
-    file_name_template: "{{ .ProjectName }}-{{ .Version }}-{{ .Arch }}"
+    file_name_template: "{{ .ConventionalFileName }}"
```
mxyng, line 32: goreleaser has some reasonable defaults for the different distros so we'll use that instead of defining overrides for everything

package-files/systemd/infra.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Infra SSH Connector
+Documentation=https://docs.infrahq.com
+Wants=network-online.service
+After=network-online.service
+ConditionPathExists=/etc/infra/connector.yaml
+
+[Service]
+Type=simple
+ExecStart=/usr/local/sbin/infra connector -f /etc/infra/connector.yaml
```
mxyng, line 10: This service should ideally run as `infra:infra` but we'll need to figure out what if it needs privilege and what those privileges are

package-files/sshd_config
```diff
@@ -0,0 +1,5 @@
+Match group infra-users
```
mxyng, line 1: Not sure this should exist in the package. Might be better to template the file in code and write it out to `/etc/ssh/sshd_config.d/`

package-files/systemd/infra.service
dnephin: The main thing is calling `useradd`, `userdel`, and `pkill` (https://github.com/infrahq/infra/blob/main/docs/dev/ssh/README.md).  All users should be able to read `/etc/passwd`, and the rest is API calls to infra.

If we don't run this as root then we'll need to call `sudo` and make sure the infra user is granted password-less sudo access to those commands.

Are there other options that allow us to run it as non-root ?

package-files/sshd_config
dnephin: We definitely shouldn't install it directly to `/etc/ssh/sshd_config.d/` in the package. I was thinking it could be installed to `/etc/infra/sshd_config` and the setup command could move it to `/etc/ssh/sshd_config.d/infra.conf`.

We could alternatively have the setup command write the file, but I think that might make it more difficult for someone to integrate with automation. By creating the file at `/etc/infra/sshd_config` the automation can just move the file into place.

What would we make a variable in this if we made it a template?  Are there other advantages to writing it out from the setup command?

package-files/systemd/infra.service
dnephin: I see `CAP_KILL` for pkill I guess, but I'm not seeing anything for adding/removing users. I guess it makes sense to keep those only for root, since it effectively lets you create other root users.

mxyng: add/remove user are file operations, not syscalls, so there won't be caps for those

package-files/sshd_config
mxyng: The location of `infra`, the group name though when I said template, I was referring broadly to writing it out to a file from code

package-files/systemd/infra.service
mxyng: I've tried this with User/Group as infra and sudoers allowing `/usr/sbin/useradd`, `/usr/sbin/userdel`, and `/usr/bin/pkill` without any issues though it probably didn't exercise all code paths

jmorganca:
This is looking good! We'll need https://github.com/infrahq/infra/pull/3791 to populate `/etc/infra/connector.yaml`. Do we need arch linux for this first iteration? Given there's a long tail of distros we might be better off providing the `infra` binary for now

dnephin:
I think it would be fine to exclude `archlinux` from the packages.

dnephin:
I tried this out with the following script in a `ubuntu:latest` container and it worked great!

```sh
#!/usr/bin/env bash

set -eu -o pipefail


# step=install-package
dpkg -i "${PACKAGE_PATH}"
apt-get update && apt-get install openssh-server
mkdir -p /run/sshd


# step=write-sshd-config
cat << EOF > /etc/ssh/sshd_config.d/infra.conf
Match group infra-users
    AuthorizedKeysFile none
    PasswordAuthentication no
    AuthorizedKeysCommand /usr/local/sbin/infra sshd auth-keys %u %f
    AuthorizedKeysCommandUser nobody
EOF


# step=write-connector.yaml
cat << EOF > /etc/infra/connector.yaml
kind: ssh
name: testing-ssh
endpointAddr: "$DESTINATION_ADDR"
server:
  url: "$INFRA_SERVER_URL"
  accessKey: "$INFRA_ACCESS_KEY"
EOF

```

PR #3766, opened 2022-11-30
maintain: optional social login domain

BruceMacD:
## Summary
In order to setup social login routing in our hosted Infra offering we need to be able to specify the URL that user is redirected to after social login.

## Checklist


ui/pages/login/redirect.js
```diff
@@ -13,14 +13,14 @@ export default function Redirect() {
   const [error, setError] = useState('')
 
   useEffect(() => {
+    // if it takes over 3 seconds to get the redirect values, something went wrong
```
BruceMacD, line 16: not related to this change, I just noticed this error flashes sometimes on redirect due to getting the cookie being async. Moved to a timed hook instead.

internal/server/configuration.go
```diff
@@ -11,5 +11,6 @@ func (a *API) GetServerConfiguration(c *gin.Context, _ *api.EmptyRequest) (*api.
 	return &api.ServerConfiguration{
 		IsEmailConfigured: email.IsConfigured(),
 		BaseDomain:        a.server.options.BaseDomain,
+		LoginRedirect:     a.server.options.LoginRedirect,
```
mxyng, line 14: Could you not key off the base domain?

BruceMacD: Made this change

ui/pages/login/redirect.js
jmorganca: Why is getting the cookie async? It should be immediately available in the browser unless it's populated by a request (doesn't seem like it here?)

internal/cmd/server.go
```diff
@@ -80,6 +80,7 @@ func newServerCmd() *cobra.Command {
 	cmd.Flags().Duration("session-inactivity-timeout", 0, "A user must interact with Infra at least once within this amount of time for their session to remain valid")
 	cmd.Flags().Bool("enable-signup", false, "Enable one-time admin signup")
 	cmd.Flags().String("base-domain", "", "base-domain for the server, eg example.com")
+	cmd.Flags().String("login-domain-prefix", "", "The path prefix on the base-domain that clients are redirected to after social login")
```
mxyng, line 83: Does this need to be a CLI flag? There's some other options in here that are probably unnecessary as flags

dnephin: Definitely would be nice to avoid adding new flags. I think most people will be configuring the server with either config files or env vars. Command line flags tend to be for interactive commands expected to be run by a human.

BruceMacD: Lets leave it here for now for consistency, all the other server options are available here. It's nice to be able to run the server locally without setting an environment variable.

ui/pages/login/redirect.js
BruceMacD: Got it, it was because the page re-rendered after I removed the cookie. Removed this timeout.

ui/pages/redirect/index.js
```diff
@@ -31,17 +23,20 @@ export default function Redirect() {
...
-      })
       // send the browser to the org specific callback URL to finish login
       router.replace(callbackURL)
     }
 
-    if (code && state) {
-      finish({ code, state })
+    const cookies = new Cookies()
+    const redirect = cookies.get('finishLogin') // the org to redirect to is stored in this cookie
+    console.log(redirect)
```
jmorganca, line 32: Remove me

jmorganca (commented):
Remove the `console.log` and we're good to go.

ui/pages/redirect/index.js
BruceMacD: thanks for catching this

jmorganca (approved):

PR #3767, opened 2022-11-30
Only update user SSH config when flag is enabled

dnephin:
## Summary

SSH destinations are in technical preview, so we don't want to update the user SSH config unconditionally right now.  This PR adds a new `--enable-ssh` flag to `infra login`, which will do the one time update of their ssh config.

Also adds more detail to the message we print when the SSH config is modified.

I think we expect to do this by default once SSH destinations are GA.

Also remove an unused arg from `updateInfraConfig`.

We could also add this to `infra use --ssh` as a way for users to re-inject the config without having to login, but we can think about that more.

Looking for suggestions on the output text and flag help text.


jmorganca (approved):
LGTM

PR #3768, opened 2022-11-30
maintain: change grants api change_by to use lower camel case

pdevine:
## Summary

The `created_by` field when creating/listing grants should be  in lower camel case (i.e. `createdBy`) similar to each of the other fields.

## Checklist


jmorganca (approved):
LGTM. Note: there's a typo in the PR name to fix before merging (`s/change_by/created_by`)

PR #3769, opened 2022-11-30
ci: only run -race on main and when there is a label

dnephin:
## Summary

The `-race` flag to `go test` enables the go race detector. The race detector is great for finding bugs, but also adds a lot of runtime to our test run (about 2x). This PR changes our CI so that we only run `-race` on `main` (or when the `actions/race` label is set on the PR). This should speed up test runs in most PRs, and still gives us a reliable way to catch data races.

I'll be testing the config a bit to get the condition right.

### DIFF
--- a/.github/labels.yaml
+++ b/.github/labels.yaml
@@ -90,3 +90,6 @@
 - name: "action/fuzz"
   color: "dddddd"
   description: Run fuzzing on this pull request.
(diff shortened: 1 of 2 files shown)


jmorganca (approved):
LGTM

dnephin:
Alright, this seems to be working now.

PR #3770, opened 2022-11-30
Check that all API parameters use camel case

dnephin:
## Summary

This PR adds some validation to our OpenAPI document generation to catch when we use the wrong form of a parameter name. All our of parameters should use `camelCase` for consistently.

The document generation always runs in CI, so we'll catch this panic as a test failure in dev before merging any PRs.

I added a temporary allow list for the two fields that don't conform to the convention, which we'll remove once they are fixed.

I believe the parameter name always comes from `getFieldName`, so it should be safe to validate it here for now.

Related to #3590, and #3768

### DIFF
--- a/internal/server/openapi.go
+++ b/internal/server/openapi.go
@@ -534,10 +534,12 @@ func removeString(seq []string, name string) []string {
 
 func getFieldName(f reflect.StructField, parent reflect.Type) string {
 	if name, ok := f.Tag.Lookup("form"); ok {
+		validateFieldName(name)
 		return name
 	}
 
 	if name, ok := f.Tag.Lookup("uri"); ok {
+		validateFieldName(name)
 		return name
 	}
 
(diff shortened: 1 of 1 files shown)


internal/server/openapi.go
```diff
@@ -548,8 +550,21 @@ func getFieldName(f reflect.StructField, parent reflect.Type) string {
...
 		return name
 	}
 
 	panic(fmt.Sprintf("field %q of struct %q must have a tag (json, form, or uri) with a name or '-'", f.Name, parent.Name()))
 }
+
+func validateFieldName(name string) {
+	// temporary allow list
+	switch name {
+	case "created_by", "unique_id":
```
dnephin, line 563: `created_by` is fixed by #3768

I think we'll be able to remove `uniqueID` soon, so it's probably not worth renaming that one. Let's remove it sometime in the next release or two.

jmorganca (approved):

ssoroka (changes requested):
Form fields are supposed to be underscore, aren't they? Hmm. Nevermind, I guess I'm mistaken.
We should probably validate the `header` fields are valid header tags with underscores.

dnephin:
> We should probably validate the header fields are valid header tags with underscores.

Right now the only headers we include in the document are hardcoded into `buildRequest` in `openapi.go`. We pass them directly to a method in a third-party library, so there's not really any place for us to do that validation.

Maybe in the future we could write something that reads the openAPI document and validates the document?

The change in this PR should catch all of the bugs we've encountered so far, so I think it's worth it to make this change now and do additional validation later.

ssoroka (approved):

internal/server/openapi.go
dnephin: I removed `created_by` from this list, since #3768 has merged.

PR #3774, opened 2022-11-30
Fix docgen; output to stdout

jmorganca:


mxyng (approved):
Change looks good though I'm wondering why it's needed.

jmorganca:
Due to `NewRootCmd` requiring a non-nil pointer now. We don't run this on every commit and so we didn't catch the crash.

PR #3776, opened 2022-12-01
maintain: allow create grants to take usernames

pdevine:
## Summary

This change fixes the `POST /api/grants` endpoint to allow it to take a user name for the subject similar to how the `PATCH /api/grants` endpoint works. Currently the endpoint will fail incorrectly with a 500 error.

## Checklist

### DIFF
--- a/internal/server/grants.go
+++ b/internal/server/grants.go
@@ -188,7 +188,7 @@ func getGrantFromGrantRequest(c *gin.Context, r api.GrantRequest) (*models.Grant
 			}
 			return nil, err
 		}
-		subject = uid.NewIdentityPolymorphicID(group.ID)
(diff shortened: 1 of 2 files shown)


jmorganca (approved):
Nice fix 😃

BruceMacD (commented):
This means creating a grant by group name will work also, correct?

dnephin (approved):
Looks good!

Edit: Nevermind! RequireOneOf is already exactly one!

pdevine:
@BruceMacD Yes! I've added a test for that, and realized there was a bug with group names where it was not creating the correct ID type. That is fixed now too (and the unit test is also fixed)

jmorganca (approved):

PR #3777, opened 2022-12-01
Use standard Kubernetes config loading instead of custom options in connector

jmorganca:


internal/connector/connector.go
```diff
@@ -63,22 +61,6 @@ type ListenerOptions struct {
 	Metrics string
 }
 
-type KubernetesOptions struct {
-	// AuthToken may be used to override the token used to authenticate with the
-	// kubernetes API server. When the connector is run in-cluster, the
-	// service account associated with the pod will be used by default.
-	// When run outside of cluster there is no default, and this value must
-	// be set to a token that has permission to impersonate users in the cluster.
-	AuthToken types.StringOrFile
```
BruceMacD, line 72: Does this change mean its not possible to run the connector outside of Kubernetes anymore?

dnephin: It should be, it's just that we have to use the kubectl config instead of our own config file.

internal/cmd/connector_test.go
```diff
@@ -72,6 +75,15 @@ func TestConnector_Run(t *testing.T) {
 	ctx := context.Background()
 	runAndWait(ctx, t, srv.Run)
 
+	kubeconfig := path.Join(dir, "kubeconfig")
+	os.Setenv("KUBECONFIG", kubeconfig)
```
dnephin, line 79:
```suggestion
	t.Setenv("KUBECONFIG", kubeconfig)
```

:grinning:

dnephin (approved):

internal/cmd/connector_test.go
jmorganca: Thanks!

internal/connector/connector.go
jmorganca: It's easier now: running `infra connector` will just read the current kubeconfig for auth info

PR #3781, opened 2022-12-01
fix: index provider user emails with identity ID

BruceMacD:
## Summary
While testing Google social login I found this index can conflict now, due to the Google login not having a provider ID. Updating the index to include the identity ID to make sure it is unique.

This index is only used by SCIM filters.

## Checklist

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -80,6 +80,7 @@ func migrations() []*migrator.Migration {
 		updateAccessKeysTimeoutColumn(),
 		addUserPubicKeysTable(),
 		addUserSSHLoginName(),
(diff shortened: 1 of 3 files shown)


dnephin (commented):
I think this makes sense, but I'm not sure I totally understand how this works without a provider ID. The primary key is `(providerId, identityID)`, so I think the providerID can't be null, right?  What goes into the providerID field?

dnephin (approved):
LGTM! I got an answer to my question out-of-band. We'll be storing a `0` for the provider_id in these cases.

I think we should consider using a hard-coded value for this, instead of a 0. I guess we'll need that if we ever add a second social login provider. But that could be done as a follow up, since this fixes a bug.

PR #3784, opened 2022-12-01
fix: more descriptions and examples for destinations, users, and groups

technovangelist:

## Summary

Added more descriptions for Destinations, Users, and Groups

## Checklist

## Related Issues

Resolves #


api/destination.go
```diff
@@ -49,11 +49,11 @@ func (r ListDestinationsRequest) ValidationRules() []validate.ValidationRule {
...
 
 type CreateDestinationRequest struct {
-	UniqueID   string                `json:"uniqueID"`
-	Name       string                `json:"name"`
-	Kind       string                `json:"kind"`
-	Version    string                `json:"version"`
-	Connection DestinationConnection `json:"connection"`
+	UniqueID   string                `json:"uniqueID" note:"Unique ID used to identify this specific destination" example:"94c2c570a20311180ec325fd56"`
+	Name       string                `json:"name" note:"Name of the destination" example:"production-cluster"`
+	Kind       string                `json:"kind" note:"Kind of destination. eg. kubernetes or ssh or postgres" example:"kubernetes"`
```
BruceMacD, line 54: nit
```suggestion
	Kind       string                `json:"kind" note:"Kind of destination. eg. kubernetes, ssh, or postgres" example:"kubernetes"`
```

api/group.go
```diff
@@ -6,10 +6,10 @@ import (
 )
 
 type Group struct {
-	ID         uid.ID `json:"id"`
+	ID         uid.ID `json:"id" note:"Group ID" example:"gauEdoYCEU"`
 	Name       string `json:"name" note:"Name of the group" example:"admins"`
-	Created    Time   `json:"created"`
-	Updated    Time   `json:"updated"`
+	Created    Time   `json:"created" note:"Date the group was created"`
```
BruceMacD, line 11: Should this have an example time to match the destination struct?

api/user.go
```diff
@@ -54,15 +54,15 @@ func (r CreateUserRequest) ValidationRules() []validate.ValidationRule {
...
+	ID              uid.ID `json:"id" note:"User ID"`
+	Name            string `json:"name" note:"Email address of the user" example:"bob@example.com"`
+	OneTimePassword string `json:"oneTimePassword,omitempty" note:"One-time password (only returned when self-hosted)" example:"password"`
 }
 
 type UpdateUserRequest struct {
 	ID          uid.ID `uri:"id" json:"-"`
-	OldPassword string `json:"oldPassword"`
-	Password    string `json:"password"`
+	OldPassword string `json:"oldPassword" note:"Old password for the user. Only required when the access key used is not owned by an Infra admin" example:"oldpassword"`
```
BruceMacD, line 64:
```suggestion
	OldPassword string `json:"oldPassword" note:"Old password for the user. Only required when the access key making this request is not owned by an Infra admin" example:"oldpassword"`
```

api/group.go
technovangelist: No, the times generate a valid example on their own. and it  seems to be good.

PR #3785, opened 2022-12-01
Fix a few problems with local user removal in ssh connector

dnephin:
## Summary

1. Instead of exiting on the first error, attempt the rest of the operations and collect the error to be returned at the end. This ensures that one user in the wrong state doesn't block the creation and delete of all other users.

2. Ignore exit status 1 from pkill. The user being removed may not be running any processes, so pkill may exit with status 1 and that's ok. In the future we can look to implement this logic in Go, which might give us more granular information about why the operation failed.

3. Kill all processes before removing any users. This may give the processes time to exit before we try to remove the user. Removing the user will fail if they still have processes running. Also use SIGKILL instead of SIGTERM, which should also speed up process exit. With the fixes above it shouldn't be a problem if the `userdel` fails. We'll attempt it again on the next iteration of the reconcile loop.

It's also important to note that removing the user is not strictly necessary to prevent them from logging in. `sshd` does a synchronous check for grants, so a user will be unable to login immediately after the grant is removed, even if the connector has trouble removing the user.

### DIFF
--- a/internal/connector/ssh.go
+++ b/internal/connector/ssh.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal/cmd/cliopts"
 	"github.com/infrahq/infra/internal/linux"
 	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/repeat"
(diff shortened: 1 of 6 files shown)


jmorganca (approved):

mchiang0610 (commented):
👍 tried on a digital ocean host

PR #3788, opened 2022-12-02
feat: google sign-up

### LINKED ISSUE #3601: Sign-up with Google
I want to sign-up with my existing Google org.

When signing up I should be able to click a "sign-up with Google" button.

BruceMacD:
## Summary
Sign-up using Google. This change modifies the sign-up API to accept either a Google social login code and redirect URL or a directly specified username, password, org name, and org sub-domain.

After authenticating with Google, an new user can pick the org name and domain.

signup

After Google authentication:
Screen Shot 2022-12-06 at 1 48 28 PM

## Checklist

## Related Issues
Resolves #3601


internal/access/organization.go
```diff
@@ -62,3 +66,16 @@ func DeleteOrganization(c *gin.Context, id uid.ID) error {
...
 	return strings.ToLower(subDomain) + "." + serverBaseDomain
 }
+
+// DomainAvailable is needed to check if an org domain is available before completing social sign-up
+func DomainAvailable(c *gin.Context, domain string) error {
+	rCtx := GetRequestContext(c)
+	_, err := data.GetOrganization(rCtx.DBTxn, data.GetOrganizationOptions{ByDomain: domain})
+	if err != nil {
+		if !errors.Is(err, internal.ErrNotFound) {
+			logging.L.Error().Err(err).Msg("failed to check if organization exists by domain")
```
mxyng, line 76: Might be safer to return an error here since we don't actually know if the domain is in use.

internal/access/signup.go
```diff
@@ -8,62 +8,153 @@ import (
...
 
 	details.Org.Domain = SanitizedDomain(details.SubDomain, baseDomain)
 
+	switch {
+	case details.User != nil:
+		email, err := email.Domain(details.User.Name)
+		if err != nil {
+			return nil, fmt.Errorf("set allowed domain from user: %w", err)
+		}
+		details.Org.AllowedDomains = []string{email}
```
mxyng, line 60: I'm not convinced this is a good idea. This can silently give much more access than the user realizes. This should at least be an option the user must confirm.

mxyng, line 60: Unless I'm mistaken, this doesn't omit `gmail.com` nor `googlemail.com` as was previously done in `handlers.go`.

BruceMacD: I'll add it to the sign-up fields

internal/access/organization.go
BruceMacD: Updated this to fail

internal/access/signup.go
BruceMacD: I tried playing with this in the sign-up flow but it was a bit confusing as something that the admin specifies as a first step along with their org name. 

Where someone from the same org being able to login is low risk (no privileges, they still need a grant) I'm going to move this to a first step in the "welcome" pop-up after sign-up in a separate PR.

BruceMacD: Fixed, thanks

BruceMacD: Another follow-up, just gonna remove allowed domains:
https://github.com/infrahq/infra/issues/3834

internal/access/signup.go
```diff
@@ -8,62 +8,154 @@ import (
...
+		adminEmail = details.User.Name
+	case details.Social != nil:
+		adminEmail = details.Social.IDPAuth.Email
+	}
+	allowedLoginDomain, err := email.Domain(adminEmail)
+	if err != nil {
+		return nil, fmt.Errorf("allowed login domain from admin email: %w", err)
+	}
+	if allowedLoginDomain != "gmail.com" && allowedLoginDomain != "googlemail.com" {
+		// if gmail or goolemal the admin will have to manually specify this later
```
mxyng, line 66:
```suggestion
		// if gmail or googlemail the admin will have to manually specify this later
```

internal/server/signup.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	keyExpires := time.Now().UTC().Add(a.server.options.SessionDuration)
+
+	var created *access.NewOrgDetails
+	switch {
+	case r.Social != nil:
+		// do social sign-up
+		if a.server.Google == nil {
+			return nil, fmt.Errorf("%w: google login is not configured, provider id must be specified for oidc login", internal.ErrBadRequest)
```
mxyng, line 43: This error isn't quite right. Probably don't need the second part

```suggestion
			return nil, fmt.Errorf("%w: google login is not configured", internal.ErrBadRequest)
```

mxyng (approved):
LGTM, some minor comments

PR #3793, opened 2022-12-02
use goreleaser to publish artifacts

mxyng:
There will soon be a need to publish more artifact. Replace the custom publishing actions with goreleaser which will also handle the other ways of publishing (AUR, Chocolatey)

### DIFF
--- a/.github/workflows/cd-binaries.yaml
+++ b/.github/workflows/cd-binaries.yaml
@@ -42,53 +42,12 @@ jobs:
         with:
           distribution: goreleaser
           version: latest
-          args: release --rm-dist --skip-publish --snapshot
+          args: release --rm-dist --snapshot ${{ startsWith(github.ref, 'refs/tags/') || '--skip-publish' }}
         env:
           RELEASE_NAME: ${{ needs.prepare.outputs.release-name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/cd-binaries.yaml
```diff
@@ -42,53 +43,12 @@ jobs:
         with:
           distribution: goreleaser
           version: latest
-          args: release --rm-dist --skip-publish --snapshot
+          args: release --rm-dist --snapshot ${{ startsWith(github.ref, 'refs/tags/') || '--skip-publish' }}
```
mxyng, line 46: This allows CI to still build the binaries and packages without publishing them. Useful for running in `main`

jmorganca:
Nice! Good to remove the custom publishing as it's less to maintain.

jmorganca (approved):

.github/workflows/cd-binaries.yaml
jmorganca: Great. Does this mean we'll have binary artifacts for every commit on `main` (great for testing)

mxyng: We had this before as well. Those artifacts only live in GitHub actions though

PR #3796, opened 2022-12-02
fix: useradd parameter order

mxyng:
## Summary

`*` applies to `-p` as in `--password`. The current format is invalid and causes useradd to not run correctly.

### DIFF
--- a/internal/linux/users.go
+++ b/internal/linux/users.go
@@ -66,7 +66,7 @@ func isRuneComma(r rune) bool {
 func AddUser(user *api.User, group string) error {
 	args := []string{
 		"--comment", fmt.Sprintf("%v,%v", user.ID, sentinelManagedByInfra),
-		"-m", "-p", "-g", group, "*", user.SSHLoginName,
+		"-m", "-p", "*", "-g", group, user.SSHLoginName,
 	}
 	cmd := exec.Command("useradd", args...)
 	cmd.Stdout = logging.L
--- a/internal/connector/ssh_test.go
+++ b/internal/connector/ssh_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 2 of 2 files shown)


internal/linux/users.go
```diff
@@ -66,7 +66,7 @@ func isRuneComma(r rune) bool {
 func AddUser(user *api.User, group string) error {
 	args := []string{
 		"--comment", fmt.Sprintf("%v,%v", user.ID, sentinelManagedByInfra),
-		"-m", "-p", "-g", group, "*", user.SSHLoginName,
+		"-m", "-p", "!", "-g", group, user.SSHLoginName,
```
dnephin, line 69: From what I've been reading `!` in the password is potentially a problem because it can lock the entire account (even if the account is not using password login). `*` sounded like it was safer. 

Any reason to change this to `!` ?

dnephin (commented):
LGTM when tests pass. I think the files lost the trailing whitespace.

internal/linux/users.go
mxyng: `!` locks the password, not the account. The canonical way of disabling an account is to set the expire date to the past.

https://man7.org/linux/man-pages/man1/passwd.1.html

       -l, --lock
           Lock the password of the named account. This option disables
           a password by changing it to a value which matches no
           possible encrypted value (it adds a ´!´ at the beginning of
           the password).

           Note that this does not disable the account. The user may
           still be able to login using another authentication token
           (e.g. an SSH key). To disable the account, administrators
           should use usermod --expiredate 1 (this set the account's
           expire date to Jan 2, 1970).

           Users with a locked password are not allowed to change their
           password.

https://man7.org/linux/man-pages/man8/usermod.8.html

       -L, --lock
           Lock a user's password. This puts a '!' in front of the
           encrypted password, effectively disabling the password. You
           can't use this option with -p or -U.

           Note: if you wish to lock the account (not only access with a
           password), you should also set the EXPIRE_DATE to 1.

dnephin: We should test this more. I was having problems using SSH with a key when the account had a `!` in the password field.

mxyng: I've tested this with Ubuntu, Fedora, and Arch. Both Ubuntu and Arch work as expected with a locked user account (`!`). Fedora required some modifications since `umask` was set incorrectly and created files sshd refused to read but once that was fixed, it worked as expected

dnephin: Awesome, thanks for testing it out! I've been testing with alpine. It's possible the problem is specific to busybox, but that's not an important use case outside of testing.  I've created #3813 for adding some automated tests, which should help catch these problems.

dnephin (approved):

internal/linux/users.go
dnephin: Ahhh, that's it! Thanks for figuring this out!

PR #3819, opened 2022-12-05
Fix macos compatibility in connector test

dnephin:
## Summary

`echo -n` was not working on macos in sh. Using 'go run' should avoid any compatibility problems.


internal/connector/ssh_test.go
```diff
@@ -47,13 +46,13 @@ func TestUpdateLocalUsers(t *testing.T) {
...
-	expected := `pkill '--signal' 'KILL' '--uid' 'three333'
-pkill '--signal' 'KILL' '--uid' 'four444'
-userdel '--remove' 'three333'
-userdel '--remove' 'four444'
-useradd '--comment' 'Ej,managed by infra' '-m' '-p' '*' '-g' 'infra-users' 'two222'
+	expected := `pkill --signal KILL --uid three333
+pkill --signal KILL --uid four444
+userdel --remove three333
+userdel --remove four444
+useradd --comment "Ej,managed by infra" -m -p * -g infra-users two222
```
dnephin, line 53: I changed it to only quote the args when there is some whitespace in the arg, so the expected values also changed.

internal/connector/ssh_test.go
```diff
@@ -47,13 +46,13 @@ func TestUpdateLocalUsers(t *testing.T) {
...
 `
-	assert.Equal(t, expected, strings.ReplaceAll(string(actual), " \n", "\n"))
```
dnephin, line 56: We no longer have trailing whitespace, so we shouldn't need this replace

internal/connector/ssh_test.go
mxyng: It might be better to keep the quotes for consistency. Otherwise, it's up to the author to figure out if/when to wrap expected values in quotes.

dnephin: By author do you mean the person changing this expected value?

You can automatically update this expected value by running tests with `-update`.

I think it reads a bit better without the extra quotes, but it's not a big problem to add them back.

What do you think about adding a comment informing the reader about `-update` flag? Would that work instead of always adding quotes?

mxyng: Yes, author as in author of a commit. I agree this reads better and easier to map to the actual command being written. Don't have to add the quote back in though it might be a good ideal to quote the `*`

dnephin: Good call, I've added a comment about how to update this value, and I've added single quotes for any values with spaces or `*`.

dnephin:
Thanks for the help debugging this one! I this I've fixed both problems now.

jmorganca (approved):

BruceMacD (approved):
works on my machine now too

PR #3825, opened 2022-12-06
fix: print selected login option

mxyng:
## Summary

With removal of login prompts, it's become confusing when using username and password login options, in particular when set with environment variables. This PR adds a few informational outputs to help the user trace which option is currently being used.

### Device flow (Default)
image

### User with password prompt
image

### User with password from env var
image

### DIFF
--- a/internal/cmd/cli.go
+++ b/internal/cmd/cli.go
@@ -87,6 +87,6 @@ func newCLI(ctx context.Context) *CLI {
 	return &CLI{
 		Stdin:  os.Stdin,
 		Stdout: os.Stdout,
-		Stderr: os.Stderr,
+		Stderr: newStderr(os.Stderr),
 	}
 }
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -123,6 +123,8 @@ func login(cli *CLI, options loginCmdOptions) error {
 	options.Server = strings.TrimPrefix(options.Server, "https://")
 	options.Server = strings.TrimPrefix(options.Server, "http://")
 
+	fmt.Fprintf(cli.Stderr, "  Logging in to %s\n", termenv.String(options.Server).Bold().String())
+
 	if len(options.TrustedCertificate) == 0 {
 		// Attempt to find a previously trusted certificate
 		for _, hc := range config.Hosts {
@@ -143,6 +145,8 @@ func login(cli *CLI, options loginCmdOptions) error {
 	case options.AccessKey != "":
 		loginReq.AccessKey = options.AccessKey
 	case options.User != "":
+		fmt.Fprintf(cli.Stderr, "  Logging in as user %s\n", termenv.String(options.User).Bold().String())
+
 		if options.Password == "" {
 			if options.NonInteractive {
 				return Error{Message: "Non-interactive login requires setting the INFRA_PASSWORD environment variable"}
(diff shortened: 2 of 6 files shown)


internal/cmd/login_test.go
```diff
@@ -492,6 +491,8 @@ func TestLoginCmd_TLSVerify(t *testing.T) {
 		expected := &ClientConfig{ClientConfigVersion: clientConfigVersion}
 		assert.DeepEqual(t, cfg, expected, cmpopts.EquateEmpty())
 
-		golden.Assert(t, bufs.Stderr.String(), t.Name())
```
mxyng, line 495: It's not possible to compare the exact outputs since the server endpoint has a variable port component

internal/cmd/login.go
```diff
@@ -456,8 +460,8 @@ func deviceFlowLogin(ctx context.Context, client *api.Client, cli *CLI) (*api.Lo
 	url := resp.VerificationURI + "?code=" + resp.UserCode
 
 	// display to user
-	cli.Output("Navigate to " + url + " and verify your code:\n")
-	cli.Output("\t\t" + resp.UserCode + "\n")
+	fmt.Fprintf(os.Stderr, "  Navigate to %s and verify your code:\n\n", termenv.String(url).Underline().String())
+	fmt.Fprintf(os.Stderr, "\t\t%s\n\n", termenv.String(resp.UserCode).Bold().String())
```
mxyng, line 464: This adds highlighting to the output to more easily identity important components

BruceMacD: What does this look like? Feels like a lot of new lines.

internal/cmd/login.go
```diff
@@ -143,6 +145,8 @@ func login(cli *CLI, options loginCmdOptions) error {
 	case options.AccessKey != "":
 		loginReq.AccessKey = options.AccessKey
 	case options.User != "":
+		fmt.Fprintf(cli.Stderr, "  Logging in as user %s\n", termenv.String(options.User).Bold().String())
```
BruceMacD, line 148:
```suggestion
		fmt.Fprintf(cli.Stderr, "  Logging in as %s\n", termenv.String(options.User).Bold().String())
```

internal/cmd/login.go
```diff
@@ -143,6 +145,8 @@ func login(cli *CLI, options loginCmdOptions) error {
 	case options.AccessKey != "":
 		loginReq.AccessKey = options.AccessKey
```
BruceMacD, line 146: Should we output something in the case of an access key environment variable too? Might make sense where it takes precedence

jmorganca:
@mxyng can you add an example here in the PR description? Thanks!

internal/cmd/login.go
mxyng: It's the same content but with some highlights. Will add some screenshots to PR description

internal/cmd/login.go
mxyng: This line and the preceding line are very similar so I wanted something more to differentiate between the two

internal/cmd/cli.go
```diff
@@ -87,6 +87,6 @@ func newCLI(ctx context.Context) *CLI {
 	return &CLI{
 		Stdin:  os.Stdin,
 		Stdout: os.Stdout,
-		Stderr: os.Stderr,
+		Stderr: newStderr(os.Stderr),
```
mxyng, line 90: Can alternatively use `terminal.NewAnsiStderr` but that doesn't seem to display underlined text correctly

BruceMacD (approved):

internal/cmd/login.go
BruceMacD: I can't really come up with anything great either, maybe just `Logging in with access key` would suffice.

internal/cmd/login.go
jmorganca, line 148: We can remove this line – the user already specified `--user`

Issue #3812, opened 2022-12-05
Require an expiration time for SSH public keys


dnephin:
Today we don't have any expiration time for public keys. When a key is uploaded we should set a default expiration time, and return it in the response.

The user doesn't really have a way of setting the expiration time (assuming the key is being generated for them from `infra ssh hosts` which is run by `ssh`). We could make it an organization setting, but for now we can set a default expiration time on all keys.

I'm thinking something like 8 hours might be a good place to start.

We need #3811 before making this change.

dnephin:
#3826 adds a TODO that needs to be resolved with this change. The CLI needs to remove local keys when it sees they are removed from the API.

PR #3826, opened 2022-12-06
Improve ssh local key management

### LINKED ISSUE #3811: Improve management of local ssh keys
### Background

Today we create a single SSH key if one doesn't exist in the local file system on the client side. We never check if the key exists in the API, or if it has expired, or deleted.  This can also cause confusing errors on the client if the key was updated to a different API server, or different org.

We need to solve this problem before we add expiration times to the public keys.

### Proposal

Use a local file `~/.ssh/infra/keys.json` for tracking the server address, and organization ID of each local key. This keys file allows us to do two things: find the appropriate key for a given server and org, and also remove keys once they are deleted or expired.

The `keys.json` file should have the following format:

```go
type keysConfig struct {
    Keys []localPublicKey
    // spaced reserved for other fields like Version 
}

type localPublicKey struct {
    serverHost string
    organizationID string
    userID string
    // publicKeyID is the ID of the user public key returned by the server when the key is uploaded.
    publicKeyID string
}
```

The public and private keys will be stored in files named `~/.ssh/infra/keys/<publicKeyID>{.pub}`.

The `provisionSSHKey` function in `internal/cmd/ssh.go` will change from using `fileExists` to the following:

1. read the `keys.json` and find any entries that match the server, orgID, and userID of the logged in user.
2. if one or more entries are found in the `keys.json` file, and keys no longer exist on the filesystem then remove the entries from the file and proceed as if no keys exist.
3. if one or more entries are found in the `keys.json` file, and the keys still exist on the local filesystem, then query the API to verify the key still exists. If it does use it, otherwise remove the entry from the file and proceed as if no keys exist. 
4. if no entries are found in `keys.json`, or if the file doesn't exist, create a new key.

Creating new keys should work almost the same as today. The one difference would be:
1. upload the public key first (before saving it to the file), and then using the returned primary key to save the keys to the filesystem
2. write an entry to the `keys.json` file once the keys are saved

Today we don't store the organizationID, and we don't get it from login, so either we'll have to query `/api/organizations/self` to get the ID, or we should return it from login.

dnephin:
## Summary

Resolves #3811

Track key pairs in a local file so that when keys are deleted or expire from the API, we can remove the local keys as well. 

This will ensure that when someone runs `ssh` for an infra destination, the Infra CLI will generate a key when necessary, and skip generating a key when one exists.

The logic in `provisionSSHKey` changes so that it uploads to the API first, and then uses the ID in the response as the filename for the key.

### DIFF
--- a/internal/cmd/ssh.go
+++ b/internal/cmd/ssh.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/exp/slices"
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal/logging"
@@ -221,10 +222,11 @@ func provisionSSHKey(ctx context.Context, opts provisionSSHKeyOptions) (string, 
 
 	keysDir := filepath.Join(opts.infraSSHDir, "keys")
 	existingKeys := matchingPublicKeys(keysCfg, opts.hostConfig, org.ID)
-	for _, existing := range existingKeys {
+	for i, existing := range existingKeys {
 		filename := filepath.Join(keysDir, existing.PublicKeyID)
 		if !fileExists(filename) || !fileExists(filename+".pub") {
 			// key doesn't exist locally
+			keysCfg.Keys = slices.Delete(keysCfg.Keys, i, i+1)
 			continue
 		}
 
@@ -235,7 +237,15 @@ func provisionSSHKey(ctx context.Context, opts provisionSSHKeyOptions) (string, 
 		}
 
 		// key doesn't exist in the API
-		// TODO: delete the local key file, and remove it from keysConfig
+		fmt.Fprintf(opts.cli.Stderr,
+			"Removing %v because it was expired or deleted from Infra", filename)
+		if err := os.Remove(filename); err != nil && !errors.Is(err, fs.ErrNotExist) {
+			return "", fmt.Errorf("removing deleted key %w", err)
+		}
+		if err := os.Remove(filename + ".pub"); err != nil && !errors.Is(err, fs.ErrNotExist) {
+			return "", fmt.Errorf("removing deleted key %w", err)
+		}
+		keysCfg.Keys = slices.Delete(keysCfg.Keys, i, i+1)
 	}
 
 	_ = os.MkdirAll(keysDir, 0o700)
--- a/internal/cmd/ssh.go
+++ b/internal/cmd/ssh.go
@@ -158,7 +158,11 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
 	infraSSHDir := filepath.Join(homeDir, ".ssh/infra")
 	_ = os.MkdirAll(infraSSHDir, 0o700)
 
-	opts, err := defaultClientOpts()
+	hostCfg, err := currentHostConfig()
+	if err != nil {
+		return err
+	}
+	opts, err := apiClientFromHostConfig(hostCfg)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


internal/cmd/ssh.go
```diff
@@ -167,72 +171,125 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
...
+		keysCfg = &keysConfig{}
+	case err != nil:
+		return "", err
+	}
 
-	// TODO: check expiration
-	// TODO: check the key exists in the API
-	if fileExists(keyFilename) {
-		return nil
+	org, err := opts.client.GetOrganizationSelf(ctx)
```
dnephin, line 217: It would be nice to get this from the `hostConfig` instead of having to query for it again. If the `LoginResponse` included this ID in addition to the name, I think it would be easy to store this in the `hostConfig`. 

I created #3827 for this.

internal/cmd/ssh.go
```diff
@@ -167,72 +171,125 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
...
+	existingKeys := matchingPublicKeys(keysCfg, opts.hostConfig, org.ID)
+	for _, existing := range existingKeys {
+		filename := filepath.Join(keysDir, existing.PublicKeyID)
+		if !fileExists(filename) || !fileExists(filename+".pub") {
+			// key doesn't exist locally
+			continue
+		}
+
+		if userPublicKeyContains(opts.user.PublicKeys, existing.PublicKeyID) {
+			// TODO: check expiration when expiry is added
```
dnephin, line 232: This will need to be done with https://github.com/infrahq/infra/issues/3812

internal/cmd/ssh.go
```diff
@@ -167,72 +171,125 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
...
+			// key exists locally and in the API
+			return filename, nil
+		}
+
+		// key doesn't exist in the API
+		// TODO: delete the local key file, and remove it from keysConfig
```
dnephin, line 238: I forgot about this one. I should resolve this TODO in this PR.

internal/cmd/ssh.go
```diff
@@ -167,72 +171,125 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
...
+		return "", err
 	}
 
-	fmt.Fprintf(cli.Stderr, "Creating a new RSA 4096 bit key pair in %v\n", keyFilename)
+	keysDir := filepath.Join(opts.infraSSHDir, "keys")
+	existingKeys := matchingPublicKeys(keysCfg, opts.hostConfig, org.ID)
+	for _, existing := range existingKeys {
+		filename := filepath.Join(keysDir, existing.PublicKeyID)
+		if !fileExists(filename) || !fileExists(filename+".pub") {
+			// key doesn't exist locally
```
BruceMacD, line 227: should we remove the key from the keys config file in this case?

internal/cmd/ssh.go
```diff
@@ -167,72 +171,125 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
...
+			// key exists locally and in the API
+			return filename, nil
+		}
+
+		// key doesn't exist in the API
+		// TODO: delete the local key file, and remove it from keysConfig
+	}
+
+	_ = os.MkdirAll(keysDir, 0o700)
+	fmt.Fprintf(opts.cli.Stderr, "Creating a new RSA 4096 bit key pair in %v\n", keysDir)
```
BruceMacD, line 242: Does the user need to know everytime this happens? Feels like it could just happen behind the scenes.

dnephin: I think it's important we tell the user what is happening for a few reasons.  The main one is that generating a new key is relatively slow. It can easily take a second or even two. If we don't tell the user what is happening then they get the impression that Infra is slow or sluggish.  By giving them information about what is happening , they can better understand why some runs of `ssh` are fast, and others appear slow.

In general I think it's always better to give the user information, than to hide it. It's definitely possible to go too far and spew tons of logs, but here it's a single informative log message. If a user ever has to debug a problem with `infra`, they will be in a better position because we've given them some basic information about how it works. If we don't do this users are left feeling helpless and that's also a bad experience.

dnephin:
Ok, I resolved the TODO that I missed, and fixes the keysConfig to remove entries where the files are missing. I made those changes in a separate commit. This should be ready for another review.

BruceMacD (approved):
looks good

PR #3836, opened 2022-12-07
fix: remove allowed domains

### LINKED ISSUE #3834: Remove allowed login domains
## Describe the bug
Allowed login domains are not needed for now. Require manual user invites instead.

BruceMacD:
## Summary
With the current implementation of Google login "allowed domains" are an unnecessary complication. Removing them for now in favor of specifically inviting users, or automatically pulling user information from Google.

## Checklist

## Related Issues

Resolves #3834

### DIFF
--- a/api/organization.go
+++ b/api/organization.go
@@ -6,12 +6,11 @@ import (
 )
 
 type Organization struct {
-	ID             uid.ID   `json:"id"`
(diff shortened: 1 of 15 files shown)


BruceMacD:
Deletion of identities/grant behavior actually means it would be better to keep this

PR #3840, opened 2022-12-09
Remove public keys when the user is deleted

dnephin:
## Summary

Branched from #3826, this PR fixes two problems:

1. we were never removing a users public keys. Now they are removed when the user is deleted.
2. the test cases in `TestProvisionSSHKey` depended on the previous test cases having run for the count of expected keys. Now that we have a way to delete the keys, the test cases no longer depend on each other.

Also two other small cleanups, renamed a data function to include the word `list` (to match other similar functions), and add a check that a `UserPublicKey` has a non-zero value for the public key before saving it.

### DIFF
--- a/internal/server/data/user_public_keys.go
+++ b/internal/server/data/user_public_keys.go
@@ -58,8 +58,8 @@ func AddUserPublicKey(tx WriteTxn, key *models.UserPublicKey) error {
 	return insert(tx, (*userPublicKeysTable)(key))
 }
 
-func deleteUserPublicKeys(tx WriteTxn, userID uid.ID) error {
-	stmt := "UPDATE user_public_keys SET deleted_at = ? WHERE user_id = ?"
+func DeleteUserPublicKeys(tx WriteTxn, userID uid.ID) error {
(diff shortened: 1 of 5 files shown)


jmorganca (approved):

dnephin (commented):
I added some more test coverage of `listUserPublicKeys` to show that it doesn't include deleted keys.

internal/server/data/user_public_keys.go
```diff
@@ -51,6 +52,14 @@ func AddUserPublicKey(tx WriteTxn, key *models.UserPublicKey) error {
...
 	case key.KeyType == "":
 		return fmt.Errorf("key type is required")
+	case key.PublicKey == "":
+		return fmt.Errorf("public key is required")
 	}
 	return insert(tx, (*userPublicKeysTable)(key))
 }
+
+func DeleteUserPublicKeys(tx WriteTxn, userID uid.ID) error {
+	stmt := "UPDATE user_public_keys SET deleted_at = ? WHERE deleted_at is null AND user_id = ?"
```
dnephin, line 62: This was missing `WHERE deleted_at is null`. I just added that now

PR #3842, opened 2022-12-09
fix: load google as a provider for identity

### LINKED ISSUE #3809: Social login not shown in user identity provider icons
## Describe the bug
When a user logs in via Google social login the "Users" page in the UI shows their provider as Infra only. No Google icon is shown.

## Expected behavior
The Google icon is displayed.

## Additional context
This feature is only in dev currently.

BruceMacD:
- pass context to list, get, and delete identities to store if google social login is enabled
- outer join provider users loading to return 0 provider ID records
- add google social provider to context

## Summary
Google social login was not being added as a name in user's identity provider name returned from the API. This was due to the `INNER JOIN` provider look-up not returning records with a 0  provider ID.

## Checklist

## Related Issues

Resolves #3809

### DIFF
--- a/internal/server/data/identity.go
+++ b/internal/server/data/identity.go
@@ -1,7 +1,6 @@
 package data
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"strings"
@@ -375,7 +374,7 @@ type ListIdentityOptions struct {
 	LoadPublicKeys         bool
 }
 
-func ListIdentities(c context.Context, tx ReadTxn, opts ListIdentityOptions) ([]models.Identity, error) {
+func ListIdentities(tx ReadTxn, opts ListIdentityOptions) ([]models.Identity, error) {
 	if len(opts.ByNotIDs) > 0 && opts.CreatedBy == 0 {
 		return nil, fmt.Errorf("ListIdentities by 'not IDs' requires 'created by'")
 	}
@@ -451,7 +450,7 @@ func ListIdentities(c context.Context, tx ReadTxn, opts ListIdentityOptions) ([]
 	}
 
 	if opts.LoadProviders {
-		if err := loadIdentitiesProviders(c, tx, result); err != nil {
+		if err := loadIdentitiesProviders(tx, result); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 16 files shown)


internal/server/data/identity.go
```diff
@@ -374,7 +375,7 @@ type ListIdentityOptions struct {
 	LoadPublicKeys         bool
 }
 
-func ListIdentities(tx ReadTxn, opts ListIdentityOptions) ([]models.Identity, error) {
+func ListIdentities(c context.Context, tx ReadTxn, opts ListIdentityOptions) ([]models.Identity, error) {
```
dnephin, line 378: There's a `context.Context` in our transaction. We could expose that with a `Context()` method, but I think we probably don't need to be handling this in `data`. We can include the google provider in the API handler.

internal/server/routes.go
```diff
@@ -250,6 +251,9 @@ func wrapRoute[Req, Res any](a *API, routeID routeIdentifier, route route[Req, R
 			Response:      &access.ResponseMetadata{},
 		}
 		c.Set(access.RequestContextKey, rCtx)
+		if a.server.Google != nil {
+			c.Set(data.GoogleProviderContextKey, a.server.Google)
```
dnephin, line 255: We need a way to include some `Server` data on `RequestContext`. I've been thinking a new field like `Server` would allow us to do that. That would remove the need to set anything on `gin.Context`. The API handler for `ListProviders` could read the field directly.

Here we would simply set `Server: a.server`.

dnephin (commented):
Does this need to be done in the `data` package? It seems like this required a lot of changes. 

What if we use add the provider from `API.ListProviders` instead, and let the data layer return empty values for the built-in google provider?

internal/server/routes.go
BruceMacD: Update to not use the context after our discussion

internal/server/data/identity.go
BruceMacD: Updated to not use the context after our discussion

dnephin (approved):
Nice! I guess the name and URL could also be constants, along with the ID. 

This looks great!

PR #3843, opened 2022-12-09
Remove some access functions that don't provide authorization

dnephin:
## Summary

This PR removes some functions from `access` package that were simply wrappers around a `data` function. These functions did not provide any authorization.

The purpose of the `access` package is to provide authorization, so a function in this package that doesn't provide authorization is misleading, and could actually be harmful. Someone reading the code might assume the operation is being authorized, when it is not.

There are still more functions in this package that aren't providing authorization, but I want to keep these PRs small, so I've limited the number being removed in each PR. Next steps are `Signup`, and the two for `PasswordReset`.

Related to #3866


internal/server/handlers.go
```diff
@@ -41,7 +41,7 @@ func (a *API) CreateToken(c *gin.Context, r *api.EmptyRequest) (*api.CreateToken
 			return nil, fmt.Errorf("%w: failed to update identity info from provider: %s", internal.ErrUnauthorized, err)
 		}
 
-		token, err := access.CreateToken(rCtx)
+		token, err := data.CreateIdentityToken(rCtx.DBTxn, rCtx.Authenticated.User.ID)
```
dnephin, line 44: TODO: add a comment at the top of this function to say why this doesn't require authorization

internal/server/destinations.go
```diff
@@ -79,7 +82,7 @@ func (a *API) UpdateDestination(c *gin.Context, r *api.UpdateDestinationRequest)
 	rCtx := getRequestContext(c)
 
 	// Start with the existing value, so that non-update fields are not set to zero.
-	destination, err := access.GetDestination(c, r.ID)
+	destination, err := data.GetDestination(rCtx.DBTxn, data.GetDestinationOptions{ByID: r.ID})
```
dnephin, line 85: This is an example of where the `access` package is not working well. Ideally we would authorize before doing any work, but we can see here that even though we were calling an `access` function first, the correct authorization does not happen until the call below here on line 98.

I think the API handlers should be authorizing first, either declaratively from something on the `route`, or imperatively as the first operation in the handler.

jmorganca (approved):

internal/server/handlers.go
jmorganca: If it's clear this route only requires authentication, then we probably don't even need the TODO since all users can call this for now.

PR #3845, opened 2022-12-09
fix: exit 1 if all attempts fail

mxyng:
## Summary

If all attempts to sync the image fails, the job should fail.

### DIFF
--- a/.github/workflows/cd-containers.yaml
+++ b/.github/workflows/cd-containers.yaml
@@ -114,5 +114,8 @@ jobs:
             argocd-image-updater run --once --match-application-label=ci.infrahq.com/component=${{ matrix.component.name }} && exit 0
             sleep $(( 3 ** $TRY ))
           done
+
+          # all attempts to sync image has failed
+          exit 1
         env:
           MAX_TRIES: 5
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-containers.yaml
```diff
@@ -114,5 +114,8 @@ jobs:
             argocd-image-updater run --once --match-application-label=ci.infrahq.com/component=${{ matrix.component.name }} && exit 0
             sleep $(( 3 ** $TRY ))
           done
+
+          # all attempts to sync image has failed
```
jmorganca, line 118:
```suggestion
          # all attempts to sync image have failed
```

dnephin (commented):
The fix makes sense to me, but there's something I don't understand about the rest of the script.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell says that the default shell is `bash -e`, so wouldn't the first failure cause it to exit before the `sleep`?

Have we seen the `for` loop work on retries?

mxyng:
> Have we seen the for loop work on retries?

Yes there was. The run logs don't load but this failed all attempts before "succeeding"

dnephin (approved):

PR #3848, opened 2022-12-09
Remove some access functions that don't provide authorization - 2

dnephin:
## Summary

Branched from #3843

More of the same, removing access functions that don't provide authorization.

Related to #3866


jmorganca:
LGTM once https://github.com/infrahq/infra/pull/3843 merges

dnephin:
The diff is not quite as nice anymore, because the API handler for signup moved to `internal/server/signup.go`, so now instead of the diff looking like a move of `internal/access/signup.go -> internal/server/signup.go` it looks like a merge of the files.

Thankfully git was mostly smart enough to figure out the merge, so all the changes that were made to `internal/access/signup.go` applied cleanly.

dnephin:
I verified the diff again by moving the file to `signup_user.go` which shows the minimal diff that looks like a file move. That's probably not a good option because we don't want a separate file for signup long term. So I'll keep it like this.

I resolve the conflict by applying the change from #3907

Issue #2796, opened 2022-08-02
api: replace most uses of gin.Context with RequestContext


dnephin:
PR #2702 introduced a new `RequestContext` type. This struct allows us to communicate details about the request using a strongly typed interface (as apposed to `gin.Context` which is mutable and uses a weakly typed `map[string]interface{}`).

This issue is to document and track converting most uses of `gin.Context` to the new `RequestContext`.

* [x] replace `gin.Context` with `RequestContext` as the first parameter to `access` functions
* [x] replace other uses of `gin.Context` in api handlers with methods or fields on `RequestContext`
* [ ] replace look ups on `API` fields in api handlers with methods or fields on the `RequestContext`
* [x] change `HandlerFunc` (the interface implemented by api handlers) to `func(c access.RequestContext, req *Req)` so that api handlers accept the type instead of looking it up from the gin.Context

When these tasks are complete we should only reference `gin.Context` in api middelware. All other code should be using either `RequestContext`, or accepting some interface that is implemented by `RequestContext`.

https://github.com/infrahq/internal/issues/19 has more context

dnephin:
We've made some good progress here, but still a little more to do.  We have our `RequestContext` type. We've started to replace `gin.Context` in some `access` functions with our new type.

Next steps would be to replaces it in all access functions and api handlers.

PR #3849, opened 2022-12-09
Remove gin.Context from handling of cookies

dnephin:
## Summary

Related to #2796

I found this on a stale branch, so thought I'd PR it. We were only using `gin.Context` as a way to get at the stdlib `http.Request` and `http.ResponseWriter`.

We could add `http.ResponseWriter` to our `RequestContext`, but it's used in so few places it might be better to handle it like this as separate args.

### DIFF
--- a/internal/server/cookie.go
+++ b/internal/server/cookie.go
@@ -5,8 +5,6 @@ import (
 	"net/url"
 	"time"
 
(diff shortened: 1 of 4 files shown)


jmorganca (approved):
Great fix!

PR #3850, opened 2022-12-09
add models for destination credentials

ssoroka:
## Summary

splitting up https://github.com/infrahq/infra/pull/3763, 1 of n

## Checklist


internal/server/data/destinationcredential.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	return r, nil
+}
+
+func CreateDestinationCredential(tx WriteTxn, cr *models.DestinationCredential) error {
+	q := querybuilder.New("INSERT INTO destination_credentials")
+	q.B("(id, organization_id, expires_at, user_id, destination_id, update_index)")
+	q.B("VALUES (?,?,?,?,?,nextval('seq_update_index'))", cr.ID, cr.OrganizationID, cr.ExpiresAt, cr.UserID, cr.DestinationID)
+
+	logging.Debugf("Should be triggering notification for credreq_%s_%s", cr.OrganizationID.String(), cr.DestinationID.String())
```
BruceMacD, line 32:
```suggestion
```

internal/server/data/destinationcredential.go
```diff
@@ -0,0 +1,95 @@
...
+	_, err := tx.Exec(q.String(), q.Args...) // will trigger destination_credential_insert_notify()
+	if err != nil {
+		return handleError(err)
+	}
+
+	return nil
+}
+
+func UpdateDestinationCredential(tx WriteTxn, cr *models.DestinationCredential) error {
```
BruceMacD, line 41: nit:
```suggestion
func AnswerDestinationCredential(tx WriteTxn, cr *models.DestinationCredential) error {
```

internal/server/models/destinationcredential.go
```diff
@@ -0,0 +1,48 @@
...
+	/////////////////////
+	// response fields //
+	/////////////////////
+
+	// // username & pw
+	// Username string
+	// Password string
+
+	// // API key
+	BearerToken sql.NullString
```
jmorganca, line 41: Should we encrypt these at rest? While ephemeral not all credentials may expire in minutes (e.g. some may expire in hours/days). Similarly, these should be deleted (hard deleted) once the user has claimed them or they expire, although I'm not sure that's in the scope of this PR

internal/server/data/credentialrequest_test.go
```diff
@@ -0,0 +1,139 @@
...
+		tx, err := db.Begin(context.Background(), nil)
+		assert.NilError(t, err)
+		tx = tx.WithOrgID(orgID)
+		defer func() { _ = tx.Rollback() }()
+
+		creds, err := ListDestinationCredentials(tx, destID)
+		assert.NilError(t, err)
+
+		// inject used updateIndex because we don't want to compare it
+		dc.UpdateIndex = creds[0].UpdateIndex
```
dnephin, line 117: I guess we should at least check this is not zero, right?

internal/server/data/migrations.go
```diff
@@ -1054,6 +1055,59 @@ func makeIdxEmailsProvidersUnique() *migrator.Migration {
...
+			_, err := tx.Exec(`
+				CREATE TABLE IF NOT EXISTS destination_credentials (
+					id 							 	bigint NOT NULL,
+					organization_id  	bigint NOT NULL,
+					expires_at 				timestamp with time zone,
+					update_index 			bigint NOT NULL,
+					user_id 					bigint NOT NULL,
+					destination_id 		bigint NOT NULL,
+					answered          bool NOT NULL DEFAULT false,
+					--
```
dnephin, line 1072: empty comment?

internal/server/data/migrations.go
```diff
@@ -1054,6 +1055,59 @@ func makeIdxEmailsProvidersUnique() *migrator.Migration {
...
+				);
+
+				CREATE INDEX IF NOT EXISTS idx_cred_req_org_dest on destination_credentials (organization_id, destination_id);
+
+				CREATE OR REPLACE FUNCTION destination_credential_insert_notify() RETURNS trigger
+					LANGUAGE PLPGSQL
+					AS $$
+				BEGIN
+					-- on insert, we notify connector listeners for this destination
+					PERFORM pg_notify(current_schema() || '.credreq_' || uidinttostr(NEW.organization_id) || '_' || uidinttostr(NEW.destination_id), NEW.id::TEXT);
```
dnephin, line 1083: The maximum postgres identifier length is 63 bytes: https://www.postgresql.org/docs/current/limits.html

Each of these ids can be up to 19 bytes when represented as a string (11 if we use the uid form).

I believe there are 10 other characters here to identify the channel and separate the ids.

That leaves 15 characters for the schema name if we used integer representation for the IDs, and 31 by using the uid representation.

That seems like a reasonable limit for the schema name.

internal/server/data/migrations.go
```diff
@@ -1054,6 +1055,59 @@ func makeIdxEmailsProvidersUnique() *migrator.Migration {
...
+				CREATE TRIGGER credreq_notify_insert_trigger AFTER insert
+					ON destination_credentials
+					FOR EACH ROW EXECUTE FUNCTION destination_credential_insert_notify();
+
+				CREATE OR REPLACE FUNCTION destination_credential_update_notify() RETURNS trigger
+					LANGUAGE PLPGSQL
+					AS $$
+				BEGIN
+					-- on update, we notify user listeners for this specific id, waiting to login
+					PERFORM pg_notify(current_schema() || '.credreq_' || uidinttostr(NEW.organization_id) || '_' || uidinttostr(NEW.id), NEW.id::TEXT);
```
dnephin, line 1097: Doesn't this channel need a different name? How is the code able to tell if the `.credreq_` prefix refers to a `<org_id>_<create_req_id>` (this notify) vs a `<org_id>_<dest_id>` (the above one).

I think we need a different channel name prefix. 

Since we renamed the type to DestinationCredentials, maybe `destcreds_id` vs `destcreds_dest` for the channel names?

internal/server/models/destinationcredential.go
```diff
@@ -0,0 +1,48 @@
...
+	// // ssh
+	// PublicKey []byte // supplied if the user is planning to connect via client-generated key pair
+
+	/////////////////////
+	// response fields //
+	/////////////////////
+
+	// // username & pw
+	// Username string
+	// Password string
```
dnephin, line 38: Let's track this in a github issue instead of large block comments in code. I think we can delete these lines, and the ones below under `Certificate` as well.

internal/server/data/migrations.go
ssoroka: yeah. just a logical separator.

internal/server/data/migrations.go
ssoroka: I hit this limit representing the integers as strings directly, thus the uid form.

internal/server/data/migrations.go
ssoroka: statistically these numbers are never going to collide, but different channel names makes sense too

internal/server/data/migrations.go
dnephin: I had to add the schema for testing, without it tests would conflict with each other because these channels are global and not scoped to a schema.  It's possible we can find a different way to make unique names.

internal/server/data/credentialrequest_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/opt"
+
+	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
+)
+
+func TestPgNotifyRequestWithTriggers(t *testing.T) {
```
dnephin, line 17: I think these test names (this one and the one below for response) should be something like `TestCreateDestinationCredential`.  When testnames don't reference the function they are testing it becomes really difficult to maintain the test suite, and hard to audit what coverage exists. It's true these are indirectly testing triggers and notify, but primarily what we care about from our application is that a create results in a notify, so I think it's really `TestCreateDestinationCredential`.

internal/server/data/migrations.go
```diff
@@ -81,8 +81,9 @@ func migrations() []*migrator.Migration {
 		addUserPubicKeysTable(),
 		addUserSSHLoginName(),
 		makeIdxEmailsProvidersUnique(),
+		addDestinationCredentials(),
```
dnephin, line 84: minor: putting this below `deviceFlowAuthRequestsAddUserIDProviderID` will ensure the migrations will be run in tests, in the same order as they did in dev and prod.

dnephin (commented):
LGTM, a couple minor suggestions, but nothing that should block a merge

dnephin (approved):

PR #3851, opened 2022-12-09
Add AuthenticationError to provide an error message in some cases

### LINKED ISSUE #3837: api: allow 401 status responses to provide an explanation
In https://github.com/infrahq/infra/blob/8c6e5f361e561de77c6de330f6c77499d9d54140/internal/server/errors.go#L44-L47

We always hide the full error message. This is generally a good idea for the reason in the comment (we don't want to leak information). But there are plenty of cases where we do want to provide more information about the error (https://github.com/infrahq/infra/pull/3739#discussion_r1043657913 is a great example).

We should create a custom error type (similar to `validate.Error`, and `data.UniqueConstraintError`), that allows us to send a 401 status code with a hard-coded message that we can validate to not contain anything sensitive.

dnephin:
Resolves #3837

This PR  adds `AuthenticationError` to send a bit more detail about authentication failures. In some cases we don't want to provide any information, but in a few cases we can safely provide a bit more information to tell the user why the request failed.

`internal.ErrUnauthorized` will continue to log the real error at `INFO` level and return no information in the response (other than the status code), where as `AuthenticationError` will log at `DEBUG` level (like other API errors), and provide some hard coded error message to the user. The few cases I found that were safe to return more information are:
1. no authentication was provided, we can say "authentication is required"
2. the access key expired, we can say that it was expired, because the key can not be used anymore

In a few cases I changed the error from `401` (not authenticated) to `403` (not authorized), because the error happens after the user was authenticated. In these cases we can safely tell the user why the operation failed, because they have already authenticated themselves.

### DIFF
--- a/internal/server/deviceflow.go
+++ b/internal/server/deviceflow.go
@@ -150,7 +150,7 @@ func (a *API) ApproveDeviceFlow(c *gin.Context, req *api.ApproveDeviceFlowReques
 	rctx := getRequestContext(c)
 
 	if !rctx.Authenticated.AccessKey.Scopes.Includes(models.ScopeAllowCreateAccessKey) {
-		return nil, fmt.Errorf("%w: access key missing scope '%s'", internal.ErrUnauthorized, models.ScopeAllowCreateAccessKey)
+		return nil, fmt.Errorf("%w: access key missing scope '%s'", access.ErrNotAuthorized, models.ScopeAllowCreateAccessKey)
 	}
 
 	dfar, err := data.GetDeviceFlowAuthRequest(rctx.DBTxn, data.GetDeviceFlowAuthRequestOptions{ByUserCode: strings.Replace(req.UserCode, "-", "", 1)})
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -32,6 +32,7 @@ func sendAPIError(c *gin.Context, err error) {
 	var uniqueConstraintError data.UniqueConstraintError
 	var authzError access.AuthorizationError
 	var overLimitError redis.OverLimitError
+	var authnError AuthenticationError
 	var apiError api.Error
 
 	log := logging.L.Debug()
@@ -47,6 +48,10 @@ func sendAPIError(c *gin.Context, err error) {
 		resp.Message = "unauthorized"
 		// log the error at info because it is not in the response
 		log = logging.L.Info()
+
+	case errors.As(err, &authnError):
+		resp.Code = http.StatusUnauthorized
+		resp.Message = authnError.Message
 
 	case errors.Is(err, data.ErrAccessKeyExpired):
 		resp.Code = http.StatusUnauthorized
@@ -137,4 +142,18 @@ func newAPIErrorForUniqueConstraintError(ucErr data.UniqueConstraintError, msg s
 		Errors:    []string{ucErr.Error()},
 	}}
 	return apiError
+}
+
+// AuthenticationError is used to respond with a 401 Unauthorized response code.
+// Unlike internal.ErrUnauthorized, AuthenticationError includes an error message
+// in the response.
+type AuthenticationError struct {
+	// Message is sent as the api.Error.Message in the response. Message should
+	// always be a hard coded string to ensure that it does not include any
+	// sensitive data.
+	Message string
+}
+
+func (a AuthenticationError) Error() string {
+	return a.Message
 }
--- a/internal/server/errors.go
+++ b/internal/server/errors.go
@@ -30,7 +30,6 @@ func sendAPIError(c *gin.Context, err error) {
 
 	var validationError validate.Error
 	var uniqueConstraintError data.UniqueConstraintError
-	var authzError access.AuthorizationError
 	var overLimitError redis.OverLimitError
 	var authnError AuthenticationError
 	var apiError api.Error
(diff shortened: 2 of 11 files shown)


internal/server/deviceflow.go
```diff
@@ -150,7 +150,7 @@ func (a *API) ApproveDeviceFlow(c *gin.Context, req *api.ApproveDeviceFlowReques
 	rctx := getRequestContext(c)
 
 	if !rctx.Authenticated.AccessKey.Scopes.Includes(models.ScopeAllowCreateAccessKey) {
-		return nil, fmt.Errorf("%w: access key missing scope '%s'", internal.ErrUnauthorized, models.ScopeAllowCreateAccessKey)
+		return nil, fmt.Errorf("%w: access key missing scope '%s'", access.ErrNotAuthorized, models.ScopeAllowCreateAccessKey)
```
dnephin, line 153: Changing from not authenticated here, to not authorized, because the user is an authorized admin, they're just using the wrong access key.

internal/server/errors.go
```diff
@@ -138,3 +142,19 @@ func newAPIErrorForUniqueConstraintError(ucErr data.UniqueConstraintError, msg s
...
+	// always be a hard coded string to ensure that it does not include any
+	// sensitive data.
+	Message string
+}
+
+func (a AuthenticationError) Error() string {
+	return a.Message
+}
+
+var errMissingAuthentication = AuthenticationError{Message: "authentication is required"}
```
dnephin, line 160: Suggestions on this wording? A user is going to see this in their API client (maybe CLI, maybe connector logs, maybe UI).

internal/server/errors.go
```diff
@@ -48,14 +48,18 @@ func sendAPIError(c *gin.Context, err error) {
...
+		resp.Code = http.StatusUnauthorized
+		resp.Message = authnError.Message
+
 	case errors.Is(err, data.ErrAccessKeyExpired):
 		resp.Code = http.StatusUnauthorized
 		// this means the key was once valid, so include some extra details
 		resp.Message = fmt.Sprintf("%s: %s", internal.ErrUnauthorized, err)
 
-	case errors.As(err, &authzError):
+	case errors.Is(err, access.ErrNotAuthorized):
```
dnephin, line 60: This now handles both `access.AuthorizationError` and `access.ErrNotAuthorized`, because `access.AuthorizationError` has a custom `Is` method that allows it to match `access.ErrNotAuthorized`.

internal/server/handlers.go
```diff
@@ -34,22 +34,21 @@ type API struct {
...
-			return nil, fmt.Errorf("%w: failed to update identity info from provider: %s", internal.ErrUnauthorized, err)
-		}
-
-		token, err := access.CreateToken(rCtx)
-		if err != nil {
-			return nil, err
-		}
+	if rCtx.Authenticated.User == nil {
+		return nil, errMissingAuthentication
+	}
```
dnephin, line 39: I flipped the condition here to make it a guard class, so the diff is not as easy to follow. The previous error message was at the end of the function.

internal/server/middleware.go
```diff
@@ -162,15 +162,15 @@ func requireAccessKey(c *gin.Context, db *data.Transaction, srv *Server) (access
 	accessKey, err := data.ValidateRequestAccessKey(db, bearer)
 	if err != nil {
 		if errors.Is(err, data.ErrAccessKeyExpired) {
-			return u, err
+			return u, AuthenticationError{Message: "access key has expired"}
```
dnephin, line 165: I think previously this would have returned a `500` response, because I don't see `data.ErrAccessKeyExpired` being handled by `sendAPIError`. Not it should be a 401 with this message to help the user understand why it's failing.

internal/server/middleware.go
```diff
@@ -162,15 +162,15 @@ func requireAccessKey(c *gin.Context, db *data.Transaction, srv *Server) (access
...
 		}
 		return u, fmt.Errorf("%w: invalid token: %s", internal.ErrUnauthorized, err)
 	}
 
 	if accessKey.Scopes.Includes(models.ScopePasswordReset) {
 		// PUT /api/users/:id only
 		if c.Request.URL.Path != "/api/users/"+accessKey.IssuedFor.String() || c.Request.Method != http.MethodPut {
-			return u, fmt.Errorf("%w: temporary passwords can only be used to set new passwords", internal.ErrUnauthorized)
+			return u, fmt.Errorf("%w: temporary passwords can only be used to set new passwords", access.ErrNotAuthorized)
```
dnephin, line 173: Changed to `403` (not authorized) because the user has authenticated successfully, but their access key only gives them permission for one operation.

internal/server/middleware.go
```diff
@@ -290,7 +290,7 @@ func reqBearerToken(c *gin.Context, opts Options) (string, error) {
 
 	// this will get caught by key validation, but check to be safe
 	if strings.TrimSpace(bearer) == "" {
-		return "", fmt.Errorf("%w: skipped validating empty token", internal.ErrUnauthorized)
+		return "", AuthenticationError{Message: "Authorization header is missing a bearer token"}
```
dnephin, line 293: This message will be sent to the user. It seems safe to let them know they have an invalid `Authorization` header, because that's user error and doesn't really leak any information.

internal/server/handlers.go
BruceMacD: this should never happen, it should fail at the authentication middleware, so I'm wondering if we should return a 500 error instead

dnephin: That seems reasonable to me. There are 2 other uses of `errMissingAuthentication` that I think are in the same category.

dnephin: Ok, I've made this change. These are now 500 errors, since they should only happen on programming errors.

BruceMacD (approved):
looks good 👍

Issue #1382, opened 2022-03-29
api: include user/group name in ListGrants response


dnephin:
The `ListGrants` api response includes a subject, but the subject is the ID of an identity.

In the connector we have to lookup group, user, group, and/or provider to get their names. Instead of requiring the API client to query for additional data we should include those names as part of the `ListGrants` response. 

I'm not sure what form this would take exactly. I imagine it might be easier to see after #1284 is done

dnephin:
This issue is may still be relevant. We did something like this for AccessKeys, we added `IssuedForName`.

PR #3863, opened 2022-12-12
fix issue with "unknown user" message

ssoroka:
## Checklist

## Related Issues

Resolves #

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -145,15 +145,13 @@ func userGrants(ctx context.Context, cli *CLI, client *api.Client, grants *[]api
 	rows := make([]row, 0, len(items))
 	for _, item := range items {
 		user, ok := mapUsers[item.User]
-		if !ok {
-			return 0, fmt.Errorf("unknown user for ID %v", item.ID)
+		if ok {
+			rows = append(rows, row{
+				User:     user.Name,
+				Role:     item.Privilege,
+				Resource: item.Resource,
+			})
 		}
-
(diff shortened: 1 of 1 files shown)


dnephin:
> We remove the grants if the last user provider & placeholder(models.Identity) is removed.

I believe we already do that here: https://github.com/infrahq/infra/blob/5fd5af7640645a888b135582f8581bf86512c7d0/internal/server/data/identity.go#L674-L685

I guess it's always possible that a user is removed after the first API call to list grants, but before the next API call to list users.

Maybe we should use the userID in the list output if we can't lookup the user name?

ssoroka:
I think most people on the team have run into this, so it's not a race condition. it's possible that it's from before we were deleting grants–I'm not sure how old that code is–but we should probably still handle this somehow if we're already deleting them and people are still managing to hit this scenario. Ignoring the record does seem to make some sense, as it's a grant that doesn't have an associated user/group, which is worthless.

dnephin:
If the problem is inconsistent data, I think we should clean that up instead of trying to patch it in the CLI. Hiding things from output, especially things that are related to access control, seems like a risky approach. It could mislead someone looking at the output to believe access doesn't exist, or make it harder for us to find other bugs in the future because they are being masked by this change.

ssoroka:
It's not really a race. Definitely #1382 will help, but you'll have the same problem: some grants won't join to user records because they're simply not there.

BruceMacD (approved):

PR #3864, opened 2022-12-12
feat: set allowed domains in the UI

BruceMacD:
- Add organization update endpoint
- Add "allowed domains" configuration to the UI.
- Add slice validation back to API
- Change "providers" settings tab to "authentication"

## Summary
Allow org admins to update their org's allowed social login domains.

When no social login is configured (self-hosted) this input is hidden.

📸
Screen Shot 2022-12-22 at 12 01 02 PM

## Checklist

## Related Issues

Part of #3261

### DIFF
--- a/api/organization.go
+++ b/api/organization.go
@@ -51,11 +51,13 @@ func (r UpdateOrganizationRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
 		validate.Required("id", r.ID),
 		validate.Required("allowedDomains", r.AllowedDomains),
-		validate.StringSliceRule{
+		// permissive validation for a domain field (with no protocol)
+		validate.SliceRule{
 			Value: r.AllowedDomains,
 			Name:  "allowedDomains",
 			ItemRule: validate.StringRule{
 				Name:      "allowedDomains.values",
+				MinLength: 2,
 				MaxLength: 254,
 				CharacterRanges: []validate.CharRange{
 					validate.AlphabetLower,
@@ -66,23 +68,6 @@ func (r UpdateOrganizationRequest) ValidationRules() []validate.ValidationRule {
 				},
 				FirstCharacterRange: validate.AlphaNumeric,
 			},
-		},
-	}
-}
-
-// ValidateDomain returns a permissive validation for a domain field (with no protocol).
-func ValidateDomain(name, value string) validate.StringRule {
-	return validate.StringRule{
-		Value:     value,
-		Name:      name,
-		MinLength: 2,
-		MaxLength: 256, // arbitrary max length
-		CharacterRanges: []validate.CharRange{
-			validate.AlphabetLower,
-			validate.AlphabetUpper,
-			validate.Numbers,
-			validate.Dash,
-			validate.Dot,
 		},
 	}
 }
--- a/api/organization.go
+++ b/api/organization.go
@@ -67,6 +67,7 @@ func (r UpdateOrganizationRequest) ValidationRules() []validate.ValidationRule {
 					validate.Dot,
 				},
 				FirstCharacterRange: validate.AlphaNumeric,
+				RequiredCharacters:  []rune{'.'},
 			},
 		},
 	}
--- a/api/organization.go
+++ b/api/organization.go
@@ -42,6 +42,51 @@ func (r CreateOrganizationRequest) ValidationRules() []validate.ValidationRule {
 	}
 }
 
+type UpdateOrganizationRequest struct {
+	ID             uid.ID   `uri:"id" json:"-"`
+	AllowedDomains []string `json:"allowedDomains"`
+}
+
+func (r UpdateOrganizationRequest) ValidationRules() []validate.ValidationRule {
+	return []validate.ValidationRule{
+		validate.Required("id", r.ID),
+		validate.Required("allowedDomains", r.AllowedDomains),
+		validate.StringSliceRule{
+			Value: r.AllowedDomains,
+			Name:  "allowedDomains",
+			ItemRule: validate.StringRule{
+				Name:      "allowedDomains.values",
+				MaxLength: 254,
+				CharacterRanges: []validate.CharRange{
+					validate.AlphabetLower,
+					validate.AlphabetUpper,
+					validate.Numbers,
+					validate.Dash,
+					validate.Dot,
+				},
+				FirstCharacterRange: validate.AlphaNumeric,
+			},
+		},
+	}
+}
+
+// ValidateDomain returns a permissive validation for a domain field (with no protocol).
+func ValidateDomain(name, value string) validate.StringRule {
+	return validate.StringRule{
+		Value:     value,
+		Name:      name,
+		MinLength: 2,
+		MaxLength: 256, // arbitrary max length
+		CharacterRanges: []validate.CharRange{
+			validate.AlphabetLower,
+			validate.AlphabetUpper,
+			validate.Numbers,
+			validate.Dash,
+			validate.Dot,
+		},
+	}
+}
+
 func (req ListOrganizationsRequest) SetPage(page int) Paginatable {
 	req.PaginationRequest.Page = page
 	return req
--- a/internal/access/organization.go
+++ b/internal/access/organization.go
@@ -89,10 +89,5 @@ func UpdateOrganization(c *gin.Context, org *models.Organization) error {
 		}
 		return data.UpdateOrganization(db, org)
 	}
-	roles := []string{models.InfraSupportAdminRole}
-	err := IsAuthorized(rCtx, roles...)
-	if err != nil {
-		return HandleAuthErr(err, "organization", "update", roles...)
-	}
-	return data.UpdateOrganization(rCtx.DBTxn, org)
+	return fmt.Errorf("%w: %s", ErrNotAuthorized, "you may only update your own organization")
 }
--- a/internal/access/organization.go
+++ b/internal/access/organization.go
@@ -78,3 +78,21 @@ func DomainAvailable(c *gin.Context, domain string) error {
 	}
 	return data.UniqueConstraintError{Table: "organization", Column: "domain", Value: domain}
 }
+
+func UpdateOrganization(c *gin.Context, org *models.Organization) error {
+	rCtx := GetRequestContext(c)
+	if user := rCtx.Authenticated.User; user != nil && user.OrganizationID == org.ID {
+		// admins may update their own org
+		db, err := RequireInfraRole(c, models.InfraAdminRole)
+		if err != nil {
+			return HandleAuthErr(err, "organization", "update", models.InfraAdminRole)
+		}
+		return data.UpdateOrganization(db, org)
+	}
+	roles := []string{models.InfraSupportAdminRole}
+	err := IsAuthorized(rCtx, roles...)
+	if err != nil {
+		return HandleAuthErr(err, "organization", "update", roles...)
+	}
+	return data.UpdateOrganization(rCtx.DBTxn, org)
+}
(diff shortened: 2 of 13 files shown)


ui/pages/settings/index.js
```diff
@@ -757,9 +1057,9 @@ export default function Settings() {
             render: <Admins />,
           },
           {
-            name: 'providers',
-            title: 'Identity Providers',
-            render: <Providers />,
+            name: 'authentication',
```
BruceMacD, line 1060: I re-named the 'providers' tab here. Let me know if this naming seems ok.

internal/server/organizations.go
```diff
@@ -66,3 +66,28 @@ func (a *API) CreateOrganization(c *gin.Context, r *api.CreateOrganizationReques
...
 }
+
+func (a *API) UpdateOrganization(c *gin.Context, r *api.UpdateOrganizationRequest) (*api.Organization, error) {
+	org, err := access.GetOrganization(c, r.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	// overwrite the existing domains to the incoming ones
+	org.AllowedDomains = []string{}
```
jmorganca, line 77: A nice layer of security other tools do for this is _only_ allow domains that existing users have emails for. This avoids adding a random domain on accident. Should we consider that here?

ui/pages/settings/index.js
```diff
@@ -657,19 +661,314 @@ function Admins() {
   )
 }
 
-function Providers() {
+function AddAllowedDomainDialog({ setOpen, org }) {
```
jmorganca, line 664: Instead of a dialog (too many modal dialogs can be exhausting!), we can use a simple "tag" (or "badge") input box here.

mxyng:
It seems odd to me this is a top level configuration but it's only applicable to Google. Ideally, an admin should enable the Google social login and configure allowed domains through an `add identity provider` work flow

internal/server/organizations.go
BruceMacD: This creates a bit of a weird flow where you need to invite a user before you can allow their domain. Happy to re-visit this though.

ui/pages/settings/index.js
BruceMacD: Updated

internal/access/organization.go
```diff
@@ -62,3 +62,21 @@ func DeleteOrganization(c *gin.Context, id uid.ID) error {
...
+	rCtx := GetRequestContext(c)
+	if user := rCtx.Authenticated.User; user != nil && user.OrganizationID == org.ID {
+		// admins may update their own org
+		db, err := RequireInfraRole(c, models.InfraAdminRole)
+		if err != nil {
+			return HandleAuthErr(err, "organization", "update", models.InfraAdminRole)
+		}
+		return data.UpdateOrganization(db, org)
+	}
+	roles := []string{models.InfraSupportAdminRole}
```
jmorganca, line 76: We can probably omit `support-admin` support in here for now and just return an error at this point

ui/pages/settings/index.js
```diff
@@ -657,19 +658,101 @@ function Admins() {
...
+              </h2>
+              <h3 className='text-sm text-gray-500'>
+                Google accounts with these domains are able to log in to your
+                organization. They will not have any infrastructure access by
+                default.
+              </h3>
+            </div>
+          </header>
+          <div className='mt-3 flex min-h-0 flex-1 flex-col'>
+            <TagsInput
```
jmorganca, line 723: The tag here needs less vertical padding.

jmorganca, line 723: We can avoid using a component here:

```
// Style this div like an input
// onSubmit (pressing enter) should add a tag
<form className='flex' onSubmit={...}>
{domains.map(d => (
  // render domains as a "tag" with an x
  <div>
    ...
  </div>
)}

// the actual input takes up the rest of the space
// backspace should remove a tag
<input className='bg-transparent focus:outline-none' onKeyDown={...} />
</form>
```

BruceMacD: made this change

ui/pages/settings/index.js
mxyng: Do you see this as the place to set things like password requirements?

api/organization.go
```diff
@@ -42,6 +42,51 @@ func (r CreateOrganizationRequest) ValidationRules() []validate.ValidationRule {
 	}
 }
 
+type UpdateOrganizationRequest struct {
```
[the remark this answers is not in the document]
BruceMacD: Agreed!

internal/access/organization.go
```diff
@@ -78,3 +78,16 @@ func DomainAvailable(c *gin.Context, domain string) error {
...
+	rCtx := GetRequestContext(c)
+	if user := rCtx.Authenticated.User; user != nil && user.OrganizationID == org.ID {
+		// admins may update their own org
+		db, err := RequireInfraRole(c, models.InfraAdminRole)
+		if err != nil {
+			return HandleAuthErr(err, "organization", "update", models.InfraAdminRole)
+		}
+		return data.UpdateOrganization(db, org)
+	}
+	return fmt.Errorf("%w: %s", ErrNotAuthorized, "you may only update your own organization")
```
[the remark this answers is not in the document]
BruceMacD: Not sure on that one, this was an existing error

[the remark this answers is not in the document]
dnephin: `Unauthorized` is actually "not authenticated" (401). I think this is correct that it's 403 Forbidden.

dnephin (approved):
API changes LGTM! Nice work on the string slice validator and the test coverage for the API endpoint!

I'm not able to review the UI changes, and I don't have a lot of context around this feature, but I think being able to update allowed domains makes sense.

api/organization.go
```diff
@@ -42,6 +42,51 @@ func (r CreateOrganizationRequest) ValidationRules() []validate.ValidationRule {
...
+					validate.Dot,
+				},
+				FirstCharacterRange: validate.AlphaNumeric,
+			},
+		},
+	}
+}
+
+// ValidateDomain returns a permissive validation for a domain field (with no protocol).
+func ValidateDomain(name, value string) validate.StringRule {
```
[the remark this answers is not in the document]
BruceMacD: This was no longer used, removed it

internal/validate/slice.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (s StringSliceRule) Validate() *Failure {
+	if len(s.Value) == 0 {
+		return nil
+	}
+
+	for _, val := range s.Value {
+		if strings.ContainsAny(val, ",") {
+			return Fail(s.Name, "list values cannot contain commas")
+		}
```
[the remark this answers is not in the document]
BruceMacD: Correct, removed this too

internal/validate/slice.go
```diff
@@ -0,0 +1,42 @@
...
+		s.ItemRule.Value = val
+		if fail := s.ItemRule.Validate(); fail != nil {
+			return fail
+		}
```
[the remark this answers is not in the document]
BruceMacD: I made this change, seems nicer, if its found to be an issue for clients we can remove it later

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+                default.
+              </h3>
+            </div>
+          </header>
+          <form
+            className='group form-input flex w-full rounded border-gray-300 py-2 px-3 leading-tight focus-within:border-blue-500 focus-within:ring-blue-500'
+            onSubmit={addDomain}
+          >
+            {allowedDomains.map(d => (
+              // render domains as a "tag" with an x
```
jmorganca, line 766:
```suggestion
```

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+              <span
+                className='mr-1 inline-block rounded bg-gray-200 py-1 px-2 pl-4 pr-2 text-xs font-medium'
```
jmorganca, line 768:
```suggestion
                className='mr-1 inline-block rounded bg-gray-200 py-1 px-2 pl-4 pr-1 text-xs font-medium'
```

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+                key={d}
+              >
+                {d}
+                <button
+                  className='px-2 font-normal hover:text-blue-600'
```
jmorganca, line 773:
```suggestion
                  className='px-1.5 font-normal'
```

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+              </h2>
+              <h3 className='text-sm text-gray-500'>
+                Google accounts with these domains are able to log in to your
+                organization. They will not have any infrastructure access by
+                default.
+              </h3>
+            </div>
+          </header>
+          <form
+            className='group form-input flex w-full rounded border-gray-300 py-2 px-3 leading-tight focus-within:border-blue-500 focus-within:ring-blue-500'
```
jmorganca, line 762:
```suggestion
            className='group form-input flex w-full rounded border-gray-300 py-2 px-3 leading-tight focus-within:border-blue-500 focus-within:ring-blue-500 focus-within:ring-1'
```

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+                Allowed Domains
+              </h2>
+              <h3 className='text-sm text-gray-500'>
+                Google accounts with these domains are able to log in to your
+                organization. They will not have any infrastructure access by
+                default.
+              </h3>
+            </div>
+          </header>
+          <form
```
jmorganca, line 761: Nice work with this (vs using a new dependency)

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+  const [error, setError] = useState('')
+
+  async function removeDomain(e) {
+    e.preventDefault()
+    let toRemove = e.target.value
+    const newAllowedDomains = allowedDomains.filter(d => d !== toRemove)
+    updateAllowedDomains(newAllowedDomains)
+  }
+
+  async function addDomain(e) {
```
jmorganca, line 696: Typing a non domain seems to work (e.g. `hi`). Should it? We may want to use a regex approach like [this](https://miguelmota.com/bytes/validate-domain-regex/), or validate this more server side

ui/pages/settings/index.js
jmorganca, line 761: Only the blue part the image below is click-eable to focus the input. Ideally it should be the entire "input box"

Screenshot 2023-01-04 at 10 46 52 PM

ui/pages/settings/index.js
```diff
@@ -675,19 +675,138 @@ function Admins() {
...
+        <>
+          <header className='my-6 flex flex-col justify-between md:flex-row md:space-y-0 md:space-x-4'>
+            <div>
+              <h2 className='mb-0.5 flex items-center font-display text-lg font-medium'>
+                Allowed Domains
+              </h2>
+              <h3 className='text-sm text-gray-500'>
+                Google accounts with these domains are able to log in to your
+                organization. They will not have any infrastructure access by
+                default.
```
jmorganca, line 757: Optional: the second sentence seems like the default expectation for the user and can be left out

```suggestion
              <h3 className='text-sm text-gray-500'>
                Users with these domains are able to log in to your
                organization.
```

ui/pages/settings/index.js
BruceMacD: Regex in our server-side validation doesn't work very well because it can't return a specific error message like other validations do, just that the pattern doesn't match. I opted to require a `.` for now, which is pretty permissive but should cover most invalid inputs.

ui/pages/settings/index.js
BruceMacD: Fixed using `grow`

ui/pages/settings/index.js
jmorganca: Sounds good. Did not mean we should use regex server side (it might be a good option client side though)

PR #3865, opened 2022-12-12
Add end-to-end testing of SSH integration

### LINKED ISSUE #3813: Add end-to-end test coverage for SSH destinations
We're working on a few changes for SSH destinations (`infra sshd auth-keys`, SSH connector, rpm/deb packaging for SSH connector, and `infra connector --setup`), all of which are difficult to test with unit tests because they depend on the environment where they are being run:
1. many of these require being run as the `root` user
2. some logic may work on one Linux distro and not others, depending on available packages, filesystem layout, standard users, etc.
3. testing it requires multiple processes, and multiple hosts (client, api server, and destination)

We should continue to write unit tests where possible, but we also need some test coverage of the whole feature to prevent some of the bugs we found last week.

### Proposal

We should create a separate test suite, in this repo, for end-to-end testing SSH destinations. The test suite should run on pushes to `main` and when a github PR has the `action/test-e2e` label (this is the same strategy we use for fuzzing and `-race`). In the future we can add other e2e tests to this suite, or create separate suites using the same pattern (whatever we find to be easier given the differences in test setup).

This test suite will require a very specific environment, so it shouldn't run as part of `go test ./...`. There are a few ways to accomplish that (ex: gate using build tags, or env vars). I generally prefer to make the suite a separate Go module. That allows someone to run all the tests using Go workspaces if they want to.

The test suite should do roughly the following:

1. Build the `infra binary`, and rpm/deb packages
2. Run containers for Infra server, postgres, and one per linux distro we want to test.
3. Once the containers have started the CI job should run the test suite to check:
     1. the destination has been registered
     2. a user without a grant fails to connect
     3. a user with a grant is able to login to infra and `ssh` to the destination.
     4. when the grant is removed the user is no longer able to connect

Outside of CI we can probably use `docker-compose` to setup the service containers, and run the same test suite by pointing at those containers.

### Considerations 

* We could run all the service containers in a single job, and have the test suite loop and check each. Maybe that would be better than a matrix that runs a separate DB and server in each?
* We could launch real cloud instances instead of containers for the SSH destination. I'd like to avoid that if possible because it adds a lot of runtime, it requires adding cloud credentials to our CI which makes it impossible to run on PRs from forks, and it's a lot more difficult to run outside of CI. We should be able to get pretty reasonable coverage by providing a no-op bash script for `systemctl`.

dnephin:
Closes #3813

This PR adds a new test suite for testing the Linux packages, and SSH connector. See the linked issue for details.


test/setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
```
BruceMacD, line 2: Will more tests go here? As far as I can tell this shell script only tests ubuntu

dnephin: Ya, we'll want to add more `docker exec` calls to this file to setup each of the destinations before we run tests.

I think we'll want to do that in parallel, so maybe this script will be replaced by a Go script when that happens.

BruceMacD (approved):
seems good 👍

PR #3869, opened 2022-12-13
Only delete access keys that are not already deleted

dnephin:
## Summary

Add the missing `WHERE deleted_at is null` clause to two access key queries, and add test coverage for this case.

I went through the other queries, and I believe all the rest are handled correctly by either querying for IDs using `List` (which already filters out deleted items), by including `WHERE deleted_at is null`, or by doing a true delete with `DELETE`.


pdevine (approved):
LGTM.

BruceMacD (approved):

PR #3870, opened 2022-12-13
fix: use a non-zero id for google

### LINKED ISSUE #3846: Set Google Social Login Provider ID to a non-zero default value
## Describe the bug
The Google Social Login Provider ID defaults to a zero value, this could cause some unexpected behavior on empty provider checks. It should default to a unique constant value.

### LINKED ISSUE #3867: Personal access keys are being deleted before they've expired or manually deleted
Manually created access keys (personal access keys) are being unexpectedly deleted. The keys were created with an expiry of 1 year and an inactivity timeout 30 days but was deleted automatically after at most 1 day.

BruceMacD:
- set google id to a constant value

- migrate old google social login users and keys to new id

- do not allow a provider user to be created without a provider id or identity id

- add exported provider to oidc client so functions using this client dont have to look-up the provider in the database

## Summary
For google social login set the ID to a default value. This revealed an issue where social sessions were not correctly synced due to their zero ID (access keys were being set with the infra provider ID as a default). This PR also fixes this behavior.

This should resolve the issue with all access keys being revoked on google social login session expiry.

## Checklist

## Related Issues

Resolves #3867
Resolves #3846

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -86,19 +86,25 @@ func ListIdentities(c *gin.Context, opts data.ListIdentityOptions) ([]models.Ide
 	return data.ListIdentities(db, opts)
 }
 
-func GetContextProviderIdentity(c RequestContext) (*models.Provider, string, error) {
+func GetContextProviderIdentity(c RequestContext, google *models.Provider) (*models.Provider, string, error) {
 	// does not need authorization check, this action is limited to the calling user
-	provider, err := data.GetProvider(c.DBTxn, data.GetProviderOptions{
-		ByID: c.Authenticated.AccessKey.ProviderID,
-	})
-	if err != nil {
-		return nil, "", fmt.Errorf("user info provider: %w", err)
-	}
+	var provider *models.Provider
+	if c.Authenticated.AccessKey.ProviderID == google.ID {
+		provider = google
+	} else {
+		var err error
+		provider, err = data.GetProvider(c.DBTxn, data.GetProviderOptions{
+			ByID: c.Authenticated.AccessKey.ProviderID,
+		})
+		if err != nil {
+			return nil, "", fmt.Errorf("user info provider: %w", err)
+		}
 
-	if provider.Kind == models.ProviderKindInfra {
-		// no external verification needed
-		logging.L.Trace().Msg("skipped verifying identity within infra provider, not required")
-		return provider, "", nil
+		if provider.Kind == models.ProviderKindInfra {
+			// no external verification needed
+			logging.L.Trace().Msg("skipped verifying identity within infra provider, not required")
+			return provider, "", nil
+		}
 	}
 
 	identity := c.Authenticated.User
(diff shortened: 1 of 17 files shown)


internal/access/identity.go
```diff
@@ -86,19 +86,25 @@ func ListIdentities(c *gin.Context, opts data.ListIdentityOptions) ([]models.Ide
...
 	// does not need authorization check, this action is limited to the calling user
-	provider, err := data.GetProvider(c.DBTxn, data.GetProviderOptions{
-		ByID: c.Authenticated.AccessKey.ProviderID,
-	})
-	if err != nil {
-		return nil, "", fmt.Errorf("user info provider: %w", err)
-	}
+	var provider *models.Provider
+	if c.Authenticated.AccessKey.ProviderID == google.ID {
+		provider = google
```
BruceMacD, line 93: skip looking this up in the database

internal/access/identity_test.go
```diff
@@ -197,10 +213,34 @@ func TestUpdateIdentityInfoFromProvider(t *testing.T) {
 		assert.NilError(t, err)
 	})
 
+	t.Run("a revoked OIDC session revokes access keys created by social login", func(t *testing.T) {
```
BruceMacD, line 216: This tests #3867

internal/access/identity.go
```diff
@@ -86,19 +86,25 @@ func ListIdentities(c *gin.Context, opts data.ListIdentityOptions) ([]models.Ide
 	return data.ListIdentities(db, opts)
 }
 
-func GetContextProviderIdentity(c RequestContext) (*models.Provider, string, error) {
+func GetContextProviderIdentity(c RequestContext, google *models.Provider) (*models.Provider, string, error) {
```
jmorganca, line 89: should this parameter be `social *models.Provider` ?

BruceMacD: When we support more forms of social login we will have to update a bunch of areas to take a list of social login providers rather than one, so as long as its just Google I'll leave it explicit to make it easier to clean up later.

internal/server/providers/oidc.go
```diff
@@ -43,6 +42,7 @@ type OIDCClient interface {
 	ExchangeAuthCodeForProviderTokens(ctx context.Context, code string) (*IdentityProviderAuth, error)
 	RefreshAccessToken(ctx context.Context, providerUser *models.ProviderUser) (accessToken string, expiry *time.Time, err error)
 	GetUserInfo(ctx context.Context, providerUser *models.ProviderUser) (*UserInfoClaims, error)
+	Provider() *models.Provider // returns the provider this client was created for
```
dnephin, line 45: I could be wrong, but it seems like we only ever use the `ProviderID` from this return value. The one place we pass the full model (`AssignIdentityToGroups`) ends up only using the ID.

Could we make this `ProviderID() uid.ID` and change `AssignIdentityToGroups` to accept just the ID? Or do you anticipate needing more than the ID in the future?

internal/access/identity.go
dnephin: This `GetContextProviderIdentity` function only has 1 caller, right?

The logic is a bit awkward right now because we pass in a provider , but only use it in some cases and we could have figured out that case from the caller.

If we moved these 2 lines and the `GetProvider` logic below to the caller I think it would be easier to follow:

```go
var provider *models.Provider
if rCtx.Authenticated.AccessKey.ProviderID == models.InternalGoogleProviderID {
    provider = a.Server.google
} else {
    var err error
    provider, err = data.GetProvider(...)
    ...
}

// This branch is in two places right now, it could be in only one of those places if we move the logic to the caller
if provider.Kind == models.ProviderKindInfra {
    ...
}

providerUser, err := data.GetProviderUser(c.DBTxn, c.Authenticated.AccessKey.ProviderID, identity.ID)
if err != nil {
    return nil, "", err
}
```

This would allow us to remove `access.GetContextProviderIdentity` entirely.

I know why we didn't do that originally, because we have this rule about "`data` functions can only be called from `access`". This is a great example of how that rule is causing is to write code that is harder to read and maintain.

The code you have here looks correct to me, so I'm ok if we want to explore that change in a separate PR.

dnephin (commented):
Looks good! One comment about the `access` package that I think we can address in a follow up, and one question about the provider in `oidcClient.ProviderModel`.

internal/server/providers/oidc.go
BruceMacD: You're right, even further we didnt need the provider ID on the client at all after making some of your suggested changes. It was a workaround for some spots where I needed to lookup the provider that dont exist now after more refactoring. Removed this.

internal/server/providers/oidc.go
```diff
@@ -61,16 +60,15 @@ func WithOIDCClient(ctx context.Context, client OIDCClient) context.Context {
 }
 
 type oidcClientImplementation struct {
-	ProviderID   uid.ID
-	Domain       string
-	ClientID     string
-	ClientSecret string
-	RedirectURL  string
+	ProviderModel models.Provider
```
dnephin, line 63:
```suggestion
```

Looks like this isn't used, but I think accepting this suggestion will require a `gofmt`.

dnephin (approved):
LGTM!

internal/server/providers/oidc.go
BruceMacD: good eye, thanks

PR #3871, opened 2022-12-13
feat: Access control

hoyyeva:
## Summary
- adding an access control table to the UI which will list all the grants in the app
- able to delete and add grants on the table as well 
- admin table to have consistent look as other table


jmorganca:
This is a fantastic change. A few early comments:

* While this can be the last change, this access control page should fully replace `/destinations/:id` pages Screenshot 2022-12-14 at 12 10 32 AM
* When there's no grants in the system the "check all" seems to be checked: Screenshot 2022-12-14 at 12 11 14 AM
* Having the "remove buttons" without confirmation is a bit startling, while I usually don't think it's necessary, this is a case where it might avoid deleting by accident: 
* Suggest renaming button to "Grant access" Screenshot 2022-12-14 at 12 12 24 AM
* This input would be even better if it would open the dropdown list by default: Screenshot 2022-12-14 at 12 12 50 AM
* Suggest we rename this to "infrastructure" Screenshot 2022-12-14 at 12 13 23 AM
* This is such a neat idea, I'm excited to see users use it! Note I think a confirmation dialog would be helpful here too: Screenshot 2022-12-14 at 12 13 59 AM
* Separating "infrastructure" and "resource" into two columns would make for a more clear table instead of showing the dot syntax. We'll most likely move away from this anyways:
    Screenshot 2022-12-14 at 12 14 43 AM

jmorganca:
Almost there!

1. Ideally this dropdown only appears when _clicking_ on the input vs when the modal appears

Screenshot 2022-12-19 at 5 17 32 PM

2. This defaults to the first destination but ideally is also blank until clicked on

Screenshot 2022-12-19 at 5 18 09 PM

3. Since this is optional, we should, as lightly as possible, signal this to the user. A simple idea could be to put `(optional)` in the input label

Screenshot 2022-12-19 at 5 18 45 PM

4. Capitalize `User`:
Screenshot 2022-12-19 at 5 20 03 PM

5. When removing, we should call this "removing access":

Screenshot 2022-12-19 at 5 20 57 PM

6. In the grant access modal, we can remove:
Screenshot 2022-12-19 at 5 21 27 PM

jmorganca:
A few small nits:

1. `Remove selected`:
Screenshot 2022-12-20 at 11 47 50 PM

2. `Group or User` as a placeholder here would be more consistent:

Screenshot 2022-12-20 at 11 48 22 PM

3. For this one, we could have no placeholder or just `infrastructure`:

Screenshot 2022-12-20 at 11 48 52 PM

4. Clicking remove on a single will "select" that row which seems a bit strange



5. "Are you sure you want to remove the selected access?"

Screenshot 2022-12-20 at 11 53 19 PM

6. It would be really helpful for when we have SSH to show the "kind" of destination in the grant dialog:

Screenshot 2022-12-20 at 11 55 27 PM

7. While we don't have to solve it in this PR, ideally roles is optional too. Not specifying a role will just give the `connect` role

Screenshot 2022-12-20 at 11 56 09 PM

ui/pages/access-control/index.js
```diff
@@ -0,0 +1,670 @@
...
+    if (roles.length > 0) {
+      setSelectedRoles([roles[0]])
+    }
+  }, [roles])
+
+  async function onSubmit(e) {
+    e.preventDefault()
+
+    try {
+      const GrantsToAdd = []
```
jmorganca, line 83:
```suggestion
      const grantsToAdd = []
```

ui/components/layouts/dashboard.js
```diff
@@ -133,6 +140,7 @@ export default function Dashboard({ children }) {
         <div className='mt-5 h-0 flex-1 overflow-y-auto'>
           <nav className='flex-1 space-y-1'>
             {navigation
+              ?.filter(n => n.hide !== true)
```
jmorganca, line 143:
```suggestion
              ?.filter(n => !n.hide)
```

ui/components/table.js
```diff
@@ -12,13 +14,17 @@ export default function Table({
   data,
   href,
   empty = 'No data',
+  deleteText = 'Remove selected',
```
jmorganca, line 17: We can remove this prop for now and just use 'Remove selected' for all tables

ui/components/table.js
```diff
@@ -41,12 +47,67 @@ export default function Table({
...
+  }, [selectedRowIds])
+
+  useLayoutEffect(() => {
+    setCheckedAll(false)
+    setIndeterminate(false)
+    setSelectedRowIds([])
+  }, [pageIndex])
+
+  function toggleAll() {
+    setSelectedRowIds(checkedAll || indeterminate ? [] : data.map(d => d.id))
```
jmorganca, line 73:
```suggestion
    setSelectedRowIds(checkedAll || indeterminate ? [] : data?.map(d => d.id))
```

ui/pages/access-control/index.js
```diff
@@ -0,0 +1,760 @@
...
+                    <Listbox
+                      value={selectedNamespaces}
+                      onChange={v => {
+                        setError({})
+                        if (v.includes(OPTION_SELECT_ALL)) {
+                          if (
+                            selectedNamespaces.length !==
+                            selectedResource?.resources?.length
+                          ) {
+                            if (selectedResource !== undefined) {
```
jmorganca, line 287:
```suggestion
                            if (selectedResource) {
```

We _may_ be able to just check if this is truth-y

ui/pages/destinations/[id]/index.js
```diff
@@ -1,971 +0,0 @@
-import { useEffect, useState, Fragment, useRef } from 'react'
```
Comment (jmorganca) [inline ui/pages/destinations/[id]/index.js:1]: Do we want to keep this file for now, if we are "silently" shipping the new page?

Comment (hoyyeva) [reply ui/pages/destinations/[id]/index.js:1]: oh i think i did a bad rebase - so it came back! removing it!

Comment (hoyyeva) [reply ui/pages/destinations/[id]/index.js:1]: oh that is true lol oopsy!

ui/components/layouts/dashboard.js
```diff
@@ -91,6 +92,12 @@ export default function Dashboard({ children }) {
   }
 
   const navigation = [
+    {
```
jmorganca, line 95: We can remove this for this PR, unless you'd like to keep it

ui/components/layouts/dashboard.js
```diff
@@ -133,6 +140,7 @@ export default function Dashboard({ children }) {
         <div className='mt-5 h-0 flex-1 overflow-y-auto'>
           <nav className='flex-1 space-y-1'>
             {navigation
+              ?.filter(n => !n.hide)
```
jmorganca, line 143: same as above

jmorganca:
LGTM once tests and linting pass

PR #3872, opened 2022-12-13
fix: return social login provider for direct identity

BruceMacD:
- get identity pre-loading providers now returns the google social login provider also

## Summary
When deleting a custom provider if a user existed via a social login also the user was removed. This was because the social login does not exist in the database and was not returned from `GetIdentity`.

## Checklist

### DIFF
--- a/internal/server/data/identity.go
+++ b/internal/server/data/identity.go
@@ -329,14 +329,19 @@ func GetIdentity(tx ReadTxn, opts GetIdentityOptions) (*models.Identity, error) 
 		if len(existsInProviders) > 0 {
 			var providerIDs []uid.ID
 			for _, relation := range existsInProviders {
-				providerIDs = append(providerIDs, relation.ProviderID)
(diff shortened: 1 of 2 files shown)


dnephin (approved):
LGTM!

I think you were right about doing this in the `data` package. My earlier idea to handle it entirely in API handlers would not work out at all.

Not for this PR, but if I look at where we are doing the addition of the google provider it seems to always be after `ListProviders`.

Instead of storing the provider as `Server.Google` I wonder if passing it to `data.DB` as a field on  `data.NewDBOptions` would allow us to handle it in only two places: `ListProvider` and `GetProvider`. That way it's still only in memory, but everything that calls those methods would have the provider it needs. 

We would need to add methods to `data.Transaction`, `data.DB`, and create a new interface with those methods to replace `ReadTxn` in `GetProvider`/`ListProvider`. That's a pretty big change,  maybe we'll find an easier way.

PR #3876, opened 2022-12-14
fix: explicitly output '' when ref is a tag

mxyng:
## Summary

The expression outputs the result of the `startsWith` when ref matches a tag which is an invalid input to goreleaser. Instead, set the `--skip-publish` flag explicitly


jmorganca (approved):
LGTM! Thanks!!

PR #3878, opened 2022-12-14
Fix a bug in config loading

dnephin:
I ran into this problem while working on #3865

A slice of structs was causing the location of the configuration value to be incorrect. There were multiple "exits" which removes the `INFRA_SERVER` item from the location. This resulting in any configuration after a slice of structs to not be loaded.

This bug impacted loading of env variables and command line flags, but only when some values were previously loaded from a config file. Using only a config file, or only env vars would have previously worked.

The test added in this commit was failing before the fix.

### DIFF
--- a/internal/cmd/cliopts/flat.go
+++ b/internal/cmd/cliopts/flat.go
@@ -67,6 +67,13 @@ func (w *flatSourceWalker) Exit(loc reflectwalk.Location) error {
 	return nil
 }
 
+func (w *flatSourceWalker) SliceElem(_ int, value reflect.Value) error {
+	if value.Type().Kind() == reflect.Struct || isPtrToStruct(value) {
+		w.location = append(w.location, "")
+	}
+	return nil
+}
+
 func (w *flatSourceWalker) Struct(value reflect.Value) error {
 	if !value.CanAddr() {
 		// TODO: what is this case?
@@ -95,6 +102,10 @@ func (w *flatSourceWalker) StructField(field reflect.StructField, value reflect.
 		w.location = append(w.location, w.fieldNameFormat(field.Name))
 	}
 	return nil
+}
+
+func (w *flatSourceWalker) Slice(reflect.Value) error {
+	return nil // no-op to satisfy an interface
 }
 
 func (w *flatSourceWalker) matchName(key string, fieldName string) bool {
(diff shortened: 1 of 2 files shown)


internal/cmd/cliopts/flat.go
```diff
@@ -97,6 +97,19 @@ func (w *flatSourceWalker) StructField(field reflect.StructField, value reflect.
...
+func (w *flatSourceWalker) Slice(reflect.Value) error {
+	return nil
+}
+
+func (w *flatSourceWalker) SliceElem(index int, value reflect.Value) error {
+	if value.Type().Kind() != reflect.Struct || isPtrToStruct(value) {
+		return nil
+	}
+
+	w.location = append(w.location, "")
```
dnephin, line 109: Above here in this file there is an `Exit` method that pops the last item off of `w.location`. What was happening was that a slice of structs was causing that pop , but there was no matching "push" so the location was wrong.

This fixes the problem by adding empty items to `location` for each item in the slice. This way the `Exit` will correctly pop the empty item, and the `location` value will be correct after iterating over the slice is complete.

internal/cmd/cliopts/flat.go
```diff
@@ -67,6 +67,13 @@ func (w *flatSourceWalker) Exit(loc reflectwalk.Location) error {
 	return nil
 }
 
+func (w *flatSourceWalker) SliceElem(_ int, value reflect.Value) error {
+	if value.Type().Kind() == reflect.Struct || isPtrToStruct(value) {
+		w.location = append(w.location, "")
```
dnephin, line 72: This `w.location` is effectively a stack that keeps track of the location within the Go structure that is receiving configuration value. 

Above here in this file there is an `Exit` method that pops the last item off of `w.location`. What was happening was that a slice of structs was causing that pop , but there was no matching "push", so the location was wrong.

This fixes the problem by pushing an empty items onto `location` for each item in the slice. This way the `Exit` will correctly pop the empty item, and the `location` value will be correct after iterating over the slice is complete.

We were already using this empty string value to indicate embedded structs, so this re-uses it for slice elements. Both of these things are treated the same way because they don't modify the path at which we expect a value.

pdevine (approved):
LGTM. It seems like this would have been a pain to track down with the reflection code.

dnephin:
It required some thinking for sure, and both the debugger and extra logging, but I've worked enough with Go reflection that it wasn't too bad.

PR #3879, opened 2022-12-15
improve: password requirements

mxyng:
## Summary

After an error, enable the "Reset Password" button once field values are updated. Otherwise, the user needs to refresh in order to submit again.

Improve the error returned by a bad password update. The errors were previously nested inside `errors.password` as a list of strings but only the first is being displayed. This is incredibly annoying as a user as the password requirements are displayed one at a time. e.g. if  the organization requires 1 lowercase letter, 1 uppercase letter, and 1 number, only first unmet requirement will be displayed.

The change to the error string first groups all errors into a single string so it can be displayed in the UI. It then populates the error string with all password requirements, not just the unmet requirements so the user knows exactly what is required. The error string also uses the correct form instead of always using the plural form

image

### DIFF
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -191,8 +191,8 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
 	}
 
 	requirements := []struct {
-		value         int
-		valueFunc     func(rune) bool
+		minCount      int
+		countFunc     func(rune) bool
 		singularError string
 		pluralError   string
 	}{
@@ -207,15 +207,15 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
 
 	valid := true
 	for _, r := range requirements {
-		switch {
-		case r.value > 1:
-			requirementError = append(requirementError, fmt.Sprintf(r.pluralError, r.value))
-		case r.value > 0:
-			requirementError = append(requirementError, fmt.Sprintf(r.singularError, r.value))
+		if !hasMinimumCount(password, r.minCount, r.countFunc) {
+			valid = false
 		}
 
-		if !hasMinimumCount(password, r.value, r.valueFunc) {
-			valid = false
+		switch {
+		case r.minCount == 1:
+			requirementError = append(requirementError, fmt.Sprintf(r.singularError, r.minCount))
+		case r.minCount > 1:
+			requirementError = append(requirementError, fmt.Sprintf(r.pluralError, r.minCount))
 		}
 	}
 
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -174,10 +174,10 @@ func isSymbol(r rune) bool {
 	return (r >= '\u0020' && r <= '\u002F') || (r >= '\u003A' && r <= '\u0040') || (r >= '\u005B' && r <= '\u0060') || (r >= '\u007B' && r <= '\u007E')
 }
 
-func hasMinimumCount(min int, password string, check func(rune) bool) bool {
+func hasMinimumCount(password string, min int, minCheck func(rune) bool) bool {
 	var count int
 	for _, r := range password {
-		if check(r) {
+		if minCheck(r) {
 			count++
 		}
 	}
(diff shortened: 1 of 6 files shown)


dnephin (commented):
Thanks for improving this!

I think in general this makes sense, but I have one concern about the implementation.

ui/lib/login.js
```diff
@@ -25,3 +25,13 @@ export function currentBaseDomain() {
   }
   return parts.join('.') // return the domain without the org
 }
+
+export function formatPasswordRequirements(requirements) {
+  if (requirements.length > 1) {
+    return `needs at least ${requirements
+      .slice(0, -1)
+      .join(', ')} and ${requirements.slice(-1)}`
```
pdevine, line 33: no oxford comma??

internal/access/credential.go
```diff
@@ -184,35 +184,68 @@ func hasMinimumCount(min int, password string, check func(rune) bool) bool {
...
+
+	format := func(n int, single, multiple string) string {
+		if n > 1 {
+			return fmt.Sprintf(multiple, n)
+		}
+
+		return fmt.Sprintf(single, n)
+	}
+
+	if s.LengthMin > 0 {
```
pdevine, line 198: This feels like it would be easier to read if you consolidated it w/ `checkPasswordRequirements()` and used a struct/slice to hold each of the requirements/format strings.

ui/lib/login.js
```diff
@@ -23,5 +23,23 @@ export function currentBaseDomain() {
   if (parts.length > 2) {
     parts.shift() // remove the org
   }
+
   return parts.join('.') // return the domain without the org
 }
+
+export function formatPasswordRequirements(requirements) {
```
jmorganca, line 30:
```suggestion
export function formatPasswordRequirements(requirements = []) {
```

internal/access/credential.go
pdevine:
```
type ReqCheck func(rune) bool

type PasswordReq struct {
        Min      int
        Singular string
        Plural   string
        Check    ReqCheck
}

func checkPasswordRequirements(db data.ReadTxn, password string) error {
	if err != nil {
		return err
	}
	
        reqs := []PasswordReq{
                {s.LowercaseMin, "%d lowercase letter", "%d lowercase letters", unicode.IsLower},
                {s.UppercaseMin, "%d uppercase letter", "%d uppercase letters", unicode.IsUpper},
                {s.NumberMin, "%d number", "%d numbers", unicode.IsDigit},
                {s.SymbolMin, "%d symbol", "%d symbols", isSymbol},
        }
        reqString := []string{}

        for _, r := range reqs {
                if !hasMinimumCount(r.Min, password, r.Check) {
                        if r.Min > 1 {
                                reqString = append(reqString, fmt.Sprintf(r.Plural, r.Min))
                        } else {
                                reqString = append(reqString, fmt.Sprintf(r.Singular, r.Min))
                        }
                }
        }
        ...
        
}
```
You would need to add the test back for the minimum length as well. You can either add it in `reqs` (but not pass it to `hasMinimumCount()`), or just do it one off. Then just pass the error back if there is one.

mxyng: This snippet has the downside where only the unmet requirements are added to the error, one of the problems this PR is trying to address but I get the idea

internal/access/credential.go
```diff
@@ -189,30 +189,38 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+		if !hasMinimumCount(password, r.minCount, r.countFunc) {
+			valid = false
+		}
 
-	if len(password) < settings.LengthMin {
-		errs["password"] = append(errs["password"], fmt.Sprintf("needs minimum length of %d", settings.LengthMin))
+		switch {
+		case r.minCount > 1:
+			requirementError = append(requirementError, fmt.Sprintf(r.pluralError, r.minCount))
+		case r.minCount > 0:
```
pdevine, line 217: small nit: this would probably read easier as `r.minCount == 1`

pdevine (approved):

PR #3880, opened 2022-12-15
Add request timeout to connector API requests

dnephin:
## Summary

@pdevine noticed that when his laptop resumed from suspend, the connector running in a kube pod would often stay in a disconnected state.

I believe the reason for this is a missing request timeout on the HTTP client used by the connector. We have a 1 minute timeout in the CLI, but the connector did not set a timeout.

This PR adds request timeouts to the Infra API operations performed by the connector. `sshd auth-keys` only makes short lived requests, so we can set a 1 minute timeout on the `http.Client` (like we do for other CLI commands).

The connector makes  some blocking requests, so for those calls we have to set a timeout using context. For updating grants I set the timeout to 6 minutes, which gives us 5 minutes to block on updates, and another 2 minutes to update the destination. For updating the destination in the Infra API I used a timeout of 2 minutes for the full operation. It makes more than one API call to both the Kube API and the Infra API, so I think we want to give it more time.

We could alternatively give each API call its own timeout, but I think that's harder to maintain, and doesn't really provide benefit over a single timeout for the full operation.

I don't know of a good way to test this in unit tests, so maybe someone can try this out with suspend to see if it fixes the problem for them.

### DIFF
--- a/internal/connector/connector.go
+++ b/internal/connector/connector.go
@@ -225,6 +225,9 @@ func runKubernetesConnector(ctx context.Context, options Options) error {
 		return syncGrantsToDestination(ctx, con, waiter, fn)
 	})
 	group.Go(func() error {
+		ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+		defer cancel()
+
 		// TODO: how long should this wait? Use exponential backoff on error?
 		waiter := repeat.NewWaiter(backoff.NewConstantBackOff(30 * time.Second))
 		for {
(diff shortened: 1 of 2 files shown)


pdevine (approved):
This looks good.

In terms of testing, I'm not sure what the best way to test it out would be, but I did see [Toxiproxy](https://github.com/Shopify/toxiproxy) which has a go client (and was written in go). I haven't used it before, but it looked interesting.

internal/connector/connector.go
```diff
@@ -471,7 +474,10 @@ func syncGrantsToDestination(
 ) error {
 	var latestIndex int64 = 1
 
-	sync := func() error {
+	sync := func(ctx context.Context) error {
+		ctx, cancel := context.WithTimeout(ctx, 7*time.Minute)
```
dnephin, line 478: I made this 7 minutes, just in case kube API requests are slower than we expect and it takes a while to apply them.

PR #3885, opened 2022-12-16
Destination credentials endpoints

ssoroka:
part 2/n

## Checklist

## Related Issues

Resolves #

### DIFF
--- a/api/destination_credential.go
+++ b/api/destination_credential.go
@@ -4,11 +4,11 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-type CreateDestinationCredential struct {
+type CreateDestinationCredentialRequest struct {
 	Destination string `form:"destination"`
 }
 
-type ListDestinationCredential struct {
+type ListDestinationCredentialRequest struct {
 	Destination     string `form:"destination"`
 	LastUpdateIndex int64  `form:"lastUpdateIndex"`
 }
@@ -49,19 +49,12 @@ type DestinationCredential struct {
 	// Token      string
 }
 
-// AnswerDestinationCredential
-type AnswerDestinationCredential struct {
-	////////////////////
-	// request fields //
-	////////////////////
+// AnswerDestinationCredentialRequest
+type AnswerDestinationCredentialRequest struct {
 	ID             uid.ID `json:"id"`
 	OrganizationID uid.ID `json:"organizationID"`
 
-	// /////////////////////
-	// // response fields //
-	// /////////////////////
-
-	// // API key
+	// API key
 	BearerToken         string `json:"bearerToken"`
 	CredentialExpiresAt Time   `json:"credentialExpiresAt"`
 }
--- a/api/destination_credential.go
+++ b/api/destination_credential.go
@@ -1,0 +1,67 @@
+package api
+
+import (
+	"github.com/infrahq/infra/uid"
+)
+
+type CreateDestinationCredential struct {
+	Destination string `form:"destination"`
+}
+
+type ListDestinationCredential struct {
+	Destination     string `form:"destination"`
+	LastUpdateIndex int64  `form:"lastUpdateIndex"`
+}
+
+type ListDestinationCredentialResponse struct {
+	Items          []DestinationCredential `json:"items"`
+	MaxUpdateIndex int64                   `json:"maxUpdateIndex"`
+}
+
+// DestinationCredential is the database DestinationCredential object, not a request/response struct
+type DestinationCredential struct {
+	////////////////////
+	// request fields //
+	////////////////////
+	ID             uid.ID `json:"id"`
+	OrganizationID uid.ID `json:"organizationID"`
+
+	UserID      uid.ID `json:"userID"`
+	Destination string `json:"destination"`
+
+	RequestExpiresAt Time `json:"requestExpiresAt"`
+
+	// Internal Fields
+	UpdateIndex int64 `json:"updateIndex"`
+
+	/////////////////////
+	// response fields //
+	/////////////////////
+
+	// API key
+	BearerToken         string `json:"bearerToken"`
+	CredentialExpiresAt Time   `json:"credentialExpiresAt"`
+
+	// Certificate
+
+	// JWT or generic headers
+	// HeaderName string
+	// Token      string
+}
+
+// AnswerDestinationCredential
+type AnswerDestinationCredential struct {
+	////////////////////
+	// request fields //
+	////////////////////
+	ID             uid.ID `json:"id"`
+	OrganizationID uid.ID `json:"organizationID"`
+
+	// /////////////////////
+	// // response fields //
+	// /////////////////////
+
+	// // API key
+	BearerToken         string `json:"bearerToken"`
+	CredentialExpiresAt Time   `json:"credentialExpiresAt"`
+}
--- a/api/destination_credential.go
+++ b/api/destination_credential.go
@@ -18,7 +18,7 @@ type ListDestinationCredentialResponse struct {
 	MaxUpdateIndex int64                   `json:"maxUpdateIndex"`
 }
 
-// DestinationCredential is the database DestinationCredential object, not a request/response struct
+// DestinationCredential represents the request for credentials to a particular destination
 type DestinationCredential struct {
 	////////////////////
 	// request fields //
--- a/internal/access/destinationcredential.go
+++ b/internal/access/destinationcredential.go
@@ -31,7 +31,6 @@ func (l ListDestinationCredentialResponse) ItemCount() int {
 func CreateDestinationCredential(c *gin.Context, destination string) (*models.DestinationCredential, error) {
 	rCtx := GetRequestContext(c)
 	// does the user have a grant to this destination?
-	// ListGrants(c, )
 	grants, err := data.ListGrants(rCtx.DBTxn, data.ListGrantsOptions{
 		ByDestination:              destination,
 		BySubject:                  uid.NewIdentityPolymorphicID(rCtx.Authenticated.User.ID),
@@ -151,11 +150,6 @@ func ListDestinationCredentials(c *gin.Context, destination string, lastUpdateIn
 		return listDestinationCredentialsWithMaxUpdateIndex(rCtx, dest.ID)
 	}
 	return blockingRequest(rCtx, listenOpts, query, lastUpdateIndex)
-}
-
-type ListDestinationCredentialRespct struct {
-	Items          []models.DestinationCredential
-	MaxUpdateIndex int64
 }
 
 func listDestinationCredentialsWithMaxUpdateIndex(rCtx RequestContext, destinationID uid.ID) (ListDestinationCredentialResponse, error) {
--- a/internal/access/destinationcredential.go
+++ b/internal/access/destinationcredential.go
@@ -1,0 +1,195 @@
+package access
+
+import (
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/infrahq/infra/api"
+	"github.com/infrahq/infra/internal"
+	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/internal/server/data"
+	"github.com/infrahq/infra/internal/server/models"
+	"github.com/infrahq/infra/uid"
+)
+
+type ListDestinationCredentialResponse struct {
+	Items          []models.DestinationCredential
+	MaxUpdateIndex int64
+}
+
+func (l ListDestinationCredentialResponse) UpdateIndex() int64 {
+	return l.MaxUpdateIndex
+}
+
+func (l ListDestinationCredentialResponse) ItemCount() int {
+	return len(l.Items)
+}
+
+func CreateDestinationCredential(c *gin.Context, destination string) (*models.DestinationCredential, error) {
+	rCtx := GetRequestContext(c)
+	// does the user have a grant to this destination?
+	// ListGrants(c, )
+	grants, err := data.ListGrants(rCtx.DBTxn, data.ListGrantsOptions{
+		ByDestination:              destination,
+		BySubject:                  uid.NewIdentityPolymorphicID(rCtx.Authenticated.User.ID),
+		IncludeInheritedFromGroups: true,
+	})
+	if err != nil {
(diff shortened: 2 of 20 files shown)


api/destination_credential.go
```diff
@@ -0,0 +1,67 @@
...
+	Destination     string `form:"destination"`
+	LastUpdateIndex int64  `form:"lastUpdateIndex"`
+}
+
+type ListDestinationCredentialResponse struct {
+	Items          []DestinationCredential `json:"items"`
+	MaxUpdateIndex int64                   `json:"maxUpdateIndex"`
+}
+
+// DestinationCredential is the database DestinationCredential object, not a request/response struct
```
BruceMacD, line 21: Shouldn't this be in the data package if its not a request response struct?

api/destination_credential.go
```diff
@@ -0,0 +1,67 @@
...
+type DestinationCredential struct {
+	////////////////////
+	// request fields //
+	////////////////////
+	ID             uid.ID `json:"id"`
+	OrganizationID uid.ID `json:"organizationID"`
+
+	UserID      uid.ID `json:"userID"`
+	Destination string `json:"destination"`
```
BruceMacD, line 30: Should this be a destination ID also?

internal/access/destinationcredential.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (l ListDestinationCredentialResponse) ItemCount() int {
+	return len(l.Items)
+}
+
+func CreateDestinationCredential(c *gin.Context, destination string) (*models.DestinationCredential, error) {
+	rCtx := GetRequestContext(c)
+	// does the user have a grant to this destination?
+	// ListGrants(c, )
```
BruceMacD, line 34:
```suggestion
```

internal/access/destinationcredential.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// commit the transaction so others can see the request
+	err = rCtx.DBTxn.Commit()
+	if err != nil {
+		return nil, err
+	}
+
+	// wait for the request to be filled; listen for the notify on the specific record req.ID
+	listener, err := data.ListenForNotify(c, rCtx.DataDB, data.ListenForNotifyOptions{
+		OrgID:                      rCtx.Authenticated.AccessKey.OrganizationID,
```
BruceMacD, line 73: I feel like Org ID should come from the transaction context rather than something that can be specified

internal/access/destinationcredential.go
```diff
@@ -0,0 +1,195 @@
...
+		logging.Debugf("could not roll back transaction")
+	}
+
+	return cr, nil
+}
+
+func ListDestinationCredentials(c *gin.Context, destination string, lastUpdateIndex int64) (ListDestinationCredentialResponse, error) {
+	rCtx := GetRequestContext(c)
+
+	roles := []string{models.InfraAdminRole, models.InfraViewRole, models.InfraConnectorRole}
```
BruceMacD, line 118: This looks like something you should only be able to do for your own credentials, it looks like this could return bearer tokens

internal/access/destinationcredential.go
```diff
@@ -0,0 +1,195 @@
...
+		DestinationCredentialsByDestinationID: dest.ID,
+		OrgID:                                 rCtx.DBTxn.OrganizationID(),
+	}
+	query := func() (ListDestinationCredentialResponse, error) {
+		return listDestinationCredentialsWithMaxUpdateIndex(rCtx, dest.ID)
+	}
+	return blockingRequest(rCtx, listenOpts, query, lastUpdateIndex)
+}
+
+type ListDestinationCredentialRespct struct {
```
BruceMacD, line 156:
```suggestion
type ListDestinationCredentialResp struct {
```

Not sure, but this looked like a typo

internal/access/longpolling.go
```diff
@@ -0,0 +1,67 @@
...
+	UpdateIndex() int64
+	ItemCount() int
+}
+
+func blockingRequest[Result updateIndexable](rCtx RequestContext, listenOpts data.ListenForNotifyOptions, query func() (Result, error), lastUpdateIndex int64) (Result, error) {
+
+	listener, err := data.ListenForNotify(rCtx.Request.Context(), rCtx.DataDB, listenOpts)
+	if err != nil {
+		// nolint: gocritic
+		return *new(Result), fmt.Errorf("listen for notify: %w", err)
```
BruceMacD, line 24: Why return a new result like this?

api/destination_credential.go
```diff
@@ -0,0 +1,67 @@
...
+	////////////////////
+	ID             uid.ID `json:"id"`
+	OrganizationID uid.ID `json:"organizationID"`
+
+	// /////////////////////
+	// // response fields //
+	// /////////////////////
+
+	// // API key
+	BearerToken         string `json:"bearerToken"`
```
BruceMacD, line 65: What is the bearer token in this context?

api/destination_credential.go
ssoroka: This isn't actually a database struct. fixed the comment. `DestinationCredential represents the request for credentials to a particular destination`

api/destination_credential.go
ssoroka: Maybe, but I think that makes it harder for the user to use the `infra credentials add dest` command.

internal/access/destinationcredential.go
ssoroka: it's actually only something the connector is supposed to be able to do. I'd love to tie this down to permissions that only the connector has.

internal/access/longpolling.go
ssoroka: AFAIK there are not a lot of ways to create a new instance of a generics type.

internal/access/destinationcredential.go
ssoroka: FYI the List function cannot be used to retrieve bearer tokens.

internal/access/destinationcredential.go
```diff
@@ -0,0 +1,189 @@
...
+	logging.Debugf("Creating DestinationCredential with id %d, orgID %d", cr.ID, cr.OrganizationID)
+
+	// commit the transaction so others can see the request
+	err = rCtx.DBTxn.Commit()
+	if err != nil {
+		return nil, err
+	}
+
+	// wait for the request to be filled; listen for the notify on the specific record req.ID
+	listener, err := data.ListenForNotify(c, rCtx.DataDB, data.ListenForNotifyOptions{
```
dnephin, line 71: This is my primary concern. This approach is what ends up requiring `SafeCommit`.

I think it's unexpected that a `POST` request blocks for up to 5 minutes. 

Bruce made this suggestion on the previous draft PR. I think we should split this up into two API calls. A short lived `POST` that submits the request, and then a long polling `GET` to receive the response. 

I think that approach would also make the API request/response structs easier to follow, without the need for big block comments about how some subset of the fields will be used.

There's another problem hidden here, which is that we wait up to 5 minutes for the write to happen, where as normally we timeout write requests at 1 minute. We could fix that by setting a shorter deadline on the context used for the first part of this request, but the separate endpoints seems like it fixes more problems.

Is there anything that becomes more difficult with separate API endpoints for this?

Does this operation not fit into the `access.blockingRequest` helper that we extracted in this PR? Would it if we split the endpoint into two?

internal/access/longpolling.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}()
+
+	result, err := query()
+	if err != nil {
+		return result, err
+	}
+
+	// The query returned results that are new to the client
+	if result.ItemCount() > 0 && result.UpdateIndex() > lastUpdateIndex {
```
dnephin, line 42: This `result.ItemCount` is new, and that seems like it introduces a bug. If all the grants are deleted, the client needs to know the response is now empty, so a response should be sent when `itemCount == 0`. 

I guess we don't have test coverage of that scenario.

Why is this new clause necessary?

internal/server/destinationcredentials_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/infrahq/infra/uid"
+)
+
+func TestDestinationCredentials(t *testing.T) {
+	srv := setupServer(t, withAdminUser, withMultiOrgEnabled)
+	db := srv.db
+	routes := srv.GenerateRoutes()
+	destName := "foo"
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
dnephin, line 25: This use of wait group and `t.Parallel` is surprising.  I think this test would be more obvious if we modeled it like a real-world use case.

https://pkg.go.dev/golang.org/x/sync/errgroup makes managing the errors from goroutines much easier, and wraps a `WaitGroup`, so it also manages waiting for them to exit.

This test could start a goroutine that acts as the connector using `errgroup` first (I guess that's what's in the ListDestinationCredentials test case). Then run create in the main goroutine, and at the end call `assert.NilError(t, group.Wait())`. 

If you set a deadline on the context, it should prevent the goroutine from running too long in the case of failure. 

I think that would make it more clear that this is testing all 3 endpoints together. We'll also want separate test cases for different response codes, and to test the full response body. Testing only the response struct that was decoded makes it easy to miss fields in the test coverage. So we will want a table test for each of the endpoints as well eventually.

dnephin (commented):
This is looking good!

I think there's a couple things here we should address before a merge:

1. Combining both the `Create` and `Get` into a single blocking POST endpoint causes a few problems. I think we should evaluate splitting that into a short-lived `Create` request, and a blocking `Get`.
2. I believe the change to `blockingRequest` (commented below) is a regression.

internal/access/longpolling.go
dnephin: I think we might be able to return `nil` since it's an interface

ssoroka:
> Combining both the Create and Get into a single blocking POST endpoint causes a few problems. I think we should evaluate splitting that into a short-lived Create request, and a blocking Get.

When the endpoints are combined you don't need to worry that the person requesting the token isn't the one who requested the credentials get created. This is also additional complexity for devs implementing clients, who just want their data and get zero value from splitting it across two requests.

internal/server/data/data.go
```diff
@@ -219,6 +219,15 @@ func (t *Transaction) Rollback() error {
...
 }
 
+// SoftCommit commits the transaction if it has not already been committed.
+// if it has been committed, it does not error.
+func (t *Transaction) SoftCommit() error {
+	if t.completed.Load() {
+		return nil
+	}
+	return t.Commit()
+}
```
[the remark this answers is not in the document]
ssoroka: I suspect it reverts to auto-commit mode, but anywhere I commit I start a new transaction for the remainder of the request. The problem seems to be tying that new transaction into the original context/db as if it was the original transaction. That sounds problematic. This means if the handler wants to commit the transaction, there's not much choice here. Alternatively you could choose to avoid trying to commit or rollback at all, but I don't really like that: What if it fails before it gets to the commit?

dnephin:
>When the endpoints are combined you don't need to worry that the person requesting the token isn't the one who requested the credentials get created. 

The request is authenticated with an access key, right? So splitting it up doesn't change that? Both requests get authenticated the same way.

> This is also additional complexity for devs implementing clients, who just want their data and get zero value from splitting it across two requests.

I think they will get value from it! Splitting it up won't matter for someone using the CLI or UI, it only matters for someone using the API directly, right?

The initial `Create` request returning lets them know the request was accepted and is being processed. If the request just hangs they have no idea what is happening, there's no feedback. If someone is using this API This allows them to do other work without having to manage background threads.

Separate endpoints also makes the experience much more consistent with our other async APIs. Device flow uses two separate endpoints and it's essentially the same idea.

internal/access/longpolling.go
ssoroka: When there are no records in the table and this guard is not in place, this causes the function to return early with no results instead of waiting. This would force the client to make another request. Returning with zero records doesn't seem to be that useful for a function that's supposed to block until _new_ data arrives.

dnephin: > When there are no records in the table and this guard is not in place, this causes the function to return early with no results instead of waiting

That's the correct behaviour. The first query should always return the state of the world so that the client knows it has the current state. The next query that comes in should block properly. Is the second query not blocking? That would be a bug.

> Returning with zero records doesn't seem to be that useful for a function that's supposed to block until new data arrives.

The zero state is a completely valid state. When a client receives this first empty request it goes from "unknown" state to a valid state with no data. That **MUST** be returned to the client for the system to behave correctly.

internal/access/destinationcredential.go
ssoroka: Ignoring the fact that under normal conditions this flow takes 300ms, why would it be unexpected that a POST blocks for 5 minutes but it's normal for a GET request? There's nothing in the http spec that would indicate a POST should respond quickly. 

The block comments are required anyway, because they come from the db table. You can split up the request/response structs currently without changing which endpoint returns with the results, so the two issues are not tied together. 

Adjusting the timeouts to match is fine, 5 minutes is not a requirement anywhere.

It doesn't fit into the existing `blockingRequest` structure because it does not use an updateIndex, it simply wants to know when the pg_notification happens.

internal/server/destinationcredentials_test.go
ssoroka: It's elegant as is without adding a ton of complexity. Table tests here are not currently needed, but we can always add this in the future. I added a comment for clarity.

internal/access/longpolling.go
ssoroka: > "No data" can also be new data (when everything is deleted).

State of the world is not useful to me; I only need pending, unanswered requests, and there are often none. I believe without this check the client thrashes, and continually hits the server immediately over and over again, so we need a solution to that before I revert this change.

dnephin: The update index is what should prevent that. It sounds like maybe you aren't using an update index, which is maybe the problem.

internal/server/destinationcredentials_test.go
dnephin: This is abusing `t.Parallel` for something it was never designed for.

internal/access/longpolling.go
dnephin: The update index doesn't have to come from the DB. You could fake it in code if you can derive it from the state of the query.

If all your concerned about is "has this been updated", then you could use the following for update index:

```
1 - pending
2 - done
```

Issue #3881, opened 2022-12-15
Check OIDC session in Device Flow


BruceMacD:
## Describe the bug
When a new access key is issued for device flow login if the OIDC session it's issued from has expired then the device flow session won't work. 

## Expected behavior
Check the OIDC session validity in the device flow, and return a better error message when it does expire.

- [x] sync user on device login
- [ ] return a better error on idp session timeout in cli

PR #3886, opened 2022-12-16
fix: check user idp session on device login

BruceMacD:
- sync user session from external IDPs in middleware
- add sync cache for 120 minutes
- remove explicit IDP sync on token creation, this will slow the propagation of IDP changes
- attempt to make user sync logic easier to follow

## Summary
If a user's session in a provider was expired on device login this wasnt checked until an `infra tokens add` command was run. Check the user's session in the IDP on device flow login via the middleware, and get them to log back in if its gone stale, then they can proceed with the device flow as normal.

1. `infra login`
2. Users session is validated with IDP
3. If session is no longer valid, user must login again (in the browser)
4. User is returned to device flow and continues as usual

## Checklist

## Related Issues

Part of #3881

### DIFF
--- a/internal/access/identity.go
+++ b/internal/access/identity.go
@@ -1,16 +1,13 @@
 package access
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/gin-gonic/gin"
 
 	"github.com/infrahq/infra/internal"
-	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
-	"github.com/infrahq/infra/internal/server/providers"
 	"github.com/infrahq/infra/uid"
 )
 
@@ -84,38 +81,4 @@ func ListIdentities(c *gin.Context, opts data.ListIdentityOptions) ([]models.Ide
 		return nil, HandleAuthErr(err, "users", "list", roles...)
 	}
 	return data.ListIdentities(db, opts)
-}
-
-// UpdateIdentityInfoFromProvider calls the identity provider used to authenticate this user session to update their current information
-func UpdateIdentityInfoFromProvider(c RequestContext, provider *models.Provider, oidc providers.OIDCClient) error {
-	// does not need authorization check, this action is limited to the calling user
-	ctx := c.Request.Context()
-
-	// added by the authentication middleware
-	identity := c.Authenticated.User
-	if identity == nil {
-		return errors.New("user does not have session with an identity provider")
-	}
-
-	db := c.DBTxn
-
-	// get current identity provider groups and account status
-	err := data.SyncProviderUser(ctx, db, identity, provider, oidc)
-	if err != nil {
-		if errors.Is(err, internal.ErrBadGateway) {
-			return err
-		}
-
-		if nestedErr := data.DeleteAccessKeys(db, data.DeleteAccessKeysOptions{ByIssuedForID: identity.ID, ByProviderID: provider.ID}); nestedErr != nil {
-			logging.Errorf("failed to revoke invalid user session: %s", nestedErr)
-		}
-
-		if nestedErr := data.DeleteProviderUsers(db, data.DeleteProviderUsersOptions{ByIdentityID: identity.ID, ByProviderID: provider.ID}); nestedErr != nil {
-			logging.Errorf("failed to delete provider user: %s", nestedErr)
-		}
-
-		return fmt.Errorf("sync user: %w", err)
-	}
-
-	return nil
 }
--- a/internal/access/identity_test.go
+++ b/internal/access/identity_test.go
@@ -1,11 +1,7 @@
 package access
 
 import (
-	"context"
-	"fmt"
-	"net/http"
 	"testing"
-	"time"
 
 	"gotest.tools/v3/assert"
 
@@ -13,7 +9,6 @@ import (
 	"github.com/infrahq/infra/internal/generate"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
-	"github.com/infrahq/infra/internal/server/providers"
 	"github.com/infrahq/infra/uid"
 )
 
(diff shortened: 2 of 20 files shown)


internal/server/handlers.go
```diff
@@ -298,8 +298,8 @@ func (a *API) Version(c *gin.Context, r *api.EmptyRequest) (*api.Version, error)
 	return &api.Version{Version: internal.FullVersion()}, nil
 }
 
-// UpdateIdentityInfoFromProvider calls the identity provider used to authenticate this user session to update their current information
-func (a *API) UpdateIdentityInfoFromProvider(rCtx access.RequestContext) error {
+// SyncRequestIdentityInfo calls the identity provider used to authenticate this user session to update their current information
+func (a *API) SyncRequestIdentityInfo(rCtx access.RequestContext) error {
```
jmorganca, line 302: nit: should this function be lowercase (i.e. private)?

internal/server/users.go
```diff
@@ -63,6 +64,27 @@ func (a *API) GetUser(c *gin.Context, r *api.GetUserRequest) (*api.User, error)
 		return nil, err
 	}
 
+	if r.Sync {
```
jmorganca, line 67: I wonder if we'd _always_ want to do this if the access key is associated to a provider. We could cache the result to avoid long requests?

BruceMacD: Im gonna take this approach

BruceMacD: Made this change

internal/access/identity.go
```diff
@@ -86,35 +87,34 @@ func ListIdentities(c *gin.Context, opts data.ListIdentityOptions) ([]models.Ide
 	return data.ListIdentities(db, opts)
 }
 
+var ErrUserSyncFailed = fmt.Errorf("sync user failed") // this means either their session expired or they were removed from the idp
```
jmorganca, line 90: Move to `internal/errors.go`?

internal/access/identity.go
```diff
@@ -86,35 +87,34 @@ func ListIdentities(c *gin.Context, opts data.ListIdentityOptions) ([]models.Ide
...
-	// does not need authorization check, this action is limited to the calling user
+func UpdateIdentityInfoFromProvider(c RequestContext, providerUser *models.ProviderUser, oidc providers.OIDCClient) error {
 	ctx := c.Request.Context()
+	tx := c.DBTxn
 
-	// added by the authentication middleware
-	identity := c.Authenticated.User
-	if identity == nil {
-		return errors.New("user does not have session with an identity provider")
+	providerUser.LastUpdate = time.Now().UTC()
```
jmorganca, line 97: nit: update this at the end of this function for a more accurate timestamp?

internal/server/data/provideruser.go
```diff
@@ -296,40 +297,31 @@ func ProvisionProviderUser(tx WriteTxn, user *models.ProviderUser) error {
...
 }
 
-func SyncProviderUser(ctx context.Context, tx WriteTxn, user *models.Identity, provider *models.Provider, oidcClient providers.OIDCClient) error {
-	providerUser, err := GetProviderUser(tx, provider.ID, user.ID)
-	if err != nil {
-		return err
-	}
-
-	accessToken, expiry, err := oidcClient.RefreshAccessToken(ctx, providerUser)
+func SyncProviderUser(ctx context.Context, tx WriteTxn, user *models.ProviderUser, oidcClient providers.OIDCClient) ([]models.Group, error) {
```
jmorganca, line 300: nit for later: we should probably move this out of `data` since it's more about making an external call to the identity provider and calling `data` functions

jmorganca (approved):
Looks good. Much cleaner! Requested @pdevine to review as well since you collaborated on this together

internal/server/routes.go
```diff
@@ -56,7 +57,14 @@ func (s *Server) GenerateRoutes() Routes {
 
 	get(a, authn, "/api/users", a.ListUsers)
 	post(a, authn, "/api/users", a.CreateUser)
-	get(a, authn, "/api/users/:id", a.GetUser)
+	// getuser needs its own route declaration because it may call database write operations on sync
+	getUserRoute := route[api.GetUserRequest, *api.User]{
+		handler: a.GetUser,
+		routeSettings: routeSettings{
+			omitFromTelemetry: true,
+		},
```
dnephin, line 65: I think this is a sign we might not have the right design for this.  

If we need to write during a `GET` request can we do that in the middleware before handing off the read-only transaction to the handler?

Is there something special about `GetUser`? Why do it in that endpoint instead of approval of the device flow (which is already a POST) ?

Could we do this in the background in the middleware? That would also fix the problem of latency communicating with the provider causing slow API requests.

internal/server/handlers.go
```diff
@@ -325,17 +330,21 @@ func (a *API) UpdateIdentityInfoFromProvider(rCtx access.RequestContext) error {
...
+	providerUser, err := data.GetProviderUser(rCtx.DBTxn, provider.ID, identity.ID)
 	if err != nil {
 		return fmt.Errorf("failed to get provider user to update: %w", err)
 	}
 
-	oidc, err := a.providerClient(rCtx.Request.Context(), provider, providerUser.RedirectURL)
-	if err != nil {
-		return fmt.Errorf("update provider client: %w", err)
+	// if provider user was updated recently, skip checking this now to avoid hitting rate limits
+	if time.Since(providerUser.LastUpdate) > providerUserUpdateThreshold {
```
dnephin, line 339: Nice! I think this is a good mechanism for throttling queries to the IDP.

internal/server/handlers.go
```diff
@@ -298,8 +298,13 @@ func (a *API) Version(c *gin.Context, r *api.EmptyRequest) (*api.Version, error)
 	return &api.Version{Version: internal.FullVersion()}, nil
 }
 
-// UpdateIdentityInfoFromProvider calls the identity provider used to authenticate this user session to update their current information
-func (a *API) UpdateIdentityInfoFromProvider(rCtx access.RequestContext) error {
+// providerUserUpdateThreshold is the duration of time that must pass before a
+// users session is attempted to be validated again with an external identity provider.
+// This prevents hitting IDP rate limits.
+const providerUserUpdateThreshold = 5 * time.Minute
```
dnephin, line 304: 5 minutes seems pretty frequent. Could we cache this for longer? Maybe a couple hours?

internal/server/data/provideruser.go
```diff
@@ -259,6 +259,7 @@ func GetProviderUser(tx ReadTxn, providerID, identityID uid.ID) (*models.Provide
 	query.B(columnsForSelect(pu))
 	query.B("FROM")
 	query.B(pu.Table())
+	query.B("LEFT OUTER JOIN providers ON provider_users.provider_id = providers.id AND providers.organization_id = ?", tx.OrganizationID())
```
dnephin, line 262: Is this related to the fix, or was this about something else?

internal/server/data/identity.go
```diff
@@ -107,38 +109,37 @@ func AssignIdentityToGroups(tx WriteTxn, user *models.Identity, provider *models
...
 
 		if len(ids) == 0 {
 			// add user to group
-			_, err = tx.Exec("INSERT INTO identities_groups (identity_id, group_id) VALUES (?, ?)", user.ID, groupID)
+			_, err = tx.Exec("INSERT INTO identities_groups (identity_id, group_id) VALUES (?, ?)", user.IdentityID, groupID)
 			if err != nil {
-				return fmt.Errorf("insert: %w", handleError(err))
+				return nil, fmt.Errorf("insert: %w", handleError(err))
 			}
+			identity.Groups = append(identity.Groups, models.Group{Model: models.Model{ID: groupID}, OrganizationMember: models.OrganizationMember{OrganizationID: identity.OrganizationID}, Name: name})
```
dnephin, line 138: This line would read a lot better if it were formatted with each field of the struct on a separate line.

internal/access/identity.go
dnephin: I think we should avoid putting more things into `internal/errors.go`. Application-wide errors are easy to mess up.  This error is specific to this one function, so defining it right next to that function is the right place for it, if we need it.

Do we need this to be an error variable? I can't find any code that checks this with `errors.Is`. Is that missing, or do you expect to need to need that later? If we aren't checking the error with `errors.Is`, I don't think we need a variable for it. We can inline it to the one place it is returned.

dnephin (commented):
Looks good! A few minor suggestions and questions, and just one concern about the transaction handling.

internal/access/identity_test.go
```diff
@@ -185,20 +185,20 @@ func TestUpdateIdentityInfoFromProvider(t *testing.T) {
...
 
 	t.Run("a revoked OIDC session revokes access keys created by provider login", func(t *testing.T) {
-		_, err = data.CreateProviderUser(db, provider, rCtx.Authenticated.User)
+		user, err := data.CreateProviderUser(db, provider, rCtx.Authenticated.User)
 		assert.NilError(t, err)
 		oidc := &fakeOIDCImplementation{UserInfoRevoked: true}
 
 		toBeRevoked := &models.AccessKey{IssuedFor: rCtx.Authenticated.User.ID, ProviderID: provider.ID}
-		_, err := data.CreateAccessKey(db, toBeRevoked)
+		_, err = data.CreateAccessKey(db, toBeRevoked)
```
pdevine, line 193: I'm confused about how this compiled in the old version.

internal/server/data/provideruser.go
BruceMacD: Should have called this one out, it was not related to the fix. Just noticed the inconsistency here between `GetProviderUser` and `ListProviderUser`. In `ListProviderUser` we do this join as a safety thing to scope the query to the current organization. We should do it in `GetProviderUser` too just in case.

internal/access/identity_test.go
dnephin: Oh ya, this is surprising! I think it's this: https://go.dev/play/p/huEC6yz8pur

There is an `err` in the parent scope, so it's fine to do `err = ...`, but it's also fine to do `err := ...` once in the new scope, even though it comes after assigning to `err` of the parent scope.

internal/server/data/identity.go
```diff
@@ -34,39 +34,41 @@ func (i *identitiesTable) ScanFields() []any {
...
-	pu.LastUpdate = time.Now().UTC()
-	if err := UpdateProviderUser(tx, pu); err != nil {
-		return fmt.Errorf("save: %w", err)
+	user.Groups = newGroups
+	user.LastUpdate = time.Now().UTC()
+	if err := UpdateProviderUser(tx, user); err != nil {
+		return nil, fmt.Errorf("save: %w", err)
 	}
 
 	// remove user from groups
```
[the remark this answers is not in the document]
dnephin: That would be good, but maybe better in a separate PR?

internal/server/routes.go
```diff
@@ -56,7 +57,14 @@ func (s *Server) GenerateRoutes() Routes {
 
 	get(a, authn, "/api/users", a.ListUsers)
 	post(a, authn, "/api/users", a.CreateUser)
-	get(a, authn, "/api/users/:id", a.GetUser)
+	// getuser needs its own route declaration because it may call database write operations on sync
+	getUserRoute := route[api.GetUserRequest, *api.User]{
```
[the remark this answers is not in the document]
dnephin: The pattern we've been using for other routes is to move this `getUserRoute` variable next to where the handler (`API.GetUser`) is defined. That way this list of routes stays concise, and it's easy to see which route settings are applied to the handler when looking at the handler.

I don't think hiding those settings in functions is a great option here. All we're doing is setting values on a struct. A variable for common settings could work well (we have `defaultRouteSettingsGet` in #2006), but I doubt we want a separate one for each permutation. I think something in the middle works well: a shared variable for the most common cases, but any route that does something unique should just define the settings for that one route.

I think it's pretty important to avoid write operations in `GET` requests (outside of a few isolated places in the "middleware"). I'm hoping we can change where this update happens to avoid the need for a custom route setting here.

internal/access/identity.go
jmorganca: Makes sense! Thanks!

BruceMacD: I meant to use this error to check that the error was expected as a sync fail, and if not log it at the error level rather than the debug level. I've made that change now.

internal/server/routes.go
BruceMacD: I did it here because its the endpoint that the UI uses to check session validity. Moved to the middleware instead from your feedback.

internal/server/handlers.go
BruceMacD: Changed to 2 hours

internal/server/routes.go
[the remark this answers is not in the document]
BruceMacD: Removed this update to the get request

internal/server/data/identity.go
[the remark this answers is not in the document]
BruceMacD: Lets do this one in a separate PR because there are some more in-depth changes in this PR now that I've moved IDP sync to the middleware.

internal/server/data/provideruser.go
mxyng: I wonder if we need to store the provider user in the database at all since it's mostly functioning a cache for the user's OIDC access token. Perhaps we can store it in Redis instead? Expiry can simply be a setting a TTL on the Redis key and when a key isn't found, call `oidcClient.RefreshAccessToken`

mxyng (approved):

internal/server/middleware.go
```diff
@@ -110,6 +111,25 @@ func authenticateRequest(c *gin.Context, route routeSettings, srv *Server) (acce
 		if err := handleInfraDestinationHeader(rCtx, c.Request.Header); err != nil {
 			return authned, err
 		}
+		if route.idpSync {
+			if route.authenticationOptional {
+				// this should be caught during development
+				return authned, fmt.Errorf("idp sync requires authentication")
+			}
+			// sync the identity info here to keep the UI session in sync with IDP session validity
+			if err := srv.syncIdentityInfo(context.Background(), tx, authned.User, authned.AccessKey.ProviderID); err != nil {
```
dnephin, line 120: We don't want to do this sync (external API call to an IDP) with an open transaction that may be holding database row locks.

I think we'll need to adjust the transactions so that we commit it before doing this step.

internal/server/middleware.go
```diff
@@ -110,6 +111,25 @@ func authenticateRequest(c *gin.Context, route routeSettings, srv *Server) (acce
...
+				deleteCookie(c.Writer, cookieAuthorizationName, c.Request.Host)
+				if errors.Is(err, ErrSyncFailed) {
+					logging.L.Debug().Err(err)
+				} else {
+					logging.L.Error().Err(err)
+				}
+				if err = tx.Commit(); err != nil {
+					logging.L.Error().Err(err)
+				}
```
dnephin, line 129: Do we need to commit here on all errors?  I think maybe `syncIdentityInfo` needs to manage its own transaction.

internal/server/data/provideruser.go
dnephin: I believe we need the provider user in the database to be able to identify which IDPs provide a particular user. It's much easier to do that with a SQL `JOIN` clause, than having to query multiple data sources and doing the join in the application.

internal/server/middleware.go
```diff
@@ -112,8 +113,40 @@ func authenticateRequest(c *gin.Context, route routeSettings, srv *Server) (acce
...
+		if err := srv.syncIdentityInfo(context.Background(), tx2, authned.User, authned.AccessKey.ProviderID); err != nil {
+			deleteCookie(c.Writer, cookieAuthorizationName, c.Request.Host)
+			if errors.Is(err, ErrSyncFailed) {
+				logging.L.Debug().Err(err)
+			} else {
+				logging.L.Error().Err(err)
+			}
+			if err = tx2.Commit(); err != nil {
+				logging.L.Error().Err(err)
+			}
```
dnephin, line 141: It's a bit unconventional to commit in an error handling block, but I guess the reason this is needed is because we may have updated the ProviderUser, or delete access keys.

I can't think of any obvious problems this would cause. We'll have to keep this in mind as we make changes. It's possible we might want some more fine grained transaction handling in `syncIdentityInfo` and the functions it calls, but let's worry about that in the future.

dnephin (approved):
LGTM!

Issue #3889, opened 2022-12-16
Allow for testing of transactional emails from API handlers


dnephin:
### Background

Today we have some test coverage for `internal/server/email` using `setupSMTPServer`, and `setupClient`. This works well for testing the templates in isolation, but does not give us any way to test that an email is sent as a result of an API request. Some test coverage of that integration would be valuable.

There are also a number of package-level global variables (`SMTPServer`, `TestMode`, `TestDataSent`, `emailTemplates`, `textTemplateList`, etc). Some are used as shims for testing, others are set from configuration options.

### Proposal

I think we can make this easier to test, by making two small changes:

1. instead of using package-level variables, create a `type Email struct` which is configured from the passed in options. Store an instance of this type on `Server`, and use that type as the first argument to the `SendEmail` functions. Avoiding package-level variables generally makes code easier to test, and much safer in general.
2. move `setupSMTPServer` and `setupClient` to somewhere like `internal/testing/emailtest`, so that they can be used to create an `Email` instance that is setup to send locally. When we test API handlers we can set this on `Server`, and use it to make assertions about the email that was sent.

PR #3890, opened 2022-12-16
Minor improvements to the ForgottenDomain data query and email

### LINKED ISSUE #3726: Sorting organization accounts via forgot domain link
https://login.infrahq.com/forgot-domain 

The received email does not sort by org name or last seen time. 

Image 2022-11-24 at 3 23 54 PM

dnephin:
## Summary

Best viewed by individual commit, there are a few related changes here.

This PR makes a few small changes to forgotten domains, and improves the test coverage at the same time:
* use more strict assertions (`DeepEqual`), and use `Equal` instead of `Assert` for better failure messages
* move the one function from `data/forgotdomain.go` to `data/organizations.go`. See the commit message in the first commit for more details.
* sort the organizations by last seen at (#3726)
* moves `models.ForgottenDomain` to `data.ForgottenDomain`. See the commit message in the third commit for more details.

Fixes #3726


internal/format/format.go
```diff
@@ -65,10 +65,14 @@ func HumanDurationWithCase(d time.Duration, useCaps bool) string {
 }
 
 func HumanTime(t time.Time, zeroValue string) string {
-	return HumanTimeWithCase(t, zeroValue, true)
+	return humanTimeWithCase(t, zeroValue, true)
 }
 
-func HumanTimeWithCase(t time.Time, zeroValue string, useCaps bool) string {
+func HumanTimeLower(t time.Time, zeroValue string) string {
```
dnephin, line 71: Functions with boolean arguments are generally not great because it's not clear what that boolean means from the caller.

We already have an exported function for using uppercase, so this changes it to handle the only other remaining case (lowercase the first letter).

internal/server/data/organization.go
```diff
@@ -158,3 +158,28 @@ func UpdateOrganization(tx WriteTxn, org *models.Organization) error {
...
 }
+
+type ForgottenDomain struct {
+	OrganizationName   string
+	OrganizationDomain string
+	// LastSeenAt is the last time the user logged into this organization.
+	LastSeenAt time.Time
+}
+
+func GetForgottenDomainsForEmail(tx ReadTxn, email string) ([]ForgottenDomain, error) {
```
dnephin, line 169: This was moved from internal/server/data/forgotdomain.go.

I also formatted the query across multiple lines so that it is easier to read and changed to use the `scanRows` helper.

internal/server/data/organization.go
```diff
@@ -158,3 +158,28 @@ func UpdateOrganization(tx WriteTxn, org *models.Organization) error {
 func CountOrganizations(tx ReadTxn) (int64, error) {
 	return countRows(tx, organizationsTable{})
 }
+
+type ForgottenDomain struct {
+	OrganizationName   string
+	OrganizationDomain string
+	// LastSeenAt is the last time the user logged into this organization.
+	LastSeenAt time.Time
```
dnephin, line 166: This type was moved from `models`. 

> This type is not really a model. I doesn't have CRUD operations, and it
doesn't act as a representation of a database table.

> Types like this (see data.DestinationsCount for another example) are
nothing more than return values for a single data function. Moving it
to the data package makes this more clear. It also makes the code much
easier to trace because you don't have to jump into a completely
separate file and package when looking up the type.

`LastSeenAt` was also changed from a `string` to a `time.Time` which made testing this with strict assertions much easier. The formatting of the time moved to the template, which is covered by a test.

internal/server/email/forgotdomain.go
```diff
@@ -1,13 +1,21 @@
...
 
 type ForgottenDomainData struct {
-	Domains []models.ForgottenDomain
+	Organizations []data.ForgottenDomain
 }
 
 func SendForgotDomainsEmail(name, address string, data ForgottenDomainData) error {
+	// TODO: this should probably be done in the caller, but we don't have a reasonable
+	// way to test it in the caller. Once we can test email sends from the API handlers
+	// consider moving this to the caller.
```
dnephin, line 16: Tracked in #3889

internal/server/email/forgotdomain.go
```diff
@@ -1,13 +1,21 @@
...
+	sort.Slice(data.Organizations, func(i, j int) bool {
+		return data.Organizations[i].LastSeenAt.After(data.Organizations[j].LastSeenAt)
+	})
```
dnephin, line 19: The fix for #3726. I went with LastSeenAt sort order assuming that the most recent one is more likely to be the one they want.

dnephin (commented):
There are a lot of very small changes here , so I tried to explain them inline. If anything does not make sense, or the rational is missing, please do ask and I can elaborate!

BruceMacD (approved):
looks good to me

internal/server/data/organization.go
```diff
@@ -158,3 +158,28 @@ func UpdateOrganization(tx WriteTxn, org *models.Organization) error {
...
+}
+
+func GetForgottenDomainsForEmail(tx ReadTxn, email string) ([]ForgottenDomain, error) {
+	var results []ForgottenDomain
+
+	rows, err := tx.Query(`
```
pdevine, line 172: The query is definitely easier to read split out vs. a single line, and the generic scanRows() is much nicer too.

pdevine (approved):
LGTM.

PR #3891, opened 2022-12-16
Remove some unused errors

dnephin:
## Summary

`ErrNotImplemented` was not used at all.

`ErrInternalServerError` doesn't need to exist. It's not handled in any special way, and it's already the default behaviour for an unhandled error.

`ErrConflict` was only used in one place. We have `data.UniqueConstraintError` for any conflict in the database, so we can use that. I added the `Value` field to that error, so that we can continue to include the conflicting value in the error message.


jmorganca (approved):
Great cleanup!

PR #3910, opened 2022-12-19
fix: update version to use semver metadata instead prerelease

mxyng:
## Summary

semver's prerelease, the part after the first `-`, is inherently a lower than the released version. `git describe` outputs `<tag>-<commits since tag>-<checksum>` so any commit after the tag will have a lower version than the tag. 

Instead, modify the output so the string is `<tag>+<commits since tag>-<checksum>` which is also a better representation of this info.

```
$ infra login infra.infrahq.lvh.me
  Logging in to infra.infrahq.lvh.me
Error: client version (0.20.0) is ahead of server version (0.20.0-30-gd394a245), download the CLI version that matches your server, reference https://infrahq.com/docs/messages/cli-versions
```

### DIFF
--- a/.github/workflows/cd-binaries.yaml
+++ b/.github/workflows/cd-binaries.yaml
@@ -26,7 +26,7 @@ jobs:
       - id: release-name
         run: |
           RELEASE_NAME=${{ startsWith(github.ref, 'refs/tags/') && github.ref_name || '$(git describe --tags)' }}
-          echo "release-name=${RELEASE_NAME#v}" >>$GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


dnephin (approved):
LGTM!

For anyone not familiar with it https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions is a good reference for the semver format. I used that to refresh my memory.

I tested it locally with this:

```
export RELEASE_NAME=$(git describe --tags); echo ${RELEASE_NAME#v} | sed 's/-/+/'$ 
0.20.0+9-g33a9bfe5
```

And that works nicely

PR #3913, opened 2022-12-20
fix: allow user to remove own password login

### LINKED ISSUE #3912: Users should be able to remove their own password if there's another form of authn configured
Currently, as a precaution against locking a user out permanently, users are not able to delete their own Infra user. However, when an identity provider is configured, the local user may be obsolete. In this case, the local user can be deleted by an admin to remove password authn for that user. 

However, an admin cannot remove their own local user.

mxyng:
If another login method exists, e.g. an identity provider, the user should be able to remove their own password. A side effect is a non-admin user can remove their own password, if another authn method exists for that user.

Resolves #3912

### DIFF
--- a/ui/pages/users/index.js
+++ b/ui/pages/users/index.js
@@ -343,9 +343,11 @@ export default function Users() {
                 }
               )
 
-              // can only delete users that exist within Infra which are not the currently logged in user
+              // can only delete Infra users where password authn is the only authn method
               if (
-                info.row.original.id === user?.id ||
+                (info.row.original.id === user?.id &&
+                  info.row.original.providerNames?.filter(n => n != 'infra')
+                    .length == 0) ||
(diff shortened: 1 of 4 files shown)


ui/pages/users/index.js
```diff
@@ -343,9 +343,11 @@ export default function Users() {
                 }
               )
 
-              // can only delete users that exist within Infra which are not the currently logged in user
+              // can only delete Infra users where password authn is the only authn method
               if (
-                info.row.original.id === user?.id ||
+                (info.row.original.id === user?.id &&
+                  info.row.original.providerNames?.filter(n => n != 'infra')
```
BruceMacD, line 349: I think you should still be able to delete Infra users when they have logged in via other sources so this restriction on `infra` being the only provider isn't quite right

mxyng:
Closing this for now since we're likely taking another approach

PR #3909, opened 2022-12-19
feat: accept user invite by logging in with google

BruceMacD:
## Summary
UI changes to the "accept invite" flow. Users can now choose to login with Google rather than setting a password.

This change also cleans up some duplicate `oidcLogin` functions that were a result of a rebase of the main branch onto the sign-up branch which was merged previously. 

Clean up duplicate logic: ae2103fffb283d87c09cde8fefa6c4a600eb745d
Accept invite: 86a8a3dec5e379faafa1b9b9326af7f0c1ff7b18

📸 
Screen Shot 2022-12-20 at 9 53 12 AM

Follow-up for the back-end of this change in #3908

## Checklist

- [x] Considered security implications of the change
- [x] Change is backwards compatible if it needs to be (user can upgrade without manual steps?)
- [x] Nothing sensitive logged
- [x] Considered data migrations for smooth upgrades

## Related Issues

Part of #3261

PR #3918, opened 2022-12-20
fix: get refresh token on google login

### LINKED ISSUE #3917: UI Login Flow does not get refresh token from Google
## Describe the bug
No refresh token is retrieved from Google on OIDC login through the UI.

## Expected behavior
The consent prompt and access_type parameters are set in redirect for Google login in the UI. 
```
if provider.Kind == "google" {
    params.Add("prompt", "consent")      // google only sends a refresh token when a user consents, always prompt so we always get the ref token
    params.Add("access_type", "offline") // specifies that we want a refresh token
}
```

BruceMacD:
## Summary
In order to extend get a refresh token from Google additional parameters must be specified on OIDC login. This was done in the CLI but the UI was out of sync here.

Add the required parameters for Google sign-up and login to get a refresh token from Google.

This change is made in 2 places. #3909 removes the duplicate code.

## Checklist

## Related Issues

Resolves #3917

### DIFF
--- a/ui/components/providers.js
+++ b/ui/components/providers.js
@@ -6,7 +6,7 @@ import { currentBaseDomain } from '../lib/login'
 import { providers as providersList } from '../lib/providers'
 
 export function oidcLogin(
-  { baseDomain, loginDomain, id, clientID, authURL, scopes },
+  { baseDomain, loginDomain, id, clientID, authURL, scopes, kind },
   next
 ) {
   if (baseDomain === '') {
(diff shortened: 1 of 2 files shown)


ssoroka (approved):

ui/components/providers.js
```diff
@@ -48,9 +48,22 @@ function oidc(id, clientID, authURL, scopes, redirectURL, next) {
...
-  )}&state=${state}`
+  const sendTo = new URL(authURL)
+  // URL searchParams add query parameters to a URL
+  sendTo.searchParams.append('redirect_uri', redirectURL)
+  sendTo.searchParams.append('client_id', clientID)
+  sendTo.searchParams.append('response_type', 'code')
+  sendTo.searchParams.append('scope', scopes.join(' '))
+  sendTo.searchParams.append('state', state)
+
+  if (kind === 'google') {
```
jmorganca, line 59: Is there any harm in _always_ adding these params?

BruceMacD: In the case of Azure it means an extra step approving permissions every time you login, so I'd like to keep it specific for now.

ui/pages/login/index.js
```diff
@@ -49,9 +49,22 @@ function oidcLogin(
...
+  sendTo.searchParams.append('client_id', clientID)
+  sendTo.searchParams.append('response_type', 'code')
+  sendTo.searchParams.append('scope', scopes.join(' '))
+  sendTo.searchParams.append('state', state)
+
+  if (kind === 'google') {
+    // google only sends a refresh token when a user consents, always prompt so we always get the ref token
+    sendTo.searchParams.append('prompt', 'consent')
+    // also need to specify offline access in the case of Google to get a refresh token
+    sendTo.searchParams.append('access_type', 'offline')
```
jmorganca, line 64: Shouldn't we always include this (for Okta, others too)?

ui/components/providers.js
jmorganca: Thanks!

jmorganca (approved):

PR #3921, opened 2022-12-20
fix: edit user password

### LINKED ISSUE #3882: User cannot update their own password using CLI
Trying to change their own password returns an error `oldPassword is required` but there's no field to enter the old password

```
$ infra info

 Server: infra.infrahq.lvh.me
   User: admin@local (hW3uKifXoG)
 Groups: (none)

```

```
$ infra users edit admin@local --password
  Enter a new password:
? Password: ********
? Confirm Password: ********
Error: validation failed: oldPassword: is required
```

```
$ infra version

 Client: 0.20.0

```

mxyng:
## Summary

`infra user edit` was never updated to prompt for the user's old password. For security reasons, the server rejects any password update request without the old password. Initial login for a new user does this implicitly since the user has already entered the password in the login request but this won't be available for an explicit password update.

Resolves #3882

### DIFF
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -275,10 +275,14 @@ func updateUser(cli *CLI, name string) error {
 	if isSelf {
 		req := &api.UpdateUserRequest{ID: config.UserID}
 
-		fmt.Fprintf(cli.Stderr, "  Enter a new password:\n")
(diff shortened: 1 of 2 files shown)


BruceMacD (approved):
good fix

PR #3924, opened 2022-12-21
fix: only set namespace if specified in use

mxyng:
## Summary

`infra use cluster` with a cluster unsets context namespace which isn't expected. Instead if the namespace isn't specified, only switch the context. If a namespace is specified, set the namespace accordingly.


dnephin (approved):
LGTM, and nice test case!

jmorganca (approved):
Thanks for fixing!

PR #3925, opened 2022-12-21
Improve error output from infra ssh

dnephin:
## Summary

@pdevine realized that if someone has an SSH key called infra, it's possible that key exists at `~/.ssh/infra`, which would prevent us from using that path as a directory. The error message would be hidden from users, so they wouldn't understand why their SSH was prompting them for either a password or to verify the host key.

This PR fixes that problem by:
1. improving the error message in this case, making it explicit that there is a file in the way
2. actually printing this error message to stderr (instead of hiding it in debug logs)

I created #3927 for the remaining problems

### DIFF
--- a/internal/cmd/ssh.go
+++ b/internal/cmd/ssh.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"text/template"
 
 	"github.com/spf13/cobra"
@@ -157,7 +158,9 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
 		return fmt.Errorf("user home directory: %w", err)
 	}
 	infraSSHDir := filepath.Join(homeDir, ".ssh/infra")
-	_ = os.MkdirAll(infraSSHDir, 0o700)
+	if err := mkdirAll(infraSSHDir); err != nil {
+		return err
+	}
 
 	hostCfg, err := currentHostConfig()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/cmd/ssh.go
```diff
@@ -43,11 +44,23 @@ func newSSHHostsCmd(cli *CLI) *cobra.Command {
 		Args:  ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			host, port := args[0], args[1]
-			if err := runSSHHosts(cli, host, port); err != nil {
+
+			err := runSSHHosts(cli, host, port)
+			switch {
+			case errors.Is(err, errNotInfraDestination) || errors.Is(err, errBeforeDestinationMatch):
```
dnephin, line 50: The second error here is category 1 from #3927. We need a better solution for those errors, but for now I've preserved the existing behaviour of hiding them.

internal/cmd/ssh.go
```diff
@@ -198,6 +217,19 @@ func setupDestinationSSHConfig(ctx context.Context, cli *CLI, destination *api.D
...
+// an error. mkdirAll is a wrapper around os.MkdirAll to provide a better error
+// message when the path already exists as a file.
+func mkdirAll(path string) error {
+	err := os.MkdirAll(path, 0o700)
+	pathError := &fs.PathError{}
+	if errors.As(err, &pathError) && errors.Is(pathError.Err, syscall.ENOTDIR) {
+		return fmt.Errorf("failed to create directory %v, the path already exists as a regular file",
+			pathError.Path)
+	}
+	return err
```
BruceMacD, line 230: I think this error should still be captured somewhere even if its an error other than one for a dir that already exists.

BruceMacD (commented):
I think the directory creation error shouldnt be swallowed when unexpected, but other than that looking good

internal/cmd/ssh.go
dnephin: For sure, I think both these errors should get printed.

The only time we hide errors are in the cases above that wrap `errBeforeDestinationMatch`, and those are printed as debug logs, so we can get them to show up when we need to.

pdevine (approved):
LGTM.

PR #3928, opened 2022-12-21
add selinux module for infra

### LINKED ISSUE #3893: `AuthorizedKeysCommand` will fail if running with SELinux mode set to `enforcing`
When running on Linux with SELinux enabled and set to `enforcing` mode (e.g. an out-of-the-box RHEL server on AWS), outbound requests from `AuthorizedKeysCommand` will fail with `permission denied`. This seems to stem from an SELinux policy around `AuthorizedKeysCommand`.

mxyng:
## Summary

Package and install a SELinux module on RHEL installs to allow Infra to connect when invoked by SSH. The policy is created by `ausearch  -c 'infra' --raw | audit2allow -M infra`. The `.te` file is:

```
module infra 1.0;

require {
        type http_port_t;
        type sshd_t;
        class tcp_socket name_connect;
}

#============= sshd_t ==============

allow sshd_t http_port_t:tcp_socket name_connect;
```

Note: `audit2allow` reports this can also be enabled with either `authlogin_yubikey` or `nis_enabled` which may be better options, particularly `nis_enabled`

Verified on RHEL 9.1 on EKS

Resolves #3893


.goreleaser.yml
```diff
@@ -47,6 +47,10 @@ nfpms:
       - src: package-files/connector.example.yaml
         dst: /etc/infra/connector.example.yaml
         type: config
+      - src: package-files/rpm/infra.pp
```
BruceMacD, line 50: Do we track the source code of this file anywhere? How do we update it?

dnephin: How difficult would it be to include the source in git, and build this `.pp` as part of the release?  We don't need goreleaser to build it, we can do it as a step before goreleaser.

dnephin (commented):
Thanks for working on this! I was hoping we could add a red-hat image to the new end-to-end tests in `tests/` that fails without this change. I'll be doing some work in that area in the new year, but if you're interested in adding it, that'd be cool too.

mxyng:
I won't be possible to test in the current E2E tests since SELinux isn't intended to run in container. The security module isn't namespaced so it's not available for the container. It'll be disabled regardless of its state in the host. To test, we'll need virtualization

dnephin:
Ah, that's too bad. I didn't realize it wouldn't work in a container

.goreleaser.yml
mxyng: We could but it's pretty annoying. It requires `checkmodule` to compile `*.te` into a `*.mod` then `semodule_package` to package `*.mod` into a `*.pp`. Each binary comes from a separate package, `checkpolicy` for `checkmodule` and `policycoreutils` for `semodule_package`, on Fedora 36. I would rather just turn on `nis_enabled` in that case

package-files/postinstall.sh
```diff
@@ -10,3 +10,8 @@ if [ -f "$HOMEDIR/connector.yaml" ]; then
     systemctl restart infra
   fi
 fi
+
+. /etc/os-release
+case $ID in
+  rhel) semodule -i /usr/share/infra/selinux/infra.pp ;;
```
BruceMacD, line 16: This may be unnecessary but where the source code of this file is pretty straightforward I wonder if we should just put the whole thing in a comment about this line in the script. 

Ex:
```
# **infra.pp**
# module infra 1.0;
# require {
#         type http_port_t;
#         type sshd_t;
#         class tcp_socket name_connect;
# }
# allow sshd_t http_port_t:tcp_socket name_connect;
rhel) semodule -i /usr/share/infra/selinux/infra.pp ;;
```

Just an idea though.

BruceMacD (approved):

.goreleaser.yml
dnephin: Installing a couple packages doesn't seem like a big problem. What am I missing? Are these packages particularly large and slow to download?

dnephin (commented):
We should definitely include the source for this in the repo somewhere.

I'd still prefer we build the `.pp` in CI, so that it's easier to maintain later. If we don't do that we at least have to document how to build it, which might end up being more work than automating it in CI.

dnephin (approved):

jmorganca:
This is fantastic! Thanks @mxyng for adding the source

Issue #3915, opened 2022-12-20
Reduce `Server.Config` to only bootstrap config


dnephin:
## Background

Today we support creating entities and removing entities from the server config file. This is implemented in [loadConfig](https://github.com/infrahq/infra/blob/e1b141ebfa9eced506a9d9304a8607fe22eebdee/internal/server/config.go#L600) , the input struct is [`Server.Config`](https://github.com/infrahq/infra/blob/e1b141ebfa9eced506a9d9304a8607fe22eebdee/internal/server/config.go#L79-L84). We support both creating and removing users, grants, access keys, infra local credentials (username/passwords), and providers.

Now that we have a [terraform provider](https://github.com/infrahq/terraform-provider-infra) most of the use cases we were solving with `loadConfig` are better solved with the terraform provider.

The one use case that can't be solved by the terraform provider is bootstrapping the server. There are two scenarios when bootsrapping is extremely valuable:
* as a "break glass" mechanism for adding an admin or support-admin when an IDP is either unavailable, or has been mis-configured in a way that prevents logins (Ex: someone accidentally deleted the integration from the provider side)
* testing - we frequently need to bootstrap the server in all layers of tests (API handler tests, end-to-end tests, CLI tests, connector tests, etc). We could write something test-only, but that would force us to build something different for break-glass, which is more code to maintain.

## Proposal

Edit: some parts of this proposal were implemented in https://github.com/infrahq/infra/pull/3930

This issue proposes that we reduce `loadConfig` to just the bit we need to satisfy these the bootstraping use case:
* remove `Provider` - that can be setup via the UI or terraform
* remove `Grants` - we only need grants for Infra, there's no use case for bootstrapping destination grants because we can do that from the API.  See below for the replacement to grants.
* remove "delete entity no longer in the config" - bootstrap config only needs to create things, it doesn't need to remove them. This allows us to remove the `CreatedBy` and `NotIDs` query parameters from our list queries, which simplifies the data layer. With this change we no longer have any special requirements on the data layer , we can use the standard options that get used everywhere else
* remove `User.Password` - generally we probably want to be creating short lived access keys for break glass and testing. We don't need to create a password that never expires

With those pieces removed we significant decrease the footprint of this feature. But we also lose a couple things that we need, so this proposal includes the addition of two other small things:

* add `User.Role` (or maybe `User.InfraRole`) - when set to a value this user is given a grant where resource=infra. This allows us to bootstrap admins and support admins
[...]

PR #3930, opened 2022-12-21
Reduce server config to only bootstrapping of user and credentials

dnephin:
## Summary

Implements most of what was proposed in #3915. 

Not done, still pending further discussion:
* `User.Password` has not been removed - it's used by helm
* `SupportAdminGroups` has not been added yet
* the access key expiration time is still set to 10 years

I'd like to address those in separate follow up PRs. 

If a self-hosted install of Infra includes `grants` or `providers`  the server will refuse to start and the error message will inform the user what needs to change.

Helm changes in https://github.com/infrahq/helm-charts/pull/17

### DIFF
--- a/internal/cmd/server_test.go
+++ b/internal/cmd/server_test.go
@@ -124,18 +124,12 @@ func TestServerCmd_LoadOptions(t *testing.T) {
 			name: "env vars with config file",
 			setup: func(t *testing.T, cmd *cobra.Command) {
 				content := `
-grants:
-  - user: user1
-    resource: infra
-    role: admin
-  - user: user2
-    resource: infra
-    role: admin
 
 users:
   - name: username
     accessKey: access-key
     password: the-password
+    role: admin
 `
 				dir := fs.NewDir(t, t.Name(),
 					fs.WithFile("cfg.yaml", content))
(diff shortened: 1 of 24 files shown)


internal/server/data/grant.go
```diff
@@ -273,12 +264,6 @@ func DeleteGrants(tx WriteTxn, opts DeleteGrantsOptions) error {
 		query.B("id = ?", opts.ByID)
 	case opts.BySubject != "":
 		query.B("subject = ?", opts.BySubject)
-	case opts.ByCreatedBy != 0:
```
BruceMacD, line 276: nice, glad to see this get cleaned up

BruceMacD (approved):

internal/server/config.go
```diff
@@ -12,62 +12,25 @@ import (
...
+type BootstrapConfig struct {
+	DefaultOrganizationDomain string
+	Users                     []User
 }
 
 type User struct {
 	Name      string
 	AccessKey string
 	Password  string
+	Role      string
```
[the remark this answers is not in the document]
dnephin: Ya, I like that! I'll change it to `InfraRole`

internal/cmd/server_test.go
```diff
@@ -124,18 +124,12 @@ func TestServerCmd_LoadOptions(t *testing.T) {
...
-    role: admin
-  - user: user2
-    resource: infra
-    role: admin
 
 users:
   - name: username
     accessKey: access-key
     password: the-password
+    infraRole: admin
```
jmorganca, line 132: Fantastic clean up in this PR. It's most likely better to visit this in a follow up PR, but could we further simplify this by removing `users:`? (can move this to a new issue if so)

On first thought, a minimal bootstrap interface could be:

1. A single `accessKey` field for bootstrapping the system via terraform and for follow-on support admin needs
2. A username+password (for bootstrapping via the dashboard) although I don't know if this is a great experience. Quite a few tools would allow just allow the user to sign up as usual (of course would require multi-org by default) after standing up the server by visiting the endpoint it exposes in the browser

dnephin:
Helm chart changes are in https://github.com/infrahq/helm-charts/pull/17. I need to test those first before merging

jmorganca:
LGTM once any dependent changes are in! (E.g. helm)

dnephin:
From https://github.com/infrahq/helm-charts/pull/17#issuecomment-1376196751, it sounds like we probably need to merge this first. The helm changes would come after we have a version of the server image to tag.

jmorganca:
That makes sense, I had it backwards. LGTM!

jmorganca (approved):

PR #3949, opened 2023-01-03
Remove old client config migrations

dnephin:
These version are now older than 6 months. The CLI upgrades the file when it is run, so someone would have to be upgrading their CLI from a very old version. In that case they would  have to remove the old file and login again.


jmorganca (approved):
Good cleanup 👍

PR #3950, opened 2023-01-03
Inline and reduce database encryption at-rest logic

dnephin:
## Summary

In #1653 secrets were extracted to a library. Part of the logic that moved to that library was sealing and unsealing of sensitive values that are stored in the database.

This PR moves just the bits used for database encryption at-rest  from `github.com/infrahq/secrets` to `./internal/server/data/encrypt`. We don't need this code in a library since it's only used in this one place, and moving it inline makes it easier to change.

At the same time the implementation is reduced a bit by removing "key providers". Instead, the "native key provider" implementation is used, which reads the root key from a file. I believe this is the only mechanism that we are using for the root key. We can support other mechanism in the future using the same mechanism we do for other secret values (load from env or file).

This change is being done in preparation for removing the rest of the secret loading from config (#2186).

## Related Issues

Related to #2186

### DIFF
--- a/internal/server/models/encryption_test.go
+++ b/internal/server/models/encryption_test.go
@@ -4,7 +4,6 @@ import (
 	"testing"
 
 	"gotest.tools/v3/assert"
-	is "gotest.tools/v3/assert/cmp"
 
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
@@ -51,10 +50,8 @@ func TestEncryptedAtRest(t *testing.T) {
 
 	assert.Assert(t, "don't tell" != result)
 	assert.Assert(t, "" != result)
-	assert.Assert(t, is.Len(result, 88)) // encrypts to this many bytes
 
 	m2 := &StructForTesting{}
-
 	err = db.QueryRow(`SELECT a_secret FROM struct_for_testings where id = ?`, id).Scan(&m2.ASecret)
 	assert.NilError(t, err)
 
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -90,6 +90,15 @@ type Options struct {
 	API  APIOptions
 
 	DB data.NewDBOptions
+
+	DeprecatedConfig
+}
+
+// DeprecatedConfig contains fields that are no longer used by server, but loading
+// values for these fields allows us to error when a config file value is no
+// longer supported.
+type DeprecatedConfig struct {
+	DBEncryptionKeyProvider string
 }
 
 type ListenerOptions struct {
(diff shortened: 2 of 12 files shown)


internal/server/server.go
```diff
@@ -55,15 +58,14 @@ type Options struct {
 	GoogleClientID     string
 	GoogleClientSecret string
 
-	DBEncryptionKey         string
-	DBEncryptionKeyProvider string
```
dnephin, line 59: To catch any potential problems with upgrades of self-hosted infra servers we could keep this field and error if it's set to anything other than `native`.

internal/server/server.go
```diff
@@ -169,13 +171,13 @@ func New(options Options) (*Server, error) {
...
 	options.DB.DSN = dsn
+	options.DB.RootKeyFilePath = options.DBEncryptionKey
 
-	dbKeyProvider, ok := server.keys[options.DBEncryptionKeyProvider]
-	if !ok {
-		return nil, fmt.Errorf("key provider %s not configured", options.DBEncryptionKeyProvider)
+	if _, err := os.Stat(options.DB.RootKeyFilePath); errors.Is(err, fs.ErrNotExist) {
+		if err := encrypt.CreateRootKey(options.DB.RootKeyFilePath); err != nil {
+			return nil, err
+		}
```
dnephin, line 179: This initializes a new root key when it's missing. This used to work at some point, but when I wrote the end-to-end test suite it wasn't working, so I had to do it explicitly here:  https://github.com/infrahq/infra/blob/c5b9fa43d6d910b5d55a75cd5a74c5ffeb6a7faf/.github/workflows/ci-core.yaml#L139

I'm not sure what was not working. If we want to initialize a key, it should be part of the server process startup logic. So I've added this here to restore the initializing of a root key when it doesn't exist.

internal/server/models/encryption_test.go
```diff
@@ -51,7 +50,9 @@ func TestEncryptedAtRest(t *testing.T) {
 
 	assert.Assert(t, "don't tell" != result)
 	assert.Assert(t, "" != result)
-	assert.Assert(t, is.Len(result, 88)) // encrypts to this many bytes
+
+	// TODO: why did this length change to 139/140 bytes?
+	// assert.Assert(t, is.Len(result, 88)) // encrypts to this many bytes
```
dnephin, line 55: TODO: look into this test failure and restore the assertion

dnephin: I think I understand this test failure now. Previously the `RootKeyID` stored in the `encryption_keys` table was `db_at_rest`. Now it's the full path to the file that holds the key. I believe that accounts for the extra 52 bytes and explains why sometimes it's 1 byte less (the random temp directory path has 1 less character).

I need to fix this to store just the root key ID.

dnephin: Looking at this more, I think this is only a bug in the test. In the production code path we were always storing the full path in this field because our defaults set a full path: https://github.com/infrahq/infra/blob/c5b9fa43d6d910b5d55a75cd5a74c5ffeb6a7faf/internal/cmd/server.go#L94

I'll remove this assertion since we check the round trip already, which is the important thing for this test.

internal/server/server.go
dnephin: I've added this in the latest commit.

jmorganca: Custom DB encryption key providers will be extremely rare. I think we're okay if we document dropping support for this in the release notes.

jmorganca: (Current approach is good too)

jmorganca (approved):
LGTM and great cleanup to simplify this.

As mentioned in a comment we can probably do without `DeprecatedConfig` – PR is good as is though since it's very little addl' code

PR #3951, opened 2023-01-03
Reduce secret loading of secrets to only config files and environment variables

dnephin:
## Summary

Branched from #3950

This PR removes secret loading from external sources. Secrets can be loaded from environment variables, which is what we use in production. Kubernetes makes it easy to load values from other sources and set them as environment variables in the container.

This allows us to simplify our configuration loading quite a bit, and allows us to remove a bunch of dependencies.

The server options that change in this PR are:
* `tls.caPrivateKey` - any values with `file:` prefix need to remove the prefix
* `tls.privateKey` - any values with `file:` prefix need to remove the prefix
* `keys` - (the top-level field) no longer exists, it was used to define key providers
* `secrets` - (the top-level field) no longer exists, it was used to define custom secret providers
* `redis.password` - no longer supports `env:` or `file:` prefix, specify the value from `INFRA_SERVER_REDIS_PASSWORD` instead
* `dbPassword` - no longer supports `env:` or `file:` prefix, specify the value from `INFRA_SERVER_DB_PASSWORD` instead
* `providers[].clientSecret` - is removed in #3930 anyway, so removing secret loading from this should not matter
* `users[].password` - no longer supports `env:` or `file:` prefix. #3915 proposes removing this field. We may need an alternative way of specifying this secret if we keep it
* `users[].accessKey` - no longer supports `env:` or `file:` prefix. We may need an alternative way of specifying this secret if we don't want it in the main config file.

#3987 implements the backwards compatibility for  `users[].password` and `users[].accessKey`, which removes the need for helm chart changes.

## Related Issues

Related to https://github.com/infrahq/infra/issues/2186


jmorganca (approved):
Great cleanup

mxyng:
Good changes overall

> do we need to implement the alternative secret config syntax from [...] for `users[].password` and `users[].accessKey`?

Do these changes allow secret loading from a file? If no, I don't see a need for an alternative. Otherwise, an alternative is probably necessary since we likely can't remove password and access keys completely.

This change will break existing deployment if they're using any secret loading. The Helm chart also needs to be updated since it's using `file:` for the user password.

dnephin:
> Do these changes allow secret loading from a file?

Ya, the only way to specify users is from a config file.

Another option (instead of implementing the alternative syntax) would be to only allow creating temporary passwords and access keys with a 1 day expiry time.

Since this is only bootstrap config, we don't need these credentials to live for long. That way we don't really need secret loading, because the values become invalid very quickly, there's very little change they can be leaked before they expire.

jmorganca (approved):
Great cleanup. How close does this get us to removing `github.com/infrahq/secrets` as a dependency (vs everything being in this repo)?

dnephin:
> How close does this get us to removing github.com/infrahq/secrets as a dependency (vs everything being in this repo)?

It's done in this PR. ` github.com/infrahq/secrets` is removed from `go.mod`.

dnephin:
With the changes in https://github.com/infrahq/infra/pull/3987 (which is branched from this PR), this should be backwards compatible with the existing helm chart. 

I looked through our docs, and I didn't find anything about configuring secret providers in the self-hosting guide. I guess we probably removed those docs a while ago.

jmorganca:
@dnephin yes we did in favor of recommending an environment variable or file-based approach

pdevine (approved):
I love this. Nice cleanup.

Issue #3944, opened 2022-12-29
`POST /api/grants`  and `POST /api/tokens` results in high latency


jmorganca:
When making post requests to `/api/grants`, the requests can take anywhere from 3-5s

ssoroka:
As far as I can tell, there are no POST calls taking more than 1s. When I filter on POST (to block out the long-polling GET requests that are supposed to take 5s), there are no log results matching this criteria

mxyng:
Some improvements have been made to the UI which should mitigate this issue. The UI, when creating grants for multiple namespaces, sends a request for each namespace. This can lead to a flood of requests which ironically if is received by the server too slowly will cause a backlog of requests and each request is blocked by the preceding request. 

The improvement is to send a bulk create request so no matter how many namespaces are selected, a single request will be sent.

A second minor improvement focuses on deleting multiple namespace grants. Similar to the earlier improvemnt, this action sends a DELETE request for each selected namespace. Additionally, each DELETE request is followed by a GET to update the UI state. Therefore, the number of requests per selected namespace is actually not 1 but 2 with all of the side effects mentioned earlier. This has been updated to a bulk delete request with a single GET once the bulk delete request has been handled.

PR #3952, opened 2023-01-03
fix(ui): bulk add grants

mxyng:
## Summary

Use PATCH /api/grants to bulk add grants. This is particularly useful when adding grants for a large number of resources such as namespaces.

Relates #3944

### DIFF
--- a/ui/pages/destinations/[id]/index.js
+++ b/ui/pages/destinations/[id]/index.js
@@ -841,63 +841,31 @@ export default function DestinationDetail() {
                       privilege,
                       selectedResources,
                     }) => {
-                      // don't add grants that already exist
                       if (selectedResources.length === 0) {
-                        if (
-                          grants?.find(
-                            g =>
-                              g.user === user &&
-                              g.group === group &&
-                              g.privilege === privilege &&
-                              g.resource === destination?.namn
-                          )
-                        ) {
-                          return false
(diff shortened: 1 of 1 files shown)


ui/pages/destinations/[id]/index.js
```diff
@@ -841,63 +841,31 @@ export default function DestinationDetail() {
                       privilege,
                       selectedResources,
                     }) => {
-                      // don't add grants that already exist
```
Comment (mxyng) [inline ui/pages/destinations/[id]/index.js:844]: Since this is no longer calling the API once for each resource, it's less critical to filter out existing grants

ui/pages/destinations/[id]/index.js
```diff
@@ -841,63 +841,31 @@ export default function DestinationDetail() {
...
-                                privilege,
-                                resource: `${destination?.name}.${resource}`,
-                              }),
-                            })
-                          }
-                        )
+                      await fetch('/api/grants', {
+                        method: 'PATCH',
+                        body: JSON.stringify({
+                          grantsToAdd,
```
Comment (hoyyeva) [inline ui/pages/destinations/[id]/index.js:863]: hmm does this work? I thought the parameter is called `GrantsToAdd` on the PATCH endpoint? so isn't it supposed to be { GrantsToAdd: grantsToAdd }

hoyyeva (approved):

PR #3953, opened 2023-01-03
fix(ui): bulk remove grants

mxyng:
## Summary

Use PATCH `/api/grants` to bulk remove grants. In contrast to adding grants, only multi-select remove uses this API. Other removals, cluster-wide and remove from the dropdown menu continue to use existing DELETE `/api/grants`.

Relates #3952
Relates #3944


ui/pages/destinations/[id]/index.js
```diff
@@ -270,10 +270,30 @@ function GrantCell({ grantsList, grant, destination, onRemove, onUpdate }) {
...
                       )
+
+                      await fetch('/api/grants', {
+                        method: 'PATCH',
+                        body: JSON.stringify({
+                          grantsToRemove: grantsToRemove.flat(),
+                        }),
+                      })
+
+                      onRemove([])
```
Comment (mxyng) [inline ui/pages/destinations/[id]/index.js:296]: This calls `onRemove` to update the table to reflect the removed grants. It passes in an empty list so it doesn't actually remove anything and just calls `mutate`

hoyyeva (approved):

PR #3957, opened 2023-01-04
Fix CI failure in acceptance tests

dnephin:
The e2e test suite was not running because its go.mod was not tidy. This workspace should inform Go to use the go.mod from the parent module.

Example failure on main: https://github.com/infrahq/infra/actions/runs/3839811288/jobs/6538079495


mxyng (approved):
If it works, LGTM

PR #3962, opened 2023-01-05
Add debian and redhat destination to SSH acceptance tests

dnephin:
## Summary

Adds `debian:11-slim` and `redhat/ubi9:latest` destinations, and runs the same tests against them.

A few small changes were required, comments inline

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -42,7 +42,7 @@ nfpms:
     bindir: /usr/local/sbin
     contents:
       - src: package-files/systemd/infra.service
-        dst: /usr/lib/systemd/system/infra.service
+        dst: /lib/systemd/system/infra.service
         type: config
       - src: package-files/connector.example.yaml
         dst: /etc/infra/connector.example.yaml
--- a/test/docker-compose.yaml
+++ b/test/docker-compose.yaml
@@ -37,11 +37,25 @@ services:
   destination_ubuntu:
     build:
       context: ./test
-      dockerfile: dockerfiles/Dockerfile.ubuntu
+      dockerfile: dockerfiles/Dockerfile.debian
+      args:
+        image_tag: ubuntu:22.10
     volumes:
       - /sys/fs/cgroup:/sys/fs/cgroup
       - ..:/work
     privileged: true
     depends_on: [server]
-    ports:
-      - "127.0.0.1:8220:22"
+    ports: ["127.0.0.1:8220:22"]
+
+  destination_debian:
+    build:
+      context: ./test
+      dockerfile: dockerfiles/Dockerfile.debian
+      args:
+        image_tag: debian:11-slim
+    volumes:
+      - ..:/work
+    privileged: true
+    depends_on: [server]
+    stop_grace_period: 1s
+    ports: ["127.0.0.1:8221:22"]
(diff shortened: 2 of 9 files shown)


.goreleaser.yml
```diff
@@ -42,7 +42,7 @@ nfpms:
     bindir: /usr/local/sbin
     contents:
       - src: package-files/systemd/infra.service
-        dst: /usr/lib/systemd/system/infra.service
+        dst: /lib/systemd/system/infra.service
```
dnephin, line 45: The old path wasn't supported by systemd on debian (https://manpages.debian.org/bullseye/systemd/systemd.unit.5.en.html).

This new path seems to work on both.

test/docker-compose.yaml
```diff
@@ -37,11 +37,25 @@ services:
...
+    ports: ["127.0.0.1:8220:22"]
+
+  destination_debian:
+    build:
+      context: ./test
+      dockerfile: dockerfiles/Dockerfile.debian
+      args:
+        image_tag: debian:11-slim
+    volumes:
+      - ..:/work
```
dnephin, line 57: For some reason using the `/sys/fs/cgroup:/sys/fs/cgroup` volume on debian was causing systemd to error. I don't think I ever found the underlying error, but it was manifesting as no dbus socket, so `systemctl` could not communicate with systemd.

So I've omitted the volume here

test/dockerfiles/install-debian.sh
```diff
@@ -2,20 +2,19 @@
...
 
 PACKAGE_PATH=/work/dist/infra_0.0.0_amd64.deb
-DESTINATION_ADDR=127.0.0.1:8220
 INFRA_SERVER_URL=test-server-1
-INFRA_ACCESS_KEY=dest000001.ubuntuubuntuubuntuubuntu
 
 
 # step=install-package
+apt-get update && apt-get install --no-install-recommends -y \
+    openssh-server procps
```
dnephin, line 13: Debian install failed because `procps` wasn't installed (at least on `slim`, in practice it's probably always installed). Our package depends on it, so normally a user would be fine to just run `apt --fix-broken install`, but we don't want to handle the error here, so I moved it before we `dpkg -i` our package.

jmorganca (approved):

dnephin:
I pushed one more commit that adds a redhat destination

test/setup.sh
```diff
@@ -4,4 +4,20 @@ set -eu -o pipefail
...
+  -e DESTINATION_NAME=debian \
+  -e DESTINATION_ADDR=127.0.0.1:8221 \
+  -e INFRA_ACCESS_KEY=dest000002.debiandebiandebiandebian \
+  test-destination_debian-1 /work/test/dockerfiles/install-debian.sh
+
+docker exec -i $TTY \
+  -e DESTINATION_NAME=redhat \
+  -e DESTINATION_ADDR=127.0.0.1:8222 \
+  -e INFRA_ACCESS_KEY=dest000003.redhatredhatredhatredhat \
+  test-destination_redhat-1 /work/test/dockerfiles/install-redhat.sh
```
dnephin, line 23: We could do these setup steps in parallel, but that ends up making the logging a bit harder to consume (when it comes time to debug a failure). We'd have to send the output to a file and save the file as an artifact.

I think we'll want to do that when we had the 4th or 5th destination, but for now it seems to run just fast enough we can do it sequentially.

mxyng (approved):

PR #3963, opened 2023-01-05
feat: add password breach checking when updating credentials

mxyng:
## Summary

It's a good idea to check if a new password is a known vulnerable value and prevent the user from setting it. It might be a good idea to make this a configuration along side the password requirements.


internal/access/credential.go
```diff
@@ -225,3 +234,43 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+	//nolint:gosec // this is required for the API
+	sum := sha1.Sum([]byte(password))
+	hexsum := hex.EncodeToString(sum[:])
+
+	request, err := http.NewRequestWithContext(context.Background(), http.MethodGet, fmt.Sprintf("https://api.pwnedpasswords.com/range/%s", hexsum[:5]), nil)
+	if err != nil {
+		return err
+	}
+
+	response, err := http.DefaultClient.Do(request)
```
dnephin, line 253: I guess we need some kind of timeout on this, probably set on the context.  The default client has no timeout set. We probably don't want to wait any more than 5-10 seconds, right?

I'm not sure if we should ignore the error on timeout (allowing someone to change their password when we can't check it), or if we should fail the request in that case. My guess is ignore the timeout error would be best.

internal/access/credential.go
```diff
@@ -225,3 +234,43 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+		return err
+	}
+	defer response.Body.Close()
+
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		return err
+	}
+
+	for _, line := range strings.Split(string(body), "\r\n") {
```
dnephin, line 264: I'd suggest https://pkg.go.dev/bufio#Scanner here instead of `ReadAll` and `Split`. 

That should use a lot less memory (because strings are read in small batches), and also avoids reading the rest of the response if something matches.

There are a few examples of `bufio.Scanner` in our repo already, I use it a bunch. It can accept the `response.Body` directly.

dnephin (commented):
This is great!  Generally I've seen other services use this API: https://haveibeenpwned.com/API/v3#BreachesForAccount instead of the "all breaches" API.  The docs for that endpoint also mention it's the most commonly used one.

What's the advantage to the "all breaches" endpoint? Is the "breaches for account" something we should look into later? 

This "all breaches" endpoint seems like it returns a lot of data, and could prevent someone from changing their password, even when the breached password was never associated with that email address.

internal/access/credential.go
```diff
@@ -225,3 +234,43 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+// Have I Been Pwned's API to perform this check. The API takes the first 5-characters prefix
+// of the SHA1 checksum of the password and returns a carriage returned list. A password is
+// breached if the remaining characters of its SHA1 checksum appears in this list.
+// Ref: https://haveibeenpwned.com/API/v3#PwnedPasswords
+func checkPasswordBreached(password string) error {
+	//nolint:gosec // this is required for the API
+	sum := sha1.Sum([]byte(password))
+	hexsum := hex.EncodeToString(sum[:])
+
+	request, err := http.NewRequestWithContext(context.Background(), http.MethodGet, fmt.Sprintf("https://api.pwnedpasswords.com/range/%s", hexsum[:5]), nil)
```
pdevine, line 248: I'm not really comfortable with the approach of calling a different service on account creation. I think we can just download the list from them and check it ourselves. I don't think this even needs to be updated all that frequently.

Also, like the password complexity stuff, this should be configurable per organization.

mxyng: I can see the concern with relying on an external service but I don't really see any reason for this to be an organization configuration.

internal/access/credential.go
```diff
@@ -225,3 +235,59 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+		return err
+	}
+
+	stat, err := file.Stat()
+	if err != nil {
+		return err
+	}
+
+	if stat.Size() == 0 {
+		request, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "https://www.ncsc.gov.uk/static-assets/documents/PwnedPasswordsTop100k.txt", nil)
```
mxyng, line 256: This downloads the bad password list and stores it in the filesystem if it doesn't already exist. Alternatively, it can be stored in Redis which would make retrieval trivial but adds another dependency on Redis

internal/access/credential.go
```diff
@@ -105,8 +112,11 @@ func updateCredential(c *gin.Context, user *models.Identity, newPassword string,
...
 	db := rCtx.DBTxn
 
-	err := checkPasswordRequirements(db, newPassword)
-	if err != nil {
+	if err := checkPasswordRequirements(db, newPassword); err != nil {
+		return err
+	}
+
+	if err := checkBadPasswords(newPassword); err != nil {
 		return err
```
BruceMacD, line 120: I feel like we should be proceeding in the case of certain errors here. For example if`ncsc.gov.uk/static-assets/documents/PwnedPasswordsTop100k.txt` ever becomes unavailable or invalid, but it has not been cached locally no one will be able to sign-up with a password.

internal/access/credential.go
BruceMacD: Any reason we can't embed this file rather than downloading it?

mxyng:
Thanks for all the feedback. This PR has changed significantly since the original commit. It now uses an environment variable to retrieve a file on disk. The content is expected to be a newline delimited list of bad passwords. The check is a noop if the file is empty.

internal/access/credential.go
```diff
@@ -225,3 +230,30 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
 
 	return nil
 }
+
+// checkBadPasswords checks if the password is a known bad password, i.e. a widely reused password.
+func checkBadPasswords(password string) error {
+	badPasswordsFile := os.Getenv("INFRA_SERVER_BAD_PASSWORDS_FILE")
```
BruceMacD, line 236: This looks like it should be documented somewhere

mxyng: What do you think is the best way to document this? It's not a standard configuration value so it can't be configured through command line parameters or configuration file. This is mostly because I don't want to propagate the config all the way from `cmd` or store it in the database.

jmorganca: A suggestion: move this out of the `access` layer and into the handler where it is used. There you'd have access to the server options which would provide an easy way to read this as a flag e.g. `--bad-passwords-file`

This new file based approach is elegant. Good stuff.

mxyng: Thing is, I don't really want to expose it to server options. Even if it is configurable from server options, the check should done here so it'll need to pass at least the value around

jmorganca: Ah that's right, this is called from within in the `access` package. RE my comment: this is okay for now until we revisit the access layer separately

internal/access/credential.go
```diff
@@ -225,3 +230,30 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+	if badPasswordsFile == "" {
+		return nil
+	}
+
+	file, err := os.OpenFile(badPasswordsFile, os.O_RDONLY|os.O_CREATE, 0o600)
```
jmorganca, line 241:
```suggestion
	file, err := os.OpenFile(badPasswordsFile, os.O_RDONLY, 0o600)
```

internal/access/credential.go
```diff
@@ -225,3 +230,30 @@ func checkPasswordRequirements(db data.ReadTxn, password string) error {
...
+	if err != nil {
+		return err
+	}
+
+	scan := bufio.NewScanner(file)
+	scan.Split(bufio.ScanLines)
+	for scan.Scan() {
+		if scan.Text() == password {
+			return fmt.Errorf("%w: cannot use a password commonly found in data breaches", internal.ErrBadRequest)
```
jmorganca, line 250:
```suggestion
			return fmt.Errorf("%w: cannot use a password using common words", internal.ErrBadRequest)
```

Nit suggestion on text to be a tad friendlier

jmorganca (commented):
Overall looks good – you may want to address the `O_CREATE` comment since it may error if mounted in a read-only volume

BruceMacD (approved):

PR #3964, opened 2023-01-05
Add a default expiration time to all user public keys

### LINKED ISSUE #3812: Require an expiration time for SSH public keys

dnephin:
## Summary

Adds a 12 hour expiration time to all SSH keys that are uploaded to the API. We could make this longer, but 12 hours seems like long enough to avoid having to wait for multiple keys to be created in a single work day.

Expired keys are never returned by the API. A new background job soft-deletes the keys once they expire.

I also noticed that I had missed an index of `user_public_keys.user_id` which is used by `listUserPublicKeys`, so I added a migration for that as well.

## Related Issues

Resolves #3812

### DIFF
--- a/internal/cmd/ssh.go
+++ b/internal/cmd/ssh.go
@@ -263,7 +263,6 @@ func provisionSSHKey(ctx context.Context, opts provisionSSHKeyOptions) (string, 
 		}
 
 		if userPublicKeyContains(opts.user.PublicKeys, existing.PublicKeyID) {
-			// TODO: check expiration when expiry is added
 			// key exists locally and in the API
 			return filename, nil
 		}
--- a/api/user.go
+++ b/api/user.go
@@ -105,4 +105,5 @@ type UserPublicKey struct {
 	KeyType string `json:"keyType"`
 	// Fingerprint is the SHA256 fingerprint of the public key.
 	Fingerprint string `json:"fingerprint" note:"SHA256 fingerprint of the key"`
+	Expires     Time   `json:"expires"`
 }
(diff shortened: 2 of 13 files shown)


internal/cmd/ssh.go
```diff
@@ -263,7 +263,6 @@ func provisionSSHKey(ctx context.Context, opts provisionSSHKeyOptions) (string,
 		}
 
 		if userPublicKeyContains(opts.user.PublicKeys, existing.PublicKeyID) {
-			// TODO: check expiration when expiry is added
```
dnephin, line 266: This was no longer necessary, because the API simply doesn't return expired keys. That way we don't have to handle them in every client.

BruceMacD (approved):
This seems good, just confirming my understanding that the open-ssh server is calling this public key endpoint on an interval and overwriting the existing keys, rather than adding to them (so expired keys disappear).

dnephin:
> the open-ssh server is calling this public key endpoint on an interval and overwriting the existing keys, rather than adding to them (so expired keys disappear).

The destination server never stores public keys. It queries the infra API for the allowed keys when a user tries to connect, and we omit expired keys from the API response. So once a key expires the servers stop trusting it immediately.

PR #3965, opened 2023-01-05
Remove validate.String function

dnephin:
## Summary

Functions with more than 3 arguments are hard to use, especially when many of the arguments are the same type.

Generally the solution to more args is to create a struct as the single argument. In this case we already have a struct, so we can simply remove the function.

Also resolve a TODO by removing the use of validate and using regular switch/case statement.

Also move some constants to the once place they are used.

### DIFF
--- a/api/deviceflow.go
+++ b/api/deviceflow.go
@@ -14,7 +14,21 @@ type ApproveDeviceFlowRequest struct {
 
 func (adfr *ApproveDeviceFlowRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
-		validate.String("userCode", adfr.UserCode, 8, 9, append(validate.DeviceFlowUserCode, validate.CharRange{Low: '-', High: '-'})),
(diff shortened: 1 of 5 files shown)


api/deviceflow.go
```diff
@@ -14,7 +14,21 @@ type ApproveDeviceFlowRequest struct {
...
+			MinLength: 8,
+			MaxLength: 9,
+			CharacterRanges: []validate.CharRange{
+				{Low: 'B', High: 'D'},
+				{Low: 'F', High: 'H'},
+				{Low: 'J', High: 'N'},
+				{Low: 'P', High: 'T'},
+				{Low: 'V', High: 'X'},
+				{Low: 'Z', High: 'Z'},
+				{Low: '-', High: '-'},
```
BruceMacD, line 29:
```suggestion
				validate.Dash,
```

BruceMacD (approved):

jmorganca (approved):

PR #3969, opened 2023-01-06
Fix latency caused by high volumes of requests from the same user

### LINKED ISSUE #3944: `POST /api/grants`  and `POST /api/tokens` results in high latency

dnephin:
## Summary

Fixes #3944

This PR fixes the request latency by avoiding the database row lock that was causing the contention. The `UPDATE` queries for `LastSeenAt` are changed to use `WHERE id IN (SELECT ... FOR UPDATE SKIP LOCKED)`.  If two requests are received at roughly the same time, the first request to run `UPDATE` will perform the update. The second request will match no rows, so won't acquire a lock. These new queries also use `WHERE updated_at = ?` to prevent clobbering previous updates.

We use this same pattern in 3 places: access keys, users, and destinations. This PR updates all 3 to use the same approach.

There is one more small bug fix in here as well.  I'll add more comments inline to explain the changes. 

The first commit in this PR attempted to fix the problem by removing the transaction. That did fix the problem, but doubled the request latency in the common case. So I restored the transaction and worked to fix the queries.

### Testing

I wasn't able to reproduce the exact scenario without introducing a sleep. I believe that's because in tests our database runs on the same host, so there is no network latency between the API server and the DB.

[dnephin/test-case](https://github.com/infrahq/infra/compare/dnephin/test-case) adds a test case and a 200ms sleep in the middle of the middleware transaction. When this test is run on `main` the elapsed times range from 20 seconds down to 200ms. When it's run on this branch the elapsed times are in the range of 500ms down to 200ms. I believe this shows the changes here fix the problem because the worse case scenario decreases from 20s to 500ms.

TODO:
* [x] more test coverage of the new `FromOrganization` query option

### DIFF
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -329,10 +329,10 @@ func updateAccessKeyLastSeenAt(tx WriteTxn, key *models.AccessKey) error {
 	query := querybuilder.New("UPDATE access_keys SET")
 	query.B(columnsForUpdate(table), table.Values()...)
 	query.B("WHERE deleted_at is null")
-	query.B("AND id = ?", table.Primary())
 	query.B("AND organization_id = ?", key.OrganizationID)
 	// only update if the row has not changed since the SELECT
 	query.B("AND updated_at = ?", origUpdatedAt)
+	query.B("AND id IN (SELECT id from access_keys WHERE id = ? FOR UPDATE SKIP LOCKED)", table.Primary())
 
 	_, err := tx.Exec(query.String(), query.Args...)
 	return handleError(err)
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -270,7 +270,7 @@ func DeleteAccessKeys(tx WriteTxn, opts DeleteAccessKeysOptions) error {
 }
 
 // TODO: move this to access package?
-func ValidateRequestAccessKey(tx *Transaction, authnKey string) (*models.AccessKey, error) {
+func ValidateRequestAccessKey(tx WriteTxn, authnKey string) (*models.AccessKey, error) {
 	keyID, secret, ok := strings.Cut(authnKey, ".")
 	if !ok {
 		return nil, fmt.Errorf("invalid access key format")
@@ -280,10 +280,8 @@ func ValidateRequestAccessKey(tx *Transaction, authnKey string) (*models.AccessK
 	if err != nil {
 		return nil, fmt.Errorf("%w: could not get access key from database, it may not exist", err)
 	}
-	tx = tx.WithOrgID(t.OrganizationID)
 
 	sum := secretChecksum(secret)
-
 	if subtle.ConstantTimeCompare(t.SecretChecksum, sum) != 1 {
 		return nil, fmt.Errorf("access key invalid secret")
 	}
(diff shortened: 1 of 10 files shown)


internal/server/data/access_key.go
```diff
@@ -297,19 +295,47 @@ func ValidateRequestAccessKey(tx *Transaction, authnKey string) (*models.AccessK
...
-		// same value as server.lastSeenUpdateThreshold.
-		if t.InactivityTimeout.Sub(origTimeout) > 2*time.Second {
-			if err := UpdateAccessKey(tx, t); err != nil {
-				return nil, err
-			}
-		}
 	}
 
-	return t, nil
+	err = updateAccessKeyLastSeenAt(tx, t)
```
dnephin, line 301: This also fixes the problem where only access keys with an `InactivityTimeout` have the correct `LastSeenAt`. With this change we update the access key even if it doesn't have an `InactivityTimeout`, so the artificial `LastSeenAt` (based on `UpdatedAt`) should be correct now.

internal/server/data/destination.go
```diff
@@ -51,6 +51,37 @@ func UpdateDestination(tx WriteTxn, destination *models.Destination) error {
 	return update(tx, (*destinationsTable)(destination))
 }
 
+// UpdateDestinationLastSeenAt sets dest.LastSeenAt to now and then updates the
+// user row in the database. Updates are throttled to once every 2 seconds.
```
BruceMacD, line 55:
```suggestion
// row in the database. Updates are throttled to once every 2 seconds.
```

BruceMacD (approved):

internal/server/data/provider.go
```diff
@@ -54,6 +54,10 @@ type GetProviderOptions struct {
 	// KindInfra instructs GetProvider to return the infra provider. There should
 	// only ever be a single provider with this kind for each org.
 	KindInfra bool
+
+	// FromOrganization is the organization ID of the provider. When set to a
+	// non-zero value the organization ID from the transaction is ignored.
+	FromOrganization uid.ID
```
jmorganca, line 60:
```suggestion
	ByOrganization uid.ID
```

internal/server/data/access_key.go
```diff
@@ -297,19 +295,47 @@ func ValidateRequestAccessKey(tx *Transaction, authnKey string) (*models.AccessK
...
+	return t, err
+}
+
+// lastSeenUpdateThreshold is the duration of time that must pass before a
+// LastSeenAt value for a user, access key, or destination is updated again.
+// This prevents excessive writes when a single user performs many requests
+// in a short period of time.
+const lastSeenUpdateThreshold = 2 * time.Second
```
jmorganca, line 309: Do we need to throttle this if we're skipping locked rows below anyways?

internal/server/data/access_key.go
jmorganca: Great fix!!

internal/server/data/destination.go
```diff
@@ -60,14 +91,24 @@ type GetDestinationOptions struct {
 	ByUniqueID string
 	// ByName instructs GetDestination to return the row matching this name.
 	ByName string
+
+	// FromOrganization is the organization ID of the provider. When set to a
```
jmorganca, line 95: > FromOrganization is the organization ID of the provider

More of a nit/suggestion: `ByOrganization` for consistency?

jmorganca (commented):
Great fix! Some minor comments

internal/server/data/destination.go
dnephin: We have some other variations for these field names (ex: `CreatedBy`, `NotIDs`).

My thinking is that we want something different for `Organization` to make it clear this field works differently from others. `FromOrganization` is an override for the `tx.OrganizationID()`, not simply a new filter being applied to the query.

internal/server/data/access_key.go
dnephin: Good question. I think each of these has a different purpose. The throttling helps reduce the total number of DB queries, where as the skip locked is to prevent queries from blocking each other. If the API requests are spaced out by 200ms or so then the skip locked doesn't do anything, but throttling will still help us skip a good number of database updates.

The throttling is an optimization, so we don't absolutely need it, but reducing the number of updates is still really nice to keep load off the database. Since it's only 3 lines I'd still like to keep it.

jmorganca (approved):

PR #3970, opened 2023-01-06
Remove unnecessary indirection in background jobs

dnephin:
`SetupBackgroundJobs` and the `jobs` package were unnecessary hops when tracing the code. These things did not do enough to warrant extracting them from the caller.

This PR removes the 1 line functions from the `jobs` package (which was the entire contents of that package), and inlines the 3 lines from `SetupBackgroundJobs` into the caller.

Also remove a stale block comment from `models.Grant`. The fields of this type are already documented, and the rest of the comment is stale and doesn't really apply anymore.


internal/server/backgroundjobs.go
```diff
@@ -7,21 +7,14 @@ import (
...
 	"github.com/infrahq/infra/internal/server/data"
-	"github.com/infrahq/infra/internal/server/jobs"
 )
 
 // BackgroundJobFunc is the interface for running periodic background jobs, like
 // garbage collection. Errors and panics from the job will be logged by
 // jobWrapper, and will not stop the goroutine running the job.
 // The job should exit when ctx is cancelled. Unlike most transactions, the
 // transaction passed to this job will not have an OrganizationID.
-type BackgroundJobFunc func(ctx context.Context, tx *data.Transaction) error
```
dnephin, line 18: I also removed the `context.Context` argument for a few reasons:
1. we don't currently use it
2. it could be misleading because the `tx` has a context itself already
3. by removing it we can use the `data` functions directly, instead of wrapping them

internal/server/backgroundjobs_test.go
```diff
@@ -29,15 +29,16 @@ func TestJobWrapper(t *testing.T) {
...
 		chReady <- signal
 		<-chRun
 
 		defer func() {
 			runID++
 		}()
 
-		tx = tx.WithOrgID(db.DefaultOrg.ID)
+		// nolint:forcetypeassert
+		tx = tx.(*data.Transaction).WithOrgID(db.DefaultOrg.ID)
```
dnephin, line 41: This was a consequence of changing the signature of the `jobs` function. Our production functions use `data.WriteTxn`, but our tests have to use `data.Transaction` so they can set the organization.  

It might be nice if `CreateIdentity` used the orgID on the passed in `Identity` instead of from the transaction. That would fix the problem without the type assert.

dnephin (commented):
Reminder to myself that this will conflict with #3964, so I'll need to rebase whichever one merges second.

internal/server/models/grant.go
```diff
@@ -16,33 +16,12 @@ const (
...
-// Expiry
-//
-//	time you want the grant to expire at
+// Grant is an access grant.
 type Grant struct {
 	Model
 	OrganizationMember
 
-	// Subject is the user or group ID the grant applies to.
+	// Subject is the user or group ID that is being access to a resource.
```
BruceMacD, line 24:
```suggestion
	// Subject is the user or group ID that is being granted access to a resource.
```

jmorganca (commented):
LGTM once comments from @BruceMacD are addressed

internal/server/backgroundjobs_test.go
jmorganca: 👍  that would make things clearer

dnephin: I forgot this already does work, so I removed the forcetypeassert and changed to setting the orgID on the identity.

mxyng (approved):

PR #3981, opened 2023-01-09
maintain: use standard go slices lib

BruceMacD:
## Summary
Update to use the [standard Go slices](https://pkg.go.dev/golang.org/x/exp/slices) package that was added as of 1.18.

## Checklist


internal/server/data/identity.go
```diff
@@ -34,6 +34,23 @@ func (i *identitiesTable) ScanFields() []any {
...
+// diff compares two string slices (x and y), it returns a slice of strings that exist in x that do not exist in y
+func diff(x, y []string) []string {
+	// this modifies the order of x and y, but its ok in this case since the order of the groups assigned to a user does not matter
+	slices.Sort(x)
+	slices.Sort(y)
+
+	difference := []string{}
+
+	for _, val := range x {
+		if !slices.Contains(y, val) {
```
dnephin: I believe that since we're using `Contains` on `y` that `y` doesn't need to be sorted.

I think an implementation that sorts both and traverses both with indexes might be slightly more efficient than using `Contains`, but I doubt it's worth the optimization at this point.

dnephin (approved):
LGTM! Some minor comments about optimizations, but I don't think we need them right now

PR #3983, opened 2023-01-09
fix: no gmail allowed domains

BruceMacD:
## Summary
To protect an org from being incorrectly configured block "gmail.com" and "googlemail.com" allowed domains.

Screen Shot 2023-01-09 at 10 23 13 AM

## Checklist

### DIFF
--- a/internal/validate/string.go
+++ b/internal/validate/string.go
@@ -30,6 +30,9 @@ type StringRule struct {
 
 	// RequiredCharacters must be present in the string
 	RequiredCharacters []rune
+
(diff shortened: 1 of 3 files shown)


internal/validate/string.go
```diff
@@ -130,6 +133,15 @@ func (s StringRule) Validate() *Failure {
 		}
 	}
 
+	if len(s.DenyList) > 0 {
```
jmorganca, line 136: Go can iterate on `nil` slices, so you can remove this `if` statement

PR #3989, opened 2023-01-09
fix: settings providers link

mxyng:
## Summary

Fix providers settings links missed in https://github.com/infrahq/infra/pull/3988


BruceMacD:
Fixed in 
https://github.com/infrahq/infra/pull/3988

mxyng:
@BruceMacD that PR missed some instances of `tab=providers`

jmorganca (approved):
Thanks for fixing!

PR #3991, opened 2023-01-09
maintain: add metrics for provider calls

BruceMacD:
## Summary
Add provider metrics to middleware to track timing of external calls.

## Checklist

### DIFF
--- a/internal/server/metrics.go
+++ b/internal/server/metrics.go
@@ -9,6 +9,13 @@ import (
 	"github.com/infrahq/infra/metrics"
 )
 
+var outboundRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace: "http",
+	Name:      "outbound_request_duration_seconds",
+	Help:      "A histogram of outbound call durations made from the server to an external source, in seconds.",
+	Buckets:   prometheus.ExponentialBuckets(0.001, 2, 15),
+}, []string{"request_kind", "action"})
+
 func setupMetrics(db *data.DB) *prometheus.Registry {
 	registry := metrics.NewRegistry(productVersion())
 	registry.MustRegister(collectors.NewDBStatsCollector(db.SQLdb(), "postgres"))
@@ -122,6 +129,8 @@ func setupMetrics(db *data.DB) *prometheus.Registry {
 			{Count: float64(count)},
 		}
 	}))
+
+	registry.MustRegister(outboundRequestDuration)
 
 	return registry
 }
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -495,7 +495,12 @@ func (s *Server) syncIdentityInfo(ctx context.Context, tx *data.Transaction, ide
 		}
 
 		// update current identity provider groups and account status
+		t := time.Now()
 		_, err = data.SyncProviderUser(ctx, tx, providerUser, oidc)
+		outboundRequestDuration.With(prometheus.Labels{
+			"request_kind": provider.Kind.String(),
+			"action":       "sync",
+		}).Observe(time.Since(t).Seconds())
 		if err != nil {
 			if errors.Is(err, internal.ErrBadGateway) {
 				return err
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"context"
 	"net/http"
 	"strconv"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/infrahq/infra/internal"
+	"github.com/infrahq/infra/uid"
 )
 
 // NewRegistry creates a new prometheus.Registry and registers common collectors and metrics.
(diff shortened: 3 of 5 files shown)


metrics/metrics.go
```diff
@@ -63,6 +99,9 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
 		count.Inc()
 		defer count.Dec()
 
+		metricContext := context.WithValue(c.Request.Context(), providerMetricsKey, outboundCallDuration)
```
BruceMacD, line 102: Adding this to the context so that we can observe the metric where it is called. The other option I could come up with was to add different call durations to the context and check them at the end.

metrics/metrics.go
```diff
@@ -54,6 +83,13 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
 		Help:      "A gauge of the number of requests currently being handled.",
 	}, []string{"blocking"})
 
+	outboundCallDuration := prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
mxyng, line 86: This metric needs to be registered

metrics/metrics.go
```diff
@@ -54,6 +83,13 @@ func Middleware(registry prometheus.Registerer) gin.HandlerFunc {
...
+		Namespace: "http",
+		Name:      "outbound_request_call_seconds",
```
mxyng, line 88: This should be namespace `http` and name `client_request_duration_seconds`. We're using this in the connector to measure outbound/upstream HTTP request. It'll be nice to follow the same format.

metrics/metrics.go
```diff
@@ -38,6 +40,33 @@ func NewRegistry(version string) *prometheus.Registry {
...
+
+type ProviderActionOpts struct {
+	ProviderID      uid.ID
+	OrgID           uid.ID
+	Action          string
+	DurationSeconds float64
+}
+
+func ProviderAction(c context.Context, opts ProviderActionOpts) {
+	m := ProviderCallMetricsFromContext(c)
```
mxyng, line 60: This is overkill. Instead, this should be moved to `provideruser.go`. You're not really reducing duplication with this function

metrics/metrics.go
mxyng: I'm not sure about setting up this metric like this. The indirection makes it harder to trace. The metrics middleware is also not the right place to create/register this metric since it's shared between the server and connector. Instead it should probably be registered in `server/metrics.go`

mxyng (changes requested):

metrics/metrics.go
```diff
@@ -38,6 +40,33 @@ func NewRegistry(version string) *prometheus.Registry {
...
+	if m != nil {
+		m.With(prometheus.Labels{
+			"provider_id":     opts.ProviderID.String(),
+			"organization_id": opts.OrgID.String(),
```
dnephin, line 64: I believe these identifiers are going to be too high cardinality for metrics. They're fine for logs, but for metrics we probably need to settle for the provider kind. In many cases the provider kind should be sufficient I guess, because all the providers of the same kind are probably hitting the same servers.

metrics/metrics.go
dnephin: Do we want to use the same metric name for outbound requests to the infra API and outbound requests to provider APIs? Do we ever expect to need to aggregate those? I think we probably want a different metric name for those so that it's a bit easier to consume.

metrics/metrics.go
BruceMacD: Added it directly in-line with the `syncIdentityInfo` function

metrics/metrics.go
BruceMacD: Went with `outbound_request_duration_seconds`

internal/server/server.go
```diff
@@ -495,7 +495,12 @@ func (s *Server) syncIdentityInfo(ctx context.Context, tx *data.Transaction, ide
 		}
 
 		// update current identity provider groups and account status
+		t := time.Now()
 		_, err = data.SyncProviderUser(ctx, tx, providerUser, oidc)
+		outboundRequestDuration.With(prometheus.Labels{
```
dnephin, line 500: This ends up tracking a bunch of operations , including DB queries, right?

We probably want this metric to be include the time for only the outbound HTTP request, not this high level operation.

dnephin (commented):
This approach generally makes sense to me

metrics/metrics.go
mxyng: After a closer look, this probably should be using `namespace=oidc` since it's not using an HTTP client directly

internal/server/server.go
mxyng: This is measurement is too broad as @dnephin mentioned. Instead, it should measure the requests sent by the OIDC client, specifically `oidcClient.RefreshAccessToken()` and `oidcClient.GetUserInfo()` in `data.SyncProviderUser()` with each measurement having a distinct label value

internal/server/metrics.go
```diff
@@ -9,6 +9,13 @@ import (
 	"github.com/infrahq/infra/metrics"
 )
 
+var outboundRequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+	Namespace: "http",
+	Name:      "outbound_request_duration_seconds",
+	Help:      "A histogram of outbound call durations made from the server to an external source, in seconds.",
+	Buckets:   prometheus.ExponentialBuckets(0.001, 2, 15),
```
mxyng, line 16: It's probably not necessary to start histogram buckets at 0.001s since outgoing requests will never take that little time. This also only produces 15 buckets so the largest bucket is only ~16s. If a request takes longer, it'll get put in to the `+Inf` bucket which doesn't tell us much.

Maybe start it at 0.1s which will also gives a max bucket size of ~1638s while still providing small enough buckets on the low end.

BruceMacD:
Revisiting this in a later PR

PR #3994, opened 2023-01-10
Use `go` for recommended local development instructions

jmorganca:
This PR replaces the recommended helm and docker-compose development workflows with one that runs everything locally with `go`. This allows for debugging and faster development loops.


dnephin (approved):
LGTM

I didn't try any of the configs, but they look right.

Does the `brew install` of postgres on MacOS create the user for the system username? I've been using the `postgres` user.

BruceMacD (approved):

jmorganca:
@BruceMacD  @dnephin thanks for looking!

@dnephin Yeah the default is to create a `postgres` database with the system username as the user

PR #3995, opened 2023-01-10
maintain: additional protection around access key removal

BruceMacD:
## Summary
Add some checks to prevent deleting more access keys than were intended, also add some logging to track down a bug where access keys are removed unexpectedly. 

## Checklist

### DIFF
--- a/internal/server/data/access_key.go
+++ b/internal/server/data/access_key.go
@@ -251,6 +251,9 @@ func DeleteAccessKeys(tx WriteTxn, opts DeleteAccessKeysOptions) error {
 	if opts.ByID == 0 && opts.ByIssuedForID == 0 && opts.ByProviderID == 0 {
 		return fmt.Errorf("DeleteAccessKeys requires an ID, IssuedForID, or ProviderID")
 	}
+	if opts.ByIssuedForID != 0 && opts.ByProviderID == 0 {
+		return fmt.Errorf("DeleteAccessKeys by IssuedForID requires ProviderID")
+	}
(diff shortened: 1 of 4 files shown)


internal/server/server.go
```diff
@@ -519,6 +519,8 @@ func (s *Server) syncIdentityInfo(ctx context.Context, tx *data.Transaction, ide
 				return err
 			}
 
+			logging.L.Info().Msg("user session expired, puring keys created for this session")
```
dnephin, line 522:
```suggestion
			logging.L.Info().Msg("user session expired, pruning keys created for this session")
```

internal/server/data/access_key.go
```diff
@@ -319,5 +323,6 @@ func RemoveExpiredAccessKeys(tx WriteTxn) error {
 	query.B("AND expires_at <= ?", time.Now().UTC().Add(-1*time.Hour)) // leave buffer so keys aren't immediately deleted on expiry.
 
 	_, err := tx.Exec(query.String(), query.Args...)
+	logging.L.Info().Msg("removed expired access key")
```
dnephin, line 326: Could do something like this to get a count. I guess this would be pretty useful in general 

```suggestion
	result, err := tx.Exec(query.String(), query.Args...)
	if err != nil {
		return err
	}
	count, err := result.RowsAffected()
	if err != nil {
		return err
	}
	if count > 0 {
		logging.L.Info().Int64("count", count).Msg("removed expired access keys")
	}
```

dnephin (approved):
LGTM, one suggestion for more logging details, but not a blocker

PR #3998, opened 2023-01-10
maintain: remove update settings endpoint

### LINKED ISSUE #3990: Replace password policy API with a standard default policy
Currently Infra provides an undocumented API to change password policies per organization. Instead, this should be replaced with a secure default that supports the NIST password requirements.

mxyng:
## Summary

This API is mainly unused but redirect it to `GetSettings` in case something is calling it. The end result is a noop since it only retrieves settings rather than update it. This does not remove organization settings or password requirements.

Resolves #3990


internal/server/routes.go
```diff
@@ -141,6 +139,18 @@ func (s *Server) GenerateRoutes() Routes {
 	a.addPreviousVersionHandlersSignup()
 	a.addPreviousVersionHandlersGrants()
 
+	// Deprecated
+	add(a, authn, http.MethodPut, "/api/settings", route[api.EmptyRequest, *api.Settings]{
```
jmorganca, line 143: Overall looks good! Is this required by older clients? If not we can do without a deprecated route

BruceMacD (approved):

PR #3999, opened 2023-01-10
delete grants for a destination when deleting the destination

### LINKED ISSUE #3800: Provide a way to filter out `infra` grants in the `/api/grants` API
Infra's clients (Dashboard and CLI) need to filter out grants that assign access to `infra` (i.e. Infra API itself), but there isn't an easy way to do this in the API.

The `GET /api/grants` endpoint should hide these grants by default, and allow for showing them via `?showSystem=true` (similar to what is used elsewhere) or `?showInfra=true`.

pdevine:
## Summary

This change removes any grants for a destination when the destination is deleted.

## Checklist

## Related Issues

Resolves #3800

### DIFF
--- a/internal/server/data/destination.go
+++ b/internal/server/data/destination.go
@@ -176,11 +176,25 @@ func ListDestinations(tx ReadTxn, opts ListDestinationsOptions) ([]models.Destin
 }
 
 func DeleteDestination(tx WriteTxn, id uid.ID) error {
+	// Delete any grants for the destination before deleting it
 	stmt := `
+		UPDATE grants SET deleted_at = ?
+		WHERE resource =
+			(SELECT name FROM destinations
+			WHERE id = ? AND organization_id = ? AND deleted_at is null)
+		AND organization_id = ? AND deleted_at is null
+	`
+
+	_, err := tx.Exec(stmt, time.Now(), id, tx.OrganizationID(), tx.OrganizationID())
+	if err != nil {
+		return handleError(err)
+	}
+
+	stmt = `
 		UPDATE destinations SET deleted_at = ?
 		WHERE id = ? AND organization_id = ? AND deleted_at is null
 	`
-	_, err := tx.Exec(stmt, time.Now(), id, tx.OrganizationID())
+	_, err = tx.Exec(stmt, time.Now(), id, tx.OrganizationID())
 	return handleError(err)
 }
 
--- a/internal/server/data/destination.go
+++ b/internal/server/data/destination.go
@@ -176,21 +176,24 @@ func ListDestinations(tx ReadTxn, opts ListDestinationsOptions) ([]models.Destin
 }
 
 func DeleteDestination(tx WriteTxn, id uid.ID) error {
-	// Delete any grants for the destination before deleting it
-	stmt := `
-		UPDATE grants SET deleted_at = ?
-		WHERE resource =
-			(SELECT name FROM destinations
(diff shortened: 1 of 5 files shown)


internal/server/data/destination.go
```diff
@@ -176,11 +176,25 @@ func ListDestinations(tx ReadTxn, opts ListDestinationsOptions) ([]models.Destin
 }
 
 func DeleteDestination(tx WriteTxn, id uid.ID) error {
+	// Delete any grants for the destination before deleting it
 	stmt := `
+		UPDATE grants SET deleted_at = ?
+		WHERE resource =
+			(SELECT name FROM destinations
+			WHERE id = ? AND organization_id = ? AND deleted_at is null)
```
dnephin, line 184: This will match `ssh` destinations fine, but only makes `kubernetes` grants for the cluster, right?  Kubernetes grants for specific namspaces have a resource of `<destinationName>.<namespace>`. 

`data.ListGrants` uses `data.grantsByDestination` to perform that query. You might be able to use that here as well.  I think if you move this new delete after the `UPDATE destinations`, you can check the result from `tx.Exec` to determine if the destination was deleted or not. That should avoid the need to `WHERE organization_id = ? AND deleted_at is null` since it's already done by the `UPDATE destinations`.

Both queries are performed in the same transaction, so I'm guessing order doesn't matter (other than switching them might make it easier to do the cleanup of grants, and matches how we do this in `DeleteProviders` and `DeleteIdentities`).

dnephin (commented):
Looks good! Just one comment about something that may be missing.

jmorganca:
Maybe out of scope: do we also want a migration to prune all "destination-less" grants?

internal/server/data/destination.go
pdevine: This is a great catch. This was a little bit tricky to sort out, due to using the destination name inside of a grant instead of the IDs. I've changed this to use the query builder, and to use the `grantsByDestination()` function, which I think is pretty straight forward.

I did not drop the `WHERE organization_id = ?` guard though, because a different org could have a destination named the same thing. I did explore using the IDs of each of the grants which would have enabled me to drop the guard, but that ended up being a lot more complicated.

pdevine:
> Maybe out of scope: do we also want a migration to prune all "destination-less" grants?

@jmorganca I think we can just leave those in. If someone wants to delete them, it'll be pretty straight forward in the UI.

internal/server/data/destination.go
```diff
@@ -176,11 +176,28 @@ func ListDestinations(tx ReadTxn, opts ListDestinationsOptions) ([]models.Destin
...
+	if err != nil {
+		return handleError(err)
+	}
+
+	query := querybuilder.New("UPDATE grants")
+	query.B("SET deleted_at = ?,", time.Now())
+	query.B("update_index = nextval('seq_update_index')")
+	query.B("WHERE organization_id = ? AND", tx.OrganizationID())
+	query.B("deleted_at is null")
+	grantsByDestination(query, dest.Name)
```
dnephin, line 189: I guess it might make sense to add a `ByDestination` field to `DeleteGrantsOptions`, and use `DeleteGrants` for this. The query looks the same, we're only missing the `ByDestination` option.

dnephin (approved):
LGTM! Nice test coverage

jmorganca (approved):

internal/server/data/destination.go
pdevine: This is a great suggestion. I went ahead and refactored and added another test.

PR #4000, opened 2023-01-10
maintain(cmd): simplify user password prompt

mxyng:
## Summary

- gotos make it harder to trace the code so replace them with loops
- remove most password validation since it will be done by the server

### DIFF
--- a/internal/cmd/users.go
+++ b/internal/cmd/users.go
@@ -60,7 +60,7 @@ $ infra users add johndoe@example.com`,
 				return err
 			}
 
-			createResp, err := createUser(ctx, client, args[0])
+			createResp, err := client.CreateUser(ctx, &api.CreateUserRequest{Name: args[0]})
 			if err != nil {
 				if api.ErrorStatusCode(err) == 403 {
 					logging.Debugf("%s", err.Error())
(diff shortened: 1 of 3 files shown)


internal/cmd/users.go
```diff
@@ -371,55 +374,26 @@ func getUserByNameOrID(client *api.Client, name string) (*api.User, error) {
...
 
-		if len(newPassword) == 0 {
-			return fmt.Errorf("Value is required")
+		if err := survey.AskOne(&survey.Password{Message: "Confirm New Password:"}, &confirm, cli.surveyIO, survey.WithValidator(survey.Required)); err != nil {
+			return "", err
 		}
 
-		if newPassword == oldPassword {
-			return fmt.Errorf("input must be different than the current password")
+		if password != confirm {
```
jmorganca, line 388: Nit to save a line: you can do something like

```
if password == confirm {
    break
}

cli.Output(" Passwords do not match. Please try again.")
```

jmorganca (commented):
Great simplification and removal of GOTOs – small nit otherwise this LGTM

internal/cmd/users.go
```diff
@@ -280,19 +280,22 @@ func updateUser(cli *CLI, name string) error {
...
+			req.Password, err = promptSetPassword(cli, req.OldPassword)
+			if err != nil {
+				return err
+			}
 
-		req.Password, err = promptSetPassword(cli, req.OldPassword)
-		if err != nil {
-			return err
-		}
+			logging.Debugf("call server: update user %s", req.ID)
```
BruceMacD, line 289:
```suggestion
```
not related to this PR, but might be worth cleaning this up while in the neighborhood

BruceMacD (approved):

PR #4001, opened 2023-01-11
fix: ui first login flashes

### LINKED ISSUE #3985: UI flashes on first login
After logging in, there's a short flash of content

hoyyeva:
## Summary
The problem seems like to be the redirect in callback component. After the login is done, it redirects to `/` which it is calling the Index component. In the Index component it checks the router and redirects the path to `/destinations` 

The fixes for this are:
1) to redirect to `/destinations` directly instead of `/`. 
2) added redirect in the `next.config.js` to make sure the `/` is always redirect to `/destinations` without going through Index component

## Related Issues
Resolves #3985

### DIFF
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -18,10 +18,6 @@ module.exports = phase => ({
             type: 'cookie',
             key: 'auth',
           },
-          // {
-          //   type: 'header',
-          //   key: 'Authorization',
-          // },
         ],
         destination: '/destinations',
         permanent: false,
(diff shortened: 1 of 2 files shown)


ui/next.config.js
```diff
@@ -9,6 +9,9 @@ module.exports = phase => ({
 
     return null
   },
+  async redirects() {
+    return [{ source: '/', destination: '/destinations', permanent: true }]
```
jmorganca, line 13: LGTM overall – will this work as well in scenarios where the user is logged out?

hoyyeva: Oh, it is a good catch! I added the conditional check in the next.config.js - however this might not work for the social login yet due to this bug - https://github.com/infrahq/infra/issues/4003

jmorganca (approved):

PR #4004, opened 2023-01-12
refactor: update user/credential

mxyng:
## Summary

`UpdateUser` is actually just `UpdateCredential` since no other user field is currently modifiable. There are two modes for `UpdateCredential`: 

1. Update the caller’s user credential. This does not require special authorization.
2. Update another user’s credential. This requires admin authorization.

To support these modes, there are 6 `isSelf` checks between `access.UpdateCredential` and its helper `access.updateCredential`. This makes it very difficult to trace the code. 

In total there are 3 use cases where update credential is needed:

1. A user wants to change their own password. The current password must be provided and verified against the database
1. A user wants to reset their password out-of-band through email verification. The current password is unknown so it cannot be verified
1. An admin wants to reset another users password. The current password is unknown. The only way to do this currently is through the CLI which generates a password and sends it in the request

There's a fourth use case where a user is invited by email. To create their account, the user claims invite and sets a password. This reuses the password reset flow so it's omitted from this list.

The third use case raises red flags. The CLI generates a random 12-character password which it sends in the `UpdateUser` request. If the password policy requires greater than 12 characters, this action will never succeed. Even if the policy requires ≤ 12 characters but has other requirements like numbers or symbols, the password generate may fail even with retries. Even if the CLI generates a valid value, allowing the admin to decide a user’s new password feels like a bad pattern. While this PR makes changes relevant to this issue, for compatibility reasons this will be addressed in a follow up.

This PR focuses on streamlining the `UpdateUser` flow. There's a single check to determine if the request is intended for the caller's user `UpdateCredential` or another user `ResetCredential`.

This PR also adds `UpdateUserResponse` which is `User` with an extra field `OneTimePassword`. If the `UpdateUser` request is for resetting a user's password and a password is not specified, a password will be generated and returned in this new field. If a password is specified in the request, it will return that value. This should be compatible with older clients.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -258,8 +258,8 @@ func (c Client) CreateUser(ctx context.Context, req *CreateUserRequest) (*Create
 	return post[CreateUserResponse](ctx, c, "/api/users", req)
 }
 
-func (c Client) UpdateUser(ctx context.Context, req *UpdateUserRequest) (*User, error) {
-	return put[User](ctx, c, fmt.Sprintf("/api/users/%s", req.ID.String()), req)
+func (c Client) UpdateUser(ctx context.Context, req *UpdateUserRequest) (*UpdateUserResponse, error) {
+	return put[UpdateUserResponse](ctx, c, fmt.Sprintf("/api/users/%s", req.ID.String()), req)
 }
 
 func (c Client) DeleteUser(ctx context.Context, id uid.ID) error {
--- a/api/user.go
+++ b/api/user.go
@@ -68,8 +68,12 @@ type UpdateUserRequest struct {
 func (r UpdateUserRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
 		validate.Required("id", r.ID),
-		validate.Required("password", r.Password),
 	}
+}
+
+type UpdateUserResponse struct {
+	User
+	OneTimePassword string `json:"oneTimePassword,omitempty"`
 }
 
 func (req ListUsersRequest) SetPage(page int) Paginatable {
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -12,156 +12,153 @@ import (
 
 	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/generate"
-	"github.com/infrahq/infra/internal/logging"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 	"github.com/infrahq/infra/internal/validate"
 )
 
-func CreateCredential(c *gin.Context, user models.Identity) (string, error) {
-	db, err := RequireInfraRole(c, models.InfraAdminRole)
+func CreateCredential(c *gin.Context, user *models.Identity) (string, error) {
+	tx, err := RequireInfraRole(c, models.InfraAdminRole)
 	if err != nil {
 		return "", HandleAuthErr(err, "user", "create", models.InfraAdminRole)
 	}
 
-	tmpPassword, err := generate.CryptoRandom(12, generate.CharsetPassword)
+	password, err := generate.CryptoRandom(12, generate.CharsetPassword)
 	if err != nil {
-		return "", fmt.Errorf("generate: %w", err)
+		return "", fmt.Errorf("crypto random: %w", err)
 	}
 
-	hash, err := bcrypt.GenerateFromPassword([]byte(tmpPassword), bcrypt.DefaultCost)
-	if err != nil {
-		return "", fmt.Errorf("hash: %w", err)
-	}
-
-	userCredential := &models.Credential{
-		IdentityID:      user.ID,
-		PasswordHash:    hash,
+	credential := &models.Credential{
 		OneTimePassword: true,
 	}
 
-	if err := data.CreateCredential(db, userCredential); err != nil {
+	if err := createCredential(tx, user, credential, password); err != nil {
 		return "", err
 	}
 
-	_, err = data.CreateProviderUser(db, data.InfraProvider(db), &user)
-	if err != nil {
-		return "", fmt.Errorf("creating provider user: %w", err)
-	}
-
-	return tmpPassword, nil
+	return password, err
 }
 
+func createCredential(tx *data.Transaction, user *models.Identity, credential *models.Credential, password string) error {
+	hash, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+	if err != nil {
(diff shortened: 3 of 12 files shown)


internal/access/credential.go
```diff
@@ -10,153 +10,134 @@ import (
...
-	if err := data.CreateCredential(db, userCredential); err != nil {
+	if err := createCredential(tx, user, credential, password); err != nil {
 		return "", err
 	}
 
-	_, err = data.CreateProviderUser(db, data.InfraProvider(db), &user)
+	return password, err
+}
+
+func createCredential(tx *data.Transaction, user *models.Identity, credential *models.Credential, password string) error {
```
BruceMacD, line 40: A lot of this function feels like it should be happening in the update user handler before the access check to set values in the database, if the structure of the code allows for that.

internal/server/users.go
```diff
@@ -137,21 +138,45 @@ func (a *API) CreateUser(c *gin.Context, r *api.CreateUserRequest) (*api.CreateU
...
+		}, nil
+	}
+
+	user, err := access.GetIdentity(c, data.GetIdentityOptions{ByID: r.ID, LoadProviders: true})
 	if err != nil {
 		return nil, err
 	}
 
-	err = access.UpdateCredential(c, identity, r.OldPassword, r.Password)
+	password, err := access.ResetCredential(c, user, r.Password)
```
BruceMacD, line 159: Does this remove the ability to reset a user's password as an admin?

dnephin (commented):
Overall I think the goal of this PR makes sense, and the 3 user flows are much easier to follow.  It's possible I've missed it, but it seems like some of the flows may not have all the same functionality they used to have (comments inline).

Seems like we may be missing a test case for `TestPasswordResetFlow` where the password they chose doesn't meet the requirements.

internal/server/users.go
dnephin, line 159: I think we might be missing "check password requirements" in this flow as well (admin sets a password for a user).

mxyng: It does not. An admin can still reset a user's password

api/user.go
```diff
@@ -68,10 +68,14 @@ type UpdateUserRequest struct {
 func (r UpdateUserRequest) ValidationRules() []validate.ValidationRule {
 	return []validate.ValidationRule{
 		validate.Required("id", r.ID),
-		validate.Required("password", r.Password),
 	}
 }
 
+type UpdateUserResponse struct {
+	User
+	OneTimePassword string `json:"oneTimePassword,omitempty"`
```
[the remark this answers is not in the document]
mxyng: I decided to omit the client changes to maintain forward compatibility. Otherwise the new CLI version will break when used against old server versions. The response is included so we can perform a seamless update in the future.

internal/server/users.go
mxyng: This is intentional. The password reset as implemented by the CLI generates a password which may not conform to the configured password policy. In that case, this request will always fail

internal/access/passwordreset.go
```diff
@@ -29,8 +29,27 @@ func VerifiedPasswordReset(c *gin.Context, token, newPassword string) (*models.I
 		}
 	}
 
-	if err := updateCredential(c, user, newPassword, true); err != nil {
+	credential, err := data.GetCredentialByUserID(tx, user.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := updateCredential(tx, credential, password); err != nil {
```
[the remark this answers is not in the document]
mxyng: That's right. I missed it initially but it's now being checked in `VerifiedPasswordReset`

internal/access/passwordreset.go
```diff
@@ -29,8 +29,27 @@ func VerifiedPasswordReset(c *gin.Context, token, newPassword string) (*models.I
...
 		return nil, err
 	}
+
+	if accessKey := rCtx.Authenticated.AccessKey; accessKey != nil {
+		for i, v := range accessKey.Scopes {
+			if v == models.ScopePasswordReset {
+				accessKey.Scopes = append(accessKey.Scopes[:i], accessKey.Scopes[i+1:]...)
```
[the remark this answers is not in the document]
mxyng: I misunderstood the intention. The name is confusing. It implies it's related to password reset but instead it's actually enforcing the user update their password when login in with an single-use password.

internal/access/credential.go
```diff
@@ -12,156 +12,152 @@ import (
...
+		return err
 	}
 
-	if isSelf {
-		// if we updated our own password, remove the password-reset scope from our access key.
-		if accessKey := rCtx.Authenticated.AccessKey; accessKey != nil {
-			accessKey.Scopes = sliceWithoutElement(accessKey.Scopes, models.ScopePasswordReset)
-			if err = data.UpdateAccessKey(db, accessKey); err != nil {
-				return fmt.Errorf("updating access key: %w", err)
+	if accessKey := rCtx.Authenticated.AccessKey; accessKey != nil {
```
BruceMacD, line 132:
```suggestion
        // if we updated our own password, remove the password-reset scope from our access key.
        if accessKey := rCtx.Authenticated.AccessKey; accessKey != nil {
```

This is confusing without the comment imo

BruceMacD (approved):
Seems good to go to me

api/client.go
```diff
@@ -258,8 +258,8 @@ func (c Client) CreateUser(ctx context.Context, req *CreateUserRequest) (*Create
 	return post[CreateUserResponse](ctx, c, "/api/users", req)
 }
 
-func (c Client) UpdateUser(ctx context.Context, req *UpdateUserRequest) (*User, error) {
-	return put[User](ctx, c, fmt.Sprintf("/api/users/%s", req.ID.String()), req)
+func (c Client) UpdateUser(ctx context.Context, req *UpdateUserRequest) (*UpdateUserResponse, error) {
```
jmorganca, line 261: Shouldn't this stay `User`? As a guideline for our API we return the Resource for `PUT` requests: https://github.com/infrahq/infra/blob/main/docs/api/guidelines.md#method-calls

mxyng: As a general rule, that's true but there have been exceptions. For example, `POST` requests are also supposed to return the resource but `POST /api/users` returns a different response since it needs to return the `OneTimePassword`. Same thing with `POST /api/access-keys` which needs to return the secret. 

In order to have the server generate the new password on reset, this change is needed. There are two alternatives if following these guidelines is required:

1. Add a separate endpoint strictly for an admin to reset passwords
2. Change `api.User` to always include `OneTimePassword` but leave it unset in non-CreateUser/-UpdateUser cases and have both `POST /api/users` and `PUT /api/users/:d` return `api.User`

PR #4006, opened 2023-01-12
fix: remove non-secure cookies on logout

### LINKED ISSUE #4003: Logging out from social login may not clear the session cookie
## Describe the bug
When a user signs in via social login (ie: Google) the cookie they get for their session may not be cleared from the browser on logout.

## Expected behavior
The cookie is cleared.

BruceMacD:
## Summary
When logging out on an `http` connection the cookie is not cleared, check if the incoming request is secure to clear this cookie.

## Checklist

## Related Issues

Resolves #4003

### DIFF
--- a/internal/server/cookie.go
+++ b/internal/server/cookie.go
@@ -47,13 +47,19 @@ func setCookie(req *http.Request, resp http.ResponseWriter, config cookieConfig)
 	})
 }
 
-func deleteCookie(resp http.ResponseWriter, name, domain string) {
+func deleteCookie(req *http.Request, resp http.ResponseWriter, name, domain string) {
+	secure := true
+	if req.TLS == nil {
+		// if the request came over HTTP, then the cookie will need to be sent unsecured
+		secure = false
+	}
+
 	http.SetCookie(resp, &http.Cookie{
 		Name:     name,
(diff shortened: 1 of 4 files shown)


internal/server/cookie.go
```diff
@@ -47,13 +47,19 @@ func setCookie(req *http.Request, resp http.ResponseWriter, config cookieConfig)
 	})
 }
 
-func deleteCookie(resp http.ResponseWriter, name, domain string) {
+func deleteCookie(req *http.Request, resp http.ResponseWriter, name, domain string) {
+	secure := true
+	if req.TLS == nil {
```
jmorganca, line 52: Is this for testing? Or another reason?

BruceMacD: Assuming that the cookies the client has are `secure` if the request is over `https`, and not otherwise because we need to set the cookie to clear it, and secure cookies cannot be set over http channels. 

`http` is only used during testing locally. Hosted Infra should always be over HTTPs.

jmorganca: Ah I see this is simply making this functionality the same as elsewhere, and that makes sense for consistency. LGTM

jmorganca (approved):

PR #4008, opened 2023-01-12
fix: groups page improvement

hoyyeva:
## Summary
- review and refactor how we handle groups user table after sending the request to server side
- add multi select on groups users table
- small improvement on access control and table components

### DIFF
--- a/ui/components/table.js
+++ b/ui/components/table.js
@@ -51,8 +51,27 @@ export default function Table({
   const [checkedAll, setCheckedAll] = useState(false)
   const [indeterminate, setIndeterminate] = useState(false)
 
-  const handleSelectedRowIds = ids => {
+  function handleSelectedRowIds(ids) {
     setSelectedRowIds(ids)
+  }
+
+  function toggleAll() {
+    if (checkedAll || indeterminate) {
+      handleSelectedRowIds([])
+    } else {
+      handleSelectedRowIds(
+        data
+          ?.filter(
+            d =>
+              d.showDeleteCheckbox === true ||
+              d.showDeleteCheckbox === undefined
+          )
+          ?.map(d => d.id)
+      )
+    }
+
+    setCheckedAll(!checkedAll && !indeterminate)
+    setIndeterminate(false)
   }
 
   useLayoutEffect(() => {
@@ -71,17 +90,6 @@ export default function Table({
     setIndeterminate(false)
     handleSelectedRowIds([])
   }, [pageIndex])
-
-  const toggleAll = () => {
-    if (checkedAll || indeterminate) {
-      handleSelectedRowIds([])
-    } else {
-      handleSelectedRowIds(data?.map(d => d.id))
-    }
-
-    setCheckedAll(!checkedAll && !indeterminate)
-    setIndeterminate(false)
-  }
 
   return (
     <div className='relative overflow-x-auto rounded-lg border border-gray-200/75'>
@@ -105,7 +113,7 @@ export default function Table({
         <thead className='border-b border-gray-200/75 bg-zinc-50/50 text-xs text-gray-500'>
           {table.getHeaderGroups().map(headerGroup => (
             <tr key={headerGroup.id}>
-              {allowDelete && data?.length > 0 && (
+              {allowDelete && data?.length > 1 && (
                 <th scope='col' className='relative w-12 px-6 sm:w-16 sm:px-8'>
                   <input
                     type='checkbox'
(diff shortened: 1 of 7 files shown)


ui/components/table.js
```diff
@@ -51,6 +51,10 @@ export default function Table({
   const [checkedAll, setCheckedAll] = useState(false)
   const [indeterminate, setIndeterminate] = useState(false)
 
+  const handleSelectedRowIds = ids => {
```
jmorganca, line 54:
```suggestion
  function handleSelectedRowIds (ids) {
```

This is completely a style choice (since we aren't referring to `this`). Regular function declarations feel more standard.

ui/components/table.js
```diff
@@ -65,11 +69,16 @@ export default function Table({
   useLayoutEffect(() => {
     setCheckedAll(false)
     setIndeterminate(false)
-    setSelectedRowIds([])
+    handleSelectedRowIds([])
   }, [pageIndex])
 
-  function toggleAll() {
-    setSelectedRowIds(checkedAll || indeterminate ? [] : data?.map(d => d.id))
+  const toggleAll = () => {
```
jmorganca, line 75:
```suggestion
  function toggleAll() {
```

ui/pages/access-control/index.js
```diff
@@ -527,6 +527,10 @@ export default function AccessControl() {
     )
   }, [allGrants])
 
+  const handleSelectedDeleteIds = deletedIds => {
```
jmorganca, line 530:
```suggestion
  function handleSelectedDeleteIds(deletedIds) {
```

ui/pages/access-control/index.js
```diff
@@ -700,30 +703,63 @@ export default function AccessControl() {
...
-                    <button
-                      type='button'
-                      onClick={() => {
-                        setSelectedDeleteIds([info.row.original.id])
-                        setOpenSelectedDeleteModal(true)
+                  <>
+                    <div className='group invisible rounded-md bg-white group-hover:visible'>
+                      <button
+                        type='button'
+                        onClick={() => {
```
jmorganca, line 719: I believe this can be one line: `() => setOpen(true)`

ui/pages/access-control/index.js
```diff
@@ -700,30 +703,63 @@ export default function AccessControl() {
...
               )
             },
           },
         ]}
       />
+      {/* bulk delete modal */}
       <DeleteModal
         open={openSelectedDeleteModal}
         setOpen={setOpenSelectedDeleteModal}
+        onCancel={() => {
```
jmorganca, line 760: I think this can also be one line `() => setSelectedDeleteIds([])`

ui/pages/groups/[id].js
```diff
@@ -123,6 +125,10 @@ export default function GroupDetails() {
   const hideRemoveGroupBtn =
     !isAdmin || (infraAdmins?.length === 1 && adminGroups.includes(group?.id))
 
+  const handleSelectedDeleteIds = deletedIds => {
```
Comment (jmorganca) [inline ui/pages/groups/[id].js:128]:
```suggestion
  function handleSelectedDeleteIds(deletedIds) {
```

ui/pages/groups/[id].js
```diff
@@ -123,6 +125,10 @@ export default function GroupDetails() {
   const hideRemoveGroupBtn =
```
Comment (jmorganca) [inline ui/pages/groups/[id].js:125]: Nit: could we make this a positively named variable (vs negative)? `showRemoveGroupBtn` ? It makes reading code that uses it a tad easier below

jmorganca (commented):
Great cleanup! A few small style comments to consider but otherwise LGTM

ui/components/table.js
```diff
@@ -65,23 +88,17 @@ export default function Table({
...
-    setIndeterminate(false)
-  }
-
   return (
     <div className='relative overflow-x-auto rounded-lg border border-gray-200/75'>
       {selectedRowIds.length > 0 && (
         <div className='absolute left-12 flex h-6 items-center py-4 sm:left-16'>
           <button
             type='button'
             onClick={() => {
```
jmorganca, line 100: Nit: `() => onDelete()`

ui/pages/access-control/index.js
```diff
@@ -527,7 +527,12 @@ export default function AccessControl() {
     )
   }, [allGrants])
 
+  function handleSelectedDeleteIds(deletedIds) {
```
jmorganca, line 530: Do we need this function, or could it just be inline below?

ui/pages/groups/[id].js
```diff
@@ -120,8 +123,30 @@ export default function GroupDetails() {
   ]
 
   // Don't allow deleting the last group
-  const hideRemoveGroupBtn =
-    !isAdmin || (infraAdmins?.length === 1 && adminGroups.includes(group?.id))
+  const showRemoveGroupBtn =
+    isAdmin && !(infraAdmins?.length === 1 && adminGroups.includes(group?.id))
```
Comment (jmorganca) [inline ui/pages/groups/[id].js:127]: I believe you can simplify this to:

```suggestion
    isAdmin && !adminGroups.includes(group?.id)
```

ui/pages/groups/[id].js
```diff
@@ -120,8 +123,30 @@ export default function GroupDetails() {
...
+
+  function handleSelectedDeleteIds(deletedIds) {
+    setSelectedDeleteIds(deletedIds)
+  }
+
+  useEffect(() => {
+    setUsersData(
```
Comment (jmorganca) [inline ui/pages/groups/[id].js:134]: Having a `useEffect` only to update another variable often means we can do this elsewhere in the code, or directly in the function here outside of the hook – could we remove it?

ui/pages/groups/[id].js
Comment (hoyyeva) [reply ui/pages/groups/[id].js:127]: no we don't want to remove the check for number of infraAdmins - Because we show the remove group btn as long as there is more than one infra admin. We want to hide the btn when there is only one infra admin and that infra admin is the current group

Comment (jmorganca) [reply ui/pages/groups/[id].js:127]: That makes sense. Shortening the expression (if we can), or at least making it clearer would make this easier to understand. One idea is to break this into two – make this `isLastAdminGroup` and then use `isAdmin` below inline, or something similar.

ui/components/table.js
```diff
@@ -68,26 +90,18 @@ export default function Table({
...
-            onClick={() => {
-              onDelete(selectedRowIds)
-            }}
+            onClick={() => onDelete()}
             className='rounded-md bg-zinc-50 px-4 py-2 text-2xs font-medium  text-red-500 hover:bg-red-100'
           >
             <div className='flex flex-row items-center'>
               <TrashIcon className='mr-1 mt-px h-3.5 w-3.5' />
-              Remoeve selected
+              Remove selected
```
mxyng, line 104: Would it be possible to move this to the right and align it with the `Remove` option of individual items?

ui/pages/access-control/index.js
```diff
@@ -700,30 +700,59 @@ export default function AccessControl() {
...
             cell: function Cell(info) {
+              const [open, setOpen] = useState(false)
+              const { id, identityId, resource, privilege } = info.row.original
+
+              const name =
+                users?.find(u => u.id === identityId)?.name ||
+                groups?.find(g => g.id === identityId)?.name
+
               return (
                 isAdmin && (
```
mxyng, line 711: I'm not sure this is a good idea. It's confusing when an option is inexplicably disabled, doubly so when the option isn't even there

ui/components/table.js
```diff
@@ -96,12 +110,12 @@ export default function Table({
         <thead className='border-b border-gray-200/75 bg-zinc-50/50 text-xs text-gray-500'>
           {table.getHeaderGroups().map(headerGroup => (
             <tr key={headerGroup.id}>
-              {allowDelete && data?.length > 0 && (
+              {allowDelete && data?.length > 1 && (
```
mxyng, line 113: I'd like the checkbox remove option even if there's a single item. The user doesn't need to change their what their doing just because the contents of the table changed. Maybe even have the checkbox available when there's no items but grey out the `Remove` button?

ui/pages/access-control/index.js
hoyyeva: this is actually a good point - i think instead of hiding it, its better to make it disabled

hoyyeva: hmm I am not sure what kind of confusion that you are referring to. The delete button will only be shown when the user is an admin. If the user is not an admin, she will not see the delete button nor be able to delete the grant. I think this is aligning with today's dashboard and on cli

ui/pages/groups/[id].js
```diff
@@ -95,18 +95,20 @@ export default function GroupDetails() {
...
-  const { data: group, mutate: mutate } = useSWR(`/api/groups/${id}`)
+  const limit = 10
+  const { user, isAdmin } = useUser()
+  const { data: group, mutate } = useSWR(`/api/groups/${id}`)
   const {
     data: { items: users, totalCount, totalPages } = {},
     mutate: mutateUsers,
   } = useSWR(`/api/users?group=${group?.id}&limit=${limit}&p=${page}`)
-  const { data: { items: allUsers } = {} } = useSWR(`/api/users?limit=999`)
+  const { data: { items: allUsers } = {} } = useSWR(`/api/users?limit=1000`)
```
Comment (BruceMacD) [inline ui/pages/groups/[id].js:105]: should the match the limit declared on line 98?
```suggestion
  const { data: { items: allUsers } = {} } = useSWR(`/api/users?limit={limit}`)
```

Comment (hoyyeva) [reply ui/pages/groups/[id].js:105]: no, we want to get all the users so that can be listed in the dropdown options. There should have a better way to do it, but we don't want to paginate the users for sure

BruceMacD (approved):
Looks good to me

ui/components/table.js
```diff
@@ -68,112 +84,128 @@ export default function Table({
...
-                        header.column.columnDef.header,
-                        header.getContext()
-                      )}
-                </th>
-              ))}
-            </tr>
-          ))}
+          {table.getHeaderGroups().map(headerGroup => {
+            const disabledBulkDeleteCheckbox =
+              data?.length === 1 && data[0].disabledDeleteCheckbox
```
jmorganca, line 107:
```suggestion
              data?.length === 1 && data?.[0]?.disabledDeleteCheckbox
```

I would recommend renaming this to `enableDeleteCheckbox` since the double negative below can be confusing to read

jmorganca (commented):
Overall looks great. Some minor comments.

jmorganca (approved):

PR #4010, opened 2023-01-12
Fix configs that used file: prefix

dnephin:
The config used by `test_acceptance` and `dev/server.yaml` had `file:` prefix for the TLS CA private key, which need to be updated for the changes from #3951


jmorganca (approved):

jmorganca:
LGTM

Issue #3761, opened 2022-11-29
Support granting SSH access with group membership


dnephin:
### Background

The initial technical preview release of the SSH connector only supports granting access directly to a user. A grant where the subject is a group does not work yet (it will not grant access). This limitation exists because we don't have a reasonable API endpoint that the SSH connector can use to get the fully resolved list of users. Today we use `/api/grants?destination=<name>`, which returns us all the grants. The subject of some of those grants may be a group which contains multiple users.

To get notified of all changes that could impact the destination we would need to long-poll on:
1. `ListGrants` by destination
2. `ListGroupMembers` for each group that is a subject in the grants (potentially `n` calls, or it must accept a list of group IDs)

We could implement long polling for `ListUsersByGroup` (`/api/users?group=x`), but that API is not ideal for the connector. It would still need to make `n+1` API calls, which is extra DB connections (see #3586 for why that's an important consideration), and extra work for the client to manage as the group grants change.

Another disadvantage to this approach is that we end up querying and serializing a bunch of data that isn't necessary for the connector. 

We need the `User.SSHLoginName` and `User.ID`, but we don't need any other details about the user, or any of  the timestamps. 

### Proposal

I propose we create a new API endpoint that is designed specifically for the connector use case.

```
GET /api/destinations/:id/users?lastUpdateIndex=<updateIndex>
```

The only request parameters are the `destination.ID` in the path, and the `lastUpdateIndex` for long polling. The response should look something like this:

```go
type response api.ListResponse[api.DestinationUsers]

type DestinationUsers struct {
    User struct {
        ID uid.ID
        SSHLoginName string
    }
    Grants []struct { // a list of these
        Privilege string // always `connect` for SSH (for now), the cluster role for kuberentes
        Resource string // not used for SSH yet, but used as namespace for kubernetes
    }
}
```

This gives us a compact response with all the details we need in the connector.

Some implementation details:
* the query should be something like `SELECT * from grants LEFT JOIN identities ON grants.user_id == identities.id WHERE grants.destination = <destination>` unioned with a similar query for getting all the users in the groups with granted access (need to think more about that query to see if we can avoid the sub-select).
* the implementation of this blocking request is a little more involved than the existing one, because it involves relationships between multiple tables (group membership and grants). We'll need to add notifications for any group membership changes so that we can listen on those changes.
[...]

PR #4011, opened 2023-01-13
Replace grants.subject with grants.subject_id and grants.subject_kind

dnephin:
## Summary

Migrates the `grants` table from an encoded `subject` field to `subject_id bigint` and `subject_kind smallint`  fields.

Also update all the places that set `models.Grant.Subject` to use the new `models.Subject` type. This removes the last use of `uid.PolymorphicID`.

## Related Issues

Related to https://github.com/infrahq/infra/issues/3761#issuecomment-1380703072


internal/server/data/grant.go
```diff
@@ -104,8 +99,9 @@ func GetGrant(tx ReadTxn, opts GetGrantOptions) (*models.Grant, error) {
 	switch {
 	case opts.ByID != 0:
 		query.B("AND id = ?", opts.ByID)
-	case opts.BySubject != "":
-		query.B("AND subject = ?", opts.BySubject)
+	case opts.BySubject.ID != 0:
```
BruceMacD, line 102: Looks like you should be requiring subject kind here?

dnephin: Good call, I was kind of lazy with these checks. I guess all of these should error if there is an ID set without a kind. The query it produces is still safe in that it matches nothing, but an explicit error will make it easier to debug when we make a mistake.

dnephin:
Tests are passing, and the migration test has an expected value, so I think this is ready for review.

This change required a few steps, so best viewed by individual commit.

internal/server/data/migrations.go
```diff
@@ -1173,3 +1174,70 @@ func addUserPublicKeyUserIDIndex() *migrator.Migration {
...
+				if err != nil {
+					return err
+				}
+			}
+
+			// Step 3 - Remove the old column, add a new index, and NOT NULL
+			_, err = tx.Exec(`
+				ALTER TABLE grants DROP COLUMN IF EXISTS subject;
+				ALTER TABLE grants ALTER COLUMN subject_id SET NOT NULL;
+				ALTER TABLE grants ALTER COLUMN subject_kind SET NOT NULL;
```
dnephin, line 1236: This final step in the migration will cause problems for the other pods (that have not yet been updated to the new code). We could save this step for a later commit, or we could accept the few errors while the pods restart.

Saving it for a later commit only works for SaaS. We'd have to spread this out over 2 releases for on-prem (or accept a few errors during upgrade).

internal/server/data/migrations.go
```diff
@@ -1173,3 +1174,70 @@ func addUserPublicKeyUserIDIndex() *migrator.Migration {
...
+					kind = 1
+				} else {
+					kind = 2
+				}
+				subjectID, err = uid.Parse([]byte(grant.OriginalSubject[2:]))
+				if err != nil {
+					return err
+				}
+
+				_, err = tx.Exec(`UPDATE grants SET subject_id=?, subject_kind=? WHERE id = ?`,
```
dnephin, line 1225: Iterating over every grant and doing an individual update is relatively slow. We can see how long this takes in dev. I suspect it'll be fine at our current scale.

mxyng (approved):

internal/server/data/migrations.go
mxyng: A few errors in dying pods is likely fine. It doesn't make sense to delay this switchover to mitigate a few seconds of errors

PR #4012, opened 2023-01-13
fix: delay creating provider user

### LINKED ISSUE #3908: Remove password generation from "add user" flow
We used a temporary password to invite users to Infra. This needs to be updated now that users can login to an org with Google.

Do not generate a password for a user when they are invited, only create a one-time use-code which expires.

mxyng:
## Summary

Right now when an admin creates a user by email invite, the provider user and credential are created immediately even though a user may not accept the invitation. With this change, only create the provider user and credential when the end user accepts the invite.

Note: this is only applicable when a SMTP server is configured.

Update the UI to allow removing a user that's been invited but hasn't accepted the invite. 

image

1. Current user - user cannot be deleted
2. Okta user - user cannot be deleted
3. Invited user, accepted - user can be delete
4. Invited user, not accepted - user can be deleted

### TODO
- [x] More tests

Resolves #3908

### DIFF
--- a/internal/access/passwordreset.go
+++ b/internal/access/passwordreset.go
@@ -1,8 +1,12 @@
 package access
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/gin-gonic/gin"
 
+	"github.com/infrahq/infra/internal"
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 )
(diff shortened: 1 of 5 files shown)


internal/access/passwordreset.go
```diff
@@ -28,13 +32,21 @@ func VerifiedPasswordReset(c *gin.Context, token, password string) (*models.Iden
...
+	if err := checkPasswordRequirements(tx, password); err != nil {
 		return nil, err
 	}
 
-	if err := checkPasswordRequirements(tx, password); err != nil {
-		return nil, err
+	credential, err := data.GetCredentialByUserID(tx, user.ID)
+	switch {
+	case errors.Is(err, internal.ErrNotFound):
+		if err := createCredential(tx, user, &models.Credential{}, password); err != nil {
```
BruceMacD, line 42: This seems potentially incorrect, if a user only exists in an identity provider this seems like it would allow them to add a password, in our current org controlled model this could be unexpected.

mxyng: You're talking about a password reset request for a user that exists in Infra but only through an IDP? That workflow will never reach this point because the password reset token is only created if an credential already exists for the user: https://github.com/infrahq/infra/blob/main/internal/server/passwordreset.go#L33

BruceMacD (approved):
looks good

Issue #674, opened 2021-11-19
security: revert logger_filter change when go header value bug is fixed


ssoroka:
when https://github.com/golang/go/pull/48979 is merged and we are using it,
revert https://github.com/infrahq/infra/pull/657

PR #4013, opened 2023-01-13
Remove custom http error logger

jmorganca:
Removes a custom error logger with a filter for headers introduced in https://github.com/infrahq/infra/pull/657/ – it is not needed anymore as of Go 1.19

Fixes https://github.com/infrahq/infra/issues/674


BruceMacD (approved):
🧹

internal/connector/connector.go
```diff
@@ -262,18 +261,14 @@ func runKubernetesConnector(ctx context.Context, options Options) error {
...
 	proxy := httputil.NewSingleHostReverseProxy(kubeAPIAddr)
 	proxy.Transport = proxyTransport
-	proxy.ErrorLog = httpErrorLog
 
 	metricsServer := &http.Server{
 		ReadHeaderTimeout: 30 * time.Second,
 		ReadTimeout:       60 * time.Second,
 		Addr:              options.Addr.Metrics,
 		Handler:           metrics.NewHandler(promRegistry),
-		ErrorLog:          httpErrorLog,
```
dnephin, line 276: I think we probably still want these `ErrorLog` set on all the `http.Server` so that the logs are sent to the same place with the same format. Since our logs generally output as JSON, I would expect zerolog to format these as JSON messages as well.  By removing the logger here they definitely won't be JSON.

jmorganca: Thanks! I was looking for this

internal/server/server.go
```diff
@@ -302,7 +302,6 @@ func registerUIRoutes(router *gin.Engine, opts UIOptions) {
 		req.URL.Scheme = remote.Scheme
 		req.URL.Host = remote.Host
 	}
-	proxy.ErrorLog = log.New(logging.NewFilteredHTTPLogger(), "", 0)
```
dnephin, line 305: I forget if we use this proxy in production or not, but maybe we want the `ErrorLog` here as well?

jmorganca: Great catch!

dnephin (approved):
Nice cleanup!

PR #4014, opened 2023-01-14
Reset local user cache after login and logout

jmorganca:
This PR properly clears the cache when logging a user in or out.

### DIFF
--- a/ui/lib/hooks.js
+++ b/ui/lib/hooks.js
@@ -4,8 +4,8 @@ import useSWR from 'swr'
 const INFRA_ADMIN_ROLE = 'admin'
 
 export function useUser({ redirectTo, redirectIfFound } = {}) {
-  const { data: user, error, isValidating, mutate } = useSWR('/api/users/self')
-  const { data: org } = useSWR('/api/organizations/self')
+  const { data: user, error, mutate } = useSWR('/api/users/self')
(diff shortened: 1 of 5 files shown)


BruceMacD (approved):
Could we fix the route behavior by telling SWR that the data from `/api/users/self` has updated at the end of login?

```
// callback.js
        const user = await login({
          oidc: {
            providerID,
            code,
            redirectURL,
          },
        })

        mutate('/api/users/self') // something like this
        router.replace(next ? decodeURIComponent(next) : '/destinations')

        window.localStorage.removeItem('next')
        saveToVisitedOrgs(window.location.host, user?.organizationName)
```

ui/lib/hooks.js
```diff
@@ -54,19 +53,16 @@ export function useUser({ redirectTo, redirectIfFound } = {}) {
       }
 
       // Clear user cache
-      mutate(undefined, false)
+      mutate(undefined, true)
```
jmorganca, line 56: This should have been set to `true` since it will force revalidation: https://github.com/vercel/swr/blob/3f904887eda14861179d70afe7ad9bbdf5c00863/_internal/utils/mutate.ts#L44

BruceMacD (approved):
This seems good to me now

jmorganca:
@BruceMacD thanks for the review!

PR #4015, opened 2023-01-14
Completely clear cache on login and logout

### LINKED ISSUE #3985: UI flashes on first login

jmorganca:
Fixes #3985 

While https://github.com/infrahq/infra/pull/4014 cleared some data, errors would still be cached causing flashes of content on logout and in some remaining cases, on login.

This PR simplifies the redirect logic as well: components can now choose where to redirect users, instead of it automatically happening in `useUser` hook.

### DIFF
--- a/ui/components/layouts/dashboard.js
+++ b/ui/components/layouts/dashboard.js
@@ -78,15 +78,19 @@ function SidebarNav({ children, open, setOpen }) {
 export default function Dashboard({ children }) {
   const router = useRouter()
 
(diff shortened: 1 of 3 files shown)


BruceMacD (approved):
this is easier to reason about

hoyyeva:
ahhh, sorry i am so late to the party! but this PR makes the login/logout logistic a lot clearer than what we used to have! thanks @jmorganca

PR #4027, opened 2023-01-16
Reduce the use of gin.Context - part 1

dnephin:
## Summary

We are currently using a fork of `gin` to support decoding of types that implement `encoding.TextUnmarshaler`.

This PR takes a few small steps to reduce our use of `gin.Context`, and to replace our fork with upstream `gin`.

Full details in the commit messages, but in summary:
1. use `context.Context` directly from the `http.Request.Context()` instead of using `gin.Context` (which implements `context.Context`). `gin.Context`'s implementation of `context.Context` is error prone, and doesn't actually provide the correct context value in some cases (which I didn't bother to debug, because there's really no reason to be using `gin.Context` as a `context.Context`).
2. remove the unnecessary indirection in `readRequest`. `ShouldBind` is replaced with the methods that were called eventually after 2 unnecessary other calls that do nothing but indirect
3. Small bug fix, adding a missing newline

## Related Issues

### DIFF
--- a/internal/server/routes.go
+++ b/internal/server/routes.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"path"
@@ -12,6 +13,7 @@ import (
 	"github.com/Masterminds/semver/v3"
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/gin-gonic/gin"
+	"github.com/gin-gonic/gin/binding"
 
 	"github.com/infrahq/infra/api"
 	"github.com/infrahq/infra/internal"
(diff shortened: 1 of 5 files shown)


internal/server/routes.go
```diff
@@ -395,16 +397,24 @@ func del[Req any, Res any](a *API, r *routeGroup, path string, handler HandlerFu
...
+		}
+		if err := binding.Uri.BindUri(params, req); err != nil {
+			return fmt.Errorf("%w: %s", internal.ErrBadRequest, err)
+		}
 	}
 
-	if err := c.ShouldBindQuery(req); err != nil {
-		return fmt.Errorf("%w: %s", internal.ErrBadRequest, err)
+	if len(c.Request.URL.Query()) > 0 {
+		if err := binding.Query.Bind(c.Request, req); err != nil {
```
dnephin, line 411: By calling `binding` functions here directly it will allow us to replace this library with a much smaller implementation.

There's no functional change here. If you trace the cals of `c.ShouldBind` that I've replaced here you'll see they eventually call these functions (and do nothing else).

BruceMacD (approved):
seems like a good first step

mxyng (approved):

pdevine (approved):

PR #4028, opened 2023-01-16
Remove `gin.Context` from `access` package

dnephin:
## Summary

Branched from #4027

This PR removes the use of `gin.Context` from all access functions.
Also removes a couple access functions that were providing no access (they were just wrapping data functions).

Removing `gin.Context` from access has to be done before we remove it from the `server` package.

## Related Issues


internal/access/access_test.go
```diff
@@ -100,28 +95,26 @@ func TestRequireInfraRole_GrantsFromGroupMembership(t *testing.T) {
...
 	grant(t, tx, admin, tomsGroup.PolyID(), models.InfraAdminRole, "infra")
 
-	authDB, err = RequireInfraRole(c, models.InfraAdminRole)
+	err = IsAuthorized(rCtx, models.InfraAdminRole)
 	assert.NilError(t, err)
-	assert.Assert(t, authDB != nil)
 }
 
-func TestRequireInfraRole(t *testing.T) {
+// TODO: mergge this with TestIsAuthorized
```
BruceMacD, line 113: Seems like combining this as part of this PR would be relatively straight-forward, so might be worth doing as part of this PR, just a suggestion though

mxyng (approved):

internal/access/passwordreset.go
```diff
@@ -1,15 +1,13 @@
...
 
 import (
-	"github.com/gin-gonic/gin"
-
 	"github.com/infrahq/infra/internal/server/data"
 	"github.com/infrahq/infra/internal/server/models"
 )
 
-func VerifiedPasswordReset(c *gin.Context, token, newPassword string) (*models.Identity, error) {
+// TODO: remove
```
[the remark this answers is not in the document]
dnephin: Exactly, this one isn't as easy as moving it because there's a call to an unexported function. I'm hoping the changes in #4004 made this easier.

internal/access/access_test.go
dnephin: As part of merging these two test cases I'd like to review all the test cases to make sure we've got everything covered. I'd like to do this as a follow up to take more time to review the cases.  I've made a note to myself to do this.

PR #4029, opened 2023-01-16
Remove use of `gin` in the connector

dnephin:
## Summary

The connector doesn't require a full framework. The stdlib provides everything we need to implement these handlers.

## Related Issues

### DIFF
--- a/internal/connector/proxy.go
+++ b/internal/connector/proxy.go
@@ -6,34 +6,49 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httputil"
+	"strconv"
 	"sync"
+	"time"
 
-	"github.com/gin-gonic/gin"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/infrahq/infra/internal/certs"
 	"github.com/infrahq/infra/internal/logging"
+	"github.com/infrahq/infra/metrics"
 )
 
 func proxyMiddleware(
 	proxy *httputil.ReverseProxy,
 	authn *authenticator,
 	bearerToken string,
-) gin.HandlerFunc {
-	return func(c *gin.Context) {
-		claim, err := authn.Authenticate(c.Request)
+) func(resp http.ResponseWriter, req *http.Request) {
+	return func(resp http.ResponseWriter, req *http.Request) {
+		start := time.Now()
+		status := http.StatusOK
+		defer func() {
+			metrics.RequestDuration.With(prometheus.Labels{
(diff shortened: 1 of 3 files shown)


metrics/metrics.go
```diff
@@ -38,23 +38,23 @@ func NewRegistry(version string) *prometheus.Registry {
 	return registry
 }
 
+var RequestDuration = prometheus.NewHistogramVec(prometheus.HistogramOpts{
```
dnephin, line 41: Exported so that we can report the same metric from the connector

internal/connector/proxy.go
```diff
@@ -6,34 +6,49 @@ import (
...
+) func(resp http.ResponseWriter, req *http.Request) {
+	return func(resp http.ResponseWriter, req *http.Request) {
+		start := time.Now()
+		status := http.StatusOK
+		defer func() {
+			metrics.RequestDuration.With(prometheus.Labels{
+				"host":   req.Host,
+				"method": req.Method,
+				"path":   "proxy",
+				"status": strconv.Itoa(status),
```
dnephin, line 33: These two values changes slightly:

* previously we'd always report an empty string for `path`, because the proxy doesn't know anything about the route path (there is none). Change to use an explicit `proxy` which shows up better in datadog.
* `status` previously would report any status, but really all we care about in this case is if we rejected the request in the proxy or not. If the backend (kubernetes) happens to return a 4xx or 5xx response, that's not really relevant to this metric because it's not a failure of the proxy.

jmorganca (approved):

internal/connector/proxy.go
```diff
@@ -6,34 +6,49 @@ import (
...
 	bearerToken string,
-) gin.HandlerFunc {
-	return func(c *gin.Context) {
-		claim, err := authn.Authenticate(c.Request)
+) func(resp http.ResponseWriter, req *http.Request) {
+	return func(resp http.ResponseWriter, req *http.Request) {
+		start := time.Now()
+		status := http.StatusOK
+		defer func() {
+			metrics.RequestDuration.With(prometheus.Labels{
```
dnephin, line 29: Ah, I think I missed registering this metric. That was previously done with the middleware. Fixing this now.

dnephin: Ok, fixed this by registering the metric in the same place we register the client output request time.

BruceMacD (approved):

mxyng (approved):

PR #4030, opened 2023-01-16
Reduce the use of `gin.Context` in `server` package

dnephin:
## Summary

Removes the use of `gin.Context` from a few places in the `server` package:
* accessing the `http.Request`
* using the `http.ResponseWriter` directly
* for log context

## Related Issues

### DIFF
--- a/internal/access/request.go
+++ b/internal/access/request.go
@@ -25,7 +25,7 @@ type RequestContext struct {
 
 	// Response is a mutable field. It can be modified by API handlers to add
 	// new response metadata.
-	Response *ResponseMetadata
+	Response *Response
 }
 
 // Authenticated stores data about the authenticated user. If the AccessKey or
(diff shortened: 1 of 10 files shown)


internal/access/request.go
```diff
@@ -36,22 +37,37 @@ type Authenticated struct {
...
+// Response is accumulated by API endpoints and used for logging and
 // reporting.
-type ResponseMetadata struct {
+type Response struct {
+	// HTTPWriter is the http.ResponseWriter that will be used to write the response.
+	// In most cases the HTTPWriter should only be used to write response headers
+	// or cookies using Header().
+	// It is only safe to call Write and WriteHeader if the API handler returns
+	// an empty response and no error.
+	HTTPWriter http.ResponseWriter
```
BruceMacD, line 48: Some of these headers can contain sensitive values (specifically the cookies), any risk of that getting logged here?

dnephin: Not that I can see. It's true that the other fields on this struct are primarily used for logging, but they're all set explicitly in the logging middleware. The `gin` wrapper around this type does capture things like size of response and status code, but the stdlib `ResponseWriter` is only for writing things out. I can't think of any reason someone would try to use anything from `ResponseWriter` in a log output.

mxyng: We still implement a HTTP log filter that omits header values so this shouldn't be a problem

BruceMacD (approved):

internal/access/request.go
dnephin: I think that was only for unknown headers, and we removed it in #4013 because Go fixed the problem. That wouldn't have caught logging of values from headers I don't think.

PR #4031, opened 2023-01-16
Use upstream version of gin, and an inline fork of gin/binding

dnephin:
Edit: #4047 might be a better approach here. I'm not sure if it makes sense to do that first, or if we should do it this way first and take more time to figure out the new implementation.

## Summary

This PR replaces our use of `gin/binding` with `internal/gin-binding` , which is a reduced implementation based on the original, with support for `TextUnmarshal` (the change we had in our fork).  

Using a fork of a single package (instead of the entire framework) has a couple of advantages:
1. we can update `gin` using dependabot, like we do everything else. The only thing not updated is our one replacement package (which is now quite small, less than 200 lines of production code).
2. it reduces our dependency on `gin` overall. Once we've removed `gin.Context` the only remaining thing we'll need from `gin` is the router, which we can likely replace with something that doesn't require us to change how we define our routes. Our requirements for a router are pretty minimal.

The full set of changes from the original can be seen in this [diff](https://github.com/dnephin/gin-binding-tmp/compare/e095a67fe97fdcaff66fac3346be31b1dfb04662...compat-with-infra).

## Related Issues

https://github.com/infrahq/internal/issues/19


internal/server/routes.go
```diff
@@ -429,10 +429,6 @@ func readRequest(c *gin.Context, req interface{}) error {
 	return nil
 }
 
-func init() {
-	gin.DisableBindValidation()
```
dnephin, line 433: Our reduced implementation no longer calls validate, so we're able to remove this.

mxyng:
Closing for now

Issue #4033, opened 2023-01-17
Removing an Infra user removes personal access keys even if the user still exists through an identity provider


mxyng:
Access keys are powerful tools for automation by allowing a script or integration to act as the owner of the access key. In Infra, there are multiple types of access keys, keys for a user account, keys for connectors, and keys for login sessions to name a few. To safeguard access keys from abuse and misuse, there are a few security measures in place. One of these measures revokes access keys when the user is removed.

The problem is a user may be defined through multiple identity providers. Deleting a user through the UI only deletes the Infra user. Since personal access keys are always associated with the Infra provider, deleting a Infra provider user also deletes all personal access keys even if the user still exists through other authentication methods.

## Steps to reproduce

1. Connect an IDP
2. Create an Infra user with the same login as the IDP user
4. Login as this user and create a personal access key
5. Login as admin and remove the Infra user
6. The personal access key for this user no longer exists even though the user still exists

mxyng:
The same logic also removes all SSH keys which is also pretty bad though new keys are created automatically

BruceMacD:
Components:
- [ ] allow removing users from your Infra org, regardless of where they came from
- [ ] remove provider from access keys

PR #4035, opened 2023-01-17
maintain: allow removing any users for an org

BruceMacD:
- allow removing users from your org regardless of where they have logged in from
- remove some previously enforced key deletion requirements which are not needed
- simplify delete identity references

## Summary
Allow removing users completely from Infra (right now you can only remove the Infra user). This is needed because if you invite someone and they sign in through Google, then you can’t actually remove the user from your Infra org without going and finding all their grants to remove.

## Checklist

## Related Issues
Part of #4033


mxyng (approved):
Looks good!

PR #4036, opened 2023-01-17
remove password policy from settings

mxyng:
## Summary

use a static password policy which satisfies NIST guidelines. for more complex password policies, the recommendation is to use an external identity provider


internal/server/models/settings.go
```diff
@@ -1,39 +1,9 @@
...
 
-import (
-	"github.com/infrahq/infra/api"
-)
-
 type Settings struct {
 	Model
 	OrganizationMember
 
 	PrivateJWK EncryptedAtRest
```
BruceMacD, line 7: Doesn't have to be part of this PR, but maybe with should move these settings to the organization entity.

BruceMacD (approved):

dnephin (approved):
Sorry for the conflicts. They should be easy to resolve, I think they are in deleted functions.

mxyng:
> Sorry for the conflicts. They should be easy to resolve, I think they are in deleted functions.

All good. Conflicts were minimal. The only part that tripped me up slightly was `getPublicJWK` getting renamed and moved.

internal/server/models/settings.go
mxyng: This will be done in #4045

PR #4038, opened 2023-01-17
Access control exist grants

hoyyeva:
## Summary
- add highlight to the new / existing grants
- when delete grants, if delete the whole page, it should redirect to the previous page until it is the first page 

## 🎥 
https://user-images.githubusercontent.com/63033505/213018738-c03f4a80-08fa-49d1-bb19-5d2db4cf49ae.mov


ui/pages/access-control/index.js
```diff
@@ -508,11 +508,25 @@ export default function AccessControl() {
   const [openCreateAccess, setOpenCreateAccess] = useState(false)
   const [openSelectedDeleteModal, setOpenSelectedDeleteModal] = useState(false)
   const [selectedDeleteIds, setSelectedDeleteIds] = useState([])
+  const [newCreatedGrants, setNewCreatedGrants] = useState([])
 
   useEffect(() => {
     setGrants(
       allGrants
         // ?.filter(g => g.resource !== 'infra')
```
BruceMacD, line 516:
```suggestion
```

ui/pages/access-control/index.js
```diff
@@ -525,7 +539,13 @@ export default function AccessControl() {
           return g
         })
     )
-  }, [allGrants])
+
+    if (allGrants?.length === 0 && totalCount === 0) {
+      page > 1
+        ? router.replace(`/access-control?p=${page - 1}`)
```
BruceMacD, line 545: Could we send them back to the `totalPages - 1` from the list response rather than going back one page? 

This seems like it could be weird if you bookmarked a page like `/access-control?p=50`, then deleted a bunch of grants. It seems like this would iteratively go back with a lot of redirecting.

hoyyeva: oh yeah, this is a very good point.

BruceMacD (approved):

jmorganca (approved):

PR #4039, opened 2023-01-17
Add API endpoint to support SSH group grants

### LINKED ISSUE #3761: Support granting SSH access with group membership

dnephin:
## Summary

Branched from #4087

Add API endpoint to return destination access. Use that new API endpoint from the SSH connector.

TODO:
* [ ] API endpoint test coverage

## Related Issues

Resolves #3761

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -387,6 +387,13 @@ func (c Client) ListDestinations(ctx context.Context, req ListDestinationsReques
 	})
 }
 
+func (c Client) ListDestinationAccess(ctx context.Context, req ListDestinationAccessRequest) (*ListDestinationAccessResponse, error) {
+	path := fmt.Sprintf("/api/destinations/%s/user-access", req.Name)
+	return get[ListDestinationAccessResponse](ctx, c, path, Query{
+		"lastUpdateIndex": {strconv.FormatInt(req.LastUpdateIndex, 10)},
+	})
+}
+
 func (c Client) CreateDestination(ctx context.Context, req *CreateDestinationRequest) (*Destination, error) {
 	return post[Destination](ctx, c, "/api/destinations", req)
 }
--- a/api/destination.go
+++ b/api/destination.go
@@ -109,3 +109,20 @@ func validateDestinationName(value string) validate.StringRule {
 	}
 	return rule
 }
+
+type ListDestinationAccessRequest struct {
+	Name string `uri:"id"` // TODO: change to ID when grants stores destinationID
+	BlockingRequest
+}
+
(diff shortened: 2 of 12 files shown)


mxyng:
Closing for now

PR #4040, opened 2023-01-18
maintain: rename access key issued_for

### LINKED ISSUE #3830: Re-visit SCIM access keys with 'SCIM' Identity
Right now access keys are issued directly for providers.

We should look at moving to to 'Provider' identity pattern to match the connector.

BruceMacD:
- rename access key issued_for to issued_for_id and add issued_for_kind
- specifically check for scim access keys

## Summary
Update access key `IssuedFor` to `IssuedForID` and add `IssuedForKind`.

## Checklist

Resolves #3830

### DIFF
--- a/internal/server/access_keys_test.go
+++ b/internal/server/access_keys_test.go
@@ -23,22 +23,42 @@ import (
 func TestAPI_CreateAccessKey(t *testing.T) {
 	type testCase struct {
 		name     string
-		setup    func(t *testing.T) io.Reader
+		setup    func(t *testing.T) (io.Reader, string)
 		expected func(t *testing.T, response *httptest.ResponseRecorder)
 		headers  http.Header
 	}
 
-	srv := setupServer(t, withAdminUser)
+	srv := setupServer(t)
 	routes := srv.GenerateRoutes()
 
-	userResp := createUser(t, srv, routes, "usera@example.com")
+	admin := createAdmin(t, srv.DB())
+	adminKey := &models.AccessKey{
+		IssuedForID:   admin.ID,
+		IssuedForKind: models.IssuedForKindUser,
+		ExpiresAt:     time.Now().Add(1 * time.Hour),
+		Scopes:        []string{models.ScopeAllowCreateAccessKey},
+	}
+	adminAccessKey, err := data.CreateAccessKey(srv.DB(), adminKey)
+	assert.NilError(t, err)
+	connector, err := data.GetIdentity(srv.DB(), data.GetIdentityOptions{ByName: "connector"})
+	assert.NilError(t, err)
+	user := &models.Identity{Name: "usera@example.com"}
+	assert.NilError(t, data.CreateIdentity(srv.DB(), user))
+	userKey := &models.AccessKey{
+		IssuedForID:   user.ID,
+		IssuedForKind: models.IssuedForKindUser,
+		ExpiresAt:     time.Now().Add(1 * time.Hour),
+		Scopes:        []string{models.ScopeAllowCreateAccessKey},
+	}
+	userAccessKey, err := data.CreateAccessKey(srv.DB(), userKey)
+	assert.NilError(t, err)
 
 	run := func(t *testing.T, tc testCase) {
-		body := tc.setup(t)
+		body, key := tc.setup(t)
 
 		// nolint:noctx
 		req := httptest.NewRequest(http.MethodPost, "/api/access-keys", body)
-		req.Header.Set("Authorization", "Bearer "+adminAccessKey(srv))
+		req.Header.Set("Authorization", "Bearer "+key)
 		req.Header.Set("Infra-Version", apiVersionLatest)
 
 		for k := range tc.headers {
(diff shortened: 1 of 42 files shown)


internal/server/authn/key_exchange_test.go
```diff
@@ -13,10 +13,10 @@ import (
 )
 
 func TestKeyExchangeAuthentication(t *testing.T) {
-	db := setupDB(t)
+	db, tx := setupDB(t)
```
dnephin, line 16: Why does this test make use of both `db` and `tx` ? I see some calls use one and some use the other, but I don't think I understand why.

internal/server/data/access_key.go
```diff
@@ -94,18 +85,23 @@ func CreateAccessKey(db WriteTxn, accessKey *models.AccessKey) (body string, err
...
-		identityIssuedFor, err := GetIdentity(db, GetIdentityOptions{ByID: accessKey.IssuedFor})
+	if accessKey.IssuedForKind != models.AccessKeyForKindProvider {
+		identityIssuedFor, err := GetIdentity(db, GetIdentityOptions{ByID: accessKey.IssuedForID})
 		if err != nil {
 			return "", fmt.Errorf("key name from identity: %w", err)
 		}
 
-		accessKey.Name = fmt.Sprintf("%s-%s", identityIssuedFor.Name, accessKey.ID.String())
+		if identityIssuedFor.Name == models.InternalInfraConnectorIdentityName {
+			accessKey.IssuedForKind = models.AccessKeyForKindConnector
```
dnephin, line 95: Do we make use of this `KindConnector`, yet? Or could we introduce that if/when connector isn't an identity in the DB?

BruceMacD: Yeah it's not used at the moment. I see it as a way to get rid of the connector identity when we choose to do so, but it also felt incorrect to create the connector key as kind `user` in general. Im still open to doing this as a separate change though if you're worried about the lack of use.

dnephin: For me the question is more about if connector is the right `Kind` here. Maybe it should be `kind=organization` so that it's not specific to a connector. Any org admin would be able to manage `kind=organization`.

internal/server/access_keys_test.go
```diff
@@ -114,14 +137,169 @@ func TestAPI_CreateAccessKey(t *testing.T) {
 				assert.DeepEqual(t, respBody.FieldErrors, expected)
 			},
 		},
+		{
```
BruceMacD, line 140: this is the main new test coverage

internal/server/data/access_key.go
BruceMacD: Made this change to `kind=organization` for connector access keys.

internal/server/authn/key_exchange_test.go
BruceMacD: This was due to creating things within an org, then making a request (it would get different org IDs). It's fixed now though, so only the transaction is used.

api/client.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/infrahq/infra/uid"
 )
 
-var apiVersion = "0.18.1"
```
BruceMacD, line 22: Just validating, this is the correct way to update the CLI's API version?

mxyng: Yes but 0.20.1 is not a valid version

internal/server/models/access_key.go
```diff
@@ -13,20 +13,20 @@ var (
...
 	OrganizationMember
-	Name string
-	/* IssuedFor is either:
-	1. The ID of the user that this access key was created for.
-	2. The ID of a provider that is doing SCIM provisioning using this access key.
-	*/
-	IssuedFor     uid.ID
+	Name          string
+	IssuedForID   uid.ID
+	IssuedForKind string
```
mxyng, line 29: Is it necessary to use string for kind? It seems the things an access key can be issued for are finite and well known so an enum is likely sufficient

internal/server/data/migrations.go
```diff
@@ -1258,3 +1259,53 @@ func removeSettingsPasswordPolicy() *migrator.Migration {
...
+			stmt := `
+				ALTER TABLE access_keys RENAME COLUMN issued_for TO issued_for_id;
+				ALTER TABLE access_keys ADD COLUMN IF NOT EXISTS issued_for_kind text DEFAULT 'user';
+			`
+			if _, err := tx.Exec(stmt); err != nil {
+				return fmt.Errorf("add access key issued_for_kind and issued_for_id: %w", err)
+			}
+			if _, err := tx.Exec(`UPDATE access_keys SET issued_for_kind = 'provider' WHERE issued_for_id = provider_id;`); err != nil {
+				return fmt.Errorf("set access key issued_for_kind provider: %w", err)
+			}
```
mxyng, line 1282: These `tx.Exec()` can likely be aggregated into a single statement

internal/server/data/migrations.go
```diff
@@ -1258,3 +1259,53 @@ func removeSettingsPasswordPolicy() *migrator.Migration {
...
+			connectorIDs, err := scanRows(rows, func(id *uid.ID) []any {
+				return []any{id}
+			})
+			if err != nil {
+				return fmt.Errorf("read connector identity rows: %w", err)
+			}
+			if len(connectorIDs) > 0 {
+				query := querybuilder.New(`UPDATE access_keys`)
+				query.B(`SET issued_for_kind = 'organization'`)
```
mxyng, line 1297: Does this impact how connector keys are presented to the user?

internal/server/models/access_key.go
dnephin: I used an int for the `SubjectID `field in grants. It's a `smallint` database type, and in Go it's this: https://github.com/infrahq/infra/blob/af7a6e6cb3ef43fbe3fa22abfe12e33809c68ee5/internal/server/models/grant.go#L39-L50, which is basically a Go version of an enum.

Database enum types could also work, but I think end up being harder to update and change.

mxyng: That's what I had in mind. Not real enums since Go doesn't understand

mxyng:
Sorry about the merge conflicts. It should be relatively simple to resolve

internal/server/data/migrations.go
BruceMacD: At the moment it is still presented by "user=connector" and not changed, but in a future change we can set a filter based on "kind=organization" instead.

mxyng (approved):

PR #4041, opened 2023-01-18
Refactor AssignIdentityToGroups to use IDs

dnephin:
## Summary

In #4039 I need all the IDs of the groups so that we can use these IDs to send `pg_notify` notifications about changes to group membership, and to set the `update_index` of these groups.

This PR extracts a function to return a mapping of group name->ID, and uses that mapping to replace the sub-select in the "remove groups" operation, and the "does group exist" lookup in the "add groups" operation.

Extracted from #4039

### DIFF
--- a/internal/server/data/identity.go
+++ b/internal/server/data/identity.go
@@ -69,8 +69,8 @@ func AssignIdentityToGroups(tx WriteTxn, user *models.ProviderUser, newGroups []
 	}
 
 	newGroups = deduplicate(newGroups)
-
 	oldGroups := user.Groups
+
 	groupsToBeRemoved := diff(oldGroups, newGroups)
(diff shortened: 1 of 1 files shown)


internal/server/data/identity.go
```diff
@@ -144,6 +123,8 @@ func AssignIdentityToGroups(tx WriteTxn, user *models.ProviderUser, newGroups []
 				return nil, fmt.Errorf("create group: %w", err)
 			}
 			groupID = group.ID
+			// add to the map so we can use the IDs below
+			addIDs[name] = groupID
```
dnephin, line 127: This is not used until #4039, but I didn't want to lose it, so kept it in the same commit

BruceMacD (approved):
looks good

PR #4042, opened 2023-01-18
Move pg_notify to the server

dnephin:
## Summary

Related to #4039. Best viewed by individual commit.

This PR changes how we send notifications, and adds support for listening for multiple notifications.

1. We need to listen on multiple channels for #4039, so this PR starts by refactoring `ListenForNotify` to accept a list of channels. The previous "does notify match" becomes a lot more complicated with multiple listens, so this required a few other changes to remove the match function.
2. The channel for "grants by destination" had to use a match function because we don't have access to the destination ID from the trigger that was sending the notification. To address this I moved all the `pg_notify` into the application. I think this makes them easier to manage, but does require us to send the notification from every place a change is made. This happens to be the same places that we have to set the `update_index`, so it ends up being not so bad to send the notify from the server.
3. We can't use the destination name in the channel name (destination name is too long), so we have to look up the destination ID from the name before we do a notify or listen.

#4043 removes the triggers and functions as a follow up PR. We should deploy them to production separately so that we don't break any active blocking requests.

### DIFF
--- a/internal/access/grant.go
+++ b/internal/access/grant.go
@@ -60,9 +60,9 @@ func ListGrants(rCtx RequestContext, opts data.ListGrantsOptions, lastUpdateInde
 		return ListGrantsResponse{}, err
 	}
 
-	listenOpts := data.ListenForNotifyOptions{
-		GrantsByDestination: opts.ByDestination,
-		OrgID:               rCtx.DBTxn.OrganizationID(),
+	listenOpts := data.ListenChannelGrantsByDestination{
+		Destination: opts.ByDestination,
+		OrgID:       rCtx.DBTxn.OrganizationID(),
 	}
 	listener, err := data.ListenForNotify(rCtx.Request.Context(), rCtx.DataDB, listenOpts)
 	if err != nil {
--- a/internal/access/grant.go
+++ b/internal/access/grant.go
@@ -55,14 +55,19 @@ func ListGrants(rCtx RequestContext, opts data.ListGrantsOptions, lastUpdateInde
 		return ListGrantsResponse{Grants: result}, err
 	}
 
+	dest, err := data.GetDestination(rCtx.DBTxn, data.GetDestinationOptions{ByName: opts.ByDestination})
+	if err != nil {
+		return ListGrantsResponse{}, err
+	}
+
 	// Close the request scoped txn to avoid long-running transactions.
 	if err := rCtx.DBTxn.Rollback(); err != nil {
 		return ListGrantsResponse{}, err
 	}
 
 	listenOpts := data.ListenChannelGrantsByDestination{
-		Destination: opts.ByDestination,
-		OrgID:       rCtx.DBTxn.OrganizationID(),
+		DestinationID: dest.ID,
+		OrgID:         rCtx.DBTxn.OrganizationID(),
 	}
 	listener, err := data.ListenForNotify(rCtx.Request.Context(), rCtx.DataDB, listenOpts)
 	if err != nil {
--- a/internal/server/data/destinationcredential.go
+++ b/internal/server/data/destinationcredential.go
@@ -1,6 +1,7 @@
 package data
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/infrahq/infra/internal/server/data/querybuilder"
@@ -33,7 +34,9 @@ func CreateDestinationCredential(tx WriteTxn, cr *models.DestinationCredential) 
 		return handleError(err)
 	}
 
-	return nil
+	stmt := `SELECT pg_notify(current_schema() || '.destCredReq.' || ?, '')`
+	_, err = tx.Exec(stmt, fmt.Sprintf("%v.%v", tx.OrganizationID(), cr.DestinationID))
+	return err
 }
 
 func AnswerDestinationCredential(tx WriteTxn, cr *models.DestinationCredential) error {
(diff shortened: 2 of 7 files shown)


internal/access/grant.go
```diff
@@ -55,14 +55,19 @@ func ListGrants(rCtx RequestContext, opts data.ListGrantsOptions, lastUpdateInde
 		return ListGrantsResponse{Grants: result}, err
 	}
 
+	dest, err := data.GetDestination(rCtx.DBTxn, data.GetDestinationOptions{ByName: opts.ByDestination})
```
BruceMacD, line 58: It looks like `ByDestination` is optional in `ListGrants` calls. Will this error when `ByDestination` isn't specified? In general it feels safer to skip this look-up if `ByDestination` isn't set.

internal/server/data/destinationcredential.go
```diff
@@ -33,7 +34,9 @@ func CreateDestinationCredential(tx WriteTxn, cr *models.DestinationCredential)
 		return handleError(err)
 	}
 
-	return nil
+	stmt := `SELECT pg_notify(current_schema() || '.dcredReq.' || ?, '')`
```
BruceMacD, line 37: Basic question, what is `.dcredReq.` here? A pattern that matches all credential long polling?

internal/server/data/notify.go
```diff
@@ -92,39 +60,50 @@ func ListenForNotify(ctx context.Context, db *DB, opts ListenForNotifyOptions) (
...
+	return fmt.Sprintf("grantsByDest.%v.%v", l.OrgID, l.DestinationID)
+}
+
+type ListenChannelDestinationCredentialsByDestinationID struct {
+	OrgID         uid.ID
+	DestinationID uid.ID
+}
+
+func (l ListenChannelDestinationCredentialsByDestinationID) Channel() string {
+	return fmt.Sprintf("dcredReq.%v.%v", l.OrgID, l.DestinationID)
```
BruceMacD, line 90: I find these formats a bit confusing. The channel name standard seems to be `abbreviated-name.org-id.destination-id`? Feels like we should document this somewhere.

internal/access/grant.go
dnephin: Good call out!

It's not super obvious from here, but we have to be very strict about which parameters are used with long polling. We only support long polling with a `destination` query parameter, and no other query parameters. That's implemented here:

https://github.com/infrahq/infra/blob/64d306a5ae388ebd3cb272b91bd6263cb69dd93b/api/grant.go#L73-L91

On line 53 above we check if `lastUpdateIndex` is 0, so the only way to get to line 58 here is for long polling. In the future if we have other supported parameter combinations we'll have to make sure this destination lookup is only done for the destination parameter.

All the code below this also requires the destination parameter. So our check for "can this continue" is line 53. I don't think it would make sense to skip this lookup here, because the rest of the code would also need to be skipped.

We could add another `if opts.ByDestination == "" { ... } ` that returns and comment that it's an extra safeguard that is normally handled by request validation.

internal/server/data/destinationcredential.go
dnephin: Good question! 

It's unfortunate, but the max length for a channel name is 63 characters (https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).

We need the schema name in here to prevent tests from interfering with each other (channels are not scoped to a schema like tables, and indexes), and that schema name can be pretty long as well, which limits the channel name even further.

We also have two base58 encoded identifiers, which each take up to 11 characters, so with the dot separators that's 24 characters.
 
We want to allow at least 22 characters for the schema name, so that only gives us ~15 or so for the channel name. Our tests regularly use schema names of 21 characters (maybe more I'll need to check again), and we don't know what schema names people will use for self-hosted installs.

This `dcredReq` is short for `DestinationCredentialRequest`, which is too long for use here.

We could change this to `destCredRequest` which only allows for ~22 characters for the schema name. That's probably ok, but is a bit close to the limit. `destCredReq` might be a little safer.

I can also adjust the names we use in tests to give us more space, but that also limits the schema name limit for self-hosted installs.

internal/server/data/notify.go
dnephin: Ya, that's right. I think the `group_membership` should probably be `GroupMembers` to match the convention and reduce the length a bit.

I'll add a comment to the interface to explain the limitations and the convention.

internal/server/data/destinationcredential.go
dnephin: I've changed these to `destCredReq` and `destCredResp`, and documented the size limit in the godoc for the `ListenChannelDescriptor` interface.

internal/server/data/notify.go
dnephin: I've documented this convention in the godoc for the `ListenChannelDescriptor` interface that these types implement.

internal/server/data/grant.go
```diff
@@ -277,8 +273,22 @@ func DeleteGrants(tx WriteTxn, opts DeleteGrantsOptions) error {
...
+
+	rows, err := tx.Query(query.String(), query.Args...)
+	if err != nil {
+		return err
+	}
+
+	grants, err := scanRows(rows, func(g **models.Grant) []any {
+		*g = &models.Grant{}
+		return (*grantsTable)(*g).ScanFields()
+	})
```
dnephin, line 287: I forgot to call this out, but this is kind of interesting as well.

We use `RETURNING` with `columnsForSelect` as part of an `UPDATE` statement to get the data about which rows change.

The `scanRows` uses a `**models.Grant` so that the return value is a `[]*models.Grant`, which is what we need to pass to `notifyGrantsChanged`.

In the future if we store destinationId on the grants table we might be able to simply use `RETURNING destination_id`.

internal/access/grant.go
BruceMacD: I think the request validation seems ok for now, thanks for the clarification.

BruceMacD (approved):
Thanks for the detailed explanations. This seems good to me now.

mxyng:
Closing this for now

PR #4043, opened 2023-01-18
Remove unused database triggers and functions

dnephin:
## Summary

Branched from #4042, this PR removes the now unused triggers and functions.

We should deploy this as a separate change to not break active long-polling requests.


internal/server/data/migrations.go
```diff
@@ -1173,3 +1174,33 @@ func addUserPublicKeyUserIDIndex() *migrator.Migration {
 		},
 	}
 }
+
+func sqlFunctionsMigration() *migrator.Migration {
```
dnephin, line 1178: This moved from another file, so that I could delete the now empty file.

BruceMacD (commented):
This seems good once the other changes go in. If this deployed with connections actively waiting on these triggers they would just return, then a new request would open another poll that would work with the new polling triggers, correct?

dnephin:
I don't think active requests will notice the removal of triggers, so I'd expect the active requests to block until the timeout, then repeat the request and be fine.

BruceMacD (approved):

mxyng:
Closing this for now

PR #4045, opened 2023-01-18
maintain: move jwk to organizations table

mxyng:
## Summary

`models.Settings` contains minimal data after removing password policy. The data contained in this table is more appropriate for the organizations table so move it there instead. 

Related #4036


internal/server/data/organization.go
```diff
@@ -33,12 +39,39 @@ func CreateOrganization(tx WriteTxn, org *models.Organization) error {
...
+		return err
+	}
+
+	pubs, err := pub.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	org.PrivateJWK = models.EncryptedAtRest(secs)
+	org.PublicJWK = pubs
```
dnephin, line 70: Not a blocker, but I guess in general we would only set these values if the fields are nil, and we'd probably set these things from the caller. It's probably fine to keep it here for now.

internal/server/server.go
```diff
@@ -200,7 +200,7 @@ func New(options Options) (*Server, error) {
 	}
 
 	if options.EnableTelemetry {
-		server.tel = NewTelemetry(server.db, db.DefaultOrgSettings.ID)
+		server.tel = NewTelemetry(server.db, db.DefaultOrg.ID)
```
dnephin, line 203: This might be a problem, the `DefaultOrg.ID` is always `1000`, so this is going to make every install look like the same one.  I think we need a different ID here.

Maybe we can keep the `settings.ID` stored as a field on `Organization`, something like `InstallID` ? For new installs we can generate a new `uid.ID` for them, so they all have different IDs.

internal/server/handlers.go
```diff
@@ -63,7 +63,7 @@ func CreateToken(c *gin.Context, r *api.EmptyRequest) (*api.CreateTokenResponse,
 	if rCtx.Authenticated.User == nil {
 		return nil, fmt.Errorf("no authenticated user")
 	}
-	token, err := data.CreateIdentityToken(rCtx.DBTxn, rCtx.Authenticated.User.ID)
+	token, err := data.CreateIdentityToken(rCtx.DBTxn, rCtx.Authenticated.Organization, rCtx.Authenticated.User)
```
dnephin, line 66: This is great! Being able to use the authenticated data in this handlers and the one below instead of querying for data we already have in memory.

internal/server/server.go
mxyng: Good catch!

internal/server/data/migrations.go
```diff
@@ -1258,3 +1259,28 @@ func removeSettingsPasswordPolicy() *migrator.Migration {
...
+				_, err := tx.Exec(`
+					ALTER TABLE organizations
+						ADD COLUMN IF NOT EXISTS private_jwk bytea,
+						ADD COLUMN IF NOT EXISTS public_jwk bytea,
+						ADD COLUMN IF NOT EXISTS install_id bigint;
+					UPDATE organizations
+						SET private_jwk = settings.private_jwk, public_jwk = settings.public_jwk, install_id = settings.id
+						FROM settings
+						WHERE organizations.id = settings.organization_id
+							AND organizations.private_jwk IS NULL
```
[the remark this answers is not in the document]
mxyng: They should be null in all migration cases. It seems like a good safeguard to not accidentally override anything but this was before the entire migration was wrapped in a `migrator.HasTable("settings")` so it's probably not needed anymore

BruceMacD (approved):

PR #4046, opened 2023-01-18
fix: linting errors unseen by exclusion rules

mxyng:
Fix and remove excluded linting rules

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,14 +23,8 @@ linters-settings:
 
 issues:
   exclude-rules:
-    - linters: [staticcheck]
-      text: "\"io/ioutil\" has been deprecated"
     - linters: [errcheck]
       text: "tx\\.Rollback"
-    - linters: [noctx]
-      path: "_test\\.go"
-    - linters: [revive]
-      text: "don't use underscores"
 
 linters:
   enable:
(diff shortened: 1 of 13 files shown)


jmorganca (approved):
Great! Thanks for fixing this!

.golangci.yml
```diff
@@ -23,14 +23,8 @@ linters-settings:
...
 issues:
   exclude-rules:
-    - linters: [staticcheck]
-      text: "\"io/ioutil\" has been deprecated"
     - linters: [errcheck]
       text: "tx\\.Rollback"
-    - linters: [noctx]
-      path: "_test\\.go"
-    - linters: [revive]
-      text: "don't use underscores"
```
dnephin, line 33: In general I think these are pretty reasonable exclude rules. We don't need to care about context usage in tests, and we are not obligated to follow some arbitrary set of linter rules. `revive` has lots of rules. Some of them are good, others are questionable, and a few are harmful. We shouldn't be shy about adding excludes here when we find rules that don't work well.

dnephin (approved):
LGTM

.golangci.yml
mxyng: I don't disagree but these are rules we already follow in most places in the code base.

PR #4052, opened 2023-01-19
Add test coverage for connector proxy

dnephin:
## Summary

This PR adds a test that fails without the changes from #4050.

I think the approach we use in the server is better in general, but that would require some major changes to the connector. This PR takes an easier approach by requiring the caller to opt-in.

### DIFF
--- a/internal/cmd/connector_test.go
+++ b/internal/cmd/connector_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/cmd/connector_test.go
```diff
@@ -198,7 +200,21 @@ func TestConnector_Run_Kubernetes(t *testing.T) {
 	})
 	assert.DeepEqual(t, fakeKube.writes, expectedWrites, cmpKubeRequest)
 
-	// TODO: check proxy is listening
+	<-opts.Addrs.Available
```
dnephin, line 203: This is necessary to prevent a data race. The `opts.Addrs` are set by a goroutine, so we need to block until they are set.

internal/cmd/connector_test.go
```diff
@@ -198,7 +200,19 @@ func TestConnector_Run_Kubernetes(t *testing.T) {
...
+	u := fmt.Sprintf("https://%v/api/v1/namespaces/default/pods/podok", opts.Addrs.HTTPS.String())
```
BruceMacD, line 204: what's at this endpoint?

dnephin: Nothing, the service behind the proxy is a fake. We only care that the proxy is handling requests

BruceMacD (approved):

jmorganca (approved):

PR #4053, opened 2023-01-19
Fix panic in infra ssh command

### LINKED ISSUE #4048: Nasty error when using the wrong key for ssh
## Describe the bug

I created a key like the docs said, then deleted it and created a new one with a better name. Accidentally I copied the original key and pasted that into the command on the ssh host. When the host didn't connect to the server, i tried ssh'ing in as before and got this:

```
ssh ubuntu@212.2.243.16
Removing /Users/matt/.ssh/infra/keys/jpzAPybW35 because it was expired or deleted from Infrapanic: runtime error: slice bounds out of range [2:1]

goroutine 1 [running]:
golang.org/x/exp/slices.Delete[...]({0x140003da800?, 0x1025849e4?, 0x1024f5e7d?}, 0x14000adf368?, 0x1?)
	/home/runner/go/pkg/mod/golang.org/x/exp@v0.0.0-20221012211006-4de253d81b95/slices/slices.go:162 +0x100
github.com/infrahq/infra/internal/cmd.provisionSSHKey({0x1031163d8, 0x14000116000}, {0x1400004c0f0, 0x1400057c750, {0x1400057e360, 0x16}, 0x1400099c8e8, 0x14000258dc0})
	/home/runner/work/infra/infra/internal/cmd/ssh.go:248 +0xe9c
github.com/infrahq/infra/internal/cmd.setupDestinationSSHConfig({0x1031163d8, 0x14000116000}, 0x1400004c0f0, 0x16f2c72f8?)
	/home/runner/work/infra/infra/internal/cmd/ssh.go:180 +0x1c0
github.com/infrahq/infra/internal/cmd.runSSHHosts(0x0?, {0x16f2c72f8, 0xc}, {0x16f2c7305, 0x2})
	/home/runner/work/infra/infra/internal/cmd/ssh.go:96 +0x144
github.com/infrahq/infra/internal/cmd.newSSHHostsCmd.func1(0x140004f6f00?, {0x14000a44b80?, 0x2?, 0x2?})
	/home/runner/work/infra/infra/internal/cmd/ssh.go:46 +0x4c
github.com/spf13/cobra.(*Command).execute(0x140004f6f00, {0x14000a44b40, 0x2, 0x2})
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:916 +0x5c8
github.com/spf13/cobra.(*Command).ExecuteC(0x140009f0300)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044 +0x35c
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/spf13/cobra.(*Command).ExecuteContext(...)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:961
github.com/infrahq/infra/internal/cmd.Run({0x1031163d8?, 0x14000116000}, {0x14000136010, 0x4, 0x4})
	/home/runner/work/infra/infra/internal/cmd/cmd.go:30 +0xb0
main.main()
	/home/runner/work/infra/infra/main.go:17 +0x64
ubuntu@212.2.243.16's password:```

I get why it did what it did, but it’s a super ugly error

## Expected behavior

Just let me know nicely that I fat fingered the key

## Screenshots

<!-- If applicable, add screenshots to help explain your problem. -->

## Environment

<!-- Infra client and server versions using `infra version` -->

```
$ infra version
```

<!-- Kubernetes version using `kubectl version` -->

```
$ kubectl version
```

<!-- Where you are running Kubernetes (e.g. EKS, AKS, GKE, Docker Desktop, Minikube, etc.) -->

## Relevant Infra Logs

<!--
Use `kubectl logs deployment/infra-server` and `kubectl logs deployment/infra-connector` and
add any suspicious logs here. Try to keep snippets short and concise.

Sensitive information should be masked/redacted.
-->

## Additional context

<!-- Add any other context about the problem here. -->

### PR DESCRIPTION
**Pull request by dnephin**
## Summary

Fixes #4048

There were two bugs in the same place that could cause the wrong files to be deleted, or in the worst case a panic.

The first bug was the `for` loop was iterating over the `existingKeys` slice, but deleting from a different slice using the loop index variable. This meant that in some cases the wrong local key file (one for a different server, or different org) would be deleted.
This bug is fixed by no longer creating a separate slice. Iterate over the main slice, and skip entries as necessary.

The second bug was that we kept deleting using that index variable, even after having deleted previous entries. If there were at least 2 deletes then our index would be out of range and cause the panic.
This bug is fixed by only incrementing the index variable when we skip entries. When we delete the current entry the index variable will be pointing at the next item already (ex: we delete index 2, variable is still at 2, next item was previously at index 3 but now is at index 2), so we don't need to increment it.

### DIFF
--- a/internal/cmd/ssh.go
+++ b/internal/cmd/ssh.go
@@ -253,8 +253,14 @@ func provisionSSHKey(ctx context.Context, opts provisionSSHKeyOptions) (string, 
 	}
 
 	keysDir := filepath.Join(opts.infraSSHDir, "keys")
-	existingKeys := matchingPublicKeys(keysCfg, opts.hostConfig, org.ID)
-	for i, existing := range existingKeys {
+	// the loop variable i is only incremented when we skip a key
+	for i := 0; i < len(keysCfg.Keys); {
+		existing := keysCfg.Keys[i]
+		// skip any keys that are for different servers, orgs, or users
+		if !publicKeyMatches(existing, opts.hostConfig, org.ID) {
+			i++
+			continue
+		}
 		filename := filepath.Join(keysDir, existing.PublicKeyID)
 		if !fileExists(filename) || !fileExists(filename+".pub") {
 			// key doesn't exist locally
--- a/internal/cmd/ssh_keys.go
+++ b/internal/cmd/ssh_keys.go
@@ -54,20 +54,12 @@ func writeKeysConfig(infraSSHDir string, cfg *keysConfig) error {
 	return fh.Close()
 }
 
-// matchingPublicKeys looks for any keys in cfg that match the hostCfg and OrgID,
-// and returns the list of matching keys. Returns nil if none match.
-func matchingPublicKeys(cfg *keysConfig, hostCfg *ClientHostConfig, orgID uid.ID) []localPublicKey {
-	var result []localPublicKey
-	for _, publicKey := range cfg.Keys {
-		if publicKey.Server != hostCfg.Host || publicKey.UserID != hostCfg.UserID.String() {
-			continue
-		}
-		if publicKey.OrganizationID != orgID.String() {
-			continue
-		}
-		result = append(result, publicKey)
-	}
-	return result
+// publicKeyMatches matches localKey against hostCfg and OrgID. Returns true
+// when the localKey matches the host, user and org.
+func publicKeyMatches(localKey localPublicKey, hostCfg *ClientHostConfig, orgID uid.ID) bool {
+	return localKey.Server == hostCfg.Host &&
+		localKey.UserID == hostCfg.UserID.String() &&
+		localKey.OrganizationID == orgID.String()
 }
 
 func userPublicKeyContains(keys []api.UserPublicKey, id string) bool {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: internal/cmd/ssh.go [hunk 1]
```diff
@@ -253,8 +253,12 @@ func provisionSSHKey(ctx context.Context, opts provisionSSHKeyOptions) (string,
 	}
 
 	keysDir := filepath.Join(opts.infraSSHDir, "keys")
-	existingKeys := matchingPublicKeys(keysCfg, opts.hostConfig, org.ID)
-	for i, existing := range existingKeys {
+	for i := 0; i < len(keysCfg.Keys); {
+		existing := keysCfg.Keys[i]
+		if !publicKeyMatches(existing, opts.hostConfig, org.ID) {
+			i++
```
BruceMacD, line 259:
```suggestion
                        // this public key is not managed by infra, we can proceed to the next index
			i++
```
Does this comment describe what is happening here? I was a bit confused.

dnephin: Almost, it's very likely still managed by infra, but it's for a different infra server, org, or user. I'll add the comment

mxyng (approved):

PR #4054, opened 2023-01-20
Filter infra grants

### LINKED ISSUE #3800: Provide a way to filter out `infra` grants in the `/api/grants` API

pdevine:
## Summary

This PR changes the `/api/grants` to filter both connector grants, and infra admin grants from the default grants listing. To display the infra and connector grants, you can append the `showSystem=true` argument to the query. This behaviour is different than how `showSystem` previously worked, which only filtered out the results for the connector grants. The API maintains backwards compatibility via the `Infra-Version` header.

## Checklist

## Related Issues

Resolves #3800

### DIFF
--- a/api/grant.go
+++ b/api/grant.go
@@ -35,13 +35,14 @@ func (r *CreateGrantResponse) StatusCode() int {
 }
 
 type ListGrantsRequest struct {
-	User          uid.ID `form:"user" note:"ID of user granted access" example:"6TjWTAgYYu"`
-	Group         uid.ID `form:"group" note:"ID of group granted access" example:"6k3Eqcqu6B"`
-	Resource      string `form:"resource" example:"production.namespace" note:"a resource name"`
-	Destination   string `form:"destination" example:"production" note:"name of the destination where a connector is installed"`
-	Privilege     string `form:"privilege" example:"view" note:"a role or permission"`
-	ShowInherited bool   `form:"showInherited" note:"if true, this field includes grants that the user inherits through groups" example:"true"`
-	ShowSystem    bool   `form:"showSystem" note:"if true, this shows the connector and other internal grants" example:"false"`
+	User             uid.ID `form:"user" note:"ID of user granted access" example:"6TjWTAgYYu"`
+	Group            uid.ID `form:"group" note:"ID of group granted access" example:"6k3Eqcqu6B"`
+	Resource         string `form:"resource" example:"production.namespace" note:"a resource name"`
+	Destination      string `form:"destination" example:"production" note:"name of the destination where a connector is installed"`
+	Privilege        string `form:"privilege" example:"view" note:"a role or permission"`
+	ShowInherited    bool   `form:"showInherited" note:"if true, this field includes grants that the user inherits through groups" example:"true"`
+	ExcludeConnector bool   `form:"-"`
+	ShowSystem       bool   `form:"showSystem" note:"if true, this shows the connector and other internal grants" example:"false"`
 	BlockingRequest
 	PaginationRequest
 }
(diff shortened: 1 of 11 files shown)


api/grant.go
```diff
@@ -35,13 +35,14 @@ func (r *CreateGrantResponse) StatusCode() int {
...
-	Privilege     string `form:"privilege" example:"view" note:"a role or permission"`
-	ShowInherited bool   `form:"showInherited" note:"if true, this field includes grants that the user inherits through groups" example:"true"`
-	ShowSystem    bool   `form:"showSystem" note:"if true, this shows the connector and other internal grants" example:"false"`
+	User             uid.ID `form:"user" note:"ID of user granted access" example:"6TjWTAgYYu"`
+	Group            uid.ID `form:"group" note:"ID of group granted access" example:"6k3Eqcqu6B"`
+	Resource         string `form:"resource" example:"production.namespace" note:"a resource name"`
+	Destination      string `form:"destination" example:"production" note:"name of the destination where a connector is installed"`
+	Privilege        string `form:"privilege" example:"view" note:"a role or permission"`
+	ShowInherited    bool   `form:"showInherited" note:"if true, this field includes grants that the user inherits through groups" example:"true"`
+	ExcludeConnector bool   `form:"-"`
```
BruceMacD, line 44: It feels like it may make more sense to just filter the connector out on the client-side rather than adding a flag for it specifically. What was the use-case here, the UI?

pdevine: It makes it easier to filter in the UI. This flag is just for backwards compatibility to the old API.

BruceMacD (commented):
Looks good, will approve once tests pass

internal/server/grants_test.go
```diff
@@ -114,7 +117,7 @@ func TestAPI_ListGrants(t *testing.T) {
 	run := func(t *testing.T, tc testCase) {
 		req := httptest.NewRequest(http.MethodGet, tc.urlPath, nil)
 		req.Header.Set("Authorization", "Bearer "+accessKey)
-		req.Header.Add("Infra-Version", "0.18.2")
+		req.Header.Add("Infra-Version", "0.21.0")
```
mxyng, line 120: If this should be testing against the latest API, it should use `apiVersionLatest` instead

internal/server/grants_test.go
```diff
@@ -280,23 +270,18 @@ func TestAPI_ListGrants(t *testing.T) {
...
-						User:      idInGroup,
-						Privilege: "custom1",
-						Resource:  "res1",
-					},
 					{
 						User:      idOther,
-						Privilege: "custom2",
-						Resource:  "res1.ns1",
+						Privilege: "connector",
+						Resource:  "res1.ns2",
```
mxyng, line 276: Nit: this grant doesn't make sense. It's not possible to grant `connector` to a Kubernetes namespace

internal/server/grants_test.go
```diff
@@ -504,6 +494,51 @@ func TestAPI_ListGrants(t *testing.T) {
 				assert.Equal(t, resp.Result().Header.Get("Last-Update-Index"), "10004")
 			},
 		},
+		"migration from <= 0.20.0 show system grants": {
+			urlPath: "/api/grants?showSystem=true",
+			setup: func(t *testing.T, req *http.Request) {
+				req.Header.Set("Authorization", "Bearer "+adminAccessKey(srv))
+				req.Header.Set("Infra-Version", "0.19.0")
```
mxyng, line 501: Why not set version to 0.20.0?

ui/lib/fetch.js
```diff
@@ -1,6 +1,6 @@
 const fetch = global.fetch
 
-const base = '0.19.1'
+const base = '0.21.0'
```
mxyng, line 3: This isn't ideal. It's setting the API version to some version in the future which may cause problems down the road

PR #4055, opened 2023-01-20
Keys remove command improved error message and connector flag

BruceMacD:
## Summary
Add a suggestion in the `infra keys remove ...` command to suggest using the username flag when no access key is found for a name.

Add a `--connector` flag to match the format of `infra keys add --connector`.

```
 $ infra keys remove test 
No access key named 'test' for the current user, another user can be specified with the '--user' flag

$ infra keys remove test --user another@example.com
No access key named 'test' for the user 'another@example.com'

$ infra keys remove --help
Delete an access key

Usage:
  infra keys remove KEY [flags]

Aliases:
  remove, rm

Flags:
      --connector     Remove a key for the connector
      --force         Exit successfully even if access key does not exist
      --user string   The name of the user who owns the key

Global Flags:
      --help                 Display help
      --log-level string     Show logs when running the command [error, warn, info, debug] (default "info")
      --skip-version-check   Skip checking if the CLI is ahead of the server version

$ infra keys remove connector-jLzw17NtqU --connector
Removed access key "connector-jLzw17NtqU"
```

## Checklist

### DIFF
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -208,8 +209,12 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
 			}
 
 			if len(keys) == 0 {
+				errMsg := fmt.Sprintf("No access key named '%s' for the current user, another user can be specified with the '--user' flag", keyName)
+				if options.UserName != "" {
+					errMsg = fmt.Sprintf("No access key named '%s' for the user '%s'", keyName, options.UserName)
+				}
 				return Error{
-					Message: "No access key named: " + keyName,
+					Message: errMsg,
 				}
 			}
 
--- a/internal/cmd/keys.go
+++ b/internal/cmd/keys.go
@@ -162,8 +162,9 @@ $ MY_ACCESS_KEY=$(infra keys add -q --name my-key)
 }
 
 type keyRemoveOptions struct {
-	Force    bool
-	UserName string
+	Force     bool
+	UserName  string
+	Connector bool
 }
 
 func newKeysRemoveCmd(cli *CLI) *cobra.Command {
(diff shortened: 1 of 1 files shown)


internal/cmd/keys.go
```diff
@@ -208,8 +215,12 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
 			}
 
 			if len(keys) == 0 {
+				errMsg := fmt.Sprintf("No access key named '%s' for the current user, another user can be specified with the '--user' flag", keyName)
+				if options.UserName != "" {
+					errMsg = fmt.Sprintf("No access key named '%s' for the user '%s'", keyName, options.UserName)
```
mxyng, line 220:
```suggestion
					errMsg = fmt.Sprintf("No access key named %q exists for the user %q", keyName, options.UserName)
```

internal/cmd/keys.go
```diff
@@ -208,8 +215,12 @@ func newKeysRemoveCmd(cli *CLI) *cobra.Command {
 			}
 
 			if len(keys) == 0 {
+				errMsg := fmt.Sprintf("No access key named '%s' for the current user, another user can be specified with the '--user' flag", keyName)
```
mxyng, line 218:
```suggestion
				errMsg := fmt.Sprintf("No access key named %q for the current user. If trying to remove access key from another user, use the '--user' flag", keyName)
```

mxyng (approved):

internal/cmd/keys.go
dnephin: I wonder if it would be worth it to look up the username and say:

> Access key {keyName} for user {user.Name} does not exist

(similar to the case below)

And maybe something like

> Use the '--user' flag to remove an access key for a different user.

internal/cmd/keys.go
dnephin: Another option:

> Access key {keyName} for user {user.Name} does not exist

dnephin (approved):
Nice improvements, some suggestions for wording to add to the options, but nothing blocking

PR #4058, opened 2023-01-20
Add level to HTTP error logs

dnephin:
Previously they were output with a nil level, this change outputs the logs at warning level.

Warning seems appropriate because we get a lot of noise in these logs for bad TLS requests. 

In the future we could add some logic to inspect the messages and use `Error` for panics, but for now I'm hoping a warning level is better than no level.

https://github.com/rs/zerolog/issues/508 is an issue to add this support to `zerolog`, but it ended up being easy enough to do from our code.

I tested this manually by curling the HTTP server without trusting the CA certificate. I saw a log message with level=WARN.

### DIFF
--- a/internal/logging/logger.go
+++ b/internal/logging/logger.go
@@ -4,6 +4,7 @@ package logging
 import (
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/logging/logger.go
```diff
@@ -69,6 +70,35 @@ func UseFileLogger(filepath string) {
...
+}
+
+func trimTrailingNewline(p []byte) []byte {
+	n := len(p)
+	if n > 0 && p[n-1] == '\n' {
+		// Trim CR added by stdlog.
+		p = p[0 : n-1]
+	}
+	return p
+}
```
dnephin, line 94: This is copied from `zerolog.Logger.Write`. We need to same cleanup applied here because we can't use `Logger.Write`. It emits logs with a nil level.

BruceMacD (approved):
I'm a bit worried that this will increase the amount of noise in the logs a lot, but seems good for now

dnephin:
> I'm a bit worried that this will increase the amount of noise in the logs a lot

How so? Now we always unconditionally write these logs. With this change we'll continue to as long as the error level is INFO, WARN, or above. Only setting log level to ERROR would hide them.

PR #4081, opened 2023-01-23
Remove kin-openapi

dnephin:
## Summary

We only used this library for the struct definitions to mirror the openapi 3 specification. In #4061 dependabot failed to update this dependency because of new unexported fields on those structures.

The commit history here is a bit messy, just saving progress as I go. Probably best viewed as a single diff, and I may remove the intermediate history when I merge.

We don't need much of the complexity in the internal structure (custom json encoding, external references, etc), and we don't use any of the other code in this library.

This PR removes the dependency, and replaces it with a few struct definitions. I omitted a few of the fringe fields that we don't use and are unlikely to ever use. There are still plenty of fields we could trim, but we could also keep them since it's more likely we use them in the future.

In a bunch of places I removed the types that aliased slices and maps, to make it easier to construct the document. I also removed the `Ref` types in a few places where we don't use them.

I also sorted the fields on the struct so that our JSON document did not change. `kin-openapi` was relying on custom json encoding to make that work.

This change only removes a couple dependencies so far, because some of the dependencies are shared by `gin`. When we remove `gin` I think we'll see a lot more of them removed.

The `TestWriteOpenAPIDocToFile` test shows that our JSON document has not changed, and that's the only place where we use these types, so we should be confident that this change hasn't accidentally modified our docs.

### DIFF
--- a/internal/openapi3/openapi3.go
+++ b/internal/openapi3/openapi3.go
@@ -68,28 +68,47 @@ type RequestBody struct {
 // MediaType is specified by OpenAPI/Swagger 3.0 standard.
 // See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#media-type-object
 type MediaType struct {
-	Schema   *openapi3.SchemaRef           `json:"schema,omitempty" yaml:"schema,omitempty"`
-	Example  interface{}                   `json:"example,omitempty" yaml:"example,omitempty"`
-	Examples map[string]*openapi3.Example  `json:"examples,omitempty" yaml:"examples,omitempty"`
-	Encoding map[string]*openapi3.Encoding `json:"encoding,omitempty" yaml:"encoding,omitempty"`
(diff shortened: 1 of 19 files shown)


internal/openapi3/openapi3.go
```diff
@@ -0,0 +1,225 @@
+package openapi3
```
BruceMacD, line 1: We should include a link to the license of the project we based this on.
https://github.com/getkin/kin-openapi/blob/master/LICENSE

dnephin: Ya, possibly. In this case I suspect the license doesn't apply. This is only data structs which are based on the openapi spec. And even those structs are heavily modified.

I'm thinking of maybe removing more of the unused fields, removing the `yaml` struct tags, and changing the comments further, at which point I think it's no longer even close to the original source.

jmorganca (approved):

PR #4084, opened 2023-01-23
fix: check password policy for username/password signup

### LINKED ISSUE #4076: First sign-up does not check password against "weak passwords" file
## Describe the bug
When a user signs-up they can choose any password.

## Expected behavior
The user must pick a password that is not in the "common passwords" file.

mxyng:
## Summary

Check password policy when signing up with username and password. The signup endpoint previously was not using the same password policy checker as every other similar endpoints. This means when the bad passwords checking was added in #3963, signup was missed.

Resolves #4076

### DIFF
--- a/internal/access/credential.go
+++ b/internal/access/credential.go
@@ -86,10 +86,16 @@ func ResetCredential(rCtx RequestContext, user *models.Identity, newPassword str
 		return "", fmt.Errorf("get credential: %w", err)
 	}
 
-	credential.OneTimePassword = true
+	hash, err := bcrypt.GenerateFromPassword([]byte(newPassword), bcrypt.DefaultCost)
+	if err != nil {
+		return "", fmt.Errorf("generate from password: %w", err)
+	}
 
-	if err := updateCredential(tx, credential, newPassword); err != nil {
-		return "", err
+	credential.OneTimePassword = true
+	credential.PasswordHash = hash
+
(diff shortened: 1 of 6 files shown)


internal/access/credential.go
```diff
@@ -86,10 +86,16 @@ func ResetCredential(rCtx RequestContext, user *models.Identity, newPassword str
 		return "", fmt.Errorf("get credential: %w", err)
 	}
 
+	hash, err := bcrypt.GenerateFromPassword([]byte(newPassword), bcrypt.DefaultCost)
```
BruceMacD, line 89: Why doesn't this hash come from the `GenerateFromPassword` function used elsewhere?

BruceMacD, line 89: This hashing feels like it might make more sense in the `server` or `data` packages rather than the `access` check, although I see it was here previously too

mxyng: The password in this case is generated so it doesn't need to be validated against the password policy

mxyng: `access` seems to be the right place. It's where a lot of decision regarding password generation and policy are made. If this hashing is done too early (in `server),`, there's not a lot that can be done wrt password policy. If it's done too late (in `data)`, many of these decision need to be delayed or exceptions has to be made.

BruceMacD (approved):

PR #4085, opened 2023-01-23
fix: invite when user exists

### LINKED ISSUE #4079: Invite sent even though user exists
Users can be re-sent an invitation even if they exist, prompting them to set their password

mxyng:
## Summary

Email invite can be sent to a user who already has a password setup. This regression introduced in #4012 which changed when a user credential was created. Previously creating a user unconditionally creates the user credential which would fix any duplicates when saving to the database. 

Since the linked changed, credentials are created for the invited user only when they accept the email invite. The missing piece is checking for duplicate credentials _before_ sending the invite.

This change does not affect re-sending an invite before the target user has accepted.

The behaviour for non-email user creation is unaffected.

Resolves #4079

### DIFF
--- a/internal/server/users.go
+++ b/internal/server/users.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"database/sql"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/server/users.go
```diff
@@ -83,24 +83,22 @@ func GetUser(c *gin.Context, r *api.GetUserRequest) (*api.User, error) {
...
-	case 1:
-		user.ID = identities[0].ID
-	default:
-		logging.Errorf("Multiple identities match name %q. DB is missing unique index on user names", r.Name)
-		return nil, fmt.Errorf("multiple identities match specified name") // should not happen
+	case err != nil:
+		return nil, fmt.Errorf("get identities: %w", err)
+	}
+
+	for _, provider := range user.Providers {
```
BruceMacD, line 98: seems like this check for if the user exists in the infra provider would make the most sense as the `default` case in the switch statement above where it doesn't need to run when the user didnt exist before.

BruceMacD (approved):

PR #4086, opened 2023-01-24
Extract BlockingRequest logic for using in other requests

dnephin:
## Summary

Branched from #4042 , this is being done in preparation for another blocking request in #4039.

This PR extracts the logic used for long-polling blocking requests into a new function that can be used by other API endpoints.

It also adds some godoc explaining the rules for how to correctly implement one of these operations.

Also rename a file from `notify.go` to `listen.go`, because the file is about listening (not notifying), and move the test case that covered this operation to `listen_test.go`.


BruceMacD (approved):

mxyng:
Closing this for now

PR #4087, opened 2023-01-24
Add data support for SSH group grants 

dnephin:
## Summary

Branched from #4086, extracted from #4039

This PR adds the database query and listen/notify support for SSH group grants.

Related to #3761 

TODO:
* [x] combine MaxUpdateIndex query into ListDestinationAccess
* [ ] test cases for listen

### DIFF
--- a/internal/server/data/destination.go
+++ b/internal/server/data/destination.go
@@ -231,6 +231,7 @@ type DestinationAccess struct {
 }
 
 func ListDestinationAccess(tx ReadTxn, destination string) ([]DestinationAccess, error) {
+	// IMPORTANT: changes to this query likely also need to be applied to DestinationAccessMaxUpdateIndex
 	query := querybuilder.New("SELECT")
 	query.B("identities.id, identities.ssh_login_name, grants.privilege, grants.resource")
 	query.B("FROM grants")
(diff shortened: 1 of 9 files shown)


internal/server/data/destination.go
```diff
@@ -221,3 +222,103 @@ func CountDestinationsByConnectedVersion(tx ReadTxn) ([]DestinationsCount, error
...
+	Privilege        string
+	Resource         string
+}
+
+type ListDestinationAccessResult struct {
+	// Items is the list of items that match the query, excluding soft-deleted rows.
+	Items []DestinationAccess
+
+	// MaxUpdateIndex is the maximum update_index from all the grants and groups
+	// that match the query. This MUST include soft-deleted rows as well.
```
BruceMacD, line 238: Why does it have to include soft-deleted rows, is it because that ensures clients get updated when a row gets deleted?

internal/server/data/destination.go
```diff
@@ -221,3 +222,103 @@ func CountDestinationsByConnectedVersion(tx ReadTxn) ([]DestinationsCount, error
...
+	// The value will be 1 if no records match the query, so that the caller can
+	// block until a record exists.
+	MaxUpdateIndex int64
+}
+
+func ListDestinationAccess(tx ReadTxn, destination string) (*ListDestinationAccessResult, error) {
+	// IMPORTANT: changes to this query likely also need to be applied to
+	// destinationAccessMaxUpdateIndex.
+	query := querybuilder.New("SELECT")
```
BruceMacD, line 247: this is pretty difficult to read, but I don't think there's a better way to do this with our table structure 👍

internal/server/data/identity.go
```diff
@@ -155,6 +155,21 @@ func AssignIdentityToGroups(tx WriteTxn, user *models.ProviderUser, newGroups []
 		}
 	}
 
+	if count := len(idsToRemove) + len(addIDs); count > 0 {
```
BruceMacD, line 158: It feels like the logic above should have used a function like `AddUsersToGroup` rather than manually needing to check if we need to update the index. That could be a future refactor though.

dnephin: Ya, I was considering that. I think I didn't take that approach because I thought there might be some subtle difference in here. I think one difference is that `AddUserstoGroup` is single group many users, where as this is single user many groups.  We can likely share more code here, but I didn't want to do that as part of this change.

internal/server/data/destination.go
```diff
@@ -221,3 +222,104 @@ func CountDestinationsByConnectedVersion(tx ReadTxn) ([]DestinationsCount, error
 func CountAllDestinations(tx ReadTxn) (int64, error) {
 	return countRows(tx, destinationsTable{})
 }
+
+type DestinationAccess struct {
+	UserID           uid.ID
+	UserSSHLoginName string
+	Privilege        string
+	Resource         string
```
BruceMacD, line 230: This might have to be updated to match #4096

PR #4090, opened 2023-01-25
Call validate in versioned handlers

dnephin:
## Summary

@BruceMacD noticed that in cases where we create a versioned API request struct we were not applying validation rules, because the versioned structs did have any rules.

To ensure that we apply the right request validation this PR adds the missing calls to `validate.Validate`,  and updates the docs so that we don't forget in the future.

## Related Issues

Resolves https://github.com/infrahq/internal/issues/41


jmorganca (approved):
LGTM

mxyng (approved):

PR #4091, opened 2023-01-25
remove unused wasCreated field value from grants response

mxyng:
## Summary

This value as a JSON field in the response is unnecessary since the HTTP status code can be used to determine if the request was effective or not. The only place where this may be useful is in the bulk request where each grant is tagged with `wasCreated`. Removing `wasCreated` from this response makes it harder to determine if a particular grant has been created but since nothing is checking that value right now, it doesn't seem very important.


dnephin (approved):
Great cleanup!

PR #4092, opened 2023-01-25
use latest api version unless testing a previous version

mxyng:
## Summary

Use `apiVersionLatest` consistently in tests unless explicitly testing a previous version since using non-latest versions can produces unexpected failures when changing the API.


dnephin (approved):
LGTM. Nice cleanup!

PR #4093, opened 2023-01-25
use latest version in api/client.go

mxyng:
## Summary

Infra API is tightly coupled with the API client since its structs are used in the API handlers and also used to generate API docs. This means any changes to the API requires changing the API client so it doesn't make sense to pin the JSON format to anything other than the version implemented by the server.

Note: this version gets set to git tag when building a release so released builds won't see any change.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -15,11 +15,10 @@ import (
 
 	"github.com/Masterminds/semver/v3"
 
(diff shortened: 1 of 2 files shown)


dnephin (approved):
LGTM!

PR #4095, opened 2023-01-25
Combine api.Client hooks so that we can remove one

dnephin:
## Summary

I noticed we had two hooks in the `api.Client`,  `ObserveFunc` and `OnUnauthorized`. Both of these hooks are called at about the same place, so I think we may be able to combine these and use only `ObserveFunc`.

I haven't done much testing on this, and I suspect we don't have much automated test coverage.

Suggestions for what needs to be tested manually?

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -40,12 +40,11 @@ type Client struct {
 	HTTP      http.Client
 	// Headers are HTTP headers that will be added to every request made by the Client.
 	Headers http.Header
-	// OnUnauthorized is a callback hook for the client to get notified of a 401 Unauthorized response to any query.
-	// This is useful as clients often need to discard expired access keys.
-	OnUnauthorized func()
 
-	// ObserveFunc is a callback to measure and record the status and duration of the request
-	ObserveFunc func(time.Time, *http.Request, *http.Response, error)
+	// ObserveFunc is called after the Client performs an API request. Callers can
+	// use this hook to emit telemetry about API calls, or to perform some action
+	// based on the response.
+	ObserveFunc func(elapsed time.Time, req *http.Request, resp *http.Response, err error)
 }
 
 // checkError checks the resp for an error code, and returns an api.Error with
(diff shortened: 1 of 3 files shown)


api/client.go
```diff
@@ -121,10 +120,6 @@ func request[Res any](client Client, req *http.Request) (*Res, error) {
 		client.ObserveFunc(start, req, resp, err)
 	}
 
-	if resp != nil && resp.StatusCode == 401 && client.OnUnauthorized != nil {
-		defer client.OnUnauthorized()
```
dnephin, line 125: The one difference in this PR is that we no longer `defer` this operation, we do it immediately.

In the connector, cancelling the context immediately seems fine, it does the shutdown a bit earlier.

In the CLI doing the logout operation earlier seems like it should be ok.

Is there something I'm missing that requires these operations to happen after reading the request body?

I guess maybe it'll change the error from 401 to either an error reading the response?

If that's the case I wonder if `ObserveFunc` should return an error, so that we avoid getting to that steps.

mxyng: The only reason `ObserveFunc` is not deferred is so the HTTP client request can be measured without measuring things like reading the request body. It seems one way to resolve this is to take the elapsed time first and pass it into a deferred `ObserveFunc`, e.g. 

```go
if client.ObserveFunc != nil {
    elapsed := time.Since(start)
    defer client.ObserveFunc(elapsed, req, resp, err)
}
```

mxyng:
Testing the connector hooks should be straight forward. A working connector will pump out HTTP client metrics as part of normal operation. To trigger the unauthorized branch, revoking the connector's access key should be sufficient.

Testing the CLI client's hooks might be more tricky. One option is to test against a custom Infra server with a low session duration (minutes). This should cause the login to expire quickly, triggering the hook. Another thing to be mindful of when testing the CLI is the agent which could also trigger the hook as its polling for new grants.

mxyng (approved):
Looking good other than what you pointed out with the deferred hook. I'll leave it up to you to find the best solution.

mxyng:
Closing for now

PR #4096, opened 2023-01-26
refactor grants data model

mxyng:
Update the grant data model to split destination name from resource. This separation creates a clearer boundary between destinations and destination-specific resources. #4099, the API will be updated to follow this pattern which allows a future PR to allow dots in destination names.

### DIFF
--- a/internal/access/access_test.go
+++ b/internal/access/access_test.go
@@ -40,9 +40,9 @@ func setupAccessTestContext(t *testing.T) RequestContext {
 	}
 
 	adminGrant := &models.Grant{
-		Subject:   models.NewSubjectForUser(admin.ID),
-		Privilege: models.InfraAdminRole,
-		Resource:  ResourceInfraAPI,
+		Subject:         models.NewSubjectForUser(admin.ID),
+		Privilege:       models.InfraAdminRole,
+		DestinationName: models.GrantDestinationInfra,
 	}
 	err = data.CreateGrant(tx, adminGrant)
 	assert.NilError(t, err)
@@ -67,11 +67,11 @@ func TestIsAuthorized(t *testing.T) {
 	err := data.CreateIdentity(db, admin)
 	assert.NilError(t, err)
 
-	grant(t, db, admin, models.NewSubjectForUser(888), "read", ResourceInfraAPI)
+	grant(t, db, admin, models.NewSubjectForUser(888), "read", models.GrantDestinationInfra)
 	can(t, db, 888, "read")
 	cant(t, db, 888, "write")
 
-	grant(t, db, admin, models.NewSubjectForUser(777), "write", ResourceInfraAPI)
+	grant(t, db, admin, models.NewSubjectForUser(777), "write", models.GrantDestinationInfra)
 	cant(t, db, 777, "read")
 	can(t, db, 777, "write")
 }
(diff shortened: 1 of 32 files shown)


internal/access/access_test.go
```diff
@@ -159,12 +159,13 @@ func TestIsAuthorized_MoreCases(t *testing.T) {
 	})
 }
 
-func grant(t *testing.T, db data.WriteTxn, createdBy *models.Identity, subject models.Subject, privilege, resource string) {
+func grant(t *testing.T, db data.WriteTxn, createdBy *models.Identity, subject models.Subject, privilege, destinationName, destinationResource string) {
```
BruceMacD, line 162: Looks to me like `destinationResource` can be removed here.

internal/server/data/destination_test.go
```diff
@@ -335,19 +335,19 @@ func TestDeleteDestination(t *testing.T) {
...
+				Privilege:       "view",
+				DestinationName: "kube",
 			},
 			{
-				Subject:   models.NewSubjectForUser(1234567),
-				Privilege: "view",
-				Resource:  "kube.namespace",
+				Subject:         models.NewSubjectForUser(1234567),
+				Privilege:       "view",
+				DestinationName: "kube", DestinationResource: "namespace",
```
BruceMacD, line 345:
```suggestion
				DestinationName: "kube",
				DestinationResource: "namespace",
```

internal/server/data/grant.go
```diff
@@ -249,10 +256,9 @@ type DeleteGrantsOptions struct {
 	ByID uid.ID
 	// BySubject instructs DeleteGrants to delete all grants that match this
 	// subject. When set other fields below this on this struct are ignored.
-	BySubject models.Subject
-	// ByDestination instructs DeleteGrants to delete all grants that match
-	// this destination in their resource, including namespaces.
-	ByDestination string
+	BySubject             models.Subject
+	ByDestinationName     string
+	ByDestinationResource string
```
BruceMacD, line 261:
```suggestion
        // ByDestinationResource selects grants which specify a given resource, it requires ByDestinationName to be specified
	ByDestinationResource string
```

internal/server/data/grant.go
```diff
@@ -223,9 +228,11 @@ type GrantsMaxUpdateIndexOptions struct {
 func GrantsMaxUpdateIndex(tx ReadTxn, opts GrantsMaxUpdateIndexOptions) (int64, error) {
 	query := querybuilder.New("SELECT max(update_index) FROM grants")
 	query.B("WHERE organization_id = ?", tx.OrganizationID())
-
-	if opts.ByDestination != "" {
-		grantsByDestination(query, opts.ByDestination)
+	if opts.ByDestinationName != "" {
+		query.B("AND destination_name = ?", opts.ByDestinationName)
+		if opts.ByDestinationResource != "" {
```
BruceMacD, line 233: We should error if `ByDestinationResource` is specified but `ByDestinationName` isn't to catch developer errors (same for the delete case).

internal/server/data/migrations.go
```diff
@@ -1329,3 +1330,58 @@ func addAccessKeyIssuedForKind() *migrator.Migration {
...
+				if err != nil {
+					return err
+				}
+
+				type grantResource struct {
+					ID          uid.ID
+					ResourceURN string
+				}
+
+				rows, err := tx.Query(`SELECT id, resource FROM grants WHERE deleted_at is null`)
```
BruceMacD, line 1353: Should we still select the deleted records? That way we can still track what they were after the `resource` column is dropped. Not sure if this could mess with the `deleted_at` timestamp though.

BruceMacD (commented):
Looking good so far

internal/server/data/migrations.go
mxyng: I was thinking about that myself. Omitting deleted grants significantly reduces how many grants are migrated but that might not be a good thing in this case. Other migrations touch all records, deleted and not deleted so I think that's what I'll do here too

mxyng:
Closing for now

PR #4097, opened 2023-01-26
fix: allow group names with commas

### LINKED ISSUE #3138: Cannot use comma in group names

BruceMacD:
## Summary
Update group tracking to use a JSON binary in the database, rather than a text field separated by commas. This allows for storing group commas.

## Checklist

## Related Issues

Resolves #3138

### DIFF
--- a/internal/server/data/migrations.go
+++ b/internal/server/data/migrations.go
@@ -1343,9 +1343,6 @@ func storeProviderUserGroupsArray() *migrator.Migration {
 				// this migration has already been performed
 				return nil
 			}
-			if _, err := tx.Exec(`ALTER TABLE provider_users ADD COLUMN IF NOT EXISTS groups_json jsonb;`); err != nil {
-				return err
-			}
 			// copy the groups string to the new groupsArray column, this could be a large operation if there are a lot of users
 			type userGroups struct {
 				IdentityID uid.ID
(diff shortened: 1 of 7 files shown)


internal/server/data/migrations.go
```diff
@@ -1329,3 +1330,52 @@ func addAccessKeyIssuedForKind() *migrator.Migration {
...
+		Migrate: func(tx migrator.DB) error {
+			var groupType string
+			if err := tx.QueryRow(`SELECT pg_typeof(groups) FROM provider_users LIMIT 1;`).Scan(&groupType); err != nil {
+				return err
+			}
+			if groupType == "jsonb" {
+				// this migration has already been performed
+				return nil
+			}
+			if _, err := tx.Exec(`ALTER TABLE provider_users ADD COLUMN IF NOT EXISTS groups_json jsonb;`); err != nil {
```
mxyng, line 1346: Does this need a temporary column? It seems you can retrieve the user groups, change the column, then reinsert the rows, e.g. add ALTER TABLE between the scanRows on line 1359 and the for loop on line 1366

BruceMacD: Good observation, made this change.

internal/server/data/migrations.go
```diff
@@ -1329,3 +1330,48 @@ func addAccessKeyIssuedForKind() *migrator.Migration {
...
+			users, err := scanRows(rows, func(u *userGroups) []any {
+				return []any{&u.IdentityID, &u.ProviderID, &u.Groups}
+			})
+			if err != nil {
+				return err
+			}
+			if _, err := tx.Exec(`ALTER TABLE provider_users DROP COLUMN IF EXISTS groups;`); err != nil {
+				return err
+			}
+			if _, err := tx.Exec(`ALTER TABLE provider_users ADD COLUMN IF NOT EXISTS groups jsonb;`); err != nil {
```
mxyng, line 1365: You should be able to use a single `ALTER TABLE`

BruceMacD: Just tried this and it turns out you can't drop and add in a single statement.

mxyng (approved):

internal/server/data/schema.sql
```diff
@@ -259,15 +259,15 @@ CREATE TABLE provider_users (
...
     last_update timestamp with time zone,
     redirect_url text,
     access_token text,
     refresh_token text,
     expires_at timestamp with time zone,
     given_name text DEFAULT ''::text,
     family_name text DEFAULT ''::text,
-    active boolean DEFAULT true
+    active boolean DEFAULT true,
+    groups jsonb
```
jmorganca, line 270: nice choice of `jsonb`

jmorganca (approved):

PR #4099, opened 2023-01-27
api(grants): split resource into name and resource

mxyng:
## Summary

Follow up to #4096. This PR updates the API to separate out grant `resource` into `destinationName` and `destinationResource` where `destinationName` is the real destination the grant applies to and `destinationResource` is the destination-specific resource, e.g. Kubernetes namespace.


mxyng:
Closing for now

PR #4118, opened 2023-02-09
cli(grants): add resource command line parameter

mxyng:
Follow up to #4099

### DIFF
--- a/internal/cmd/grants.go
+++ b/internal/cmd/grants.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	survey "github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/cobra"
 
 	"github.com/infrahq/infra/api"
@@ -23,6 +24,7 @@ type grantsCmdOptions struct {
 	Role        string
 	Force       bool
 	Inherited   bool
+	DryRun      bool
 }
 
 func newGrantsCmd(cli *CLI) *cobra.Command {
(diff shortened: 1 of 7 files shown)


internal/cmd/list.go
```diff
@@ -46,39 +51,24 @@ func list(cli *CLI) error {
...
+		})
 	}
 
-	var rows []row
-	for _, k := range resources {
-		v, ok := grantsByResource[k]
-		if !ok {
-			// should not be possible
-			return fmt.Errorf("unexpected value in grants: %s", k)
+	sort.SliceStable(rows, func(i, j int) bool {
```
mxyng, line 61: This changes the output slightly. Instead of aggregating access for a particular destination/resource to a single row, this keeps it simple and gives each grant their own row.

internal/cmd/grants.go
```diff
@@ -286,35 +293,78 @@ func removeGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
...
+		}
+
+		if cmdOptions.Role != "" {
+			fmt.Fprintf(&sb, " role %q", cmdOptions.Role)
+		}
+
+		sb.WriteString(". Continue?")
+
+		var confirmRemove bool
+		if err := survey.AskOne(&survey.Confirm{Message: sb.String()}, &confirmRemove, cli.surveyIO); err != nil {
```
mxyng, line 324: This confirmation is useful since otherwise it will remove multiple grants automatically which may be a surprise to the user

internal/cmd/grants.go
```diff
@@ -286,35 +293,78 @@ func removeGrant(cli *CLI, cmdOptions grantsCmdOptions) error {
...
 
 	for _, g := range grants.Items {
 		logging.Debugf("call server: delete grant %s", g.ID)
-		err := client.DeleteGrant(ctx, g.ID)
-		if err != nil {
-			if api.ErrorStatusCode(err) == 403 {
-				logging.Debugf("%s", err.Error())
-				return Error{
-					Message: "Cannot revoke grants: missing privileges for DeleteGrant",
+		if !cmdOptions.DryRun {
```
mxyng, line 335: `--dryrun` allows the user to preview changes before actually running it against the server. It's particularly useful with removals since it's a destructive process but can be useful for `add` as well

mxyng:
Closing for now

